PR #2, opened 2024-07-26
Give decent error message if model not found

rhatdan:
Rename podman_build.sh to build.sh


ci.sh
```diff
@@ -11,14 +11,14 @@ main() {
     apt install -y python3-autopep8
   fi
 
-  ./podman-build.sh
```
ericcurtin, line 14: The thing is about this is, it doesn't build ramalama, it only builds container images...

Maybe container-build.sh ?

Not a blocking issue for this PR anyway

ericcurtin: We can catch in a follow on PR if needs be

ci.sh
```diff
@@ -7,18 +7,22 @@ available() {
 main() {
   set -ex -o pipefail
 
-  if ! available autopep8 && available apt; then
-    apt install -y python3-autopep8
+  if ! available autopep8; then
```
ericcurtin, line 10: I already fixed the sudo thing, it's one of the reasons we have conflicts in this PR

ericcurtin (approved):
LGTM, waiting on green build

rhatdan:
@ericcurtin can we use the small model for ci rather then granite?

ericcurtin:
We just need a chmod a+x or whatever to fix CI here I think, assuming python style is autopep8 auto-formatted, etc.

ericcurtin:
We can do the moondream change is this PR if needs be too

ericcurtin:
"tinyllama" is even smaller

https://ollama.com/library/tinyllama

PR #5, opened 2024-07-27
Remove container from diagram

ericcurtin:
We want to remove OCI containers as a strict dependancy of ramalama, so running in containers is options, reflecting this in the diagram.


rhatdan:
LGTM

PR #6, opened 2024-07-27
If podman or docker is available and we are not in a container

ericcurtin:
We should re-exec this script inside a llama.cpp container (incomplete
in this commit), truth is these AI runtime evironment are complex to
set up by say installing a few packages, etc. The often require a
certain version of mesa, llama.cpp, kernel drivers, ROCm, the list goes
on and on depending on hardware. Although if one does want
to run outside a container and or say they are running on macOS we
should support that.

### DIFF
--- a/ramalama
+++ b/ramalama
@@ -5,6 +5,7 @@ import sys
 import subprocess
 import json
 import hashlib
+import shutil
 
 x = False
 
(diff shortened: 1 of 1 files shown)


ramalama
```diff
@@ -160,6 +160,26 @@ def get_ramalama_store():
...
+
+
+def available(command):
+    return shutil.which(command) is not None
+
+
+def select_container_manager():
+    if available("podman"):
+        return "podman"
+    elif available("docker"):
```
rhatdan, line 177: sed s/elif/if

ericcurtin: Pushing with this change

rhatdan:
LGTM

PR #7, opened 2024-07-27
macOS build

ericcurtin:
To ensure things work on macOS.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -16,22 +16,6 @@ amd_lshw() {
   lshw -c display -numeric -disable network | grep -q 'vendor: .* \[1002\]'
 }
 
-gpu_check() {
-  if available lspci && lspci -d '10de:' | grep -q 'NVIDIA'; then
-    nvidia_available="true"
-  elif available lshw && nvidia_lshw; then
-    nvidia_available="true"
-  elif available nvidia-smi; then
-    nvidia_available="true"
-  fi
-
-  if available lspci && lspci -d '1002:' | grep -q 'AMD'; then
-    amd_available="true"
-  elif available lshw && amd_lshw; then
-    amd_available="true"
-  fi
-}
-
 download() {
   local curl_cmd=("curl" "--globoff" "--location" "--proto-default" "https")
   curl_cmd+=("-o" "$from" "--remote-time" "--retry" "10" "--retry-max-time")
@@ -68,8 +52,6 @@ main() {
   install -D -m755 "$from" "$bindir/"
 
   if false; then # to be done
-    local nvidia_available="false"
-    local amd_available="false"
     gpu_check
   fi
 }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,15 +1,18 @@
 name: ci
-on:
-  push:
-    branches: [ "main" ]
-  pull_request:
-    branches: [ "main" ]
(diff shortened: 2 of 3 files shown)


install.sh
```diff
@@ -16,21 +16,21 @@ amd_lshw() {
...
 
-gpu_check() {
-  if available lspci && lspci -d '10de:' | grep -q 'NVIDIA'; then
-    nvidia_available="true"
-  elif available lshw && nvidia_lshw; then
-    nvidia_available="true"
-  elif available nvidia-smi; then
-    nvidia_available="true"
-  fi
+# gpu_check() {
```
rhatdan, line 19: Why are you commenting this out?

install.sh
```diff
@@ -16,21 +16,21 @@ amd_lshw() {
...
+  # if available lspci && lspci -d '10de:' | grep -q 'NVIDIA'; then
+  #  nvidia_available="true"
+  # elif available lshw && nvidia_lshw; then
```
rhatdan, line 22: This line looks wrong.

Shouldn't this be 
if available nvidia_lshw && nvidia_lshw; then?

install.sh
```diff
@@ -16,21 +16,21 @@ amd_lshw() {
...
 
-  if available lspci && lspci -d '1002:' | grep -q 'AMD'; then
-    amd_available="true"
-  elif available lshw && amd_lshw; then
-    amd_available="true"
-  fi
-}
+  # if available lspci && lspci -d '1002:' | grep -q 'AMD'; then
+    # amd_available="true"
+  # elif available lshw && amd_lshw; then
```
rhatdan, line 30: Ditto
if available amd_lshw && amd_lshw; then?

install.sh
ericcurtin: I never completed the GPU checking stuff... Trying to get the design correct first before bringing on the maintenance of all the GPU backends

And because I started running shellcheck on everything, shellcheck was correctly identifying this stuff as unused shell scripts (correctly)...

Maybe I should just delete this stuff...

install.sh
ericcurtin: nvidia_lshw is a bash function a couple of lines up

lshw is a binary required to query for GPU type, not always installed.

ericcurtin:
Deleting commented out lines for now.

When we get around to enhancing this more here are some decent reference scripts, we may have to write in a portable POSIX-like manner to be compatible with bash and zsh for macOS.

Eventually we will have to rpm package properly, but it's really not straightforward for these AI/LLM tools, optional dependencies based on hardware, instable APIs that need to be changed regularly. The AI world is kinda young also, so you really want to be up-to date with things like llama.cpp to be compatible with the latest LLMs like Granite Code, Llama 3 as examples.

https://github.com/mudler/LocalAI/blob/master/docs/static/install.sh
https://github.com/ollama/ollama/blob/main/scripts/install.sh
https://github.com/ericcurtin/podman-ollama/blob/main/ollama-install.sh
https://github.com/ericcurtin/podman-ollama/blob/main/install.sh

install.sh
rhatdan: I am fine with commenting it out.

install.sh
rhatdan: Thanks.

rhatdan:
LGTM

PR #8, opened 2024-07-30
Implement ramalama run/serve

ericcurtin:
Now we can run "ramalama run/serve granite-code", if not using
container, one must at least build/install llama.cpp. Added huggingface
support.

### DIFF
--- a/ramalama
+++ b/ramalama
@@ -6,6 +6,7 @@ import subprocess
 import json
 import hashlib
 import shutil
+import re
 
 x = False
 
@@ -55,7 +56,7 @@ def run_command(args):
     if x:
         print(*args)
 
-    subprocess.run(args, check=True)
+    return subprocess.run(args, check=True, stdout=subprocess.PIPE)
 
 
 def run_curl_command(args, filename):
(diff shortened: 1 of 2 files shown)


ramalama
```diff
@@ -68,9 +68,9 @@ def run_curl_command(args, filename):
             sys.exit(e.returncode)
 
 
-def pull_ollama_manifest(ramalama_store, manifests, accept, registry_head, model_tag):
+def pull_ollama_manifest(repos_ollama, manifests, accept, registry_head, model_tag):
```
rhatdan, line 71: Why not just call it repos.  No need to keep repeating _ollama

ramalama
```diff
@@ -183,14 +202,26 @@ def select_container_manager():
...
+
     if len(sys.argv) < 2:
         usage()
 
-    ramalama_store = get_ramalama_store()
     command = sys.argv[1]
     if command == "pull" and len(sys.argv) > 2:
-        pull_cli(ramalama_store + "/repos/ollama",
-                 ramalama_store + "/models/ollama", sys.argv[2])
+        pull_cli(ramalama_store, sys.argv[2])
```
rhatdan, line 220: switch command {
case "pull":
case "run":
}

rhatdan: Do you have commands that handle 2 options?

ramalama
ericcurtin: My thinking was shortly I will bring in hugging face support again and there will likely be "repos_hf", "repos_ollama" directory is:

~/.local/share/ramalama/repos/ollama

shortly there will be:

~/.local/share/ramalama/repos/huggingface

ramalama
ericcurtin: Yeah there will be:

ramalama list/ls

as an example that lists all models downloaded and available is coming.

Will change to switch, apparently they call it match in python for some reason

ramalama
rhatdan: I think you should embed the location of the store inside of the functions, and not force it to be different.

Once we have multiple pull_manifests() functions, then we can start to consildate functionality and get as much reuse as possible.

ramalama
ericcurtin: How would we feel about going back to if/else if/else? The python switch equivalent in Python 3.10 (2021) and macOS build is failing here, I'm guessing the python3 version on macOS is older

ericcurtin: Going back to if, else, etc. fixed it anyway

rhatdan:
Ok merging, we can cleanup later.

PR #11, opened 2024-07-31
Add ramalama list

ericcurtin:
To list all downloaded models

### DIFF
--- a/ramalama
+++ b/ramalama
@@ -1,12 +1,15 @@
 #!/usr/bin/python3
 
 import os
+import glob
 import sys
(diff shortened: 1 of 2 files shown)


ramalama
```diff
@@ -225,6 +287,7 @@ def usage():
     print("  run MODEL        Run a model")
     print("  pull MODEL       Pull a model")
     print("  serve MODEL      Serve a model")
+    print("  list             List models")
```
rhatdan, line 290: Alphabetize list.

PR #12, opened 2024-07-31
Refactoring, huggingface

ericcurtin:
Trying to keep functions no larger than a 80*43 terminal.


ericcurtin:
> LGTM I think we should start to think about breaking this up into separate python files.
> 
> ramalama.py ollama.py huggingface.py oci.py
> 
> We should also look into using some kind of classes/objects although it has been a number of years since I dealt with python.

Yeah I'm in the same boat as regards writing python. I can write it, but I'm no pythonista.

Definitely long-term this has to be split up

PR #14, opened 2024-07-31
Use dict for commands

rhatdan:


ericcurtin:
Just needs a:

autopep8 -i ramalama

PR #15, opened 2024-07-31
Use container right now

ericcurtin:
ramalama run/serve right now require the container, it has the version of llama.cpp that works.

Long-term we may be able to remove this.

### DIFF
--- a/ramalama
+++ b/ramalama
@@ -215,6 +215,7 @@ def list_cli(ramalama_store, args):
 
 
 funcDict["list"] = list_cli
+funcDict["ls"] = list_cli
 
 
 def pull_huggingface(model, ramalama_store):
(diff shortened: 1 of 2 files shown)


ramalama
```diff
@@ -342,6 +342,11 @@ def main(args):
     conman = select_container_manager()
     ramalama_store = get_ramalama_store()
 
+    if conman:
```
rhatdan, line 345: Ok so we default to running a container for use with the server.

rhatdan:
LGTM, merge when tests pass.

ramalama
ericcurtin: Sounds good to me, the other problem is we need to add:

pip install "huggingface_hub[cli]==0.24.2"

to the install script but that's no biggie.

ericcurtin: Re-pushed, it now does the re-exec in a container thing for just run/serve and installs the huggingface dependency in install script

ramalama
```diff
@@ -215,6 +215,7 @@ def list_cli(ramalama_store, args):
 
 
 funcDict["list"] = list_cli
+funcDict["ls"] = list_cli
```
ericcurtin, line 218: Ollama has ls as an alias to list, so trying to keep with "norms" from other tools

ericcurtin:
I think it should work this time, just wanted to get a test in to prevent run/serve breakage again

If we can get the llama-cpp-python library working well, it might help reduce dependencies on containers.

But these LLM environments in general can get complex

ericcurtin:
Merging, people might try this out in the next few minutes, would hope they don't install the broken version

ericcurtin:
We should create a stable branch soon

PR #17, opened 2024-07-31
Support non-default namespace for Ollama

ericcurtin:
Anything that wasn't in the default namespace didn't work before, like:

ben1t0/tiny-llm

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -39,6 +39,7 @@ main() {
   set -o pipefail
 
   ./ramalama pull tinyllama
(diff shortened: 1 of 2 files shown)


ericcurtin:
With this PR, one of our suggestions in the README:

ramalama run instructlab/merlinite-7b-lab

will actually work.

It doesn't work right now because of no support for non-default namespaces like "instructlab/"

rhatdan:
LGTM

PR #20, opened 2024-08-01
Change installer to stable branch

ericcurtin:
The s branch is the stable branch, we've already had a few breaking PRs merged, naturally enough. This avoids those breakages hitting users.


rhatdan:
LGTM
But we need to update the stable branch.

ericcurtin:
Updated I just pushed there

PR #30, opened 2024-08-02
Require containers for Linux always if podman is available

ericcurtin:
Less confusing UX, long-term we may want to remove the dependancies on containers for macOS support and those that want to deploy without containers.

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -13,7 +13,7 @@ main() {
   fi
 
   local os
-  os="$(uname)"
+  os="$(uname -s)"
   if ! available autopep8; then
     if [ "$os" = "Linux" ]; then
       if available apt; then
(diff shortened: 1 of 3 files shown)


ericcurtin:
This is the UX related one @rhatdan I'll let you decide if we want to merge this one...

README.md
```diff
@@ -51,11 +51,6 @@ use a container engine like Podman to pull an OCI container image with the
 appropriate code and libraries to run the AI Model. This can take a long time to setup, but only on the first run.
 ```
 $ ramalama run instructlab/merlinite-7b-lab
-Copying blob 5448ec8c0696 [--------------------------------------] 0.0b / 63.6MiB (skipped: 0.0b = 0.00%)
```
ericcurtin, line 54: We would still see this if on first run, the user ran:

$ ramalama run instructlab/merlinite-7b-lab

and they don't do the pull explicitly

rhatdan, line 54: Why remove this, users will still see it?

rhatdan:
I think we should simplify and just force containers to start and then if users want to do it without containers we can move to it.  For now this would keep the complexity of the app less.

README.md
ericcurtin: I'll put it back in, what would happen in reality is they would see this, then the

```
##########
```

progress bar from curl downloading the model. Then:

```
>
```

But, I'll put it back in :)

PR #32, opened 2024-08-07
Add dryrun option

ericcurtin:
Useful for printing what we will run via podman in "podman run" format

### DIFF
--- a/ramalama
+++ b/ramalama
@@ -306,7 +306,7 @@ def serve_cli(ramalama_store, args):
 funcDict["serve"] = serve_cli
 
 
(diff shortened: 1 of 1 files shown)


ramalama
```diff
@@ -351,11 +351,20 @@ def main(args):
     ramalama_store = get_ramalama_store()
 
     try:
+        dryrun = False
+        if args[0] in ("--dryrun", "-d"):
```
rhatdan, line 355: nit: I would not loose the -d short option, this will not be done enough to warrant, and at some point in the future we might want to use -d for a more common option.

PR #33, opened 2024-08-08
intial push to quay

cooktheryan:


ericcurtin (approved):

ericcurtin:
@cooktheryan 

Need to sign the commit via something like:

```
git commit -a --amend -s
```

for the DCO build to be green. I think that's a "containers" org thing.

ericcurtin (approved):

ericcurtin:
@cooktheryan feel free to merge if this is ready :)

Ideally we would build new containers and push to quay on:

- every merge to main
- nightly

important to note long term we will probably have container images for:

- latest-rocm
- latest-cuda
- latest
- stable-rocm
- stable-cuda
- stable

stable would be built via the "s" branch, nightly and on merge to "s" branch also.

cooktheryan:
Merging we will fix the above as we proceed

ericcurtin:
And the "stable" ones probably shouldn't have nightly builds at all I guess... Just in case, but re-build and push on merge to stable yes...

PR #35, opened 2024-08-08
py: fix huggingface repos layout

### LINKED ISSUE #23: huggingface two identically named .ggufs different repos
To be honest I don't think we are handling huggingface right yet, for the case where there is two .gguf files, same name different repo. Need to add a test for that, opening an issue.

Ollama support seems fine though.

tarilabs:
Hi 👋 
I'm familiarizing with the project code.

It looks to me:

a file from hf:// would have downloaded straigh into {ramalama_store}/repos/huggingface
and possibly colliding with other model files
having same name coming from other repos.

the huggingface_download() was actually called
twice, fixed by calling once.

This PR addresses both points.
If that was intended behaviour, kindly explain because I might be missing something 😅 

Edit: as I was later browsing open issues, this one seems indeed resolves #23 

# Before (today)

```
./ramalama pull huggingface://tarilabs/mnist/README.md
# ...

tree ~/.local/share/ramalama
/Users/mmortari/.local/share/ramalama
├── models
│   ├── huggingface
│   │   └── tarilabs
│   │       └── mnist
│   │           └── README.md -> ../../../../repos/huggingface/README.md
│   └── ollama
└── repos
    ├── huggingface
    │   └── README.md    # <-- notice the model file is downloaded in root here? 🤔 
    └── ollama

```

# After

```
./ramalama pull huggingface://tarilabs/mnist/README.md
# ...

tree ~/.local/share/ramalama 
/Users/mmortari/.local/share/ramalama
├── models
│   ├── huggingface
│   │   └── tarilabs
│   │       └── mnist
│   │           └── README.md -> ../../../../repos/huggingface/tarilabs/mnist/README.md
│   └── ollama
└── repos
    ├── huggingface
    │   └── tarilabs    # <-- HF org/user/ns
    │       └── mnist    # <-- HF repo
    │           └── README.md
    └── ollama
```

### DIFF
--- a/ramalama
+++ b/ramalama
@@ -9,6 +9,7 @@ import hashlib
 import shutil
 import time
 import re
+import logging
 from pathlib import Path
 
 x = False
@@ -135,11 +136,10 @@ def init_pull(repos_ollama, manifests, accept, registry_head, model_name, model_
 
 
 def huggingface_download(ramalama_store, model, directory, filename):
-    return run_cmd(["huggingface-cli", "download", directory, filename, "--cache-dir", ramalama_store + "/repos/huggingface/.cache", "--local-dir", ramalama_store + "/repos/huggingface"])
+    return run_cmd(["huggingface-cli", "download", directory, filename, "--cache-dir", ramalama_store + "/repos/huggingface/.cache", "--local-dir", ramalama_store + "/repos/huggingface/" + directory])
 
 
 def try_huggingface_download(ramalama_store, model, directory, filename):
-    huggingface_download(ramalama_store, model, directory, filename)
     proc = huggingface_download(ramalama_store, model, directory, filename)
     return proc.stdout.decode('utf-8')
 


ericcurtin:
Thanks, I also had this done on a branch but never pushed:

https://github.com/containers/ramalama/pull/36

Lets merge this one

ericcurtin:
I also did a:

os.makedirs

but looks like we don't need it,

ericcurtin (approved):

rhatdan:
I think we need to start defining mechanisms for short names and a way to default images to be pulled from a specific model repository.

ramalama
```diff
@@ -135,11 +136,10 @@ def init_pull(repos_ollama, manifests, accept, registry_head, model_name, model_
 
 
 def huggingface_download(ramalama_store, model, directory, filename):
-    return run_cmd(["huggingface-cli", "download", directory, filename, "--cache-dir", ramalama_store + "/repos/huggingface/.cache", "--local-dir", ramalama_store + "/repos/huggingface"])
+    return run_cmd(["huggingface-cli", "download", directory, filename, "--cache-dir", ramalama_store + "/repos/huggingface/.cache", "--local-dir", ramalama_store + "/repos/huggingface/" + directory])
 
 
 def try_huggingface_download(ramalama_store, model, directory, filename):
-    huggingface_download(ramalama_store, model, directory, filename)
```
ericcurtin, line 142: Actually I missed this, this may be a bug.

The first time you download the model the stdout content is much more than just a filename

ericcurtin: Actually seems like it works anyway, nevermind

ericcurtin:
> I think we need to start defining mechanisms for short names and a way to default images to be pulled from a specific model repository.

I agree @rhatdan I don't see any reason to not copy podman initially:

/etc/containers/registries.conf.d/000-shortnames.conf

an eventual fallback to ollama is kinda nice too though, it's kinda built-in to ollama

PR #40, opened 2024-08-12
Run server on alternative port

ericcurtin:
No need for calculated_checksum variable

Simplify verify_checksum function

Some refactoring


rhatdan:
LGTM

PR #44, opened 2024-08-14
hacking to get nightly to run

cooktheryan:
Going to iterate on this a few times and use the manual launch of the workflow. Will probably have multiple PRs on this.


cooktheryan:
Because this is only around CI I would like to self merge the PRs surrounding this

PR #48, opened 2024-08-15
Implement `ramalama push` command for pushing models to remote reposi…

ericcurtin:
…tories

Added `push_cli` function to handle the `ramalama push` command. Updated `funcDict` to include the `push` command.
Ensured proper validation of local model existence before pushing. Added error handling for push failures with descriptive messages.

### DIFF
--- a/docs/source/markdown/ramalama-push.1.md
+++ b/docs/source/markdown/ramalama-push.1.md
@@ -1,0 +1,17 @@
+% ramalama-push 1
+
+## NAME
+ramalama - Push AI Model (OCI-only at present)
+
+## SYNOPSIS
+**ramalama push** [*options*] *model*
+
+## DESCRIPTION
+Push specified AI Model (OCI-only at present)
+
+## SEE ALSO
+**[ramalama(1)](ramalama.1.md)
+
+## HISTORY
+Aug 2024, Originally compiled by Eric Curtin <ecurtin@redhat.com>
+
--- a/docs/source/markdown/ramalama-push.1.md
+++ b/docs/source/markdown/ramalama-push.1.md
@@ -4,7 +4,7 @@
 ramalama - Push AI Model (OCI-only at present)
 
 ## SYNOPSIS
-**ramalama push** [*options*] *model*
+**ramalama push** [*options*] *model* *target*
 
 ## DESCRIPTION
 Push specified AI Model (OCI-only at present)
--- a/container-images/ramalama/latest/Containerfile
+++ b/container-images/ramalama/latest/Containerfile
@@ -12,7 +12,7 @@ RUN dnf install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.n
 
 RUN /usr/bin/python3 --version
 RUN pip install "huggingface_hub[cli]==0.24.2"
-RUN pip install "omlmd==0.1.2"
+RUN pip install "omlmd==0.1.4"
 
 ENV LLAMA_CCACHE=0
 
--- a/docs/source/markdown/ramalama-pull.1.md
+++ b/docs/source/markdown/ramalama-pull.1.md
@@ -1,13 +1,13 @@
 % ramalama-pull 1
 
 ## NAME
-ramalama - Pull and AI Models into local storage
+ramalama - Pull AI Models into local storage
 
 ## SYNOPSIS
-**ramalama pull** [*options*] *model* [*model*...]
+**ramalama pull** [*options*] *model*
 
 ## DESCRIPTION
-Pull specified AI Models into local storage
+Pull specified AI Model into local storage
 
 ## SEE ALSO
 **[ramalama(1)](ramalama.1.md)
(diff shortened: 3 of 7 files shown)


ericcurtin:
Completely untested, mostly boilerplate, if you could tell me how this is wrong @tarilabs I'd appreciate it, just had a spare 10 minutes to throw this together.

tarilabs:
I'm on limited connectivity on national holiday atm, I think from cli a metadata file (either yaml or json) is mandatory [(example)](https://github.com/containers/omlmd/blob/c5c2797ca0aa517a444f5ab93a89c387f2245352/e2e/test_cli.sh#L15) but can be an emptye one!

Or if instead you modify ramalama to allow use of the py wheels, in the omlmd py sdk the metadata parameters can be [optional](https://github.com/containers/omlmd/blob/c5c2797ca0aa517a444f5ab93a89c387f2245352/omlmd/helpers.py#L39) [(full example)](https://containers.github.io/omlmd/#push).

As soon as I get back to my laptop I shall make an escape flag for omlmd CLI to deliberately ignore metadata, at least for these initial iterations...

docs/source/markdown/ramalama-push.1.md
```diff
@@ -0,0 +1,17 @@
+% ramalama-push 1
+
+## NAME
+ramalama - Push AI Model
+
+## SYNOPSIS
+**ramalama push** [*options*] *model*
+
+## DESCRIPTION
+Push specified AI Model
```
rhatdan, line 10: If this is only supported for OCI, you should specify this in the man page.

ramalama
```diff
@@ -333,6 +333,48 @@ def pull_cli(ramalama_store, args, port):
...
+def push_oci(model, ramalama_store):
+    # Remove the prefix and extract target details
+    target = re.sub(r'^oci://', '', model)
+    registry, reference = target.split('/', 1)
+    registry, reference = ("docker.io", target) if "." not in registry else (
+        registry, reference)
+    reference_dir = reference.replace(":", "/")
+
+    # Validate the model exists locally
+    local_model_path = os.path.join(
```
ericcurtin, line 345: To do: we should be able to push any model to oci registries, including huggingface and ollama

ramalama
```diff
@@ -333,6 +333,48 @@ def pull_cli(ramalama_store, args, port):
 funcDict["pull"] = pull_cli
 
 
+def push_oci(model, ramalama_store):
+    # Remove the prefix and extract target details
+    target = re.sub(r'^oci://', '', model)
+    registry, reference = target.split('/', 1)
+    registry, reference = ("docker.io", target) if "." not in registry else (
```
rhatdan, line 340: We should not force docker.io.  Users need to fully specify on push.

tarilabs: We aren't forcing indeed,
this is more to comply with the fact if I indicate `my-artifact:v1` as a reference,
it is intended to be `docker.io/library/my-artifact:v1`.

I believe most user flow however are in the complete form, such as `quay.io/mmortari/my-artifact:v1` anyway.

If the reference convention is not desired, we can always make it display an error message and fail, for instance ~"you must supply OCI registry in your target reference".

wdyt?

rhatdan: I would prefer the latter.

ericcurtin:
@tarilabs I'd be happy to let you take on this PR in general, you know this better than me. We need to support a simple:

`ramalama push MODEL`

and that should push any model we have locally. We can have additional flags for more complex use cases also.

Feel free to open a new PR and we can close this one.

tarilabs:
@ericcurtin let me add the `--no-metadata` flag to omlmd, then happy to take a contribution at this 👍

ramalama
tarilabs: @rhatdan ack 👍

tarilabs: I cross checked. This is for consistency of Pull semantic for OCI reference, where if not specified (eg.: `mmortari/model:v1`) is intended as coming from docker.io.

tarilabs (approved):
thank you so much @ericcurtin et al., 

I've rebased due to conflict with default main branch,
and added as discussed.

with:
```
./ramalama list                                          
NAME                                                                MODIFIED        SIZE  
oci://quay.io/mmortari/gguf-py-example/v1/example.gguf              1 second ago    4.0K
```

now both:
```
./ramalama push oci://quay.io/mmortari/gguf-py-example/v1/example.gguf quay.io/mmortari/gguf-py-example:ramalama
```

and:
```
./ramalama push oci://quay.io/mmortari/gguf-py-example/v1/example.gguf oci://quay.io/mmortari/gguf-py-example:ramalama
```

works, demo:
Screenshot 2024-08-22 at 10 22 46

ramalama
rhatdan: That is the way Docker works but not Podman.  Podman allows you to set a priority on container registries.  We should not hard code a preference for storing at any container registry.  Especially not a fully free one.

tarilabs: TIL, and thank you for the explanation; will adapt to it ! 👍

tarilabs: 👉  290a52b536c1ee693b760551f7a8388ef0bd4b33

rhatdan:
LGTM

PR #49, opened 2024-08-16
Add whisper.cpp to Containerfile

ericcurtin:
We should be able to auto-detect and run/serve asr models

### DIFF
--- a/container-images/ramalama/latest/Containerfile
+++ b/container-images/ramalama/latest/Containerfile
@@ -27,3 +27,12 @@ RUN git clone -b ramalama https://github.com/ericcurtin/llama.cpp.git && \
     done; \
     cd / && \
     rm -rf llama.cpp
+
+RUN git clone https://github.com/ggerganov/whisper.cpp.git && \
+    cd whisper.cpp && \
+    git reset --hard 22fcd5fd110ba1ff592b4e23013d870831756259 && \
+    make -j $(nproc) && \
+    mv main /usr/bin/whisper-main && \
+    mv server /usr/bin/whisper-server && \
+    cd / && \
+    rm -rf whisper.cpp


ericcurtin:
@hemajv and @aakankshaduggal I copied this from one of your Containerfiles, hope it's correct

container-images/ramalama/latest/Containerfile
```diff
@@ -27,3 +27,11 @@ RUN git clone -b ramalama https://github.com/ericcurtin/llama.cpp.git && \
     done; \
     cd / && \
     rm -rf llama.cpp
+
+RUN git clone --depth 1 --branch v1.6.2 https://github.com/ggerganov/whisper.cpp.git && \
```
rhatdan, line 31: Probably should move this to a multi-phase build.  
But LGTM
I also not thrilled with the hard coded --branch since we probably want latest release.  Not sure if there is a way to differentiate between HEAD and Release.  Perhaps renovate could help here.

@lmilbaum WDYT?

ericcurtin: We can do a multi-phase build, but these two binaries were like ~3M add to the image in the end though, so it's not bulking up the build time or image size much.

I changed to commit 22fcd5fd110ba1ff592b4e23013d870831756259 because there hasn't been a release in a while. Not sure about just tracking the main branch, etc. Because @cooktheryan has this set up to build nightly and a breakage could easily slip into the container images overnight.

I'm not sure what out complete container story should be, maybe it could be:

latest
latest-cuda
latest-rocm
stable
stable-cuda
stable-rocm

? @cooktheryan  and @lmilbaum would know best.

rhatdan:
LGTM

container-images/ramalama/latest/Containerfile
p5: I have done some fancy Renovate things in the past, where you can set the branch (or commit sha) as an ARG, and Renovate will automatically version this.  I'll try and throw something similar in a PR in the next day or two.

It will look something like this:

```
// renovate: datasource=github-releases depName=ggerganov/whisper.cpp
ARG WHISPER_SHA="22fcd5fd110ba1ff592b4e23013d870831756259"
RUN git clone --branch ${WHISPER_SHA} ...
```
With Renovate detecting the version via regex and creating the PRs as it would with any others

PR #50, opened 2024-08-21
Convert install script to python3

ericcurtin:
python3 is a common interpreter between macOS and Linux, it's easier to maintain that trying to write a script that supports Linux and macOS shell.


ericcurtin:
We won't support macOS on completion of this PR, but we will be closer.

rhatdan:
LGTM

PR #52, opened 2024-08-21
Rename ramalama to ramalama.py

rhatdan:
Move ci.sh into test subdir.

This is beginning to break ramalama into a library.


ericcurtin:
We will still need a ramalama file, even if it's just:

```
#!/bin/sh

exec ramalama.py
```

Or I guess it could be just a symlink too, or another python script or whatever

rhatdan:
ramalama.py will still be installed as /usr/bin/ramalama

ericcurtin (approved):

PR #53, opened 2024-08-21
Convert to UBI9

ericcurtin:
We are hoping to base as many images as possible on UBI9, rather than
CentOS Stream 9.


ericcurtin:
Re-purposing this PR as a UBI9 PR, GPU passthough with podman-machine will come

As requested by @fatherlinux

rhatdan:
LGTM

PR #55, opened 2024-08-21
Change ramalama to be used as a library

rhatdan:


rhatdan:
This will require changes to install.py to make this work correctly.

rhatdan:
@ericcurtin This is ready for review.

rhatdan:
@tarilabs PTAL

rhatdan:
install.py is fixed.

install.py
```diff
@@ -60,7 +60,7 @@ def main():
 
     tmp_dir = tempfile.mkdtemp()
     try:
-        binfile=ramalama
```
ericcurtin, line 63: Luckily this one didn't make our stable branch... So it's not effecting anyone right now

ramalama.py
```diff
@@ -423,20 +29,24 @@ def main(args):
 
         if conman:
             home = os.path.expanduser('~')
-            conman_args = [conman, "run", "--rm", "-it", "--security-opt=label=disable", f"-v{ramalama_store}:/var/lib/ramalama", f"-v{home}:{home}", "-v/tmp:/tmp",
-                           f"-v{__file__}:{__file__}", "-e", "RAMALAMA_HOST", "-p", f"{host}:{port}", "quay.io/ramalama/ramalama:latest", __file__] + args
+            pwd = os.getcwd()
+            conman_args = [conman, "run", "--rm", "-it", "--security-opt=label=disable", f"-v{store}:/var/lib/ramalama", f"-v{home}:{home}", "-v/tmp:/tmp",
+                           f"-v{pwd}:{pwd}", "-e", "RAMALAMA_HOST", "-p", f"{host}:{port}", "quay.io/ramalama/ramalama:latest", __file__] + args
```
ericcurtin, line 34: {pwd}, will only work when executing like this:

./ramalama.py

when calling like this:

ramalama

pwd could be anywhere...

one way to avoid listing all the files individually could be to put all the python files somewhere like:

/usr/libexec/ramalama

and share that whole dir if it exists. I really don't know how imports find the directory to load though... And pwd is useful for development

install.py
ericcurtin: Temping to take this one in a separate PR to fix main...

ramalama.py
rhatdan: We could mount it at /usr/lib/python3.12/site-packages/ramalama.  If we were to install this for real, then I would have /usr/bin/ramalama and /usr/lib/python3.12/site-packages/ramalama

tarilabs:
out of curiosity, planning to having this eventually made into a poetry project or similar?

rhatdan:
I don't know what a poetry project is?

tarilabs:
> I don't know what a poetry project is?

a build system for python: https://python-poetry.org, we're using it in RHOAI/KF Model Registry and omlmd

ramalama.py
rhatdan: I have a new fix to make this work based on this location, but waiting for this to merge first.

rhatdan:
Sure, I think we should package this as RPMs for Fedora, and make it available for pypy to be downloaded. I believe @ericcurtin wants to make it installable from the git repo as well via a simple single command.

ericcurtin:
As long as it's installable in a single command (that's one of the goals of this project), I don't mind the packing or install approach used too much, we need to be installable on macOS, CentOS Stream/RHEL 9/10, Fedora, Ubuntu 22.04 also...

ericcurtin:
And we want to avoid a 10 page getting started guide...

if nvidia do this

if amd do this

etc.

rhatdan:
Understood.  

Might be best to change setup/install script to checkout git repository on initialization.

ericcurtin:
Funnily enough... macOS is probably our best working platform at present:

This:

  brew install llama.cpp

package makes it easy, it has all the Apple GPU support built in as it can just assume Apple. On Linux, we have to build llama.cpp per GPU as it's poorly moduarized.

PR #56, opened 2024-08-21
Change to upstream llama.cpp

ericcurtin:
Use current git commit from macOS brew, with this PR, we have native accelerated macOS support.


ericcurtin:
With this we should be able to run on macOS native with GPU acceleration once this makes the s (stable) branch

PR #57, opened 2024-08-22
Implement rm in ramalama

rhatdan:
We must delete the symlink and garbage collect any model that does not
have an associated symlink.

We do garbage collection of everything, because some old sha's etc. may
no longer have symlink references due to updates.

Co-Authored-By: Daniel J Walsh <dwalsh@redhat.com>
Signed-off-by: Eric Curtin <ecurtin@redhat.com>

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -54,9 +54,9 @@ def init_cli():
     logout_parser(subparsers)
     pull_parser(subparsers)
     push_parser(subparsers)
+    rm_parser(subparsers)
     run_parser(subparsers)
     serve_parser(subparsers)
-    version_parser(subparsers)
     # Parse CLI
     args = parser.parse_args()
     if args.version:
@@ -184,11 +184,8 @@ def list_parser(subparsers):
 
 
 def list_cli(args):
-    if not args.noheading and not args.json:
-        print(f"{'NAME':<67} {'MODIFIED':<15} {'SIZE':<6}")
     mycwd = os.getcwd()
     os.chdir(f"{args.store}/models/")
-
     models = []
     # Collect model data
     for path in list_files_by_modification():
@@ -208,11 +205,12 @@ def list_cli(args):
     if args.json:
         print(json.dumps(models))
         return
+
     # Calculate maximum width for each column
     name_width = max(len("NAME"), max(len(model["name"]) for model in models))
     modified_width = max(len("MODIFIED"), max(len(model["modified"]) for model in models))
     size_width = max(len("SIZE"), max(len(model["size"]) for model in models))
-    if not args.noheading:
+    if not args.noheading and not args.json:
         print(f"{'NAME':<{name_width}} {'MODIFIED':<{modified_width}} " f"{'SIZE':<{size_width}}")
 
     for model in models:
@@ -231,7 +229,7 @@ def help_cli(args):
 
 
 def pull_parser(subparsers):
-    parser = subparsers.add_parser("pull", help="Pull AI model from model registry to local storage")
+    parser = subparsers.add_parser("pull", help="Pull AI Model from model registry to local storage")
     parser.add_argument("MODEL")  # positional argument
     parser.set_defaults(func=pull_cli)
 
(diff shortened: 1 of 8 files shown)


rhatdan:
@ericcurtin this is currently broken because of the way we are storing ollama models.  Why aren't we storing like we do huggingface models?

ericcurtin:
@rhatdan

for Ollama support, we wrote our own client basically, using various curl commands (http requests).

for huggingface we use the official huggingface client:

huggingface_hub[cli]==0.24.2

for oci we use the omlmd client:

omlmd==0.1.4

These clients store the .gguf's in their respective directories here (but each client does it differently):

~/.local/share/ramalama/repos/

This dir stores the models in a more structured way (as symlinks):

~/.local/share/ramalama/models/

this is all open to change... The clients for each protocol are fundamentally different right now...

ericcurtin:
But I see what you mean...

Maybe we should store like huggingface...

That sha256: name is really useful though (that's the filename in Ollama), you can validate if the file is fully downloaded and correct.

ericcurtin:
omlmd could be completely different again, haven't tried it yet.

ericcurtin:
Huggingface client seems to create directories like this with some metadata:

.huggingface
.cache/huggingface

so maybe they don't need the sha256:

ericcurtin:
> huggingface-cli delete-cache

can also delete GB's of data

ericcurtin:
I completely forgot a super important reason for the sha256: filename being important... De-duplication, you will never download duplicate files this way as you will notice the checksum exists already

rhatdan:
@ericcurtin Since you understand how this is supposed to be, do you want to take over this PR?

ericcurtin:
Sure I'll get to it tomorrow

ericcurtin:
I will get to this Monday @rhatdan haven't forgotten...

If it unblocks you we could just rm the symlink for now and do a follow on PR that actually rm's the real file when appropriate

ericcurtin:
I rebased this and fixed some conflicts

ericcurtin:
I have no idea right now why this is failing CI... Will try and figure out tomorrow, seems fine locally

rhatdan:
The mac_steps and linux_steps part of the ci.sh are screwing up the system, those should not be run within tests.

https://paste.centos.org/view/af6ab85c

Here are the other changes required. ^^

ramalama/cli.py
```diff
@@ -256,6 +257,17 @@ def version_parser(subparsers):
     parser.set_defaults(func=version_cli)
 
 
+def rm_parser(subparsers):
+    parser = subparsers.add_parser("rm", help="Remove AI model from local storage")
+    parser.add_argument("model", help="Name of the model to remove")
```
rhatdan, line 262: Capitalize model

ramalama/cli.py
```diff
@@ -256,6 +257,17 @@ def version_parser(subparsers):
...
+    parser.set_defaults(func=rm_cli)
+
+
+def rm_cli(args):
+    model = New(args.model)
```
rhatdan, line 267: Capitalize model

test/ci.sh
```diff
@@ -40,6 +40,12 @@ tests() {
   ./${binfile} ls | grep tiny-vicuna-1b
   ./${binfile} ls | grep NAME
   ./${binfile} ls | grep oci://quay.io/mmortari/gguf-py-example/v1/example.gguf
+  ./${binfile} rm ollama://ben1t0/tiny-llm:latest
```
rhatdan, line 43: Should add this as a test to bats.

test/ci.sh
```diff
@@ -72,13 +78,13 @@ main() {
     linux_steps
   fi
 
+  $maybe_sudo rm -rf /usr/share/ramalama /opt/homebrew/share/ramalama /usr/local/share/ramalama
```
rhatdan, line 81: The problem with this is it modifies the host system when users run `make test`
 If we want to test this, we should just do it in the ci/cd system not as part of the `make test` job.

ericcurtin: We have to at least run:

mac_steps -> ./install.py

or some equivalent... To install the python dependancies for macOS...

ramalama/cli.py
ericcurtin: There's a lot of similar references, do I capitalise everywhere? :

```
ramalama/cli.py
80:    model = New(str(transport))
97:    model = New(str(transport))
210:    model = New(args.model)
226:    model = New(args.model)
239:    model = New(args.model)
251:    model = New(args.model)
267:    model = New(args.model)

ramalama/model.py
9:    model = ""
15:        self.model = model
```

ericcurtin: I think I see what you mean now, I'll try to be consistent with your recent PR

rhatdan:
Ok lets get this merged, and then I can start turning on tests and probably add some more options.

rhatdan:
LGTM

ramalama/cli.py
ericcurtin: Done

ramalama/cli.py
ericcurtin: Done

ericcurtin:
I'm fighting with builds at the moment, it might be green this one

PR #59, opened 2024-08-22
Get version of python for installed library

rhatdan:


ericcurtin:
I did another version of this in:

https://github.com/containers/ramalama/pull/60

I don't mind which one we go with

ericcurtin:
It iterates through sys.path until it finds the ramalama directory

ericcurtin:
There isn't really a way of determining guest container python version, which is tricky, I think ubi9 might be 3.9, but I don't know

ericcurtin:
Well you could figure it out by doing a quick "podman run --rm" but that starts to get messy and each podman run degrades start time

ericcurtin:
Each podman run is about 150ms from my timings... Maybe that's not too bad... GPU initialisation can take a seconds sometimes anyway

ericcurtin:
We could end up having a mix of Fedora/UBI/Stream containers also for some GPUs... But only if we really need to

ramalama/common.py
```diff
@@ -9,8 +9,15 @@
...
+def in_container():
+    if os.path.exists("/run/.containerenv") or os.path.exists("/.dockerenv") or os.getenv("container"):
+        return True
+
+    return False
+
+
 def container_manager():
-    if sys.platform == "darwin":
+    if in_container() or sys.platform == "darwin":
```
ericcurtin, line 20: Good idea, avoiding nested containers

ramalama.py
```diff
@@ -32,22 +32,26 @@ def main(args):
...
             conman_args = [conman, "run",
                            "--rm",
                            "-it",
                            "--security-opt=label=disable",
                            f"-v{store}:/var/lib/ramalama",
                            f"-v{home}:{home}",
                            "-v/tmp:/tmp",
-                           f"-v{sys.argv[0]}:/usr/bin/ramalama",
-                           f"-v{wd}:{libpath}",
+                           f"-v{__file__}:/usr/bin/ramalama:ro",
```
ericcurtin, line 53: Also a good idea to add ro

ramalama.py
```diff
@@ -32,22 +32,26 @@ def main(args):
         if host != port:
             port = host.rsplit(':', 1)[1]
 
+        syspath = '/usr/share/ramalama/python'
```
ericcurtin, line 35: Clever :)

I hope this works and we can exec this before we search for:

```
import errno
import os
from pathlib import Path
import subprocess
import ramalama
import sys
```

rhatdan:
Ok lets go with this one.

ramalama.py
ericcurtin: We could always write a wrapper python script that did the equivalent of:

```
#!/bin/sh

export PYTHONPATH="/usr/share/ramalama/python:${PYTHONPATH}"
exec ramalama
```

that has basically no imports

PR #60, opened 2024-08-22
Use __file__ over sys.argv[0]

ericcurtin:
`__file__` uses absolute path, it's safer, if we call ramalama like:

ramalama

argv[0] will be "ramalama" which is not the location of the python script.

Also fixed a typo s/podman/ramalama/g

### DIFF
--- a/ramalama.py
+++ b/ramalama.py
@@ -31,10 +31,17 @@ def main(args):
 
         if conman:
             home = os.path.expanduser('~')
-            cwd = os.getcwd()
-            wd = os.path.join(cwd, "ramalama")
-            if not os.path.exists(wd):
-                wd = "/usr/lib/python3.12/site-packages/podman"
+            wd = "ramalama"
+            for path in sys.path:
+                path = os.path.join(path, 'ramalama')
+                # Check if the "ramalama" directory exists in the current path
+                if os.path.exists(path):
+                    wd = path
+                    break
+                else:
+                    print("ramalama not found in any sys.path directories")
+
+            # To do, unsafe, this may not be a valid path in the guest
             libpath = "/usr/lib/python3.12/site-packages/ramalama"
             conman_args = [conman, "run",
                            "--rm",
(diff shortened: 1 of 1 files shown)


ericcurtin:
sys.path is this FWIW:

['', '/usr/lib64/python312.zip', '/usr/lib64/python3.12', '/usr/lib64/python3.12/lib-dynload', '/usr/local/lib64/python3.12/site-packages', '/usr/local/lib/python3.12/site-packages', '/usr/lib64/python3.12/site-packages', '/usr/lib/python3.12/site-packages']

ramalama.py
```diff
@@ -31,10 +31,17 @@ def main(args):
...
-                wd = "/usr/lib/python3.12/site-packages/podman"
+            wd = "ramalama"
+            for path in sys.path:
+                path = os.path.join(path, 'ramalama')
+                # Check if the "ramalama" directory exists in the current path
+                if os.path.exists(path):
+                    wd = path
+                    break
+                else:
+                    print("ramalama not found in any sys.path directories")
```
rhatdan, line 42: This is looking for the library not the executable.  We should only look in the current directory for developers or the system wide path.

ericcurtin: This list is:

['', '/usr/lib64/python312.zip', '/usr/lib64/python3.12', '/usr/lib64/python3.12/lib-dynload', '/usr/local/lib64/python3.12/site-packages', '/usr/local/lib/python3.12/site-packages', '/usr/lib64/python3.12/site-packages', '/usr/lib/python3.12/site-packages']

so it will first check for:

```
ramalama (this is the git repo case, it will stop here)
/usr/lib64/python312.zip/ramalama
/usr/lib64/python3.12/ramalama
/usr/lib64/python3.12/lib-dynload/ramalama
/usr/local/lib64/python3.12/site-packages/ramalama
/usr/local/lib/python3.12/site-packages/ramalama
/usr/lib64/python3.12/site-packages/ramalama
/usr/lib/python3.12/site-packages/ramalama (this is where it will stop in the installed case, because this will exists)
```

This mimics how python3 actually searches for libraries/imports, if we were running outside a container...

rhatdan:
I got this via my PR.

PR #61, opened 2024-08-22
When calling ramalama with no args, display usage

ericcurtin:
Previously it was displaying this:

$ ./ramalama.py
pop from empty list


rhatdan:
LGTM

PR #68, opened 2024-08-22
GPU enablement and bug fixes

ericcurtin:
Expose /dev/dri and /dev/kfd if they are present.

--instruct option was removed from llama.cpp so we need to do:

"-cnv", "-p", "You are a helpful assistant"

now. -cnv is conversation mode.

Also fixed a bug where return was missing.

### DIFF
--- a/ramalama.py
+++ b/ramalama.py
@@ -43,6 +43,7 @@ def main(args):
                 if os.path.exists(target):
                     wd = target
                     break
+
             conman_args = [conman, "run",
                            "--rm",
                            "-it",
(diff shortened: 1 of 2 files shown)


ericcurtin:
I chose "You are a helpful assistant" as the system prompt as that's what huggingface UI recommends, but it could be anything...

ramalama.py
```diff
@@ -51,10 +52,17 @@ def main(args):
...
-                           "-p", f"{host}:{port}",
-                           "quay.io/ramalama/ramalama:latest", __file__] + args
+                           f"-v{wd}:{syspath}:ro"]
+            if os.path.exists("/dev/dri"):
+                conman_args += ["--device", "/dev/dri"]
+
+            if os.path.exists("/dev/kfd"):
+                conman_args += ["--device", "/dev/kfd"]
+
+            conman_args += ["-e", "RAMALAMA_HOST",
```
rhatdan, line 62: This would be cleaner if  you moved these options to original conman_args

ramalama.py
```diff
@@ -51,10 +52,17 @@ def main(args):
...
+                            "-p", f"{host}:{port}",
+                            "quay.io/ramalama/ramalama:latest", __file__]
```
rhatdan, line 64: This __file__ should be "/usr/bin/ramalama" i believe.

ericcurtin: Or if you run from the git repo, it ends up being something like:

/Users/ecurtin/git/ramalama/./ramalama.py

it covers both the installed case and the development case when using `__file__`...

ramalama.py
ericcurtin: Ok... I do need to inject the --device kargs in the middle somewhere though, I can do it just before quay.io/ramalama/ramalama:latest maybe

rhatdan:
Oops accidantly merged.

PR #79, opened 2024-08-23
Translate into class structure 

rhatdan:
Also changes default TRANSPORT to OCI rather then ollama

### DIFF
--- a/ramalama.py
+++ b/ramalama.py
@@ -57,6 +57,9 @@ def main(args):
                            f"-v{home}:{home}",
                            "-v/tmp:/tmp",
                            f"-v{__file__}:/usr/bin/ramalama:ro",
+                           "-e", "RAMALAMA_HOST",
+                           "-e", "RAMALAMA_TRANSPORT",
+                           "-p", f"{host}:{port}",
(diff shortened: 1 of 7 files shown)


ericcurtin (approved):

ericcurtin:
Will will have to rewrite the README.md also, practically all the examples assume ollama as default

ericcurtin:
We probably need to implement shortnames soon, so we can do OCI equivalents of:

`ramalama run granite-code`

ericcurtin:
I wouldn't mind doing fallbacks either... If oci doesn't work, try ollama, if ollama doesn't work try huggingface... Ollama registry has a decent database of shortname convenient to use models

rhatdan:
We can go back to ollama by default, but since we are trying to get people to OCI, that seems counterproductive.  Bottom line it is a one line change.

The next thing I want to work on is getting bats testing, and adding tests for OCI Registries.

We need to improve CI/CD systems.

We also go to get ollama and huggingface push working.

ericcurtin:
Yeah I agree, there's a few bugs I've found yesterday, more tests would help

Issue #73, opened 2024-08-22
Confusion after running serve


cdrage:
After running serve:

```
ramalama serve llama3
########################################################################################################################## 100.0%
INFO [                    main] build info | tid="0x1f51d4f40" timestamp=1724351882 build=3614 commit="a1631e53"
INFO [                    main] system info | tid="0x1f51d4f40" timestamp=1724351882 n_threads=8 n_threads_batch=-1 total_threads=10 system_info="AVX = 0 | AVX_VNNI = 0 | AVX2 = 0 | AVX512 = 0 | AVX512_VBMI = 0 | AVX512_VNNI = 0 | AVX512_BF16 = 0 | FMA = 0 | NEON = 1 | SVE = 0 | ARM_FMA = 1 | F16C = 0 | FP16_VA = 1 | WASM_SIMD = 0 | BLAS = 1 | SSE3 = 0 | SSSE3 = 0 | VSX = 0 | MATMUL_INT8 = 0 | LLAMAFILE = 1 | "
INFO [                    main] HTTP server is listening | tid="0x1f51d4f40" timestamp=1724351882 port="8080" n_threads_http="9" hostname="127.0.0.1"
INFO [                    main] loading model | tid="0x1f51d4f40" timestamp=1724351882 port="8080" n_threads_http="9" hostname="127.0.0.1"
llama_model_loader: loaded meta data with 22 key-value pairs and 291 tensors from /Users/cdrage/.local/share/ramalama/models/ollama/llama3:latest (version GGUF V3 (latest))
llama_model_loader: Dumping metadata keys/values. Note: KV overrides do not apply in this output.
llama_model_loader: - kv   0:                       general.architecture str              = llama
llama_model_loader: - kv   1:                               general.name str              = Meta-Llama-3-8B-Instruct
llama_model_loader: - kv   2:                          llama.block_count u32              = 32
llama_model_loader: - kv   3:                       llama.context_length u32              = 8192
llama_model_loader: - kv   4:                     llama.embedding_length u32              = 4096
llama_model_loader: - kv   5:                  llama.feed_forward_length u32              = 14336
llama_model_loader: - kv   6:                 llama.attention.head_count u32              = 32
llama_model_loader: - kv   7:              llama.attention.head_count_kv u32              = 8
llama_model_loader: - kv   8:                       llama.rope.freq_base f32              = 500000.000000
llama_model_loader: - kv   9:     llama.attention.layer_norm_rms_epsilon f32              = 0.000010
[...]

PR #87, opened 2024-08-24
Put Running Models at the top of README.md

ericcurtin:
This behaves like a hello world example for ramalama, so it makes sense to have that after the installtion steps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,28 +36,6 @@ support for Apple Silicon, AMD, Nvidia, etc.
 
 ## Usage
 
-### Listing Models
-
-You can `list` all models pulled into local storage.
-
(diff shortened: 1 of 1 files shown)


ericcurtin:
This is to avoid confusion like:

https://github.com/containers/ramalama/issues/73

One difference between ramalama and most other inferencing tools is it does not require a client and server to be set up for simple local usage.

So if one wants to just do a quick daemonless prompt as a test, etc. "ramalama run" is the place to start

ericcurtin:
Run autodownloads models, etc. like podman

rhatdan:
LGTM

PR #88, opened 2024-08-26
Add system tests bats suite from podman

rhatdan:
This is just a quick hack.  Would love to get better testing in ramalama.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
-__pycache__
+ramalama/__pycache__
 docs/*.1
 build
 *.egg-info
 *.patch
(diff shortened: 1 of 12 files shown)


ericcurtin:
`brew install coreutils` installs timeout on macOS... Can be added to ci.sh most likely...

I think it's important we get this test suite running on macOS if we do want to continue with native macOS support.

rhatdan:
@ericcurtin lets get this in, and start requiring any changes to ramalama other then doc changes to require the addition of a test for the change.

ericcurtin (approved):

PR #90, opened 2024-08-26
Add --help, --json, --noheading support for ramalama list

rhatdan:
Also move to argparse

### DIFF
--- a/ramalama.py
+++ b/ramalama.py
@@ -1,14 +1,18 @@
 #!/usr/bin/python3
 
-import errno
-import os
-from pathlib import Path
 import subprocess
-import ramalama
+from pathlib import Path
+import os
+import errno
 import sys
 
 
 def main(args):
+    syspath = '/usr/share/ramalama'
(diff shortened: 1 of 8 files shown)


ramalama.py
```diff
@@ -29,23 +33,22 @@ def main(args):
...
         if host != port:
             port = host.rsplit(':', 1)[1]
 
-        syspath = '/usr/share/ramalama/python'
-        sys.path.insert(0, syspath)
-
         if conman:
             home = os.path.expanduser('~')
-            wd = "ramalama"
+            wd = "./ramalama"
```
ericcurtin, line 51: I think there's a bug on line 53, I would expect wd in that case to become:

/usr/lib/python3.12/site-packagesramalama

line 53 should be:

target = os.path.join(p, "ramalama")

ericcurtin: probably won't need this change then

ericcurtin (approved):
LGTM, it's failing the build because of partially implemented features

PR #92, opened 2024-08-27
Complete port to argparser

rhatdan:


docs/ramalama.1.md
```diff
@@ -7,9 +7,17 @@ ramalama - Simple management tool for working with AI Models
...
 ## DESCRIPTION
-Ramalama : The goal of ramalama is to make AI boring.
+Ramalama : The goal of ramalama is to make AI boring. Ramalama can pull an AI
+Model from model registires and start a chatbot or serve as a rest API from a
+simple single command. It treats Models similar to the way that Podman or
+Docker treat container images.
 
-Ramalama supports multiple types of model registries. Currently the following types of AI Model registries (transports):
+Ramalama runs models with a specially designed container image containing all
+of the tooling required to run the Model. Users d ont need to pre-configure
```
ericcurtin, line 16: s/d ont/don't/g

I see REST API spelt with capitals more commonly but no big deal.

ramalama/common.py
```diff
@@ -17,12 +17,6 @@ def in_container():
 
 
 def container_manager():
```
ericcurtin, line 19: In some ways it's a pity this will attempt to use podman on macOS by default...

macOS native will probably outperform Linux guest VM from using native APIs (and just macOS being very mature here from lots of usage, llama.cpp for example was first created to run on macOS and Apple Silicon) and will have access to more memory... And it's makes ramalama more complex to start on macOS... One has to start/create a podman-machine with enough memory, disk, etc. Could potentially kill the "boring" aspect on macOS.

I do think we should leave this in until we are at least sure we work with ramalama via podman-machine on macOS out of the box accelerate, we aren't far away from that, but I don't think we've tested that end to end.

I did get good feedback from everyone that used macOS native ramalama so far.

Strong chance this is why the build is failing on macOS but didn't check thoroughly.

ericcurtin: `ramalama currently only works on linux and macOS`

should be the below string...

PR #97, opened 2024-08-29
Add basic vllm support

ericcurtin:
This will only work with --nocontainer for now. Requires vllm > 6.0 .

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -39,6 +39,12 @@ def init_cli():
     parser.add_argument("--store", default=get_store(), help="store AI Models in the specified directory")
     parser.add_argument("--dryrun", action="store_true", help="show container runtime command without executing it")
     parser.add_argument(
+        "--runtime",
+        default="llama.cpp",
+        choices=["llama.cpp", "vllm"],
+        help="specify the runtime to use, valid options are 'llama.cpp' and 'vllm'",
+    )
+    parser.add_argument(
         "--nocontainer",
         default=not use_container(),
         action="store_true",
@@ -316,7 +322,7 @@ def run_parser(subparsers):
     parser.add_argument("--prompt", dest="prompt", action="store_true", help="modify chatbot prompt")
     parser.add_argument("-n", "--name", dest="name", help="name of container in which the Model will be run")
     parser.add_argument("MODEL")  # positional argument
-    parser.add_argument("ARGS", nargs="*", help="Additional options to pass to the AI Model")
+    parser.add_argument("ARGS", nargs="*", help="additional options to pass to the AI Model")
     parser.set_defaults(func=run_cli)
 
 
--- a/docs/ramalama.1.md
+++ b/docs/ramalama.1.md
@@ -77,6 +77,9 @@ show this help message and exit
 do not run ramalama in the default container (default: False)
 use environment variable "RAMALAMA_IN_CONTAINER=false" to change default.
 
+#### **--runtime**
+specify the runtime to use, valid options are 'llama.cpp' and 'vllm' (default: llama.cpp)
+
 #### **--store**=STORE
 store AI Models in the specified directory (default rootless: `$HOME/.local/share/ramalama`, default rootful: `/var/lib/ramalama`)
 
--- a/ramalama/model.py
+++ b/ramalama/model.py
@@ -102,5 +102,8 @@ class Model:
 
     def serve(self, args):
         symlink_path = self.pull(args)
-        exec_args = ["llama-server", "--port", args.port, "-m", symlink_path] + self.common_params
+        exec_args = ["llama-server", "--port", args.port, "-m", symlink_path]
+        if args.runtime == "vllm":
+            exec_args = ["vllm", "serve", "--port", args.port, symlink_path]
+
         exec_cmd(exec_args)


ericcurtin:
I think we make a breaking change here, "vllm serve" only serves .gguf file correctly with .gguf extension, so we might be better off storing files like:

~/.local/share/ramalama/models/ollama/granite-code:latest

like this instead:

~/.local/share/ramalama/models/ollama/granite-code:latest.gguf

rhatdan:
Can we play games with symbolic links?  I am not that concerned about breaking changes at this point, since we have not released anything yet.

ericcurtin:
Above are symlinks, so the cleanest way is to make a breaking change and change this symlink:

~/.local/share/ramalama/models/ollama/granite-code:latest

to this one:

~/.local/share/ramalama/models/ollama/granite-code:latest.gguf

Also logged a bug with vllm:

https://github.com/vllm-project/vllm/issues/7993

It's supposed to be possible to do this without the extension by using flags like "--quantization gguf --load-format gguf" but it doesn't work. GGUF inference is like 3 weeks old in vllm, they were using other formats before.

ericcurtin:
ramalama would solve this problem for many:

https://github.com/vllm-project/vllm/issues/1441

This is one of the most commented issues in vllm... vllm focuses more on enterprise grade GPUs, so it frustrates people they don't support Apple Devices...

Whereas with ramalama, one could test locally via:

ramalama --runtime llama.cpp serve, etc.

and if they want to use vllm on a deployment server, they could do:

ramalama --runtime vllm serve, etc.

potentially....

ericcurtin:
Some of these things are bug fixes I'm gonna split those out into another PR

kannon92:
Excited for this! https://github.com/ollama/ollama/issues/3953

There was some interest in that issue in the feature so we should make sure to highlight it as a feature.

ericcurtin:
vllm support will probably be completed, once there is a new release of vllm, to get some PRs like:

https://github.com/vllm-project/vllm/pull/8056

vllm release pretty frequently though, so it shouldn't be too far away...

matbee-eth:
> I think we may need to make a breaking change here, "vllm serve" only serves .gguf file correctly with .gguf extension, so we might be better off storing files like:
> 
> 
> 
> ~/.local/share/ramalama/models/ollama/granite-code:latest
> 
> 
> 
> like this instead:
> 
> 
> 
> ~/.local/share/ramalama/models/ollama/granite-code:latest.gguf
> 
> 

I'm not sure how feasible it is for this project, but a FuSE filesystem for dynamically managing models instead of symlinks could be an option

ericcurtin:
> > I think we may need to make a breaking change here, "vllm serve" only serves .gguf file correctly with .gguf extension, so we might be better off storing files like:
> > ~/.local/share/ramalama/models/ollama/granite-code:latest
> > like this instead:
> > ~/.local/share/ramalama/models/ollama/granite-code:latest.gguf
> 
> I'm not sure how feasible it is for this project, but a FuSE filesystem for dynamically managing models instead of symlinks could be an option

Not sure we wanna go here right now. For one, we support macOS native. This problem in particular will go away in the next version of vllm when the above PR gets merged.

But... I'd still be interested in hearing a more detailed example/explanation of why FuSE filesystem would help us here...

Although it would be preferably to have a fs-agnostic solution if at all possible...

ericcurtin:
This is very basic support for the --nocontainer case, but may unblock other efforts like:

#150

ramalama/cli.py
```diff
@@ -38,6 +38,12 @@ def init_cli():
     )
     parser.add_argument("--store", default=get_store(), help="store AI Models in the specified directory")
     parser.add_argument("--dryrun", action="store_true", help="show container runtime command without executing it")
+    parser.add_argument(
+        "--runtime",
+        default="llama.cpp",
+        choices=["llama.cpp", "vllm"],
+        help="Specify the runtime to use, valid options are 'llama.cpp' and 'vllm'",
```
rhatdan, line 45: We are defaulting to lowercase help messages now.

rhatdan: Could you also fix:

    parser.add_argument("ARGS", nargs="*", help="Additional options to pass to the AI Model")


While you are in there?

ericcurtin: Done

rhatdan:
LGTM

ericcurtin:
@kannon92 basic support for this was merged

cdoern:
@ericcurtin from an ilab side of things I have been trying to see if we can get operations like generate on a mac using vllm. Could this help with that? it seems like the discourse here is saying that vllm works on the mac using ramalama?

ericcurtin:
What this PR enables is the option to do:

```
$ ramalama --runtime vllm serve some-model
```

or

```
$ ramalama --runtime llama.cpp serve some-model
```

llama.cpp and vllm have different primary design goals. llama.cpp was written primarily with commodity grade systems in mind like a macbook, mac mini, etc. vllm was written with almost the opposite priority, to run on huge enterprise-grade systems with lots of vram and powerful gpus.

So with this change, one could test on macOS with llama.cpp as a runtime but deploy to some enterprise server with vllm as a runtime.

PR #98, opened 2024-08-29
Bug fixes and aliases

ericcurtin:
It should be self.pull(args).

Also added 'ls' alias for 'list'.

### DIFF
--- a/hack/xref-helpmsgs-manpages
+++ b/hack/xref-helpmsgs-manpages
@@ -178,7 +178,7 @@ sub xref_by_help {
 
   OPTION:
     for my $k (sort keys %$help) {
-        next if $k =~ /^_/;             # metadata ("_desc"). Ignore.
+        next if $k =~ /^_/ || $k eq "ls"; # metadata ("_desc"). Ignore.
 
         if (! ref($man)) {
             # Super-unlikely but I've seen it
--- a/ramalama/oci.py
+++ b/ramalama/oci.py
@@ -20,7 +20,7 @@ class OCI(Model):
             conman_args.extend(["--password", args.password])
         if args.passwordstdin:
             conman_args.append("--password-stdin")
-        conman_args.append(self.transport)
+        conman_args.append(args.transport)
         return exec_cmd(conman_args)
 
     def logout(self, args):
--- a/docs/ramalama-list.1.md
+++ b/docs/ramalama-list.1.md
@@ -6,6 +6,8 @@ ramalama\-list - List all AI models in local storage
 ## SYNOPSIS
 **ramalama list** [*options*]
 
+**ramalama ls** [*options*]
+
 ## DESCRIPTION
 List all the AI Models in local storage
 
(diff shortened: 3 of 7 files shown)


ericcurtin:
The ls alias is because ollama has the "rollama ls" alias for "ollama list", it's also kinda ingrained into a lot of people's muscle memory to type ls.

docs/ramalama-ls.1.md
```diff
@@ -0,0 +1,45 @@
+% ramalama-ls 1
```
rhatdan, line 1: We should just use man page aliasing.

cat podman/docs/source/markdown/links/podman-container-ls.1 
.so man1/podman-ps.1

cat ramalama-ls.1
.so man1/ramalama-ls.1

rhatdan: Than add ramalama-ls to ramalama-list.md man page info

rhatdan: Let's not duplicate the number of man pages, but link them together.

Everything else in this PR is great.

hack/man-page-checker
```diff
@@ -19,7 +19,11 @@ cd $(dirname $0)/../docs/ || die "Please run me from top-level libpod dir"
 
 rc=0
 
-for md in *.1.md;do
+for md in *.1.md; do
```
ericcurtin, line 22: The podman version has a more complex way of skipping aliases, it patches a perl script, etc. Would rather use this way for now

docs/ramalama-ls.1.md
ericcurtin: Done

hack/man-page-checker
```diff
@@ -19,7 +19,11 @@ cd $(dirname $0)/../docs/ || die "Please run me from top-level libpod dir"
...
+    if [ "$md" = "ramalama-ls.1.md" ]; then
```
rhatdan, line 23: If the link is not an md file then this is not necessary.

docs/ramalama-ls.1.md
```diff
@@ -0,0 +1 @@
+.so man1/ramalama-list.1
```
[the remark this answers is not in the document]
ericcurtin: Yup that works, thanks for spotting, re-pushed with changes.

hack/man-page-checker
ericcurtin: Yup that works, thanks for spotting, re-pushed with changes.

hack/xref-helpmsgs-manpages
```diff
@@ -178,7 +178,7 @@ sub xref_by_help {
 
   OPTION:
     for my $k (sort keys %$help) {
-        next if $k =~ /^_/;             # metadata ("_desc"). Ignore.
```
ericcurtin, line 181: We still need this skip now, but only this one.

ramalama/oci.py
```diff
@@ -20,7 +20,7 @@ def login(self, args):
             conman_args.extend(["--password", args.password])
         if args.passwordstdin:
             conman_args.append("--password-stdin")
-        conman_args.append(self.transport)
+        conman_args.append(args.transport)
```
ericcurtin, line 23: Added another small bugfix here

rhatdan:
LGTM

PR #99, opened 2024-09-01
Resolve undefined errors in `oci.py`

swarajpande5:
Resolves some undefined errors I found while exploring the codebase.


ericcurtin (approved):
LGTM

ericcurtin:
@swarajpande5 writing a test for pulling and oci image model would be useful to make sure the oci image path doesn't break long-term

PR #100, opened 2024-09-02
Make it possible to install ramalama from any branch

ericcurtin:
ramalama installer is hardcoded to run from 's' branch (the stable one), with this change, we can do things like:

```
export BRANCH=installer-select-branch
curl -fsSL https://raw.githubusercontent.com/containers/ramalama/$BRANCH/install.py | sudo BRANCH=$BRANCH python3
```

to install from the "main" branch alternatively, or any other branch on github.

Also added "-f" option to curl, so we fail on non-200 response codes like 404

### DIFF
--- a/install.py
+++ b/install.py
@@ -36,7 +36,7 @@ def amd_lshw():
 
 def download(url, to):
     curl_cmd = [
-        "curl", "--globoff", "--location", "--proto-default", "https",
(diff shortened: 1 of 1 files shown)


ericcurtin:
Gonna merge this one, makes it easier to test #101

PR #101, opened 2024-09-03
Fix installer to account for all files

ericcurtin:
The python script was split into multiple files, account for this in the installer.

### DIFF
--- a/install.py
+++ b/install.py
@@ -42,47 +42,77 @@ def download(url, to):
     subprocess.run(curl_cmd, check=True)
 
 
-def main():
+def check_platform():
     if sys.platform == 'darwin':
         if os.geteuid() == 0:
             print("This script is intended to run as non-root on macOS")
-            sys.exit(1)
+            return 1
         if not available("brew"):
             print("Please install Homebrew before continuing install on macOS")
-            sys.exit(2)
+            return 2
     elif sys.platform == 'linux':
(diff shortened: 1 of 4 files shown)


install.py
```diff
@@ -42,47 +42,79 @@ def download(url, to):
...
+
+def setup_ramalama(bindir, tmp_dir):
+    binfile = "ramalama"
+    from_file = binfile + ".py"
+    host = "https://raw.githubusercontent.com"
+    branch = os.getenv('BRANCH', 's')
+    url = f"{host}/containers/ramalama/{branch}/{from_file}"
+    to_file = os.path.join(tmp_dir, from_file)
+    download(url, to_file)
+    if sys.platform == 'darwin':  # macOS
```
rhatdan, line 80: How about:
```
ramalama_bin = os.path.join(bindir, binfile)
syspath = '/usr/share/ramalama'
if sys.platform == 'darwin':  # macOS
        install_mac_dependencies()
        sharedirs = ["/opt/homebrew/share",
                     "/usr/local/share"]
        syspath = next((d for d in sharedirs if os.path.exists(d)), None)
        syspath += "/ramalama"
```

rhatdan (approved):

install.py
ericcurtin: Yeah sure, one less comparison would be nice, will catch in the morning and merge

PR #104, opened 2024-09-04
Add codespell support

rhatdan:


ericcurtin:
If we merge #107 and rebase this PR and push, codespell might work then

Makefile
```diff
@@ -38,13 +38,19 @@ docs:
 	make -C docs
 
 .PHONY: test
-test:
```
ericcurtin, line 41: Are you sure you want pip install in the Makefile?

Like we only need to install that once

ericcurtin: test/ci.sh in the Makefile seems a little questionable too...

I'd expect one to be able to run "make test" locally without it attempting to install loads of things...

ericcurtin: But just to test this out, need to put the pip install's before codespell

ericcurtin: If it was me I'd do dependency installs in:

test/ci.sh

this could call:

make test

and:

make install (and this would only install ramalama python scripts, not dependancies)

rhatdan: I think we can make the requirement for codespel and autopep8 smarter in the Makefile.  I would like to get to the point where the CI/CD system just runs make test. Then running it locally or in the CI/CD system make sense.

What I don't understand now is why the install.py is blowing up.

ericcurtin (approved):

PR #105, opened 2024-09-04
Port logic is buggy, just support --port for now.

ericcurtin:
Just support --port for now to use port other than 8080. Also we also
want only want commands to claim a port that actually use a port, in
other words, right now only "serve" should claim a port like 8080.


ericcurtin:
This is a fix for a bug we saw in the demo today

rhatdan:
LGTM

PR #107, opened 2024-09-04
Add codespell and autopep via pip

ericcurtin:
We may want to hardcode to a version like:

pip install "codespell==some-version"
pip install "autopep8==some-version"

if builds start to become unstable by always using the latest version.

### DIFF
--- a/test/ci.sh
+++ b/test/ci.sh
@@ -9,16 +9,7 @@ mac_steps() {
 }
 
 linux_steps() {
-  if ! available autopep8; then
-    if available apt; then
-      $maybe_sudo apt install -y python3-autopep8
-    else
-      $maybe_sudo dnf install -y python3-autopep8
-    fi
-  fi
-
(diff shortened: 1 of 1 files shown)


test/ci.sh
```diff
@@ -9,14 +9,6 @@ mac_steps() {
 }
 
 linux_steps() {
-  if ! available autopep8; then
```
ericcurtin, line 12: Changing to pip, that way we can lock in on a common version for Fedora, RHEL, Ubuntu, macOS, etc.

These are only build tools anyway...

rhatdan: Not sure if this is going to be allowed on RHEL and Fedora builds, but fine.

ericcurtin: Yeah on real builds, like rpm packages, maybe won't don't need to run codespell or autopep8 at all, or we can use the packages....

PR #108, opened 2024-09-04
Fix installing of ramalama

rhatdan:


ericcurtin:
There are subtle differences between install locations on Linux, macOS x86 and macOS aarch64, just something to be aware of

test/ci.sh
```diff
@@ -73,6 +73,12 @@ main() {
     linux_steps
   fi
 
+  go install github.com/cpuguy83/go-md2man@latest
+  tmpdir=$(mktemp -d)
+  make install DESTDIR=${tmpdir} PREFIX=/usr
+  find ${tmpdir}
```
ericcurtin, line 79: This could probably be removed, but can catch in a follow-on PR:

```
find ${tmpdir}
```

PR #112, opened 2024-09-05
[WIP] Make ci install test work for forked branches

ericcurtin:
Was only working before for branches pushed to containers repository

### DIFF
--- a/test/ci.sh
+++ b/test/ci.sh
@@ -55,16 +55,13 @@ main() {
   os="$(uname -s)"
   binfile=ramalama.py
   chmod +x ${binfile} install.py
+  env
   uname -a
   /usr/bin/python3 --version
-
-  export BRANCH="main"
-  if false; then # This doesn't work for forked repos, will revisit
-    if [ -n "$GITHUB_HEAD_REF" ]; then
-      export BRANCH="$GITHUB_HEAD_REF"
-    elif [ -n "$GITHUB_REF_NAME" ]; then
-      export BRANCH="$GITHUB_REF_NAME"
-    fi
+  if [ -n "$GITHUB_HEAD_REF" ]; then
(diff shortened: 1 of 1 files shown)


test/ci.sh
```diff
@@ -55,16 +55,13 @@ main() {
   os="$(uname -s)"
   binfile=ramalama.py
   chmod +x ${binfile} install.py
+  env
```
rhatdan, line 58: What is this doing?  Is this a leftover?

test/ci.sh
```diff
@@ -55,16 +55,13 @@ main() {
...
   /usr/bin/python3 --version
-
-  export BRANCH="main"
-  if false; then # This doesn't work for forked repos, will revisit
-    if [ -n "$GITHUB_HEAD_REF" ]; then
-      export BRANCH="$GITHUB_HEAD_REF"
-    elif [ -n "$GITHUB_REF_NAME" ]; then
-      export BRANCH="$GITHUB_REF_NAME"
-    fi
+  if [ -n "$GITHUB_HEAD_REF" ]; then
```
rhatdan, line 61: Won't this break my PRs again?

test/ci.sh
ericcurtin: This is just testing the CI environment

test/ci.sh
ericcurtin: Yes it will, not intending to merge this as is...

Just needed to open a PR to kick off the CI, so I can read the CI log and fix properly.

test/ci.sh
ericcurtin: It prints all the environment variables, I want to see what they are, I want to use some

test/ci.sh
rhatdan: OK Just WIP the PR, and I won't pay attention.

PR #114, opened 2024-09-05
Set Ollama as the default

ericcurtin:
In future, we will likely set OCI as the default again (with Ollama as a fallback) so Ollama users won't notice a breakage.

OCI as a default today isn't very useful, we haven't built up a catalog of models as OCI artefacts yet.

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -349,4 +349,4 @@ def New(model):
     if transport == "oci":
         return OCI(model)
 
-    return OCI(model)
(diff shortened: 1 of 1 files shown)


ericcurtin:
@rhatdan if you don't want to merge this I understand

ericcurtin:
Longer-term, definately want to make it fallback to Ollama if OCI fails

rhatdan:
LGTM

PR #116, opened 2024-09-05
Update llama.cpp to 8962422b1c6f9b8b15f5aeaea42600bcc2d44177

ericcurtin:
The latest granite-code:3b isn't working in the version before this commit, time to upgrade.


tarilabs (approved):
Update to commit hash

/lgtm

PR #117, opened 2024-09-06
Set context size to 2048

ericcurtin:
The granite-code:3b model most recently pushed to Ollama registry is broken, it has an unreasonably high context size of 128000, set context size to 2048 so it's not unrealistically high.


ericcurtin:
Gonna merge this as I'm fearful people will run one of the granite-code:3b examples and this we are badly broken

rhatdan:
Should we add an option for this?

PR #118, opened 2024-09-06
Set columns to be as wide as required

ericcurtin:
Columns are now dynamically sized


rhatdan:
LGTM

rhatdan:
One thing we might want to do is to truncate certain fields as well and then add a --notruncate option.

PR #120, opened 2024-09-06
Remove curl dependency from Ollama models and improve progress bar

### LINKED ISSUE #115: Remove curl dependancy and better progress bar for Ollama models
We could use tqdm for this potentially:

```
#!/usr/bin/python3

import requests
from tqdm import tqdm

url = "https://registry.ollama.ai/v2/library/llama3/blobs/sha256:6a0746a1ec1aef3e7ec53868f220ff6e389f6f8ef87a01d77c96807de94ca2aa"
headers = {
    "Accept": "application/vnd.docker.distribution.manifest.v2+json"
}
output_path = "/home/curtine/sha256:6a0746a1ec1aef3e7ec53868f220ff6e389f6f8ef87a01d77c96807de94ca2aa"

# Try to get the content-length from the HEAD request
response = requests.head(url, headers=headers)
total_size = response.headers.get('content-length')

if total_size is None:
    print("Warning: Could not retrieve content length from the server. Progress bar may not be accurate.")
    total_size = 0
else:
    total_size = int(total_size)

# Define a custom bar format
bar_format = "Pulling {desc}: {percentage:3.0f}% ▕{bar:20}▏ {n_fmt}/{total_fmt} {rate_fmt} {remaining}"

# Perform the download
with requests.get(url, headers=headers, stream=True) as response:
    response.raise_for_status()  # Check for HTTP errors

    # Initialize the progress bar
    with tqdm(total=total_size, unit='B', unit_scale=True, unit_divisor=1024,
              desc=output_path[-16:], bar_format=bar_format) as progress_bar:

        # Open the output file in binary write mode
        with open(output_path, "wb") as file:
            for chunk in response.iter_content(chunk_size=8192):
                if chunk:  # Filter out keep-alive chunks
                    file.write(chunk)
                    progress_bar.update(len(chunk))  # Update progress bar

print(f"Downloaded file to {output_path}")
```

swarajpande5:
Resolves #115

Example run:
```
(venv) ramalama (feat*) $ python ramalama.py pull ollama://tinyllama       
Pulling tinyllama/latest: 100% ▕####################▏ 852/852 2.47MB/s 00:00
Pulling 6897109396245362: 100% ▕####################▏ 483/483 753kB/s 00:00
Pulling 83e42c043d6c7816:  22% ▕####4               ▏ 134M/608M 4.57MB/s 01:48 
```

### DIFF
--- a/container-images/ramalama/latest/Containerfile
+++ b/container-images/ramalama/latest/Containerfile
@@ -4,6 +4,8 @@ FROM registry.access.redhat.com/ubi9/ubi:9.4-1214.1726694543
 ARG HUGGINGFACE_HUB_VERSION=0.25.1
 # renovate: datasource=github-releases depName=containers/omlmd extractVersion=^v(?<version>.*)
 ARG OMLMD_VERSION=0.1.4
+# renovate: datasource=github-releases depName=tqdm/tqdm extractVersion=^v(?<version>.*)
+ARG TQDM_VERSION=4.66.5
 # renovate: datasource=git-refs depName=ggerganov/llama.cpp packageName=https://github.com/ggerganov/llama.cpp gitRef=master versioning=loose type=digest
 ARG LLAMA_CPP_SHA=32b2ec88bc44b086f3807c739daf28a1613abde1
 # renovate: datasource=git-refs depName=ggerganov/whisper.cpp packageName=https://github.com/ggerganov/whisper.cpp gitRef=master versioning=loose type=digest
@@ -22,6 +24,7 @@ RUN dnf install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.n
 RUN /usr/bin/python3 --version
 RUN pip install "huggingface_hub[cli]==${HUGGINGFACE_HUB_VERSION}"
 RUN pip install "omlmd==${OMLMD_VERSION}"
+RUN pip install "tqdm==${TQDM_VERSION}"
 
 ENV GGML_CCACHE=0
 
--- a/install.py
+++ b/install.py
@@ -78,6 +78,7 @@ directory containing brew to the PATH before continuing to install RamaLama
 def install_mac_dependencies():
     subprocess.run(["pip3", "install", "huggingface_hub[cli]"], check=True)
     subprocess.run(["pip3", "install", "omlmd==0.1.4"], check=True)
+    subprocess.run(["pip3", "install", "tqdm"], check=True)
     subprocess.run(["brew", "install", "llama.cpp"], check=True)
 
 
@@ -100,6 +101,7 @@ def setup_ramalama(bindir, tmp_dir):
     syspath += "/ramalama"
     subprocess.run(["install", "-m755", "-d", syspath], check=True)
     subprocess.run(["install", "-m755", to_file, ramalama_bin], check=True)
+
     python_files = [
         "cli.py",
         "huggingface.py",
--- a/ramalama/ollama.py
+++ b/ramalama/ollama.py
@@ -1,34 +1,58 @@
 import os
-import subprocess
+import urllib.request
 import json
-
-from ramalama.common import run_cmd, run_curl_cmd
+from ramalama.common import run_cmd
 from ramalama.model import Model
+
+bar_format = "Pulling {desc}: {percentage:3.0f}% ▕{bar:20}▏ {n_fmt}/{total_fmt} {rate_fmt} {remaining}"
+
+
+def download_file(url, dest_path, headers=None):
+    from tqdm import tqdm
+
+    request = urllib.request.Request(url, headers=headers or {})
+    with urllib.request.urlopen(request) as response:
+        total_size = int(response.headers.get("Content-Length", 0))
+        chunk_size = 8192  # 8 KB chunks
+        with open(dest_path, "wb") as file, tqdm(
+            desc=dest_path[-16:],
+            total=total_size,
+            unit="B",
+            unit_scale=True,
+            unit_divisor=1024,
+            bar_format=bar_format,
+            ascii=True,
+        ) as progress_bar:
+            while True:
+                chunk = response.read(chunk_size)
+                if not chunk:
+                    break
+                file.write(chunk)
+                progress_bar.update(len(chunk))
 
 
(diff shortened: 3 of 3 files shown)


ericcurtin:
Nice...

We should probably add this to the Containerfile in this repo also for Linux...

And if you could hardcode whatever version of tqdm you are on in the pip install that would be great... Our container builds run nightly, don't want them to sporadically break due to a tqdm upgrade.

ericcurtin:
We should also try to remove curl everywhere, so we only have a dependency on one http client

swarajpande5:
Sure, I'm trying to figure out why these tests are failing. 
I'll add the the same in the Containerfile and will remove curl everywhere. If I'm not wrong the only `curl` we are using after this is in the `install.py`'s `download(url, to)` function.

ericcurtin:
I think this should fix CI:

https://github.com/containers/ramalama/pull/121

container build must be run first, so everything has access to new python modules like tqdm

swarajpande5:
Thanks, I will add `check=True` in the subprocess.run command for tqdm once the above PR is merged.

ericcurtin:
I merged the other one to unblock you, rebase should hopefully fix it

swarajpande5:
I rebased but the test still failed unfortunately

ericcurtin:
@swarajpande5 I think if you put download into it's own download.py file, it should be fine.

We call some python code outside the container and some inside the container.

Creating download.py or something similar will ensure we only need this python module inside the container.

ericcurtin:
Or maybe actually just put it in ollama.py file, it's only relevant there anyway...

swarajpande5:
Sure, sounds good!

container-images/ramalama/latest/Containerfile
```diff
@@ -13,6 +13,7 @@ RUN dnf install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.n
 RUN /usr/bin/python3 --version
 RUN pip install "huggingface_hub[cli]==0.24.2"
 RUN pip install "omlmd==0.1.4"
+RUN pip install "tqdm==4.66.5"
```
rhatdan, line 16: Why are we picking particular versions?  This is going to be very difficult to get into Fedora unless these python tools are package into Fedora.

swarajpande5: I see that we have [huggingface-hub](https://packages.fedoraproject.org/pkgs/python-huggingface-hub/python3-huggingface-hub/) and [tqdm](https://packages.fedoraproject.org/pkgs/python-tqdm/python3-tqdm/) in fedora packages, however I cannot find omlmd. 

@ericcurtin Please help me in answering why particular versions have been picked, and the further steps to fix this.

ericcurtin: We have options when it comes to packaging. I'm open minded.

Right now the way things are, we are heavily reliant on the container image. Container images don't have Fedora Rawhide, CentOS Stream to act as a pipeline (at least the way it's done in this repo right now), the container image is automatically rebuilt every night.

So without versions specified, we can easily break overnight because of an API breaking change, etc.

But we should think about what will be ramalama dependencies in the rpm. Maybe it's just python3? Maybe it's just python3 and podman?

What we can't do is make everything a strict dependency in the rpm like the nvidia and amd stuff. Because then we have to pull 100G worth of complex dependencies for every use case.

rhatdan: Ok, but we need to have renovate watching these and updating them, or we will end up with ancient versions potentially loaded with CVEs.

ericcurtin: This is where @p5 may be able to help us out.

Don't let conversations like this block you btw @swarajpande5 we can always change things up in a new PR after this one.

p5: Hey
So if we need to version things that aren't natively supported by Renovate, we can introduce something called a "regex manager".  This can be configured to read the version numbers from any arbitrary file given it matches a particular pattern (that we configure), and provide it with some metadata about where to find version information about these dependencies.

I originally tried to PR a universal version of this into the [containers/automation](https://github.com/containers/automation/pull/225) repository, but the consensus between the developers there were this belongs inside individual repos first, and if it is demonstrated to be valuable across multiple places, they may consider adopting it as a central configuration.

I'll create a PR tonight or tomorrow with an implementation of this, and you can decide whether you want to use it or something else.

ericcurtin:
You'll need to add:

```
subprocess.run(["pip3", "install", "--break-system-packages", "tqdm==4.66.5"], check=True)
```

to install.py for macOS build @swarajpande5

ericcurtin:
@swarajpande5 I think you can import inside a function call on in python too, I dunno if that helps you here...

swarajpande5:
@ericcurtin I have already called the `subprocess.run()` for `tqdm` in install.py. Yeah, importing tqdm in the function wherein its required sounds really good. I'll make this change.

ericcurtin:
@swarajpande5 I see your issue now...

We should have no dependancy on tqdm in install.py , just in the actual ramalama application itself is where we want to use tqdm.

It's ok to use curl in the install.py. We can't have dependancies in install.py that aren't on a base macOS operating system already.

swarajpande5:
@ericcurtin I have not added tqdm as a dependency in install.py. I have removed the curl dependency in `download(url, to)` of install.py without using tqdm. Can you please check again, thanks.

swarajpande5:
@ericcurtin As per your suggestion I have reverted the `download(url, to)` function in install.py to using curl. If in future, we decide to fully remove curl, then we can add the implementation back.

rhatdan:
What is going on with this PR?

swarajpande5:
@rhatdan So I'm still working on this. Trying to resolve the ci tests on macos, thinking to either write a basic implementation of a macos image. Would give some more time on this weekend.

ericcurtin:
@rhatdan what is nice about the hard-coded versions... Is since @swarajpande5 kindly added the renovate support... It will auto-detect newer versions... But not only that, it runs everything through a CI build before updating a given version... There is some value in having renovate automatically run a CI build before we press the button to update the version...

rhatdan:
LGTM
Thanks @swarajpande5

PR #127, opened 2024-09-08
Update README.md

ericcurtin:
llama.cpp image is a broken link now. Some people may just end up using this with vllm anyway. Giving kudos to several projects now instead.


rhatdan:
LGTM

PR #130, opened 2024-09-10
We were getting ugly output if no models were present on the system

ericcurtin:
Ugly output like this:

  File "/usr/local/bin/ramalama", line 36, in <module>
    main(sys.argv[1:])
  File "/usr/local/bin/ramalama", line 20, in main
    ramalama.init_cli()
  File "/usr/share/ramalama/ramalama/cli.py", line 71, in init_cli
    args.func(args)
  File "/usr/share/ramalama/ramalama/cli.py", line 210, in list_cli
    name_width = max(len("NAME"), max(len(model["name"]) for model in models))

now it's cleaned up so it looks like:

NAME MODIFIED SIZE

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -187,6 +187,7 @@ def list_cli(args):
     mycwd = os.getcwd()
     os.chdir(f"{args.store}/models/")
     models = []
+
     # Collect model data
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

But could you add a test in the future.

PR #132, opened 2024-09-10
chore: add Renovate dependency management to Containerfile

p5:
I'm currently trying to get Renovate to run in my personal repository to confirm this is working, but hopefully you get the idea of how it works from this PR

Git history will be cleaned once this is working

PR #133, opened 2024-09-10
chore: add Renovate dependency management to Containerfile

p5:
Recreation of #132 because Git is hard...  

This PR adds generic Renovate regex managers to this repo to support updating versions in files not natively supported in Renovate.

This also inherits the containers org's default Renovate settings, so will start versioning everything else in this repo.

I have confirmed this setup works using my own repository (see PRs created [here](https://github.com/p5/containers-ramalama/pulls?q=sort%3Aupdated-desc+is%3Apr+is%3Aopen)), and I have confirmed a Podman build still completes successfully with the correct substitutions.

The important bits to note are the Renovate "query string" as I call it (not official name) must be on the line directly above the ARG/ENV.  Parameters must be defined in the correct order, because otherwise the Regex would have gotten way too complex.

### DIFF
--- a/container-images/ramalama/latest/Containerfile
+++ b/container-images/ramalama/latest/Containerfile
@@ -1,5 +1,14 @@
 FROM registry.access.redhat.com/ubi9/ubi:9.4
 
+# renovate: datasource=github-releases depName=huggingface/huggingface_hub extractVersion=^v(?<version>.*)
+ARG HUGGINGFACE_HUB_VERSION=0.24.2
+# renovate: datasource=github-releases depName=containers/omlmd extractVersion=^v(?<version>.*)
+ARG OMLMD_VERSION=0.1.4
+# renovate: datasource=git-refs depName=ggerganov/llama.cpp packageName=https://github.com/ggerganov/llama.cpp gitRef=master versioning=loose type=digest
+ARG LLAMA_CPP_SHA=32b2ec88bc44b086f3807c739daf28a1613abde1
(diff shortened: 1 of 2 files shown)


container-images/ramalama/latest/Containerfile
```diff
@@ -1,5 +1,14 @@
 FROM registry.access.redhat.com/ubi9/ubi:9.4
 
+# renovate: datasource=github-releases depName=huggingface/huggingface_hub extractVersion=^v(?<version>.*)
+ARG HUGGINGFACE_HUB_VERSION=0.24.2
```
p5, line 4: Renovate will go to the huggingface/huggingface_hub GitHub releases and grab the "latest" (non-prerelease) version.

Python packages do not contain the `v` in the version number, whereas the GitHub releases do.  We need to tell Renovate to ignore the `v` in the version string.

rhatdan:
LGTM
@ericcurtin PTAL

ericcurtin:
LGTM, I'm gonna defer this one to @cooktheryan 

I don't know renovate very well, but I assume this will run a whole CI build like in this PR and merge only on green?

p5:
> I don't know renovate very well, but I assume this will run a whole CI build like in this PR and merge only on green?

This configuration will run the CI pipelines automatically, but will not merge.  
Somebody with write access to the repository will still need to approve and merge the PRs, though Renovate can be configured to automatically merge if desired.

ericcurtin:
Thanks for the explanation @p5

Issue #124, opened 2024-09-06
module 'ramalama' has no attribute 'init_cli'


pineapple-vision:
Ubuntu 20.04.6 LTS
Python 3.8.10

Let me know if you need anything else to diagnose.
Receiving the following error:

```
~$ sudo ramalama list
Traceback (most recent call last):
  File "/usr/bin/ramalama", line 37, in <module>
    main(sys.argv[1:])
  File "/usr/bin/ramalama", line 21, in main
    ramalama.init_cli()
AttributeError: module 'ramalama' has no attribute 'init_cli'
```

ericcurtin:
@my-digital-plug thanks for trying this out... Wanna do a quick reinstall:

```
curl -fsSL https://raw.githubusercontent.com/containers/ramalama/s/install.py | sudo python3
```

and see if it still occurs?

And would you like guidance to contribute the fix yourself if it's still broken on your platform?

Ubuntu 20.04 isn't the most useful platform for this kinda stuff, the GPU/AI enablement stuff is a little old, but we can try...

ericcurtin:
Could you also run:

which podman; which docker

on this system?

my-digital-plug:
Still getting it after reinstall.

`~$ which podman; which docker
/usr/bin/docker`

Forgot I was using docker instead of podman, so maybe that is the issue.
Regardless, had to destroy my VM (sshd user error, very dumb mistake on my end) and will setup a new one.  Ill see if I get the same issue.

ericcurtin:
You are better off using at least 22.04 in terms of Ubuntu, I'm pretty sure that's what CI uses.

RHEL9 (and soon 10) will always be a first class citizen. Or one can use CentOS Stream 9 also, although I wouldn't recommend that in production. podman is also best on RHEL-derived distros for obvious reasons.

And yes, we only really test on podman, although docker may work fine.

In terms of Ubuntu, 24.04 the latest LTS release is probably best.

When people start using ramalama in real life, the latest LTS of a given distro will always be best... You want a relatively recent kernel, etc. to take advantage of all the latest AI/GPU enablement that's been done recently.

We do most of the testing on RHEL/podman in other words.

ericcurtin:
Things like Ubuntu/docker will be upto the community to maintain. Contributions always welcome.

my-digital-plug:
[...]

PR #143, opened 2024-09-12
Bug fix was missing the separator between p and ramalama

ericcurtin:
os.path.join function adds necessary '/' characters when required


ericcurtin:
I wanna merge this one as we are badly broken right now without this:

https://github.com/containers/ramalama/issues/124

rhatdan:
Please add tests in the future.  We need to test an installed system and sit in a directory with an empty ramalama subdir and see what happens.

PR #144, opened 2024-09-13
Add pull tests

rhatdan:
As part of this test, I also removed the tests from ci.sh that are now implemented in bats.


ericcurtin (approved):
LGTM

Issue #38, opened 2024-08-10
Implement shortnames


ericcurtin:
podman has this concept of shortnames, which looks like this:

/etc/containers/registries.conf.d/000-shortnames.conf

```
$ head /etc/containers/registries.conf.d/000-shortnames.conf
[aliases]
  # almalinux
  "almalinux" = "docker.io/library/almalinux"
  "almalinux-minimal" = "docker.io/library/almalinux-minimal"
  # Amazon Linux
  "amazonlinux" = "public.ecr.aws/amazonlinux/amazonlinux"
  # Arch Linux
  "archlinux" = "docker.io/library/archlinux"
  # centos
  "centos" = "quay.io/centos/centos"
  ```

Create something similar `/etc/ramalama/registries.conf.d/000-shortnames.conf` that can do the same for ramalama

Suggested by @rhatdan in https://github.com/containers/ramalama/pull/35

rhatdan:
I think this is a good idea.

ericcurtin:
The podman file for this is:

$ grep -v "^#" /etc/containers/registries.conf
unqualified-search-registries = ["registry.fedoraproject.org", "registry.access.redhat.com", "docker.io"]
short-name-mode = "enforcing"

so I guess the ramalama equivalent would be:

/etc/ramalama/registries.conf

rhatdan:
Yes I think that makes sense, although perhaps we should use /usr/share/ramalama/registries.conf and then allow sysadmins to override in /etc/ramalama.

PR #146, opened 2024-09-13
Implement shortnames feature

ericcurtin:
Add support for resolving model shortnames. Parse shortnames from /etc/ramalama/registries.conf.d/000-shortnames.conf . Modify New() function to use resolved shortnames. Improve user experience by allowing use of familiar shorthand names for models. This change allows users to use shortnames instead of full model names, similar to podman functionality. The feature enhances usability while maintaining compatibility with existing model naming conventions.

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -11,6 +11,7 @@ import string
 import subprocess
 import sys
 import time
+import configparser
 
 from ramalama.huggingface import Huggingface
 from ramalama.common import in_container, container_manager, exec_cmd, run_cmd
@@ -66,6 +67,12 @@ def init_cli():
         return version(args)
     # create stores directories
     mkdirs(args.store)
+    if hasattr(args, "MODEL"):
+        resolved_model = resolve_shortname(args.MODEL)
+        if resolved_model:
+            args.UNRESOLVED_MODEL = args.MODEL
+            args.MODEL = resolved_model
+
     if run_container(args):
         return
 
@@ -390,14 +397,15 @@ def get_store():
 
 
 def find_working_directory():
-    wd = "./ramalama"
-    for p in sys.path:
-        target = os.path.join(p, "ramalama")
-        if os.path.exists(target):
-            wd = target
-            break
+    if os.path.exists("./ramalama"):
+        return "./ramalama"
 
-    return wd
+    for p in sys.path:
+        p = os.path.join(p, "ramalama")
+        if os.path.exists(p):
+            return p
+
+    return ""
 
 
 def run_container(args):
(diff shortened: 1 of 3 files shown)


ramalama/cli.py
```diff
@@ -467,7 +468,23 @@ def run_container(args):
     exec_cmd(conman_args)
 
 
+def load_shortnames():
+    config = configparser.ConfigParser()
+    config_file = "/etc/ramalama/registries.conf.d/000-shortnames.conf"
```
ericcurtin, line 473: I should probably iterate through all the files in "/etc/ramalama/registries.conf.d/"

ericcurtin:
@rhatdan it's not clear to me how podman/bats handles tests that depend on certain files like "/etc/containers/registries.conf.d/000-shortnames.conf"

I can see tests in podman/bats testing the shortnames features, but they seem to test failure cases, rather testing successful cases.

ericcurtin:
Linked issue: https://github.com/containers/ramalama/issues/38

ericcurtin:
I still have to do directory precedence style code here. For example on macOS rootless installs are more common, so probably don't want to edit /etc at all:

etc/ramalama/registries.conf.d/000-shortnames.conf

etc/ramalama/registries.conf.d/000-shortnames.conf
```diff
@@ -0,0 +1,4 @@
+[aliases]
+  "granite" = "huggingface://instructlab/granite-7b-lab-GGUF/granite-7b-lab-Q4_K_M.gguf"
```
rhatdan, line 2: Should add multiple
granite7b (and or granite-7b) ( I would love it if we stored these in cri-o by default, will talk about this next week)
granite-lab

rhatdan: This file should also be in /usr/share.  Leave /etc for sysadmins. /usr/share for distributions.

ramalama/cli.py
```diff
@@ -439,6 +451,7 @@ def run_container(args):
         f"-v{home}:{home}",
         f"-v{sys.argv[0]}:/usr/bin/ramalama:ro",
         f"-v{wd}:/usr/share/ramalama/ramalama:ro",
+        f"-v{ed}:/etc/ramalama:ro",
```
rhatdan, line 454: We got to think about this one.

rhatdan:
We want to read these files in /usr/share/ramalama/aliases.conf, /etc/ramalama/aliases.conf $HOME/.config/ramalama/aliases.conf

For the in the container use case we should prepare a file out of this content with the last ALIASES winning and then mount it into the container.

cat  /usr/share/ramalama/aliases.conf /etc/ramalama/aliases.conf $HOME/.config/ramalama/aliases.conf > /tmp/aliases.conf and then process it, removing duplicates.

ericcurtin:
Makes sense to me, although instead of this one:

`$HOME/.config/ramalama/aliases.conf`

we could do:

`$HOME/.local/share/ramalama/aliases.conf`

`$HOME/.local/share/ramalama` is a directory we already use and it kinda fits with `/usr/share/ramalama/aliases.conf `

But I'm easy about directory structures, happy to go completely as you described also @rhatdan ...

etc/ramalama/registries.conf.d/000-shortnames.conf
ericcurtin: What ollama does... Which is kinda interesting is this format (like container images):

granite-code:8b
granite-code:3b
granite-code:latest (alias for 3b tag or whatever tag one selects)

rhatdan: Yup that would make sense.

rhatdan:
Well $HOME/.local/share is for app content, whereas $HOME/.config/ is for user custom configuration.

As far as the mount within the container,  I think we will need to process all of the aliases on the users host system into one aliases.conf file which we mount into the container.  So assemble it from /usr, /etc/ and $HOME into something in /tmp and then mount into container.  I don't think we want to mount all of the mounts from the host into the container. This way the container itself could have some aliases.conf files which are not on the users host system.

etc/ramalama/registries.conf.d/000-shortnames.conf
ericcurtin: At the moment I have:

"granite-lab-7b" and  "merlinite-lab-7b"

the ':' characters don't parse well in the .conf file with the python parser, but can try again to make ':' work if we really want the ':' character to be present.

ericcurtin:
> Well $HOME/.local/share is for app content, whereas $HOME/.config/ is for user custom configuration.
> 
> As far as the mount within the container, I think we will need to process all of the aliases on the users host system into one aliases.conf file which we mount into the container. So assemble it from /usr, /etc/ and $HOME into something in /tmp and then mount into container. I don't think we want to mount all of the mounts from the host into the container. This way the container itself could have some aliases.conf files which are not on the users host system.

I moved the resolving of short-names to long-names to occur outside of the container, so now no temporary files required, seemed cleaner... Less difference between container and no-container case and no temporary files to clean up. And now no bind mounts also.

ramalama/ollama.py
```diff
@@ -41,7 +41,7 @@ def init_pull(repos, manifests, accept, registry_head, model_name, model_tag, mo
             manifest_data = json.load(f)
     except subprocess.CalledProcessError as e:
         if e.returncode == 22:
-            raise KeyError((f"{model}:{model_tag} not found"))
+            raise KeyError((f"{model} not found"))
```
ericcurtin, line 44: {model} already has {model_tag}  so it made the output look weird, duplicate tag at the end

rhatdan:
LGTM

PR #147, opened 2024-09-13
Use aliases=[] for ramalama list command

rhatdan:


ericcurtin:
Ah much nicer, didn't know about that feature. I think I've learned more about python in the last few weeks than I have in my whole career :sweat_smile:

PR #157, opened 2024-09-19
Cleanup and install shortnames files

rhatdan:


ramalama/shortnames.py
```diff
@@ -0,0 +1,32 @@
+import configparser
+import os
+
+
+class Shortnames:
+    """Model super class"""
```
ericcurtin, line 6: Shortnames utility class? (or whatever)

Seems like a copy/paste carryover

ramalama/shortnames.py
```diff
@@ -0,0 +1,32 @@
...
+
+    shortnames = {}
+
+    def __init__(self):
+        config = configparser.ConfigParser()
+        sn = {}
+        file_paths = [
+            "/usr/share/ramalama/shortnames.conf",  # for development
```
ericcurtin, line 14: This first directory was:

`usr/share/ramalama/shortnames.conf`

it was intended for when one calls ramalama.py in the git repo for testing. Not required, but useful.

Now that it has a slash, it's just a duplicate entry in the list.

ramalama/shortnames.py
```diff
@@ -0,0 +1,32 @@
...
+    def __init__(self):
+        file_paths = [
+            "/usr/share/ramalama/shortnames.conf",
+            "/etc/ramalama/shortnames.conf",
+            os.path.expanduser("~/.config/ramalama/shortnames.conf"),
+            "./shortnames/shortnames.conf",  # for development
+            "./shortnames.conf",  # for development
+        ]
+
+        for file_path in file_paths:
```
ericcurtin, line 19: If we do this:

```
        for file_path in file_paths:
            config = configparser.ConfigParser(delimiters=('='))  # Only use '=' as a delimiter
            config.read(file_path)
            if "shortnames" in config:
                self.shortnames.update(config["shortnames"])
```

I think we can have shortnames in this format:  "granite-lab:7b", "merlinite-lab:7b"

But I haven't tested. It wasn't working before because python's config parser by default treats both '=' and ':'  as delimiters by default.

But we can do it in the next PR.

The colon separator for the 3b, 7b, etc. was popularised by Ollama.

ericcurtin (approved):
This is approved, not sure what's the state of our packit builds if they are supposed to be red right now or not.

ericcurtin (approved):

ericcurtin (approved):

rhatdan:
granite:7b is now supported.
Local config can override global config as well.

PR #160, opened 2024-09-20
RPM: metaprovides `ramalama`

lsm5:
This will make it possible for users to `dnf install ramalama`.


rhatdan:
LGTM

lsm5:
https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/build/8043703/ <- should allow you to `dnf install ramalama`

PR #165, opened 2024-09-22
Consistency changes

ericcurtin:
The default help text in the python parsers starts with a lower-case, the only way to make everything consistent is to make everything start with a lower-case:

  -h, --help            show this help message and exit

Also, various other consistency changes.

Also make Model capital everywhere.

Some typo fixes like "containersis".


ericcurtin:
Maybe there's a strong preference for caps in some places, I dunno, if there is just shout... It was just noticeable that in options, often:

-h, --help show this help message and exit

started with a lower-case and nothing else.

Did this in transit while I had no internet access.

rhatdan:
This is fine.
LGTM

PR #166, opened 2024-09-22
Makefile: separate install-program target

juhp:
This factors out a new `install-program` make target from `install`.

Currently the rpm/spec file does not package the main ramalama program so this helps towards that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,16 +33,19 @@ help:
 	@echo
 
 .PHONY:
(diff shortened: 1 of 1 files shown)


juhp:
Also afaict `RAMALAMA_VERSION` is not used currently, should it be removed?

rhatdan:
Yes
LGTM

PR #174, opened 2024-09-23
Switch ramalama serve to run containers in detach mode

rhatdan:


ericcurtin (approved):
It's kinda like sshd/httpd style now...

It's debatable... Maybe even when run as --nocontainer we may want to run as a background process, repeating myself, like sshd/httpd

rhatdan:
Can not run in background without a container, we could attempt to do detach by using shell, but I think the user can do this.

PR #175, opened 2024-09-23
Cleanup execution of the container

rhatdan:
No reason to leak /tmp, $HOME into the container.
The container is not run interactively, so we should not use --it Always name the container

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,9 +2,15 @@ name: ci
 on: [push, pull_request]
 jobs:
   linux:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - name: install bats
+        shell: bash
+        run: |
+	   sudo apt update
+	   sudo apt -y install bats
+	apt-get bash
       - name: Run a one-line script
         run: make test
 
@@ -17,5 +23,8 @@ jobs:
         run: brew install go
       - name: Run a one-line script
         shell: bash
-        run: make test
+        run: make validate
+      - name: Run ci
+        shell: bash
+        run: make ci
 
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -322,7 +322,9 @@ def _name():
 def run_parser(subparsers):
     parser = subparsers.add_parser("run", help="run specified AI Model as a chatbot")
     parser.add_argument("--prompt", dest="prompt", action="store_true", help="modify chatbot prompt")
-    parser.add_argument("-n", "--name", dest="name", help="name of container in which the Model will be run")
+    parser.add_argument(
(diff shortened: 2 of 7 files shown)


ericcurtin (approved):
CI is complaining about styling again

ericcurtin:
But we may need "-it" for the "ramalama run" case right? In bash this can be a good way of detecting if we need the -t flag:

T=
if [ -t 1 ]; then
  T="-t"
fi

Apparently this is the python3 equivalent:

```
import sys

T = ""
if sys.stdout.isatty():
    T = "-t"
```

rhatdan:
Yes in run case we need it.

I think we are not running tests on PRs.

ramalama/cli.py
```diff
@@ -442,15 +448,15 @@ def run_container(args):
...
-        name = args.name
-
     conman = container_manager()
     if conman == "":
         return False
 
-    home = os.path.expanduser("~")
+    if hasattr(args, "name"):
+        name = args.name
+    else:
```
ericcurtin, line 457: I'm starting to learn python doesn't have `if` scope, variables declared inside the if scope can be used outside the scope... I must use if/else more in python to avoid setting things twice...

test/system/030-run.bats
```diff
@@ -28,7 +28,7 @@ load helpers
 
 @test "ramalama run granite with prompt" {
     run_ramalama run --name foobar granite "How often to full moons happen"
-    is "$output" ".*Moon" "should include some info about the Moon"
+    is "$output" ".*month" "should include some info about the Moon"
```
ericcurtin, line 31: This seems like a test that will always be flakey... Since the models are not so deterministic...

I wonder should we check for, at least x characters used or at least x lines printed...

ericcurtin: I find flakey CI can really slow down development

ericcurtin (approved):

test/system/030-run.bats
rhatdan: Yup, I did not know of a better way to test this out.

rhatdan: But at least we will have tests firing now.

ericcurtin: If it gets flakey I guess we can try and test for x characters that are not whitespace, etc.

We must find a way of getting both the system prompt into the model "You are a helpful assistant" and the user prompt:

```
$ ./ramalama.py run granite-code hello
-world)
$ ./ramalama.py run granite-code hello
;
```

without the "You are a helpful assistant", the output is weird...

.github/workflows/ci.yml
```diff
@@ -2,9 +2,15 @@ name: ci
 on: [push, pull_request]
 jobs:
   linux:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     steps:
```
ericcurtin, line 6: I think our CI builds have stopped running since this MR, github/gitlab yaml is not fun

PR #179, opened 2024-09-24
Start of quadlet generator

ericcurtin:


ericcurtin:
There's gonna be a lot of if's around this part of the code... But it might be nice to put podman run/quadlet/docker compose/k8s generation all in the same place, so that if somebody changes one, they know to change all the others

ericcurtin:
Happy to throw this away now, FWIW it was printing like this:

```
[Container]
Environment=RAMALAMA_TRANSPORT
Label="RAMALAMA container"
SecurityLabelDisable=true
ContainerName=ramalama_MNKEe5YUZI
Volume=/home/curtine/.local/share/ramalama::/var/lib/ramalama
Volume=./ramalama.py:/usr/bin/ramalama:ro
Volume=/home/curtine/git/ramalama/ramalama:/usr/share/ramalama/ramalama:ro
PublishPort=8080:8080
AddDevice=/dev/dri
Image=quay.io/ramalama/ramalama:latest
Exec=/usr/bin/ramalama serve granite-code
```

PR #182, opened 2024-09-24
change install dir preferences

boredsquirrel:
priorities: /opt/homebrew/share then /usr/local/share then /usr/share/ramalama

Co-Authored-By: boredsquirrel [113100745+boredsquirrel@users.noreply.github.com](mailto:113100745+boredsquirrel@users.noreply.github.com)
Signed-off-by: Eric Curtin [ecurtin@redhat.com](mailto:ecurtin@redhat.com)

### DIFF
--- a/install.py
+++ b/install.py
@@ -90,12 +90,11 @@ def setup_ramalama(bindir, tmp_dir):
     to_file = os.path.join(tmp_dir, from_file)
     download(url, to_file)
     ramalama_bin = os.path.join(bindir, binfile)
-    syspath = "/usr/share/ramalama"
+    sharedirs = ["/opt/homebrew/share", "/usr/local/share", "/usr/share"]
+    syspath = next((d for d in sharedirs if os.path.exists(d)), None)
+    syspath += "/ramalama"
     if sys.platform == "darwin":
         install_mac_dependencies()
-        sharedirs = ["/opt/homebrew/share", "/usr/local/share"]
-        syspath = next((d for d in sharedirs if os.path.exists(d)), None)
-        syspath += "/ramalama"
 
     subprocess.run(["install", "-m755", "-d", syspath], check=True)
     syspath += "/ramalama"
--- a/ramalama.py
+++ b/ramalama.py
@@ -7,11 +7,8 @@ import sys
 
 
 def main(args):
-    syspath = "/usr/share/ramalama"
-    if sys.platform == "darwin":
-        sharedirs = ["/opt/homebrew/share/ramalama", "/usr/local/share/ramalama"]
-        syspath = next((d for d in sharedirs if os.path.exists(d)), None)
-
+    sharedirs = ["/opt/homebrew/share/ramalama", "/usr/local/share/ramalama", "/usr/share/ramalama"]
+    syspath = next((d for d in sharedirs if os.path.exists(d)), None)
     sys.path.insert(0, syspath)
 
     import ramalama


ericcurtin:
@lsm5 another occurrence

install.py
```diff
@@ -88,9 +88,11 @@ def setup_ramalama(bindir, tmp_dir):
     syspath = "/usr/share/ramalama"
     if sys.platform == "darwin":
         install_mac_dependencies()
-        sharedirs = ["/opt/homebrew/share", "/usr/local/share"]
+        sharedirs = ["/opt/homebrew/share", "/usr/local/share", "/usr/share/ramalama"]
         syspath = next((d for d in sharedirs if os.path.exists(d)), None)
         syspath += "/ramalama"
+    if sys.platform == "darwin":
```
rhatdan, line 94: You are already in a "darwin" if block?

ericcurtin: This should be:

```
    sharedirs = ["/opt/homebrew/share", "/usr/local/share", "/usr/share/ramalama"]
    syspath = next((d for d in sharedirs if os.path.exists(d)), None)
    syspath += "/ramalama"
    if sys.platform == "darwin":
        install_mac_dependencies()
```

only one if statement

ericcurtin: After getting a new macbook... I realise now, modern macOS ships a really old bash binary (the latest GPLv2 version)... But after installing chrome it pulls in /usr/bin/python3 somehow (I think there's some extended macOS tools that can be dragged in like python3)... Maybe we should have left this install script as /bin/bash

rhatdan: Excellent, now you can play with ramalama and podman-machine (Desktop).

ericcurtin: Yup, I remember you brought up in a discussion today podman-machine vs native macOS support... IMO we should do both, unless one starts to get hard to maintain...

native macOS performs better, people notice this and podman-machine is Linux and closer to a server environment...

ericcurtin:
You'll need to sign your commit also @boredsquirrel 

`git commit --amend -s # to sign an old commit`

install.py
rhatdan: I think this becomes the --nocontainer switch. We add a Environment variable to cause this to be default.

rhatdan: We also need to work with Sergio to get better performance.

boredsquirrel:
I am struggling for whatever reason. Created a GPG key, same credentials, synced my fork of the repo on github, pulled to local, switched to my branch, `git pull --rebase`, changed the file, added the changes, did a signed commit, trying to push, always the same error.

```
 ! [rejected]        nonsystem-install-dirs -> nonsystem-install-dirs (non-fast-forward)
error: failed to push some refs to 'github.com:boredsquirrel/ramalama'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. If you want to integrate the remote changes,
hint: use 'git pull' before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
```

install.py
ericcurtin: We should but it's gonna be hard to beat macOS... llama.cpp's initial goal was to run well on macOS and Apple Silicon, a ton of projects are using it that way, so that's a lot of combined engineering efforts, Ollama being one tool that uses it this way...

There's also the whole making sure the Linux VM allocates huge amounts of RAM to the VM, etc.

I do think I will port all this back to bash...

It's just the install experience on macOS really sucks via python3...

And bash is probably easier for this kind of thing anyway...

ericcurtin:
On your branch locally, I would recommend:

```
git fetch
git reset --hard origin/main
git cherry-pick 8db81d2afedddf2c8ac5578a4f2b0cfc7572b106
git commit -a --amend -s
git push -f
```

that should get you back in business, I think...

rhatdan:
LGTM

boredsquirrel:
thanks guys!

Git can be really confusing.

I think I got it to work correctly with @ericcurtin s guide.

Thanks for being kind on the noobs :D

PR #185, opened 2024-09-24
Also suggest adding brew dir to PATH

ericcurtin:
After one installs brew, there is one more step that the installer doesn't do, add /opt/homebrew/bin to the PATH.


ericcurtin:
@lsm5 it happened again :(

rhatdan:
LGTM

PR #187, opened 2024-09-25
llama-cli made a change

ericcurtin:
Now instead of --log-disable we need to redirect stderr to /dev/null:

https://github.com/ggerganov/llama.cpp/pull/9418

### DIFF
--- a/ramalama/common.py
+++ b/ramalama/common.py
@@ -34,9 +34,14 @@ def available(cmd):
     return shutil.which(cmd) is not None
 
 
-def exec_cmd(args):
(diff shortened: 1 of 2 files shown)


ericcurtin:
Noticed on macOS (because we currently just install the lastest llama.cpp from brew), it is worrying llama-cli regularly makes breaking changes... We may eventually have to stop using the llama.cpp examples like llama-cli, llama-server at some point and write our own application against llama.cpp like Ollama, etc.

rhatdan:
LGTM

PR #186, opened 2024-09-24
Fix CI

ericcurtin:
Use ubuntu 22.04, maybe we can move to 24.04 also in another PR.

PR #188, opened 2024-09-25
Fix CI

rhatdan:
Replaces: https://github.com/containers/ramalama/pull/186

Use ubuntu 22.04, maybe we can move to 24.04 also in another PR.


ericcurtin:
working CI 🍾

rhatdan:
More coming.

PR #189, opened 2024-09-25
pip3 on macOS can install in funny directories

ericcurtin:
Like:

/Users/ecurtin/Library/Python//3.9/bin/omlmd

### DIFF
--- a/ramalama/oci.py
+++ b/ramalama/oci.py
@@ -2,15 +2,23 @@ from pathlib import Path
 import os
 import re
 import subprocess
+import sys
 
 from ramalama.model import Model
-from ramalama.common import run_cmd, exec_cmd, perror
+from ramalama.common import run_cmd, exec_cmd, perror, available
 
 
 class OCI(Model):
     def __init__(self, model):
         super().__init__(model.removeprefix("oci://").removeprefix("docker://"))
         self.type = "OCI"
+        if available("omlmd"):
+            self.omlmd = "omlmd"
+        else:
+            for i in sys.path:
+                self.omlmd = f"{i}/../../../bin/omlmd"
+                if os.path.exists(self.omlmd):
+                    break
 
     def login(self, args):
         conman_args = [self.conman, "login"]
@@ -53,7 +61,7 @@ class OCI(Model):
         model_file = Path(local_model_path).resolve()
         try:
             # Push the model using omlmd, using cwd the model's file parent directory
-            run_cmd(["omlmd", "push", target, str(model_file), "--empty-metadata"], cwd=model_file.parent)
+            run_cmd([self.omlmd, "push", target, str(model_file), "--empty-metadata"], cwd=model_file.parent)
         except subprocess.CalledProcessError as e:
             perror(f"Failed to push model to OCI: {e}")
             raise e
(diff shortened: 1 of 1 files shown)


ericcurtin:
@tarilabs omlmd doesn't really work great for me at least on macOS, which is a pity, it's a very popular platform for model inferencing...

$ ./ramalama.py pull oci://quay.io/mmortari/gguf-py-example:v1
Downloading quay.io/mmortari/gguf-py-example:v1...
/Users/ecurtin/Library/Python/3.9/lib/python/site-packages/urllib3/__init__.py:35: NotOpenSSLWarning: urllib3 v2 only supports OpenSSL 1.1.1+, currently the 'ssl' module is compiled with 'LibreSSL 2.8.3'. See: https://github.com/urllib3/urllib3/issues/3020
  warnings.warn(

tarilabs:
That's weird because I made all the demos of omlmd on Mac OSX 👉 https://containers.github.io/omlmd/demos/demo/

tarilabs:
Is the installation you are doing in a pyenv or venv?
I think in general, or at least my experience, using the global Mac OS Python environment is _in general_ discouraged.

For another instance this is also analogously highlighted for ilab, here: https://github.com/instructlab/instructlab#:~:text=NOTE%3A%20The%20following%20steps%20in%20this%20document%20use%20Python%20venv%20for%20virtual%20environments.

Happy to follow-up online to double check which steps you followed and potential more robust alternatives?
lmk!

ericcurtin:
Maybe it's just a harmless warning and https://quay.io/repository/mmortari/gguf-py-example isn't a real model

tarilabs:
> and https://quay.io/repository/mmortari/gguf-py-example isn't a real model

if I recall I've taken that from the gguf example of llama.cpp repo and a valid gguf file but indeed not a Large Language Model

tarilabs:
> Maybe it's just a harmless warning

I don't get those warning anyways on Mac OSX during my demos, so I think it's a matter of how the library/dependencies are installed maybe?

tarilabs:
fwiw, just tried straight from repo root now and with:

`./ramalama.py pull oci://quay.io/mmortari/gguf-py-example:v1` 

I get just-and-only:

```
Downloading quay.io/mmortari/gguf-py-example:v1...
```

which then shows in my storage:

```
% ./ramalama.py list                                          
NAME                                                   MODIFIED    SIZE
oci://quay.io/mmortari/gguf-py-example/v1/example.gguf 1 month ago 4.0K
```

image


I believe the warn from https://github.com/containers/ramalama/pull/189#issuecomment-2374391662 maybe depends on the local system environment and configuration used then?

Possibly some `pip install ...` was done against the Mac native _global_ Python environment?
On Mac, my understanding that is discouraged,
and I suggest using a Python environment like [example from InstructLab](https://github.com/instructlab/instructlab#:~:text=NOTE%3A%20The%20following%20steps%20in%20this%20document%20use%20Python%20venv%20for%20virtual%20environments),  
or use `pipx`

my2c and happy to discuss further!

tarilabs:
...I think I see now; seems to me install.py is making use of the default pip3 from the default Mac global python environment:

https://github.com/containers/ramalama/blob/07555d59a1d85aa43136fd4f4d2d2be6c7d62943/install.py#L79-L80

what about:

1. using pipx instead of installing in the Mac global python environment
i.e. make for Mac `pipx` a depedency as homebrew `brew`

2. just "shading" omlmd sources inside the ramalama
after all is just an sdk

3. I could check what distributing omlmd through brew entails

please notice 2, 3 wouldn't however solve the same issue for the other dependency `huggingface_hub[cli]` or any future dependencies, which imho should not be installed on the end-user default Mac global python environment my2c

hence, I believe option 1 is the most preferable for the dependencies or ramalama when installed/used on Mac.

always happy to discuss further !

ericcurtin:
@tarilabs if you could try and package ramalama and omlmd via brew that would be appreciated! We should migrate to there eventually...

tarilabs:
> could try and package ramalama and omlmd via brew that would be appreciated! We should migrate to there eventually...

noted @ericcurtin eventually this will also impact however other dependencies of ramalama itself, as I noted in https://github.com/containers/ramalama/pull/189#issuecomment-2376546887

PR #194, opened 2024-09-26
Rename project to RamaLama

ericcurtin:
There was a mix of Ramalama and RamaLama. ramalama is the executable name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,16 +1,16 @@
 # ramalama
 
-The Ramalama project's goal is to make working with AI boring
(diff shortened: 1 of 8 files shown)


rhatdan:
LGTM. I think there are still ramalama in the code which should be RamaLama.

ericcurtin:
Yeah it's not always clear when to call it ramalama vs RamaLama... But anyway...

One example that should probably change is:

The ramalama uses the Ollama registry transport by default.

to

RamaLama uses the Ollama registry transport by default.

PR #196, opened 2024-09-26
Fix issues in quadlet generation

rhatdan:
No reason to volume mount ramalama into container.

### DIFF
--- a/docs/ramalama-serve.1.md
+++ b/docs/ramalama-serve.1.md
@@ -48,32 +48,44 @@ CONTAINER ID  IMAGE                             COMMAND               CREATED   
 
 Generate a quadlet for running the AI Model service
 ```
-$ ramalama serve --generate=quadlet granite
+$ ramalama serve --name MyGraniteServer --generate=quadlet granite > $HOME/.config/containers/systemd/MyGraniteServer.container
+$ cat $HOME/.config/containers/systemd/MyGraniteServer.container
 
 [Unit]
 Description=RamaLama granite AI Model Service
 After=local-fs.target
 
 [Container]
-Device=+/dev/dri
-Device=+/dev/kfd
-Environment=RAMALAMA_TRANSPORT=HuggingFace
+AddDevice=-/dev/dri
+AddDevice=-/dev/kfd
 Exec=llama-server --port 8080 -m /home/dwalsh/.local/share/ramalama/models/huggingface/instructlab/granite-7b-lab-GGUF/granite-7b-lab-Q4_K_M.gguf
 Image=quay.io/ramalama/ramalama:latest
-Label=RAMALAMA container
-Name=ramalama_YcTTynYeJ6
-SecurityLabelDisable=true
-Volume=/home/dwalsh/ramalama/ramalama:/usr/bin/ramalama/ramalama:ro
-Volume=./ramalama.py:/var/lib/ramalama:ro
+Volume=/home/dwalsh/.local/share/ramalama/models/huggingface/instructlab/granite-7b-lab-GGUF/granite-7b-lab-Q4_K_M.gguf:/home/dwalsh/.local/share/ramalama/models/huggingface/instructlab/granite-7b-lab-GGUF/granite-7b-lab-Q4_K_M.gguf:ro,z
+ContainerName=MyGraniteServer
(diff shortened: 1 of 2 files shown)


rhatdan:
@ygalblum PTAL

ramalama/model.py
```diff
@@ -128,14 +128,10 @@ def quadlet(self, args, exec_args):
 [Container]
 Device=+/dev/dri
```
ericcurtin, line 129: We should only include these if they exist on the system I think:

/dev/dri
/dev/kfd

not sure if the container will start if they are specified and they are not present.

ramalama/model.py
```diff
@@ -128,14 +128,10 @@ def quadlet(self, args, exec_args):
...
 Device=+/dev/kfd
-Environment=RAMALAMA_TRANSPORT=%s
 Exec=%s
 Image=%s
-Label=RAMALAMA container
-%s
 SecurityLabelDisable=true
-Volume=%s:/usr/bin/ramalama/ramalama:ro
-Volume=%s:/var/lib/ramalama:ro
```
ericcurtin, line 138: Without bind mounting /var/lib/ramalama the container won't have access to models, in the rootless case we do something like:

```
-v/var/home/ecurtin/.local/share/ramalama:/var/lib/ramalama
```

rhatdan: Good point.  We should just volume mount in the image

ericcurtin (approved):
I'm not sure about the code here, but assuming these quadlets were generated and tested with a couple of curl commands to show the model is responsive. That's good with me

docs/ramalama-serve.1.md
```diff
@@ -48,7 +48,7 @@ CONTAINER ID  IMAGE                             COMMAND               CREATED
 
 Generate a quadlet for running the AI Model service
 ```
-$ ramalama serve --generate=quadlet granite
+$ ramalama serve --name MyModel --generate=quadlet granite
```
ygalblum, line 51: A few questions:

1. Can't the user chose the model? Or is this what `granite` is supposed to mean? If yes, then shouldn't the name fit the one used in the output?
2. Using `MyModel` as the value of the Container name is a bit confusing. I initially thought it was the name of the model

ramalama/model.py
```diff
@@ -106,18 +106,18 @@ def serve(self, args):
...
         if hasattr(args, "port"):
             port_string = f"PublishPort={args.port}"
 
         name_string = ""
-        if hasattr(args, "name") and args.name != "":
-            name_string = f"Name={args.name}"
+        if hasattr(args, "name") and args.name:
+            name_string = f"ContainerName={args.name}"
 
         print(
```
ygalblum, line 122: Not blocking, but have you considered using `f""` instead of `%`?

rhatdan: I have used both in the code,  Eric tends to use the f"" and I tend to use the %s method. But we should go through the code and use one for consistency.

rhatdan:
@ericcurtin I am not thrilled with what we have here so far, but this should work.  I think we should have an example with OCI Artifact models which would continue to use the Artifact name rather then a path to the model for the generated quadlet.  Perhaps generating two quadlets, a QUADLET.volume and QUADLET.container.

ramalama/model.py
ericcurtin: I understand the desire for printf-style, that's normally what I lean towards in new languages. But I am starting to prefer f"", you don't have to count out the args. This code is a good example, I really have to count the %'s and the args to figure out which is the 5th parameter. Also with f"" one doesn't have to worry about type.

I'm leaning towards f""

These python style {} formatted strings actually made the C++ standard recently via std::format (which was previously https://github.com/fmtlib/fmt). It's actually faster than printf family of functions.

ericcurtin:
Happy to merge on green @rhatdan ... We can always revisit in other PRs anyway :)

I'll be offline shortly

ramalama/model.py
ericcurtin: Genuinely I don't mind using both either where appropriate. flake8 and black provides us some standards, not everything has to be standardised on one solution. Like maybe sometimes one does want to explicitly represent something as a %s and that's fine with me!

PR #197, opened 2024-09-26
Turn on serve and stop test for CI/CD system

rhatdan:


rhatdan:
@edsantiago @Luap99 Any ideas on what is going on in the CI/CD system?

The following tests is failing.  The run_ramalama command is basically doing a `podman run -d` under the covers.

```
@test "ramalama serve and stop" {
    skip_if_nocontainer

    model=ollama://tiny-llm:latest
    container1=c_$(safename)
    container2=c_$(safename)

    run_ramalama serve --name ${container1} --detach ${model}
    cid="$output"
    run podman wait --condition=running $cid
    run podman inspect $cid

    run_ramalama ps
    is "$output" ".*${container1}" "list correct for for container1"

    run_ramalama containers --noheading
    is "$output" ".*${container1}" "list correct for for container1"
    run_ramalama stop ${container1}

    run_ramalama serve --name ${container2} -d ${model}
    cid="$output"
    run podman wait --condition=running $cid
    run_ramalama containers -n
    is "$output" ".*${cid:0:10}" "list correct with cid"
    run_ramalama ps --noheading
    is "$output" ".*${container2}" "list correct with cid and no heading"
    run_ramalama stop ${cid}
    run_ramalama ps --noheading
    is "$output" "" "all containers gone"
}
```

Seems like the podman wait is successful'

But the next command 
ramalama ps is failing to see the container.

ramalama ps is doing a podman ps --filter label="ramalama container" under the covers.

Bottom line this is working locally but failing 100% of the time in CI.  Also I find it strange that I don't see any of the podman commands in the output in the ci system?

edsantiago:
The first thing I see is a bunch of `run podman`. Bats is tricky, and this is one of the most evil parts of Bats: `run` does not do what you think it does. In particular, `run` happily ignores errors. 

If you have a new enough version of Bats, you can use `run -0 podman ...`, meaning, `run this command and expect 0 exit status, barfing if it fails`. Otherwise you have to do `run podman ...` and then check `$status`.

HTH

test/system/040-serve.bats
```diff
@@ -84,14 +83,14 @@ verify_begin="podman run --rm -i --label \"RAMALAMA container\" --security-opt=l
 
     run_ramalama serve -p ${port1} --detach ${model}
     cid="$output"
-    run podman wait --condition=running $cid
+    run -0 podman wait --condition=running $cid
```
Luap99, line 86: FYI, this seems rather pointless if you do `podman run -d` followed by that. The container will always be running here unless it immediately exited but even then the wait would just hang.

rhatdan: Wouldn't there be a starting stage before Running?

Luap99: Well in podman container startup logic sure but a `run -d` will only return after the container state has been changed to running (but well that just means we exec'ed the pid 1 in the container so that does not necessary mean that the container process is ready to do whatever it want).

rhatdan: Yup, but makes the check `podman wait --condition running` rather useless.

PR #200, opened 2024-09-28
Add command completion to Ramalama

rhatdan:
I had to move some things around to process in ramalama.py to make this work.

To try this out, you will need to install it and recreate your shell.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
         shell: bash
         run: |
            sudo apt-get update
-           sudo apt-get install bats bash codespell
+           sudo apt-get install bats bash codespell python3-argcomplete
            pipx install omlmd black flake8
       - name: run test
         run: make test
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ OS := $(shell uname;)
 SELINUXOPT ?= $(shell test -x /usr/sbin/selinuxenabled && selinuxenabled && echo -Z)
 PREFIX ?= /usr/local
 BINDIR ?= ${PREFIX}/bin
-SHAREDIR ?= ${PREFIX}/share/ramalama
+SHAREDIR ?= ${PREFIX}/share
 PYTHON ?= $(shell command -v python3 python|head -n1)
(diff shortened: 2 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -9,8 +9,8 @@ jobs:
         shell: bash
         run: |
            sudo apt-get update
-           sudo apt-get install bats bash
```
ericcurtin, line 12: Do we need this installed twice?

argcomplete

rhatdan: Nope, I think the issue I am seeing is because argcomplete is not in the Container.

Makefile
```diff
@@ -40,6 +40,18 @@ install-requirements:
 	pip install "huggingface_hub[cli]==0.25.1"
 	pip install "omlmd==0.1.4"
 
+.PHONY:
+install-completions:
+	install ${SELINUXOPT} -d -m 755 $(DESTDIR)${SHAREDIR}/bash-completion/completions
+	register-python-argcomplete --shell bash ramalamay > $(DESTDIR)${SHAREDIR}/bash-completion/completions/ramalama
```
ericcurtin, line 46: Typo s/ramalamay/ramalama/g

Can catch in a follow on PR if needs be

rpm/python-ramalama.spec
```diff
@@ -57,13 +58,19 @@ Provides: %{pypi_name} = %{version}-%{release}
...
+%{__make} DESTDIR=%{buildroot} PREFIX=%{_prefix} install-completions
 
 %files -n python%{python3_pkgversion}-%{pypi_name} -f %{pyproject_files}
 %license LICENSE
 %doc README.md
 %dir %{_datadir}/%{pypi_name}
 %{_datadir}/%{pypi_name}/shortnames.conf
 %{_mandir}/man1/ramalama*.1*
+%{_datadir}/bash-completion/completions/%{pypi_name}
+%{_datadir}/fish/vendor_completions.d//%{pypi_name}.fish
```
ericcurtin, line 70: Double // , very minor nit

PR #209, opened 2024-09-30
Convert to fully use setup.py for PyPi installation 

rhatdan:
Also fix up requirements.txt

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,16 @@ $ cat /usr/share/ramalama/shortnames.conf
 
 ## Install
 
+## Install via PyPi
+
+Ramalama is available via PyPi [https://pypi.org/project/ramalama](https://pypi.org/project/ramalama)
+
+```
+pip install ramalama
+```
+
+## Install by script
+
 Install RamaLama by running this one-liner (on macOS run without sudo):
 
 Linux:
--- a/bin/ramalama
+++ b/bin/ramalama
@@ -8,10 +8,10 @@ import sys
 
 
 def main(args):
-    sharedirs = ["/opt/homebrew/share/ramalama", "/usr/local/share/ramalama", "/usr/share/ramalama"]
-    syspath = next((d for d in sharedirs if os.path.exists(d)), None)
-    sys.path.insert(0, syspath)
-
+    sharedirs = ["./", "/opt/homebrew/share/ramalama", "/usr/local/share/ramalama", "/usr/share/ramalama"]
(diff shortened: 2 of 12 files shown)


rhatdan:
https://github.com/AdamWill PTAL @lsm5 
Copied from podman-python package.

Don't really know what I am doing here.

README.md
```diff
@@ -56,6 +56,16 @@ $ cat /usr/share/ramalama/shortnames.conf
 
 ## Install
 
+## Install via PyPy
+
+Ramalama is available via PyPy [https://pypi.org/project/ramalama](https://pypi.org/project/ramalama)
```
QuLogic, line 61: It's PyPI, not PyPy (the latter is a different Python implementation).

rhatdan: Thanks

QuLogic: PS, the PR title also says PyPy.

bin/ramalama
```diff
@@ -8,7 +8,7 @@
 
 
 def main(args):
-    sharedirs = ["/opt/homebrew/share/ramalama", "/usr/local/share/ramalama", "/usr/share/ramalama"]
+    sharedirs = ["./", "/opt/homebrew/share/ramalama", "/usr/local/share/ramalama", "/usr/share/ramalama"]
```
ericcurtin, line 11: I'm not sure this will work on an installed system. The way this code works is, go through the list one by one and select the first directory that exists.

"./" will always exists.

I think this would work better:

```
sharedirs = ["ramalama", "/opt/homebrew/share/ramalama", "/usr/local/share/ramalama", "/usr/share/ramalama"]
```

This way if we run from the base of the git repo, it will use "ramalama" as a syspath.

setup.cfg
```diff
@@ -0,0 +1,20 @@
+[metadata]
+name = ramalama
+description = RamaLama is a command line tool for working with AI LLM models.
+long_description = file: README.md
+long_description_content_type = text/markdown
+url = https://github.com/containers/ramalama
+platforms = any
+project_urls =
+    Bug Tracker = https://github.com/containers/ramalamay/issues
```
ericcurtin, line 9: s/ramalamay/ramalama/g

bin/ramalama
rhatdan: I changed it to look for the subdir ramalama, although

ericcurtin (approved):

setup.cfg
```diff
@@ -0,0 +1,20 @@
+[metadata]
```
QuLogic, line 1: This file should not be necessary; you can put all the config in `pyproject.toml`.

Issue #211, opened 2024-10-01
rpm: huggingface-hub, omlmd dependencies needed


lsm5:
Looks like there are new / unpackaged dependencies.

```
$ sudo dnf update python3-ramalama
Updating and loading repositories:
Repositories loaded.
Failed to resolve the transaction:
Problem: cannot install the best update candidate for package python3-ramalama-0-1.20240930152001048385.main.fc42.noarch
  - nothing provides python3.13dist(huggingface-hub[cli]) needed by python3-ramalama-0-1.20241001125821155338.main.fc42.noarch from copr:copr.fedorainfracloud.org:rhcontainerbot:podman-next
  - nothing provides python3.13dist(omlmd) needed by python3-ramalama-0-1.20241001125821155338.main.fc42.noarch from copr:copr.fedorainfracloud.org:rhcontainerbot:podman-next
You can try to add to command line:
  --no-best to not limit the transaction to the best candidates
```

rhatdan:
@lsm5 thise packages are being pulled in as Requires in the spec file, We would like to make them obtional and then have the code realize when it does not have them and suggest to the user to install them.  Or just run the commands within a container, where we will garantee they exists.

lsm5:
This issue is back on rawhide at least. I think it would be good to have a TMT test to run a basic installation check.
```
Repositories loaded.
Failed to resolve the transaction:
Problem: problem with installed package
  - cannot install the best update candidate for package python3-ramalama-0.0.14-1.20241009142409027341.main.0.ga86594a.fc42.noarch
  - nothing provides python3.13dist(huggingface-hub[cli]) needed by python3-ramalama-0-1.20241001125821155338.main.fc42.noarch from copr:copr.fedorainfracloud.org:rhcontainerbot:podman-next
  - nothing provides python3.13dist(omlmd) needed by python3-ramalama-0-1.20241001125821155338.main.fc42.noarch from copr:copr.fedorainfracloud.org:rhcontainerbot:podman-next
  - nothing provides python3.13dist(huggingface-hub[cli]) needed by python3-ramalama-0-1.20241001134012872881.main.fc42.noarch from copr:copr.fedorainfracloud.org:rhcontainerbot:podman-next
  - nothing provides python3.13dist(omlmd) needed by python3-ramalama-0-1.20241001134012872881.main.fc42.noarch from copr:copr.fedorainfracloud.org:rhcontainerbot:podman-next
[...]

PR #212, opened 2024-10-01
Make huggingface[cli]  and omlmd optional

rhatdan:
Fixes: https://github.com/containers/ramalama/issues/211

### DIFF
--- a/ramalama/huggingface.py
+++ b/ramalama/huggingface.py
@@ -1,7 +1,13 @@
 import os
-from ramalama.common import run_cmd, exec_cmd
+from ramalama.common import run_cmd, exec_cmd, perror
 from ramalama.model import Model
 
+missing_hugginface="""
+Huggingface models requires the huggingface-cli and tldm modules.
+These modules can be installed via PyPi tools like pip, pip3, pipx or via
+distribution package managers like dnf or apt. Example:
+pip install huggingface_hub[cli] tldm
+"""
 
 def download(store, model, directory, filename):
     return run_cmd(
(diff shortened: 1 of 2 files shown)


lsm5:
as long as installation from the packit job rpm works, lgtm from me. I'll try this out in a bit.

lsm5 (approved):
rpm update with packit rpm works fine

```
sudo dnf install -y dnf-plugins-core
sudo dnf copr enable packit/containers-ramalama-212
sudo dnf install -y python3-ramalama-0-1.20241001151639838794.pr212.fc42.noarch
```

LGTM

rhatdan:
These requires are coming from the pyproject.toml, @lsm5 do you know if there is a way we could add them there but remove them from the spec requires.

It would be nice if we could get them automatically when people install via PyPi

ramalama/huggingface.py
```diff
@@ -1,7 +1,12 @@
 import os
-from ramalama.common import run_cmd, exec_cmd
+from ramalama.common import run_cmd, exec_cmd, perror
 from ramalama.model import Model
 
+missing_hugginface="""
+Huggingface models requires the huggingface-cli and tldm modules.
+These modules can be installed via:
+pip install huggingface_hub[cli] tldm
```
ericcurtin, line 8: I don't think we need tldm, huggingface_hub[cli] should be fine

ericcurtin:
```
huggingface-cli: command not found
```

would have been fine too. A user may want to install via pip, pip3, pipx, dnf, apt. Maybe we should let them decide how to install. Some people don't recommend pip.

rhatdan: The issue is we don't know if those packages exists via other means.  I can add, 
`or modules might be available via pip3, pipx, dnf, apt or other install tools.`

ericcurtin (approved):

ramalama/huggingface.py
ericcurtin: approved nonetheless :)

PR #223, opened 2024-10-02
Fix CI

ericcurtin:
Fix GitHub yaml

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,8 +20,9 @@ jobs:
       - uses: actions/checkout@v4
       - name: install golang
         shell: bash
-        run: brew install go
-	run: make install-requirements
+        run: |
+           brew install go
+           make install-requirements
       - name: Run a one-line script
         shell: bash
         run: make validate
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -20,8 +20,9 @@ jobs:
       - uses: actions/checkout@v4
       - name: install golang
         shell: bash
-        run: brew install go
-	run: make install-requirements
+        run: |
+           brew install go
+           make install-requirements
```
rhatdan, line 25: This can only be done on Linux right now?

ericcurtin:
@rhatdan when github yaml is wrong, it doesn't trigger CI at all, it's happened a few times, so all builds will be green as CI hasn't been run

.github/workflows/ci.yml
ericcurtin: Not if we use pipx...

Ubuntu 24.04 and macOS bail when using pip and recommend pipx

ericcurtin:
@rhatdan you might be better off taking up this PR, you are most familiar with the recent changes

.github/workflows/ci.yml
ericcurtin: This is why CI stopped running FWIW:

run:

followed by

run:

PR #224, opened 2024-10-02
Mount distinfo into container if it exists

rhatdan:
Currently PyPi and make install is creating a distinfo file which is needed in order to retrieve the version of ramalama in use inside of the container.

Also handle shortnames.conf file created via PyPi which is installed in $HOME/.local/share/ramalama/shortnames.conf


ericcurtin (approved):
I don't like merging without CI, but this seems good

PR #228, opened 2024-10-03
Creating a hack/rm.sh script

ericcurtin:
Want to try and install via pipx and wrote a script to remove old install. Also converted python install script back to shell, the python experience in a new mac is just not nice in comparison to bash.

### DIFF
--- a/bin/ramalama
+++ b/bin/ramalama
@@ -1,21 +1,59 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 
-import argcomplete
+import glob
 import os
 import errno
 import subprocess
 import sys
 
+def add_pipx_venvs_to_syspath():
+    """
+    Adds all available pipx virtual environments' site-packages directories to sys.path.
+    This function looks for any venv in ~/.local/pipx/venvs and appends their site-packages directory to sys.path.
+    """
+    python_version = f'{sys.version_info.major}.{sys.version_info.minor}'
+    pipx_venvs_path = os.path.expanduser(f'~/.local/pipx/venvs/*/lib/python{python_version}/site-packages')
+    matched_paths = glob.glob(pipx_venvs_path)
+    if not matched_paths:
+        return
+
+    should_we_return = False
+    for path in matched_paths:
+        should_we_return = True
+        sys.path.insert(0, path)
+
+    if should_we_return:
+        return
+
+    pipx_venvs_path = os.path.expanduser('~/.local/pipx/venvs/*/lib/python*/site-packages')
+    matched_paths = glob.glob(pipx_venvs_path)
+    if not matched_paths:
+        return
+
+    for path in matched_paths:
+        sys.path.insert(0, path)
 
 def main(args):
     sharedirs = ["./", "/opt/homebrew/share/ramalama", "/usr/local/share/ramalama", "/usr/share/ramalama"]
     syspath = next((d for d in sharedirs if os.path.exists(d+"/ramalama/cli.py")), None)
     if syspath:
         sys.path.insert(0, syspath)
-    import ramalama
+
+    add_pipx_venvs_to_syspath()
+    try:
+        import ramalama
+    except:
+        print(f"ramalama module not found in sys.path: {sys.path}", file=sys.stderr)
+        raise
 
     parser, args = ramalama.init_cli()
-    argcomplete.autocomplete(parser)
+
+    # if autocomplete doesn't exist, just do nothing, don't break
+    try:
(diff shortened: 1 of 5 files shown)


rhatdan:
You removed install.py and it is referenced in the README.md

ericcurtin:
> You removed install.py and it is referenced in the README.md

It seems silly, but this is deliberate, the stable branch still has the python version

ramalama/version.py
```diff
@@ -5,10 +5,14 @@
 
 def version():
     try:
-        return importlib.metadata.version("ramalama")
+        ver = importlib.metadata.version("ramalama")
```
ericcurtin, line 8: I have to head offline @rhatdan but I'm having diffculties with this version function, if you have an idea what would help, let me know, otherwise I'll poke at it again tomorrow.

bin/ramalama
```diff
@@ -1,21 +1,47 @@
...
 
 def main(args):
-    sharedirs = ["./", "/opt/homebrew/share/ramalama", "/usr/local/share/ramalama", "/usr/share/ramalama"]
+    sharedirs = ["./", "/opt/homebrew/share", "/usr/local/share", "/usr/share"]
     syspath = next((d for d in sharedirs if os.path.exists(d+"/ramalama/cli.py")), None)
     if syspath:
         sys.path.insert(0, syspath)
-    import ramalama
+
+    sys.path.insert(0, get_pipx_dir())
```
rhatdan, line 29: If there are non found we should not do this.

bin/ramalama
```diff
@@ -1,21 +1,47 @@
...
 
     parser, args = ramalama.init_cli()
-    argcomplete.autocomplete(parser)
+
+    # if autocomplete doesn't exist, just do nothing, don't break
+    try:
+        import argcomplete
+        argcomplete.autocomplete(parser)
+    except Exception:
+        None
```
rhatdan, line 44: We need a way to debug this, perhaps add a debugging level to RamaLama.

bin/ramalama
```diff
@@ -1,21 +1,47 @@
...
+    if os.path.exists(argcompletion_path):
+        return argcompletion_path
+
+    # Use glob to find the matching directory for any Python version
+    path_pattern = '~/.local/pipx/venvs/argcomplete/lib/python*/site-packages'
+    matched_paths = glob.glob(path_pattern)
+
+    # Select the first matching path (if multiple matches are found, you can refine this)
+    if matched_paths and os.path.exists(matched_paths[0]):
+        return matched_paths[0]
```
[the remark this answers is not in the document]
ericcurtin: I'm guessing the default in python is to return False or None if we get to the end of the function without a return.

But an explicit:

`return ""`

at the end of the function would do no harm just in case.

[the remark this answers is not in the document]
ericcurtin: It would be nice if we could enforce returns in our functions using some python static analyzer like C/C++ does with modern compliers and that warning/error for functions that don't return in all cases.

bin/ramalama
ericcurtin: SGTM, I'd like to catch this in a new PR though, really want to mimise the time macOS installs are broken.

bin/ramalama
ericcurtin: Done

bin/ramalama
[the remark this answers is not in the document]
ericcurtin: Done

bin/ramalama
```diff
@@ -1,21 +1,51 @@
...
-import argcomplete
+import glob
 import os
 import errno
 import subprocess
 import sys
 
+def get_pipx_dir():
+    python_version = f'{sys.version_info.major}.{sys.version_info.minor}'
+    argcompletion_path = os.path.expanduser(f'~/.local/pipx/venvs/argcomplete/lib/python{python_version}/site-packages')
```
rhatdan, line 11: Should we just add
I think this is wrong and would only find argcomplete, and not other packages installed by pipx? Should we just add .local/pipx/venvs to sys.paths?

ericcurtin: argcomplete is our only dependency, we call CLI versions of omlmd and huggingface so the python imports directory is irrelevant

ericcurtin: We'd have to loop though and add every:

~/.local/pipx/venvs/*/lib/python{python_version}/site-packages'

dir, which might be worth it

ericcurtin: Added a much more generic version

rhatdan:
LGTM

PR #229, opened 2024-10-03
use basename to avoid attempting to make invalid volume names

jhjaggars:
When running the base command like:
```
bin/ramalama
```

I got the following error:
```
rror: creating named volume "bin/ramalama": running volume create option: names must match [a-zA-Z0-9][a-zA-Z0-9_.-]*: invalid argument
```

This PR takes the basename of the first argument to force this volume to be created with the same name as the script itself.


ericcurtin (approved):
Thanks @jhjaggars

PR #231, opened 2024-10-03
turn on --nocontainer tests

rhatdan:


rhatdan:
@lsm5 any idea why the centos builds are failing for building rpms.

rhatdan:
Since this is turning on more tests, and now passes, merging.

lsm5:
> @lsm5 any idea why the centos builds are failing for building rpms.

some bad dnf metadata got synced to all mirrors. It's been fixed but mirror propagation could take a while.

PR #232, opened 2024-10-03
Replace install.py with install.sh and add rm.sh

rhatdan:
Creating a hack/rm.sh script

Want to try and install via pipx and wrote a script to remove old install. Also converted python install script back to shell, the python experience in a new mac is just not nice in comparison to bash.


rhatdan:
Replaces: https://github.com/containers/ramalama/pull/228

ericcurtin (approved):

PR #238, opened 2024-10-06
We now have a fedoraproject.org Matrix

ericcurtin:
For people to reach out

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,6 +238,7 @@ so if you like this tool, give some of these repos a :star:, and hey, give us a 
 
 ## Community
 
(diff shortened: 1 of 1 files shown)


ericcurtin:
Do we want to announce in the Discord that Matrix will become the primary place for collaboration... Or just leave things be organically?

rhatdan:
Can we link them together?

ericcurtin:
I think @smooge said we can't, but just tagging him to confirm

smooge:
In order to link them someone needs to run and maintain a bridge server. It can turn into a full time job

rhatdan:
@smooge so you think we should just move to Matrix and drop Discord?

PR #240, opened 2024-10-07
Add rocm Containerfile

ericcurtin:
To build rocm Container Image for use with AMD GPUs

### DIFF
--- a/container-images/cuda/Containerfile
+++ b/container-images/cuda/Containerfile
@@ -1,4 +1,4 @@
-FROM nvidia/cuda:12.6.1-devel-ubi9
+FROM docker.io/nvidia/cuda:12.6.1-devel-ubi9
 
 # renovate: datasource=github-releases depName=huggingface/huggingface_hub extractVersion=^v(?<version>.*)
 ARG HUGGINGFACE_HUB_VERSION=0.25.2
--- a/ramalama/model.py
+++ b/ramalama/model.py
@@ -27,7 +27,7 @@ class Model:
 
     def __init__(self, model):
         self.model = model
-        if sys.platform == "darwin":
+        if sys.platform == "darwin" or os.getenv("HIP_VISIBLE_DEVICES") or os.getenv("CUDA_VISIBLE_DEVICES"):
             self.common_params += ["-ngl", "99"]
 
     def login(self, args):
--- a/container-images/rocm/Containerfile
+++ b/container-images/rocm/Containerfile
@@ -1,0 +1,37 @@
+FROM quay.io/ramalama/ramalama:latest
+
+RUN /usr/bin/python3 --version
+
+ARG ROCM_VERSION=6.2.2
+ARG AMDGPU_VERSION=6.2.2
+
+COPY amdgpu.repo /etc/yum.repos.d/
+COPY rocm.repo /etc/yum.repos.d/
+
+RUN dnf config-manager --add-repo https://mirror.stream.centos.org/9-stream/AppStream/$(uname -m)/os/
+RUN curl -o /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-Official http://mirror.centos.org/centos/RPM-GPG-KEY-CentOS-Official
+RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-Official
+
+RUN dnf install -y rocm-dev hipblas-devel rocblas-devel && \
+    dnf clean all && \
+    rm -rf /var/cache/*dnf*
+
+RUN git clone https://github.com/ggerganov/llama.cpp && \
+    cd llama.cpp && \
+    git reset --hard ${LLAMA_CPP_SHA} && \
+    cmake -B build -DCMAKE_INSTALL_PREFIX:PATH=/usr -DGGML_CCACHE=0 \
+      -DGGML_HIPBLAS=1 && \
+    cmake --build build --config Release -j $(nproc) && \
+    cmake --install build && \
+    cd / && \
+    rm -rf llama.cpp
+
+RUN git clone https://github.com/ggerganov/whisper.cpp.git && \
+    cd whisper.cpp && \
+    git reset --hard ${WHISPER_CPP_SHA} && \
+    make -j $(nproc) GGML_HIPBLAS=1 && \
+    mv main /usr/bin/whisper-main && \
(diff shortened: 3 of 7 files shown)


container-images/ramalama/latest-rocm/Containerfile
```diff
@@ -0,0 +1,54 @@
+FROM quay.io/ramalama/ramalama:latest
+
+RUN /usr/bin/python3 --version
+
+ARG ROCM_VERSION=6.2.2
+ARG AMDGPU_VERSION=6.2.2
+
+RUN bash -c 'cat > /etc/yum.repos.d/amdgpu.repo <<EOF
```
rhatdan, line 8: Lets create these files on the host system and copy them in rather then attempting to use <<EOF which is not supported on older versions of podman/buildah and maybe Docker.

container-images/ramalama/latest-rocm/Containerfile
```diff
@@ -0,0 +1,54 @@
...
+[amdgpu]
+name=amdgpu
+baseurl=https://repo.radeon.com/amdgpu/$AMDGPU_VERSION/rhel/9.4/main/x86_64/
+enabled=1
+priority=50
+gpgcheck=1
+gpgkey=https://repo.radeon.com/rocm/rocm.gpg.key
+EOF'
+
+RUN bash -c 'cat > /etc/yum.repos.d/rocm.repo <<EOF
```
rhatdan, line 18: Ditto:

COPY rocm.repo /etc/yum..repos.d/

russellb:
For your disk space issue, here's an action you can copy to free up some space. It may get you enough. https://github.com/instructlab/instructlab/blob/main/.github/actions/free-disk-space/action.yml

tiran:
There is an easier, faster way if you just need more disk space for building a container with Docker. The `/mnt` partition has plenty of disk space.

https://github.com/tiran/instructlab-containers/blob/main/.github/workflows/containers.yaml#L82-L93

```shell
                  sudo mkdir /mnt/docker
                  jq '. + {"data-root": "/mnt/docker"}' < /etc/docker/daemon.json | tee /tmp/daemon.json
                  sudo mv /tmp/daemon.json /etc/docker/daemon.json
                  sudo systemctl restart docker.service
```

ramalama/model.py
```diff
@@ -13,6 +13,9 @@ class Model:
 
     def __init__(self, model):
         self.model = model
+        amd_card = False
+        if available("rocm-smi"):
+--showmeminfo vram --json
```
rhatdan, line 18: This looks like a tab issue?

ericcurtin: I didn't mean to push this, sorry

rhatdan:
@ericcurtin you need to change the location of the docker storage.

ericcurtin:
This is the logic I plan on using to decide whether to pull a ROCm image or not FWIW:

```
#!/usr/bin/python3

import glob

i = 0
gpu_num = False
gpu_bytes = 0

# Loop through each file matching the pattern
for filepath in sorted(glob.glob('/sys/bus/pci/devices/*/mem_info_vram_total')):
    # Open the file and read its content
    with open(filepath, 'r') as file:
        content = int(file.read())
        if content > 1073741824 and content > gpu_bytes:
            gpu_bytes = content
            gpu_num = i

    i += 1

print(gpu_num)
print(gpu_bytes)
```

If we have an AMD/ROCm card with greater than 1GB use the AMD/ROCm card with the most VRAM. On the system I am testing with theirs a measly 512MB integrated graphics card on the AMD CPU, but an external graphics card with 8G VRAM...

ramalama/model.py
```diff
@@ -13,7 +13,7 @@ class Model:
 
     def __init__(self, model):
         self.model = model
-        if sys.platform == "darwin":
+        if sys.platform == "darwin" or os.getenv("HIP_VISIBLE_DEVICES") or os.getenv("CUDA_VISIBLE_DEVICES"):
```
ericcurtin, line 16: When using a ROCm or CUDA GPU, in the case where there are multiple GPUs you can select one specifically using HIP_VISIBLE_DEVICES or CUDA_VISIBLE_DEVICES environment variable.

How do we select a GPU in vulkan/kompute @slp , just curious how we will adapt this for the kompute case...

rhatdan: Should we allow users to select the GPU on the command line as well?

ericcurtin: Yeah there's lots of options like that we should/could add... Select a specific GPU type (AMD/Nvidia), select a specific GPU framework (ROCm/Nvidia/Vulkan/Kompute/OpenCL/etc.), select a specific GPU (AMD GPU 1, AMD GPU 2)... I just wonder should we do this up front or wait until someone requests the feature...

What we could also do is... If HIP_VISIBLE_DEVICES or CUDA_VISIBLE_DEVICES is already set externally don't look for the GPU with the most VRAM on the system...|

That's kinda the simple heuristic for now for auto-detecting the primary GPU, if we have a GPU that has more than 1GB VRAM, run on the GPU with the most VRAM...

But this PR only does it for AMD GPUs for now...

rhatdan: I was thinking they would actually give us the device(s)
--gpu /dev/dri/card0 or something like that.

ericcurtin: The problem is, it's not always straightforward to infer what /dev/dri/card0 is Nvidia/AMD or something else

ericcurtin: /dev/dri/card0 is not deterministic also, so that should not be used in a quadlet for example. I fixed that bug in multiple graphics frameworks as a recent kernel change made card0, card1, etc. less deterministic than ever...

https://github.com/libsdl-org/SDL/pull/5865
https://swiftshader.googlesource.com/SwiftShader/+/afbfd351374d93484f8eecf532afbd0dac084865

container-images/cuda/Containerfile
```diff
@@ -1,4 +1,4 @@
-FROM nvidia/cuda:12.6.1-devel-ubi9
```
ericcurtin, line 1: podman doesn't seem to like the shortened version, I think this was tested on docker initially

ericcurtin:
This is ready @rhatdan will be reworked a bit when the Kompute stuff gets merged in

ericcurtin:
containah pushed https://quay.io/repository/ramalama/rocm

rhatdan:
LGTM

PR #257, opened 2024-10-08
Fix subdir tree of Containerfiles

rhatdan:
Subtree structure was adding nothing to information about Containerfiles.

I see us ending up with 4 subdirs

nocpu
cuda
rocm
gaudi

And then perhaps adding others in the future.


ericcurtin:
nocpu should be either nogpu or cpu.

nocpu doesn't make sense.

Are we sure we want to do gaudi right now? vulkan makes more sense to me at least...

But maybe you know something about gaudi devices I don't

rhatdan:
nocpu was a typo, should have been cpuonly.

Changing to vulcan as well.

ericcurtin:
s/vulcan/vulkan/g

.github/workflows/latest.yml
```diff
@@ -50,7 +50,7 @@ jobs:
             image: quay.io/ramalama/ramalama
             tags: latest
             containerfiles: |
-              container-images/ramalama/latest/Containerfile
+              container-images/cpuonly/Containerfile
```
ericcurtin, line 53: The intent for ramalama/latest was that if you replace the '/' with ':' we have a natural structure to represent name:tag syntax

We can change that but what's the new technique to define name and tag?

rhatdan: I think it is the same, just that we always create in the ramalama repository and we are always creating latest version.

ericcurtin: Just to define a format, would it be something like

llama.cpp-cpuonly:latest
llama.cpp-cuda:latest
llama.cpp-rocm:latest
llama.cpp-vulkan:latest
vllm-cpuonly:latest
vllm-cuda:latest
vllm-rocm:latest
vllm-vulkan:latest

or in other words:

quay.io/ramalama/runtime-accelerator:tag

?

ericcurtin: I was thinking this:

quay.io/ramalama/ramalama:runtime-accelerator-tag

and putting all the details in the tag, just because, as far as I know, every time we need a need a repository we have to start clicking in quay.io and that can make introducing a new runtime for example a pain in terms of quay.io clicking. But maybe I'm just bad at the efficient tricks around pushing new repos. OCI registry management isn't exactly a strength of mine...

rhatdan: Yes, I think we just do:

llama.cpp-cpuonly:latest
llama.cpp-cuda:latest
llama.cpp-rocm:latest
llama.cpp-vulkan:latest
vllm-cpuonly:latest
vllm-cuda:latest
vllm-rocm:latest
vllm-vulkan:latest

I would think more about combining vllm and lama.cpp inside of the same image, if the size is not huge to limit the number of images we need to support.

ericcurtin (approved):
Will need follow on PRs... Thanks for doing this gradually... Will need to ensure images are pushed to quay.io in the new format before changing a lot of the rest.

ericcurtin:
This has simple conflicts to fix also, it's the renovate update I guess...

PR #261, opened 2024-10-08
Add -image option to override defaut OCI Image

rhatdan:
By default ramalama attempts to use the best AI image based on GPU on the local system. The --image option allows users to override the default.


ericcurtin (approved):
This will be useful for testing, need to fix a simple conflict

ericcurtin (approved):

PR #269, opened 2024-10-09
Add tdqm as a requirement

rhatdan:
I think this will break the RPM build requirements but it is needed to may pip install work properly on Mac.

we need to figure a way to fix the RPM to not use the requirements defined in the pyproject. package.


rhatdan:
Going to merge this to get CI/CD system back running.

PR #271, opened 2024-10-09
[WIP] Supposedly this works on a Mac

rhatdan:


slp:
Installing packages from brew behind the user's back feels... weird. Why not ship ramalama itself through brew? That way it can depend on llama.cpp, python-argparse, and whatever other dependencies.

rhatdan:
Yes I would like to get that done as well.  I am thinking of raising an error when users try to run lama_server if it does not install and telling them to install it or run the work load within a container.

PR #277, opened 2024-10-10
Change directory to fix CI

ericcurtin:
The old locations was messing with CI, all the other docs are man pages.


lsm5 (approved):
copr jobs are successful. So, LGTM

PR #278, opened 2024-10-10
Fix hack/man-page-checker in CI

ericcurtin:
Formatting problems


ericcurtin:
@josiahbryan I've had to partially remove your PR because it doesn't agree with out man page builders. I was able to keep the main part of your pull request though, the:

## REST API ENDPOINTS

section

ericcurtin:
Merging to fix CI, before it infects any more branches

PR #279, opened 2024-10-10
Remove useless `cli run` parameter `--prompt`

arouene:
`--prompt` arg is not used, if ARGS exists it will be used as a prompt.

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -351,10 +351,9 @@ def _name():
 
 def run_parser(subparsers):
     parser = subparsers.add_parser("run", help="run specified AI Model as a chatbot")
-    parser.add_argument("--prompt", dest="prompt", action="store_true", help="modify chatbot prompt")
     parser.add_argument("-n", "--name", dest="name", help="name of container in which the Model will be run")
     parser.add_argument("MODEL")  # positional argument
-    parser.add_argument("ARGS", nargs="*", help="additional options to pass to the AI Model")
+    parser.add_argument("ARGS", nargs="*", help="Overrides the default prompt")
     parser.set_defaults(func=run_cli)
 
 
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -353,7 +353,7 @@ def run_parser(subparsers):
     parser = subparsers.add_parser("run", help="run specified AI Model as a chatbot")
     parser.add_argument("-n", "--name", dest="name", help="name of container in which the Model will be run")
     parser.add_argument("MODEL")  # positional argument
-    parser.add_argument("ARGS", nargs="*", help="Overrides the default prompt")
+    parser.add_argument("ARGS", nargs="*", help="Overrides the default prompt, and the output is returned without entering the chatbot")
     parser.set_defaults(func=run_cli)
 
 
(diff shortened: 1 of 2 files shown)


ericcurtin:
@arouene needs a rebase and a re-push to run through CI properly... Would like to see a green build here... For example we may have tests that break because of removal of this, which may need altering

rhatdan:
You also need to remove --prompt definition in docs/ramalama-run.1.md man page.

ramalama/cli.py
```diff
@@ -351,10 +351,9 @@ def _name():
 
 def run_parser(subparsers):
     parser = subparsers.add_parser("run", help="run specified AI Model as a chatbot")
-    parser.add_argument("--prompt", dest="prompt", action="store_true", help="modify chatbot prompt")
     parser.add_argument("-n", "--name", dest="name", help="name of container in which the Model will be run")
     parser.add_argument("MODEL")  # positional argument
-    parser.add_argument("ARGS", nargs="*", help="additional options to pass to the AI Model")
+    parser.add_argument("ARGS", nargs="*", help="Overrides the default prompt")
```
ericcurtin, line 356: It actually does more than this, it turns off conversation mode, so a response is given and ramalama is exited.

We also may want a --system-prompt like CLI option in future to tell the model how to behave but that's kinda irrelevant to this PR.

@rhatdan introduced this option, so may have opinions on the description.

rhatdan: The goal of this option was to have a command executed against the AI Model and then exit.  

Something like the following:

ramalama run MODEL Generate a test framework for python > /tmp/output

As opposed to standard where we end up in a chatbot.

So if I give an argument to the chatbot, process it and exit.  If I don't then go into interactive mode.

Handling this in a different manner would be fine with me as well.

arouene: > We also may want a --system-prompt like CLI option in future to tell the model how to behave but that's kinda irrelevant to this PR.

I saw an issue in llama.cpp that consider removing `--system-prompt`: https://github.com/ggerganov/llama.cpp/issues/9811

arouene: I changed the help message for ARGS, to be a bit more specific

arouene: What could be neat, but in another PR. It's accepting stdin like so : `cat file.py | ramalama run model "check this following code"`

ericcurtin: When I say system prompt, I mean this in llama-cli: `-p "You are a helpful assistant"`. The -p option is effectively a system prompt when used like this.

ericcurtin: I would definately appreciate if this feature was added in a PR @arouene :

> What could be neat, but in another PR. Is accepting stdin like so : cat file.py | ramalama run model "check this following code"

we should also consider this syntax if it makes implementation/usage simpler:

cat file.py | ramalama run model "check this following code" -

Some tools use '-' to denote, read from stdin

ericcurtin (approved):

ramalama/cli.py
rhatdan: I think it is fine to have ramalama act differently if STDIN is not a TTY and read from stdin.

PR #281, opened 2024-10-10
Report error when AI Service is not available on host

rhatdan:
Currently we throw an exception if lama.cpp is not installed on the host This PR should make it easier for users to discover how to install the services.

### DIFF
--- a/ramalama/model.py
+++ b/ramalama/model.py
@@ -1,6 +1,20 @@
 import os
 import sys
-from ramalama.common import container_manager, exec_cmd, default_image
+from ramalama.common import container_manager, exec_cmd, default_image, in_container
+
+
+file_not_found = """\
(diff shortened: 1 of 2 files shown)


ramalama/model.py
```diff
@@ -1,6 +1,20 @@
 import os
 import sys
-from ramalama.common import container_manager, exec_cmd, default_image
+from ramalama.common import container_manager, exec_cmd, default_image, in_container
+
+
+file_not_found = """\
+RamaLama requires the "%s" command to be installed on the host when running with --nocontainer.
```
ericcurtin, line 7: I like that we just said %s command is to be installed (and not suggested a particular package manager). I think in similar python related messages we suggested apt, dnf, pip, pipx (and that's leaving out other popular ones like zypper)

Better to let the user decide what's best, maybe that's `make; make install` for them, who knows

PR #282, opened 2024-10-10
Add pipx paths to PATH environment variable

rhatdan:
When installing via pipx, requirement executables of RamaLama are
stored in the venvs/ramalama/bin directory.  Have to modify the
system PATH so ramalama can find them.

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -11,7 +11,15 @@ import sys
 import time
 
 from ramalama.huggingface import Huggingface
-from ramalama.common import in_container, container_manager, exec_cmd, run_cmd, default_image, find_working_directory, perror
+from ramalama.common import (
+    in_container,
+    container_manager,
+    exec_cmd,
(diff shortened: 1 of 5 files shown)


ramalama/cli.py
```diff
@@ -405,7 +416,7 @@ def _stop_container(args, name):
 
     conman_args = [conman, "stop", "-t=0"]
     if args.ignore:
-        conman_args += [ "--ignore", str(args.ignore)]
+        conman_args += ["--ignore", str(args.ignore)]
```
ericcurtin, line 419: Surprised flake8/black didn't pick this up

ericcurtin (approved):

ramalama/cli.py
rhatdan: I see some of our CI/CD system not working properly.  I have a failing that these are causing flake8/black to fail, but we are not catching the errors in ci/cd system

PR #284, opened 2024-10-11
Ramalama to RamaLama

ericcurtin:
Just a consistency change


.codespelldict
```diff
@@ -1,6 +1,7 @@
 hugginface->huggingface
 ramalamay->ramalama
 ramamlama->ramalama
+Ramalama->RamaLama
```
ericcurtin, line 4: Not sure if codespell does case-sensitive fixes

rhatdan:
LGTM

PR #287, opened 2024-10-11
Disable automatic detach for the serve command

### LINKED ISSUE #285: Serve command fails silently
Hello,

Running the `serve` command with a wrong model name outputs a container ID that dies instantly without a reason.
Here is an example reproducer:

```ShellSession
$ ramalama serve huggingface://ibm-granite/granite-7b-instruct
6aa6dc6cca217b28656990f69a7c602c28c40f64289d8cdb5c55e6ecccdc0828
$ ramalama ps
CONTAINER ID  IMAGE       COMMAND     CREATED     STATUS      PORTS       NAMES
$
```

I tried to add an helpful message by calling the pull command before serve, but it looks like there is another issue in the glob `f"{args.store}/models/*/{model}"` because the Model class doesn't have a str method, resulting in the path to never match because it looks like:

```raw
ramalama/models/*/<ramalama.huggingface.Huggingface object at 0x7fa1c6e659a0>
```

How would you recommend to fix that issue?

TristanCacqueray:
Fix #285

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -376,7 +376,7 @@ def run_cli(args):
 def serve_parser(subparsers):
     parser = subparsers.add_parser("serve", help="serve REST API on specified AI Model")
     parser.add_argument(
-        "-d", "--detach", action="store_true", default=True, dest="detach", help="run the container in detached mode"
+        "-d", "--detach", action="store_true", dest="detach", help="run the container in detached mode"
     )
     parser.add_argument("-n", "--name", dest="name", help="name of container in which the Model will be run")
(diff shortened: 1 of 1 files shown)


ericcurtin:
@TristanCacqueray 

Please sign your commit and re-push with:

```
git commit --amend -s
git push -f
```

The DCO build failed, which means the commit wasn't signed.

ericcurtin (approved):
Tested, this runs great, might be best to run by Dan though

rhatdan:
I am fine with hit, I was just thinking the most common case would be to run the service in detached mode, but users can do this via bash or use the --detach flag.

rhatdan:
Thanks @TristanCacqueray

PR #290, opened 2024-10-11
Remove requirements.txt, pyproject.toml handles it

rhatdan:
fix huggingface_hub since this package exists as well.

make install  use --no-deps


rhatdan:
@lsm5 PTAL

PR #291, opened 2024-10-11
On Mac, run in container ...

rhatdan:
if podman is configured with krunkit,

or if Docker is installed.

### DIFF
--- a/docs/ramalama.1.md
+++ b/docs/ramalama.1.md
@@ -26,6 +26,11 @@ behaviour.
 
 RamaLama supports multiple AI model registries types called transports. Supported transports:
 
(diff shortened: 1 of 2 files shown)


ericcurtin (approved):

rhatdan:
Have you built one?  IE Can you push one there?

ericcurtin:
Unfortunately I can't by end of day, would have to set up podman farm

ericcurtin:
We really have to get multi-arch container builds automated

ericcurtin:
Even for contributors, it's not reasonable to expect them to do test builds for multiple CPU arches

PR #292, opened 2024-10-11
Take into account pipx version of shortnames.conf

ericcurtin:
Installed in the ~/.local/pipx/venvs/ramalama location


ericcurtin:
There's a possibility we could get ramalama to re-exec itself early in a venv also... But this works for now...

rhatdan:
LGTM

PR #293, opened 2024-10-12
Enable resumable downloads in ollama.py

swarajpande5:
Resolves #258 

Example run: 
```
(venv) spande-mac:ramalama (resumable-ollama-pulls*) $ ./bin/ramalama pull ollama://llama3.2:1b
Warning: podman needs to be configured to use krunkit for AI Workloads,
running without containers

Pulling rary/llama3.2/1b: 100% ▕####################▏ 861/861 3.40MB/s 00:00
Pulling d54c02d2e1b1118d: 100% ▕####################▏ 485/485 547kB/s 00:00
Pulling bed452548fa88d45:   7% ▕#4                  ▏ 92.9M/1.23G 3.68MB/s 05:32^C 


(venv) spande-mac:ramalama (resumable-ollama-pulls*) $ ./bin/ramalama pull ollama://llama3.2:1b
Warning: podman needs to be configured to use krunkit for AI Workloads,
running without containers

Manifest file /Users/spande/.local/share/ramalama/repos/ollama/manifests/https://registry.ollama.ai/library/llama3.2/1b seems corrupted, deleting and retrying download.
Pulling rary/llama3.2/1b: 100% ▕####################▏ 861/861 1.90MB/s 00:00
File https://registry.ollama.ai/v2/library/llama3.2/blobs/sha256:4f659a1e86d7f5a33c389f7991e7224b7ee6ad0358b53437d54c02d2e1b1118d already fully downloaded.
Pulling bed452548fa88d45:  10% ▕#9                  ▏ 122M/1.23G 3.44MB/s 05:47^C
```

PR #294, opened 2024-10-13
Add checksum verification in ollama.py

swarajpande5:
Adds checksum check after model blobs pulls in ollama.py
Reference: #293 

Example run: 
```
(venv) spande-mac:ramalama (checksum-verification*) $ ./bin/ramalama pull tiny
Warning: podman needs to be configured to use krunkit for AI Workloads,
running without containers

Pulling tinyllama/latest: 100% ▕####################▏ 852/852 2.62MB/s 00:00
File https://registry.ollama.ai/v2/library/tinyllama/blobs/sha256:6331358be52a6ebc2fd0755a51ad1175734fd17a628ab5ea6897109396245362 already fully downloaded.
Pulling 83e42c043d6c7816: 100% ▕####################▏ 608M/608M 3.54MB/s 00:00
Checksum verified for blob /Users/spande/.local/share/ramalama/repos/ollama/blobs/sha256:2af3b81862c6be03c769683af18efdadb2c33f60ff32ab6f83e42c043d6c7816.
```

@ericcurtin If you want checksum verification to be silent (i.e. not printing the "Checksum verified for blob" string), I can amend the commit.

### DIFF
--- a/ramalama/ollama.py
+++ b/ramalama/ollama.py
@@ -1,7 +1,7 @@
 import os
 import urllib.request
 import json
-from ramalama.common import run_cmd
+from ramalama.common import run_cmd, verify_checksum
 from ramalama.model import Model
 
 bar_format = "Pulling {desc}: {percentage:3.0f}% ▕{bar:20}▏ {n_fmt}/{total_fmt} {rate_fmt} {remaining}"
@@ -67,7 +67,6 @@ def pull_manifest(repos, manifests, accept, registry_head, model_tag):
     # Attempt to redownload the manifest if already exists,
     # otherwise json.JSONDecodeError exception is thrown when reading the manifest
     if os.path.exists(manifests):
-        print(f"Manifest file {manifests} seems corrupted, deleting and retrying download.")
         os.remove(manifests)
 
     download_file(url, manifests, headers=headers)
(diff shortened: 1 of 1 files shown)


ramalama/ollama.py
```diff
@@ -87,6 +86,16 @@ def pull_blob(repos, layer_digest, accept, registry_head, models, model_name, mo
...
 
+    # Verify checksum after downloading the blob
+    if not verify_checksum(layer_blob_path):
+        print(f"Checksum mismatch for blob {layer_blob_path}, retrying download...")
+        os.remove(layer_blob_path)
+        download_file(url, layer_blob_path, headers=headers)
+        if not verify_checksum(layer_blob_path):
+            raise ValueError(f"Checksum verification failed for blob {layer_blob_path}")
+    else:
+        print(f"Checksum verified for blob {layer_blob_path}.")
```
rhatdan, line 97: This is going to be noisy.  I think humans will just assume what they pulled was correct, and this message will not add anything.

ericcurtin:
Please be silent... But we can have these messages under a debug or verbose option if needs be... Also we only need one progress bar... The model one... The other http request/responses can be silent...

But of course print on errors...

swarajpande5:
> Please be silent... But we can have these messages under a debug or verbose option if needs be... Also we only need one progress bar... The model one... The other http request/responses can be silent...
> 
> But of course print on errors...

@ericcurtin I've observed that 3 separate files are downloaded whenever a model is pulled, two of them are generally really small and the final one is the large one ( which is the model file I assume :) ).
Shall I create a separate issue for this, such that it can be resolved under a separate PR ?

ericcurtin:
> > Please be silent... But we can have these messages under a debug or verbose option if needs be... Also we only need one progress bar... The model one... The other http request/responses can be silent...
> > But of course print on errors...
> 
> @ericcurtin I've observed that 3 separate files are downloaded whenever a model is pulled, two of them are generally really small and the final one is large ( which is the model file I assume :) ). Shall I create a separate issue for this, such that it can be resolved under a separate PR ?

Yeah sure it could be handled in another PR... The second of those files is unnecessary to be downloaded (I had this change in a branch somewhere but maybe I lost that branch)... The first one downloaded doesn't need to be stored in a file, it can just be stored in a variable in python... The only one we need to store is the money

swarajpande5:
@ericcurtin Got it!, I'll start to implement this in a separate PR. Thanks for the suggestions.

PR #295, opened 2024-10-13
Handles stdin input

arouene:
Be able to pass a prompt or a file by stdin like :
```
cat file | ramalama run model prompt
```

Works by prepending the input to the prompt as I didn't managed to make stdin for llama.cpp work.
Prepending the input to the prompt is inspired by shell_gpt, that's the way they do it.

### DIFF
--- a/docs/ramalama-run.1.md
+++ b/docs/ramalama-run.1.md
@@ -34,6 +34,14 @@ ramalama run merlinite "when is the summer solstice"
 The summer solstice, which is the longest day of the year, will happen on June ...
 ```
 
+Run command with a custom prompt and a file passed by the stdin
+```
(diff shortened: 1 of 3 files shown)


ericcurtin (approved):
Thanks, I am already thinking of a couple of use cases for this ☺️

Would you mind opening an issue in llama.cpp upstream for llama-cli to support reading from stdin properly?

I reckon we will run into character limits etc. maybe invalid character sequences depending on:

-p "prompt text"

ericcurtin:
Would appreciate some tests too, we have an automated test framework

arouene:
Ok, I will open an issue tu llama-cli, and yes we may need to sanitize the stdin/prompt... I will also try to put some tests!

PR #303, opened 2024-10-14
Check for "krunkit" or "libkrun" strings

ericcurtin:
I think the latest version of podman machine/krunkit displays libkrun here.


ericcurtin:
I'm gonna treat #304 as a LGTM from @maxandersen 😄

rhatdan:
libkrun was the correct value to check, my mistake.

PR #304, opened 2024-10-14
latest podman report libkrun not krunkit

maxandersen:
fixes #303


maxandersen:
never mind - #303 was a PR already - read it as issue :)

PR #306, opened 2024-10-14
No whitespace in container label

ericcurtin:
Apparently docker doesn't allow it, going with RAMALAMA as container is sort of redundant anyway.


Ben-Epstein:
tested @ericcurtin, this works 👍

ericcurtin:
Thanks for testing @Ben-Epstein !

@rhatdan PTAL when you get back

rhatdan:
LGTM

PR #309, opened 2024-10-14
Changes to start trimming rocm image

ericcurtin:
So it can fit in CI and it's easier to download

We removed support for gfx9. Which are roughly AMD GPUs prior to
2019. This saves us a ton of container image size.

gfx11 corresponds to the RDNA 3 architecture.
gfx10 corresponds to the RDNA and RDNA 2 architecture.
gfx9 corresponds to the Vega and some Navi-based architectures.


rhatdan:
Test failures have nothing to do with this change.
LGTM

Issue #302, opened 2024-10-14
zsh autocomplete now working on macOS


ericcurtin:
zsh is the default shell on macOS. autocomplete is not working, reported by @rhatdan

PR #315, opened 2024-10-15
Add zsh completions

rhatdan:
Fixes: https://github.com/containers/ramalama/issues/302


ericcurtin:
It's so hard to get a green build, might be worth turning off cs10

ericcurtin (approved):

PR #317, opened 2024-10-16
Check if podman machine is running before attempting to run

ericcurtin:
We just get an ugly failure if podman machine isn't running and we attempt to use it.


ericcurtin:
Just made "main" branch protected as I accidentally pushed to it

ericcurtin:
@rhatdan I think we need to merge this, I pushed to main by mistake and the build is broken

PR #318, opened 2024-10-16
Make changes to spec file to better pass Fedora packaging guidelines

smooge:
- Make License: MIT to match code
- Make URL for the downloads match the tag
- Change Build to what worked in 0.0.16 and before to work in mock.

### DIFF
--- a/rpm/python-ramalama.spec
+++ b/rpm/python-ramalama.spec
@@ -7,6 +7,9 @@
 
 %global _python_dist_allow_version_zero 1
 
+# Temporary removal while we work on getting omlmd package.
+%global __requires_exclude omlmd
+
 Name: python-%{pypi_name}
 # DO NOT TOUCH the Version string!
 # The TRUE source of this specfile is:
--- a/rpm/python-ramalama.spec
+++ b/rpm/python-ramalama.spec
@@ -8,7 +8,11 @@
 %global _python_dist_allow_version_zero 1
 
 # Temporary removal while we work on getting omlmd package.
+%if 0%{?rhel}
+%global __requires_exclude omlmd|huggingface-hub
+%else
 %global __requires_exclude omlmd
+%endif
 
 Name: python-%{pypi_name}
 # DO NOT TOUCH the Version string!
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM 
Although tests are not happy.

smooge:
I don't know why the tests failed.. it is trying to 'find' omlmd but it isn't listed as a Recommends or a Requires but must be getting pulled in due to searching inside the code. Not sure how to get that to turn off. Will look at what is needed for omlmd

Updating and loading repositories:
Repositories loaded.
 https://download.copr.fedorainfracloud 100% |   6.0 MiB/s |  67.5 KiB |  00m00sFailed to resolve the transaction:
Problem: conflicting requests
  - nothing provides python3.13dist(omlmd) needed by python3-ramalama-0.0.18-1.20241016125916525014.pr318.4.ga22e300.fc42.noarch from @commandline
You can try to add to command line:
  --skip-broken to skip uninstallable packages

smooge:
Ok problem now is that 'nothing provides python3.12dist(huggingface-hub) needed by python3-ramalama-0.0.18-1.20241016140057496492.pr318.8.gb7a2831.el10.noarch from @commandline for CS10'. I don't know what is needed to get that available

rhatdan:
Yes we want to avoid this requirement which is why we were trying to avoid the python install stuff.

I think we need to just do a manual make install and ignore python altogether.

The Requires: hugginface_hub is coming from the python call and the line in pyproject.toml.

Which is somehow being realized within the python code and generating it.
```
dependencies = [
  "argcomplete",
  "tqdm",
  "omlmd",
  "huggingface_hub",
]
```
We want these requirements for PyPi but not for rpm.

rpm/python-ramalama.spec
```diff
@@ -1,9 +1,18 @@
...
+%global source_url https://github.com/containers/%{pypi_name}
 %global pypi_name ramalama
 %global desc RamaLama is a command line tool for working with AI LLM models.
+%global _name python%{python3_pkgversion}-%{pypi_name}
 
-%define _python_dist_allow_version_zero 1
+%global _python_dist_allow_version_zero 1
+
+# Temporary removal while we work on getting omlmd package.
+%if 0%{?fedora}
```
rhatdan, line 11: Excellent glad you found something to simplify this.

PR #319, opened 2024-10-16
Fix handling of completions on MacOS

rhatdan:
Currently we are not installing command completions via PyPi. For some reason python is ignoring the build directory. Moving the completions from build to completions fixes the issue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,42 +34,45 @@ help:
 	@echo
 
 
-.PHONY:
+.PHONY: install-requirements
 install-requirements:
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -34,42 +34,45 @@ help:
 	@echo
 
 
-.PHONY:
+.PHONY: install-requirements
 install-requirements:
 	pipx install tqdm black flake8 argcomplete wheel omlmd huggingface_hub codespell
 
-.PHONY:
```
ericcurtin, line 41: Seems like you are missing a space here

`.PHONY:install-completions`

to

`.PHONY: install-completions`

or at least all other versions have a space.

PR #320, opened 2024-10-16
Add support for testing with docker

rhatdan:


test/system/helpers.bash
```diff
@@ -918,7 +918,7 @@ function skip_if_nocontainer() {
 }
 
 function skip_if_darwin() {
-    [ "$(uname)" == "darwin" ] || skip "Not supported on darwin"
```
ericcurtin, line 921: Darwin is capital on my machine, so this comparison might not work if it's case sensitive

```
$ uname
Darwin
```

rhatdan: Yup, I just figured that out.

ericcurtin (approved):

PR #322, opened 2024-10-16
Fix erroneous output in CUDA containerfile

bmahabirbu:
Updated the containerfile to fix incorrect outputs during model execution with certain prompt sizes.
Ensured compatibility with Docker and Podman. Related to issue [#9848](https://github.com/ggerganov/llama.cpp/issues/9848).

### DIFF
--- a/container-images/cuda/Containerfile
+++ b/container-images/cuda/Containerfile
@@ -22,32 +22,50 @@ RUN /usr/bin/python3 --version
 RUN pip install "huggingface_hub==${HUGGINGFACE_HUB_VERSION}"
 RUN pip install "omlmd==${OMLMD_VERSION}"
 
-# Build wouldn't complete couldn't find libcuda so made a systemlink
-# But this didn't work
+# CUDA_DOCKER_ARCH = 
+# Hopper GPUs (e.g., H100): Use 90
+# Ampere GPUs (e.g., RTX 30 Series, A100): Use 80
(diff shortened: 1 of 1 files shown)


ericcurtin (approved):

ericcurtin:
We can't push this one to quay.io like the other images due to Nvidia licensing restrictions... It's tempting to get ramalama to podman/docker build this Containerfile when it detects and Nvidia GPU... That would be obeying the rules I think...

p5:
> We can't push this one to quay.io like the other images due to Nvidia licensing restrictions... It's tempting to get ramalama to podman/docker build this Containerfile when it detects and Nvidia GPU... That would be obeying the rules I think...

Sorry to chime in, but could you elaborate why this is the case, when other projects also distribute these CUDA assets via OCI images? 
Is it because RedHat is a for-profit company?

ericcurtin:
> > We can't push this one to quay.io like the other images due to Nvidia licensing restrictions... It's tempting to get ramalama to podman/docker build this Containerfile when it detects and Nvidia GPU... That would be obeying the rules I think...
> 
> Sorry to chime in, but could you elaborate why this is the case, when other projects also distribute these CUDA assets via OCI images? Is it because RedHat is a for-profit company?

Not sure of the specifics, but all I know is we don't have clearance to do it in the RamaLama case, at least not yet, hopefully it changes in future.

rhatdan:
LGTM

PR #324, opened 2024-10-17
Add hf:// as an alias to huggingface://

ericcurtin:
Also check start of string contains :// , by checking if start of the string simply starts with things like "hf", "ollama", etc. we are kinda polluting the namespace and not allowing models to start with these terms.

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -193,7 +193,7 @@ def login_parser(subparsers):
 def login_cli(args):
     transport = args.TRANSPORT
     if transport != "":
-        transport = os.getenv("RAMALAMA_TRANSPORT")
+        transport = os.getenv("RAMALAMA_TRANSPORT") + "://"
     model = New(str(transport))
     return model.login(args)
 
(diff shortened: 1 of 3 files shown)


ericcurtin:
Haven't run tests locally, must keep an extra eye on the builds

ramalama/common.py
```diff
@@ -69,7 +69,11 @@ def run_cmd(args, cwd=None, ignore_stderr=False):
     if ignore_stderr:
         stderr=subprocess.PIPE
 
-    return subprocess.run(args, check=True, cwd=cwd, stdout=subprocess.PIPE, stderr=stderr)
+    try:
+        return subprocess.run(args, check=True, cwd=cwd, stdout=subprocess.PIPE, stderr=stderr)
+    except Exception:
+        perror(f"subprocess.run({args}, check=True, cwd={cwd}, stdout=subprocess.PIPE, stderr=stderr)")
```
rhatdan, line 75: This line here is for debugging?  It is causing tests to fail. The docker error is caught above and ignored.

ericcurtin: I'll remove it, it was for debugging, I was hoping to keep it in, but will remove for now

rhatdan:
LGTM

PR #325, opened 2024-10-17
There's many cases where macOS support is broken

ericcurtin:
By centralizing the check in container_manager() function. We fix a lot of these.


rhatdan:
LGTM

Issue #183, opened 2024-09-24
Add podman serve --generate kube MODEL


rhatdan:
`podman serve --generate quadlet` was just added, and generates a quadlet which outputs a quadlet file to run the model as a service.

PR #327, opened 2024-10-17
Add kubernetes.YAML support to ramalama serverve

rhatdan:
Fixes: https://github.com/containers/ramalama/issues/183

### DIFF
--- a/docs/ramalama-serve.1.md
+++ b/docs/ramalama-serve.1.md
@@ -23,8 +23,13 @@ The default is TRUE. The --nocontainer option forces this option to False.
 
 Use the `ramalama stop` command to stop the container running the served ramalama Model.
 
-#### **--generate**=quadlet
+#### **--generate**=type
 Generate specified configuration format for running the AI Model as a service
+
+| Key       | Description                                                      |
+| --------- | ---------------------------------------------------------------- |
+|  quadlet  | Podman supported container definition for running AI Model under systemd |
+|  kube     | Kubernetes YAML definition for running the AI MOdel as a service |
 
 #### **--help**, **-h**
 show this help message and exit
@@ -36,10 +41,9 @@ Name of the container to run the Model in.
 port for AI Model server to listen on
 
 ## EXAMPLES
-
-Run two AI Models at the same time, notice that they are running within Podman Containers.
-
+### Run two AI Models at the same time. Notice both are running within Podman Containers.
 ```
+
 $ ramalama serve -p 8080 --name mymodel ollama://tiny-llm:latest
 09b0e0d26ed28a8418fb5cd0da641376a08c435063317e89cf8f5336baf35cfa
 
@@ -52,8 +56,7 @@ CONTAINER ID  IMAGE                             COMMAND               CREATED   
 3f64927f11a5  quay.io/ramalama/ramalama:latest  /usr/bin/ramalama...  17 seconds ago  Up 17 seconds  0.0.0.0:8082->8082/tcp  ramalama_YMPQvJxN97
 ```
 
-Generate a quadlet for running the AI Model service
-
+### Generate a quadlet for running the AI Model service
 ```
 $ ramalama serve --name MyGraniteServer --generate=quadlet granite > $HOME/.config/containers/systemd/MyGraniteServer.container
 $ cat $HOME/.config/containers/systemd/MyGraniteServer.container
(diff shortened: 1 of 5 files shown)


rhatdan:
@ygalblum @umohnani8 @mrunalp @haircommander PTAL to make sure my kube YAML looks ok

docs/ramalama-serve.1.md
```diff
@@ -91,6 +94,42 @@ CONTAINER ID  IMAGE                             COMMAND               CREATED
...
+$ ramalama --nocontainer serve --name tini --generate kube tiny
+# Save the output of this file and use kubectl create -f to import
+# it into Kubernetes.
+#
+# Created with ramalama-0.0.17
+apiVersion: v1
+kind: Pod
+metadata:
+  labels:
+    app: tini-pod
```
ygalblum, line 108: Why postfix the name with `-pod`?

docs/ramalama-serve.1.md
```diff
@@ -91,6 +94,42 @@ CONTAINER ID  IMAGE                             COMMAND               CREATED
...
+  name: tini-pod
+spec:
+  containers:
+  - name: tini
+    image:   quay.io/ramalama/ramalama:latest
+    command: [ "llama-server" ]
+    args:    ['--port', '8080', '-m', '/run/model']
+    ports:
+    - containerPort: 8080
+      hostPort: 8080
```
ygalblum, line 118: I think it's best not to the `hostPort`

docs/ramalama-serve.1.md
```diff
@@ -91,6 +94,42 @@ CONTAINER ID  IMAGE                             COMMAND               CREATED
...
+    volumeMounts:
+        - mountPath: /run/model
```
ygalblum, line 120: This indentation works, but we should keep the way arrays are indented consistent

docs/ramalama-serve.1.md
```diff
@@ -91,6 +94,42 @@ CONTAINER ID  IMAGE                             COMMAND               CREATED
...
+          name: model
+        - mountPath: /dev/dri
+          name: dri
+  volumes:
+  - name model
+    hostPath:
```
ygalblum, line 126: How is this expected to work on K8S?

haircommander:
LGTM

ramalama/model.py
```diff
@@ -171,3 +175,78 @@ def quadlet(self, model, args, exec_args):
...
+                    volumes
+                    + f""""
+  - name {dev}
+    hostPath:
+      path: /dev/{dev}"""
+                )
+
+        return mounts + volumes
+
+    def kube(self, model, args, exec_args):
```
ygalblum, line 223: My experience with generating YAML files with templates has always been a poor one. There was always some indentation that got botched. I think a better solution is to create a dict and dump it using the `pyyaml` library by calling `print(yaml.dump(d))`

docs/ramalama-serve.1.md
rhatdan: So only do hostPort if different?

docs/ramalama-serve.1.md
rhatdan: Yes this is the goal.

ramalama/model.py
rhatdan: Cool, do you have an example of a template for a kube.yaml?

docs/ramalama-serve.1.md
ygalblum: No, these are two different entities

docs/ramalama-serve.1.md
ygalblum: In most cases `hostPort` is not really used in K8S. Endpoints are exposed via Services. Maybe consider setting the port's name

docs/ramalama-serve.1.md
ygalblum: I don't think `hostPath` will be used for these type of data. It will probably use a persistent volume

docs/ramalama-serve.1.md
```diff
@@ -91,6 +94,42 @@ CONTAINER ID  IMAGE                             COMMAND               CREATED
...
 
+### Generate a kubernetes YAML file named tini
+```
+$ ramalama --nocontainer serve --name tini --generate kube tiny
+# Save the output of this file and use kubectl create -f to import
+# it into Kubernetes.
+#
+# Created with ramalama-0.0.17
+apiVersion: v1
+kind: Pod
```
ygalblum, line 105: Come to think about it I think a `Deployment` will be more useful than a `Pod`

docs/ramalama-serve.1.md
rhatdan: name: tini-port?

docs/ramalama-serve.1.md
rhatdan: Yes i am hoping to change this to use an image at some point.  But if you could give me an example of what it should look like?

docs/ramalama-serve.1.md
rhatdan: Changed

docs/ramalama-serve.1.md
rhatdan: Fixed

docs/ramalama-serve.1.md
rhatdan: Removed -pod

ericcurtin (approved):

PR #330, opened 2024-10-18
fix: remove duplicate GitHub Actions workflow runs in PRs

p5:
Removes the duplicate CI pipeline runs during pull requests.

This works by configuring the workflows on `push` to only happen on the main branch.


rhatdan:
LGTM

PR #331, opened 2024-10-18
Break up tests: build, bats, bats-nocontainer, docker, mac-nocontainer

rhatdan:
Should speed up testing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,15 +7,15 @@ on:
       - main
 
 jobs:
-  linux:
+  build:
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
-      - name: install bats
+      - name: install
         shell: bash
         run: |
            sudo apt-get update
-           sudo apt-get install podman bats bash codespell python3-argcomplete pipx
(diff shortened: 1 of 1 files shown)


ericcurtin (approved):

.github/workflows/ci.yml
```diff
@@ -57,43 +71,32 @@ jobs:
...
-           echo '{"data-root": "/mnt/docker"}' > /tmp/daemon.json
-           sudo mv /tmp/daemon.json /etc/docker/daemon.json
-           cat /etc/docker/daemon.json
-           sudo systemctl restart docker.service
-           df -h
+           pip install tqdm --break-system-packages
+           make bats-nocontainer
 
-      - name: Print disk space after cleanup
+  docker:
```
p5, line 90: X-post from my PR
I think this job needs the container images to be built before bats-docker is ran, within the same job.  Otherwise we will be pulling container images from the registry (and CUDA images will fail).  

This means we are not testing changes from the PR, but rather always testing against the images built from the main branch.

p5: Discussed in the other PR so closing this

PR #332, opened 2024-10-18
Build container images only on changes

p5:
This PR splits the container image builds into a separate GHA workflow so we only need to run these when the ./container-images/ directory is changed.

We could have done this inside the existing workflow, but it would have either involved some verbose Git commands and/or introducing a new third party dependency (e.g. [tj-actions/changed-files](https://github.com/tj-actions/changed-files)), but I feel we benefit more to having minimal dependencies or complexity to help with keeping the repo secure and easy to contribute to.

There are further optimizations which can be done, the ones I can think of are parallelising the builds and detecting which specific Containerfile is updated and only building that one.  For now, this seems like a good start.

### DIFF
--- a/.github/workflows/ci-images.yml
+++ b/.github/workflows/ci-images.yml
@@ -56,6 +56,6 @@ jobs:
         run: |
           df -h
 
-      - name: run bats
+      - name: Build Images
         run: |
           make build
--- a/.github/workflows/ci-images.yml
+++ b/.github/workflows/ci-images.yml
@@ -1,0 +1,61 @@
+# Split the image builds into a separate workflow so we can easily run it
+# when only the Containerfiles have changed without introducing new external
+# GitHub Actions dependencies, such as the `tj-actions/changed-files` action.
+name: ci images
+
+on:
+  pull_request:
+    paths:
+      - "**/container-images/**"
+  push:
+    branches:
+      - main
+    paths:
+      - "**/container-images/**"
+
+jobs:
+  build:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: install
+        shell: bash
+        run: |
+          sudo apt-get update
+          sudo apt-get install podman bash
+          make install-requirements
+
+      - name: Print disk space before cleanup
+        shell: bash
+        run: |
+          df -h
+
+      - name: Free Disk Space Linux
+        shell: bash
+        run: |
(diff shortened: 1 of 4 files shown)


ericcurtin:
It is important we build the cpuonly container to be used in any of the builds where we run tests or execute ramalama, rather than pull a container...

Because we want to test that Containerfile changes actually work...

p5:
> It is important we build the cpuonly container to be used in any of the builds where we run tests or execute ramalama, rather than pull a container...
> 
> Because we want to test that Containerfile changes actually work...

This PR is based on Dan's branch, which has split the builds into separate jobs.  I have only moved an entire job from one workflow to another.

rhatdan:
Can we just build the ramalama cpuonly container and not the Cuda one which takes forever.

p5:
@rhatdan @ericcurtin 
I've pushed a new commit ([f62705a](https://github.com/containers/ramalama/pull/332/commits/f62705a33eece87ab37b2dbbf3218ca7bc636bf0)) which updates the container_build.sh script to support specifying a single container, plus adds the cpuonly build as a prerequisite to the bats-docker build.

I'm trying to figure out how to test the workflow, so will leave this as a draft for now.

ericcurtin:
> Can we just build the ramalama cpuonly container and not the Cuda one which takes forever.

Yeah, cuda and rocm container builds aren't important unless those Containerfile's are actually changed.

The cpuonly one is because we use that for testing in CI.

ericcurtin:
Like one thing that is cool about the renovate setup you did @p5 is that when we version bump a PR is automatically opened, which means that version bump can be tested before we merge. It's better than blindly trusting version bumps.

p5:
Yeah, the reasoning behind wanting the builds makes complete sense.

So if I can figure out the correct Makefile syntax, do you think my latest commit would solve the requirement?  Changes to any of the Containerfiles or related config files cause all images to be built, but the regular builds will only build the cpuonly Containerfile just before running docker-bats

p5:
We're also able to split each container build into separate jobs so they run in parallel (which could reduce the time waiting for CI by 50%), though I'm aware there have been some limitations with the number of concurrent GitHub Actions available within the org.  
I'll not implement this for now, but please let me know if you want this.

ericcurtin:
What's here so far LGTM 😄

p5:
Rebased my changes onto the latest main branch, hence for force-push.  
Think this PR is ready to go.  We can see the `ci/docker` workflow building only the "ramalama" image before running the tests.  The other docker builds should run once this PR is merged.

rhatdan:
LGTM

PR #334, opened 2024-10-20
Kompute Containerfile

ericcurtin:
Getting this build and pushed so we can use for testing.

Co-authored-by: Sergio Lopez <slp@redhat.com>
Signed-off-by: Eric Curtin <ecurtin@redhat.com>


ericcurtin:
Trying to split out the Containerfile part of:

https://github.com/containers/ramalama/pull/235

and get that part done.

ericcurtin:
It's named ramalama for now, although it's tempting to call it kompute because we want to merge the other vulkan implementation also and somehow portray that both images can use vulkan

rhatdan:
@ericcurtin needs a rebase.

ericcurtin:
@rhatdan rebased... Might as well get this in before we do a multi-arch rebuild for granitemoe support I guess...

PR #340, opened 2024-10-21
Update llama.cpp to fix granite3-moe models

ericcurtin:
This was added recently to llama.cpp:

https://github.com/ggerganov/llama.cpp/pull/9438


ericcurtin:
Will need to rebuild and re-push the container images to fix granitemoe models @rhatdan

rhatdan:
LGTM

PR #351, opened 2024-10-21
[WIP] Add support for vllm container image

rhatdan:


rhatdan:
@ericcurtin First attempt to get vllm based container for running ramalama.  Sadly blows up after pulling Model. Not sure what is going wrong.

ericcurtin:
Maybe we should reach out to IBM and ask them to rebase their fork against upstream:

https://github.com/IBM/vllm

.gguf support only made vllm very recently. Could try using upstream also to see if it makes a difference.

It would also be interesting to query some IBM folk as to why they fork.

dtrifiro:
IBM is currently using the version here: https://github.com/opendatahub-io/vllm, the main Dockerfile is `Dockerfile.ubi`

danielezonca:
@dtrifiro 
Can you help with this PR?
The goal is to add vLLM (mainly CPU+cuda I would say, we can iterate and other accelerator later) to ramalama.
I don't think it should require special customization but maybe there are options that can help

@ericcurtin 
That IBM fork is old and not used anymore

ericcurtin:
Closing, we are using the OpenShift AI one now

PR #356, opened 2024-10-22
Added a check to the zsh completions generation step

ericcurtin:
Ensuring that zsh is actually supported

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,8 +64,13 @@ completions:
 	mkdir -p completions/fish/vendor_completions.d
 	register-python-argcomplete --shell fish ramalama > completions/fish/vendor_completions.d/ramalama.fish
 
-	mkdir -p completions/zsh/vendor-completions
-	register-python-argcomplete --shell zsh ramalama > completions/zsh/vendor-completions/_ramalama
+	tmpfile=$$(mktemp); \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -64,8 +64,10 @@ completions:
 	mkdir -p completions/fish/vendor_completions.d
 	register-python-argcomplete --shell fish ramalama > completions/fish/vendor_completions.d/ramalama.fish
 
-	mkdir -p completions/zsh/vendor-completions
-	register-python-argcomplete --shell zsh ramalama > completions/zsh/vendor-completions/_ramalama
+	if ! (register-python-argcomplete --shell zsh ramalama > completions/zsh/vendor-completions/_ramalama); then \
```
rhatdan, line 67: How about?
```
	tmpfile=$(mktemp)
	if ! (register-python-argcomplete --shell zsh ramalama > $i{tmpfile}); then
		mkdir -p completions/zsh/vendor-completions; \
		mv ${tmpfile} completions/zsh/vendor-completions/_ramalama; \
	else
		rm ${tmpfile}
	fi
```

ericcurtin: I put in some Makefile equivalent that seems to work

rhatdan:
LGTM.

PR #357, opened 2024-10-22
Allow removal of models specified as shortnames

rhatdan:


ericcurtin:
I see the issue you are encountering... How I solved this in the past in:

https://github.com/containers/ramalama/pull/146

is by always resolving to the long-names outside the container and only using long-names within the container.

We must make sure to take quadlet, k8s generators into account.

We have two options:

1. Do all name resolution outside the container (which is what we initially did).
2. Somehow copy the shortnames file or bind-mount it into the container.

Whatever works well, I was favouring the first technique but we can do either.

PR #358, opened 2024-10-22
Allow pushing of oci images

rhatdan:


ramalama/oci.py
```diff
@@ -53,16 +55,47 @@ def _target_decompose(self, model):
...
+                f"""\
+FROM {args.image} as builder
+RUN mkdir -p /run/model; cd /run/model; ln -s {model_name} model.file
+
+FROM scratch
+COPY --from=builder /run/model /
+COPY {model} /{model_name}
+"""
+            )
+        run_cmd([self.conman, "build", "-t", target, "-f", containerfile.name, contextdir], stdout=None)
```
tarilabs, line 77: wouldn't this require `conman` to be available a system tool/dependency? 🤔

rhatdan: Being able to create and push to a OCI Registry requires an OCI Container Engine.

rhatdan:
Need to add tests using local oci registries to RamaLama.

tarilabs (commented):
I believe pushing a non-executable OCI Image with only containing model is a valid alternative way, given more recent K8s KEP updates.

I _also_ believe we should seize this time as an opportunity to give relevance to Metadata of the model, see comment below.

This will also compound for benefit of putting as well Metadata under scope of Signature and Attestation.

wdyt?

ramalama/oci.py
```diff
@@ -53,16 +55,47 @@ def _target_decompose(self, model):
...
+        # Open the file for writing.
+        with open(containerfile.name, 'w') as c:
+            c.write(
+                f"""\
+FROM {args.image} as builder
+RUN mkdir -p /run/model; cd /run/model; ln -s {model_name} model.file
+
+FROM scratch
+COPY --from=builder /run/model /
+COPY {model} /{model_name}
```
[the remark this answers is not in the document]
rhatdan: I would prefer to get this in, and then open PRs to add more metadata to the container image since this is just a simple Containerfile, it should be simple to enhance it.

ericcurtin (approved):

PR #368, opened 2024-10-24
Fix Makefile to be less stringent on failues of zsh

smooge:
I am making some lines optional in the Makefile and added fixes from the RPM review.


rhatdan:
Didn't @ericcurtin do this yesterday?

smooge:
He did but I missed it in my merge. I tried it in f39 and it broke in the make so I have went with my original design. If the other mode is still needed I can try to figure out why f39 Make hates shell

rhatdan:
LGTM

PR #369, opened 2024-10-24
Start an Asahi version

ericcurtin:
Asahi has a forked version of mesa while it upstream.

### DIFF
--- a/container-images/asahi/Containerfile
+++ b/container-images/asahi/Containerfile
@@ -1,0 +1,25 @@
+FROM fedora:41
+
+# renovate: datasource=github-releases depName=containers/omlmd extractVersion=^v(?<version>.*)
+ARG OMLMD_VERSION=0.1.6
+ARG LLAMA_CPP_SHA=1329c0a75e6a7defc5c380eaf80d8e0f66d7da78
+# renovate: datasource=git-refs depName=ggerganov/whisper.cpp packageName=https://github.com/ggerganov/whisper.cpp gitRef=master versioning=loose type=digest
+ARG WHISPER_CPP_SHA=0377596b77a3602e36430320cbe45f8c305ef04a
+
+RUN dnf install -y dnf-plugins-core && \
+    dnf copr enable -y @asahi/fedora-remix-branding && \
+    dnf install -y asahi-repos && \
+    dnf install -y mesa-vulkan-drivers vulkan-headers vulkan-loader-devel \
+      vulkan-tools spirv-tools glslc glslang git procps-ng vim cmake gcc-c++ \
+      python3-pip python3-argcomplete && \
+    dnf clean all && \
+    rm -rf /var/cache/*dnf*
+
+RUN /usr/bin/python3 --version
+RUN pip install "omlmd==${OMLMD_VERSION}"
+
+COPY ../scripts /scripts
+RUN chmod +x /scripts/*.sh && \
+    /scripts/build_llama_and_whisper.sh "$LLAMA_CPP_SHA" "$WHISPER_CPP_SHA" \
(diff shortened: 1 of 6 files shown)


container-images/ramalama/Containerfile
```diff
@@ -4,8 +4,6 @@ FROM registry.access.redhat.com/ubi9/ubi:9.4-1214.1729773476
 ARG HUGGINGFACE_HUB_VERSION=0.26.2
```
rhatdan, line 4: We don't need this anymore.

container-images/asahi/Containerfile
```diff
@@ -0,0 +1,32 @@
+FROM fedora:41
+
+# renovate: datasource=github-releases depName=huggingface/huggingface_hub extractVersion=^v(?<version>.*)
+ARG HUGGINGFACE_HUB_VERSION=0.26.2
```
rhatdan, line 4: Don't need this anymore.

container-images/cuda/Containerfile
```diff
@@ -32,8 +30,7 @@ RUN dnf install -y python3 python3-pip && \
     rm -rf /var/cache/*dnf*
 
 # Install Python packages in the runtime image
-RUN pip install "huggingface_hub==${HUGGINGFACE_HUB_VERSION}" \
-      "omlmd==${OMLMD_VERSION}"
+RUN pip install "omlmd==${OMLMD_VERSION}"
```
rhatdan, line 33: Don't need this either.

container-images/ramalama/Containerfile
```diff
@@ -24,9 +20,7 @@ RUN dnf install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.n
     rm -rf /var/cache/*dnf*
 
 RUN /usr/bin/python3 --version
-RUN pip install "huggingface_hub==${HUGGINGFACE_HUB_VERSION}"
 RUN pip install "omlmd==${OMLMD_VERSION}"
```
rhatdan, line 23: Don't need this.

rhatdan:
Does this mean on Asahi we can't use the standard RamaLama image?

ericcurtin:
Fedora Asahi Remix is 99% percent Fedora... The two main packages it forks are the kernel and mesa (mesa being where the userspace GPU driver is).

This Vulkan driver "Honeykrips" is weeks old, so the patches have not made their way to mainline Fedora mesa yet:

https://www.phoronix.com/news/Mesa-HoneyKrisp-October

anyway, with the other images being UBI based, even besides this mesa fork, we would have had to make a special container image for Asahi because Asahi is only enabled for Fedora, not RHEL.

So yeah, we need a specific container image for Asahi.

I don't expect to make this exception often, but the Asahi community is so vibrant and kinda is a flagbearer for ARM Linux efforts and ARM GPU drivers. I think it's worth it.

ericcurtin:
@conan-kudo expressed an interesting in running llama.cpp outside of containers which is fine too, as long as you are using the forked mesa and building/using llama.cpp built with -DGGML_VULKAN=1 it should work

rhatdan:
We should be able to build the asahi image from the RamaLama Containerfile.

podman build --from fedora:41 -t quay.io/ramalama/asahi:latest containers-images/ramalama

That way we don't need to update multiple containerfiles.

ericcurtin:
> We should be able to build the asahi image from the RamaLama Containerfile.
> 
> podman build --from fedora:41 -t quay.io/ramalama/asahi:latest containers-images/ramalama
> 
> That way we don't need to update multiple containerfiles.

Other than:

rhatdan:
No, I was thinking they were the same Containerfiles, other then the From, I will merge and we can revisit later.

PR #371, opened 2024-10-25
Change to None instead of ""

ericcurtin:
Usage help was showing `(default: )` when we used "" which was not clear.


rhatdan:
LGTM

PR #372, opened 2024-10-25
Fix shortname paths

swarajpande5:


ericcurtin (approved):

ericcurtin:
We could actually do a test here sort of like a HEAD http request to see if all these things actually exists

PR #373, opened 2024-10-25
Add --debug option to show exec_cmd and run_cmd commands

rhatdan:
Also fixed a bug I found when testing pulling a bogus image.

Hugginface pulls should have a similar error, but currently we execute the command line and it blows up. When we convert to using internal http we need to catch pulling a bogus image.

### DIFF
--- a/ramalama/common.py
+++ b/ramalama/common.py
@@ -59,9 +59,9 @@ def available(cmd):
     return shutil.which(cmd) is not None
 
 
-def exec_cmd(args, stderr=True):
-    if x:
-        print(*args)
+def exec_cmd(args, stderr=True, debug=False):
+    if debug:
+        perror("exec_cmd: ", *args)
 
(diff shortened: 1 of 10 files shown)


ericcurtin (approved):

ramalama/common.py
```diff
@@ -59,9 +59,9 @@ def available(cmd):
     return shutil.which(cmd) is not None
 
 
-def exec_cmd(args, stderr=True):
-    if x:
-        print(*args)
+def exec_cmd(args, stderr=True, debug=False):
+    if debug:
+        perror("exec_cmd: ", *args)
```
ericcurtin, line 64: Should probably do

```
    if debug:
        stderr = True
        perror("exec_cmd: ", *args)
```

but that can be for another PR.

ericcurtin (approved):

PR #377, opened 2024-10-26
Support new hf.co/ format

ericcurtin:
This new format for hf has been implemented by Ollama.

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -688,13 +688,38 @@ def dry_run(args):
     print()
 
 
+def convert_model_format(input_string):
+    # Split the input string to extract the relevant parts
+    parts = input_string.split('/')
+    if len(parts) != 3 or not parts[2]:
+        raise ValueError("Input string is not in the expected format")
+
+    username = parts[1]
+    model_info = parts[2]
+
+    # Check if quantization is specified, else default to Q4_K_M
+    if ':' in model_info:
+        model_name, quantization = model_info.split(':')
+    else:
+        model_name = model_info
+        quantization = 'Q4_K_M'  # Default quantization scheme
+
+    # Construct the new format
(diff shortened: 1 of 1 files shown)


ericcurtin:
This doesn't work that well, but this is up for grabs if someone can figure it out

ramalama/cli.py
```diff
@@ -688,13 +688,38 @@ def dry_run(args):
...
+
+    # Check if quantization is specified, else default to Q4_K_M
+    if ':' in model_info:
+        model_name, quantization = model_info.split(':')
+    else:
+        model_name = model_info
+        quantization = 'Q4_K_M'  # Default quantization scheme
+
+    # Construct the new format
+    return f"hf://{username}/{model_name}/{model_name}-{quantization}.gguf"
```
swarajpande5, line 708: @ericcurtin I would like to take this up, the new format which is being returned might not always be correct. 
For example, running the following, 
```
ramalama pull hf.co/instructlab/merlinite-7b-lab-GGUF
```
will use 
```
hf://instructlab/merlinite-7b-lab-GGUF/merlinite-7b-lab-GGUF-Q4_K_M.gguf
```
And the above path does not exist.

ericcurtin: Yeah... Maybe there's a REST API to get the real filename if you pass it instructlab/merlinite-7b-lab-GGUF and Q4_K_M or something like that but I'm really guessing...

swarajpande5: Yes, there should be one to get the complete path. I'll explore this :)

rhatdan:
Any update on this?

rhatdan:
@ericcurtin @swarajpande5 Should this close?

ericcurtin:
Closing...

PR #379, opened 2024-10-27
Drop in config

ericcurtin:
For setting configs in a config file

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Running in containers eliminates the need for users to configure the host system
 
 RamaLama then pulls AI Models from model registries. Starting a chatbot or a rest API service from a simple single command. Models are treated similarly to how Podman and Docker treat container images.
 
-When both Podman and Docker are installed, RamaLama defaults to Podman, The `RAMALAMA_CONTAINER_ENGINE=docker` environment variable can override this behavior. When neather are installed RamaLama will attempt to run the model with software on the local system.
+When both Podman and Docker are installed, RamaLama defaults to Podman, The `RAMALAMA_CONTAINER_ENGINE=docker` environment variable can override this behaviour. When neither are installed RamaLama will attempt to run the model with software on the local system.
 
 RamaLama supports multiple AI model registries types called transports.
 Supported transports:
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -21,6 +21,7 @@ from ramalama.common import (
 from ramalama.oci import OCI
 from ramalama.ollama import Ollama
 from ramalama.shortnames import Shortnames
+from ramalama.toml_parser import TOMLParser
 from ramalama.version import version, print_version
 
 shortnames = Shortnames()
@@ -52,8 +53,43 @@ class ArgumentParserWithDefaults(argparse.ArgumentParser):
         super().add_argument(*args, **kwargs)
 
 
+def load_config():
+    """Load configuration from a list of paths, in priority order."""
+    config = {}
+    config_paths = [
+        "/usr/share/ramalama/ramalama.conf",
+        "/etc/ramalama/ramalama.conf",
+        os.path.expanduser("~/.config/ramalama/ramalama.conf"),
+    ]
+
+    parser = TOMLParser()
+    # Load configuration from each path
+    for path in config_paths:
+        if os.path.exists(path):
+            # Load the main config file
+            config = parser.parse_file(path)
+        if os.path.isdir(path + ".d"):
+            # Load all .conf files in ramalama.conf.d directory
+            for conf_file in sorted(Path(path + ".d").glob("*.conf")):
+                config = parser.parse_file(conf_file)
+
+    return config
+
+
 def init_cli():
-    description = """\
+    """Initialize the RamaLama CLI and parse command line arguments."""
+    description = get_description()
+    config = load_and_merge_config()
+    parser = create_argument_parser(description, config)
+    configure_subcommands(parser)
+    args = parse_arguments(parser)
+    post_parse_setup(args)
+    return parser, args
+
+
+def get_description():
+    """Return the description of the RamaLama tool."""
+    return """\
 RamaLama tool facilitates local management and serving of AI Models.
 
 On first run RamaLama inspects your system for GPU support, falling back to CPU support if no GPUs are present.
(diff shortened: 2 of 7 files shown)


ericcurtin:
@conan-kudo PTAL , haven't tested or anything yet

ericcurtin:
Tests to fix

ramalama/cli.py
```diff
@@ -52,8 +52,63 @@ def add_argument(self, *args, help=None, default=None, **kwargs):
         super().add_argument(*args, **kwargs)
 
 
+def parse_config_file(file_path, config):
```
rhatdan, line 55: Lets move the config parsing to a separate file config.py?

Isn't there a standard toml parser written in Python that we can use?

rhatdan:
```
import tomllib

with open("/usr/share/containers/containers.conf", "rb") as f:
    data = tomllib.load(f)

print(data)

```

ericcurtin: tomllib is available since 3.11, not available on all platforms.

Conan-Kudo:
I'd recommend using `tomlkit` here if you want to use TOML for configuration.

ericcurtin:
The thing is, if we write a simple toml parser like below, which is easy to maintain. We don't necessarily have to worry about whether somebody has that python library available on platform X:

```
import re

class TOMLParser:
    def __init__(self, toml_string=""):
        self.data = {}
        if toml_string:
            self.parse(toml_string)

    def parse(self, toml_string):
        current_section = self.data
        for line in toml_string.splitlines():
            line = line.strip()
            if not line or line.startswith("#"):
                continue

            if line.startswith("[") and line.endswith("]"):
                section_name = line[1:-1].strip()
                current_section = self._create_section(section_name)
            elif "=" in line:
                key, value = line.split("=", 1)
                key = key.strip()
                value = self._parse_value(value.strip())
                current_section[key] = value
            else:
                raise ValueError(f"Invalid TOML line: {line}")
        return self.data

    def _create_section(self, section_name):
        keys = section_name.split(".")
        section = self.data
        for key in keys:
            if key not in section:
                section[key] = {}
            section = section[key]
        return section

    def _parse_value(self, value):
        if value.startswith('"') and value.endswith('"'):
            return value[1:-1]
        elif value.startswith("[") and value.endswith("]"):
            return [self._parse_value(v.strip()) for v in value[1:-1].split(",")]
        elif re.match(r'^\d+$', value):
            return int(value)
        elif re.match(r'^\d+\.\d+$', value):
            return float(value)
        elif value.lower() in {"true", "false"}:
            return value.lower() == "true"
        else:
            raise ValueError(f"Unsupported value type: {value}")

    def get(self, key, default=None):
        keys = key.split(".")
        value = self.data
        for key in keys:
            value = value.get(key)
            if value is None:
                return default
        return value
```

ericcurtin:
tomlkit is only in EPEL which isn't ideal. toml and tomli is in AppStream. But thinking outside of RHEL/Fedora I'd rather just write a small toml parser to cover macOS and other random Linux distros.

README.md
```diff
@@ -15,7 +15,7 @@ Running in containers eliminates the need for users to configure the host system
 
 RamaLama then pulls AI Models from model registries. Starting a chatbot or a rest API service from a simple single command. Models are treated similarly to how Podman and Docker treat container images.
 
-When both Podman and Docker are installed, RamaLama defaults to Podman, The `RAMALAMA_CONTAINER_ENGINE=docker` environment variable can override this behavior. When neather are installed RamaLama will attempt to run the model with software on the local system.
+When both Podman and Docker are installed, RamaLama defaults to Podman, The `RAMALAMA_CONTAINER_ENGINE=docker` environment variable can override this behaviour. When neather are installed RamaLama will attempt to run the model with software on the local system.
```
ericcurtin, line 18: Today I learned behavior vs behaviour is an US English vs UK English thing. The UK version for whatever reason is more prominent for this word. Can default to US English though if needs be.

rhatdan:
Will your patches work on Windows?

We need to have multi-levels of defaults.

```
Unix Descriptions
            Built-in defaults
            Overwritten by:
                     /usr/share/ramamalam/containers.conf
            Overwritten by:
                     /etc/ramamalam/containers.conf
            Overwritten by:
                     /usr/share/ramamalam/containers.conf.d/*
            Overwritten by:
                     $HOME/.config/ramamalam/containers.conf
            Overwritten by:
                     Environment Variables
            Overwritten by command line options:
                    --engine=foobar
```


            Overwritten by:

ericcurtin:
> Will your patches work on Windows?
>

This will work using WSL2. There's a lot of things in RamaLama that won't work on Windows natively. For one, we use symlinks, which there isn't an exact equivalent for in Windows, although there are similar things.

> We need to have multi-levels of defaults.
> 
> ```
> Unix Descriptions
>             Built-in defaults
>             Overwritten by:
>                      /usr/share/ramamalam/containers.conf
>             Overwritten by:
>                      /etc/ramamalam/containers.conf
>             Overwritten by:
>                      /usr/share/ramamalam/containers.conf.d/*
>             Overwritten by:
>                      $HOME/.config/ramamalam/containers.conf
>             Overwritten by:
>                      Environment Variables
>             Overwritten by command line options:
>                     --engine=foobar
> ```
> 
> ```
>         Overwritten by:
> ```

SGTM

ericcurtin:
Although. This seems funny to me:

```
            Overwritten by:
                     /usr/share/ramamalam/containers.conf
            Overwritten by:
                     /etc/ramamalam/containers.conf
            Overwritten by:
                     /usr/share/ramamala/containers.conf.d/*
            Overwritten by:
                     $HOME/.config/ramamala/containers.conf
            Overwritten by:
                     $HOME/.config/ramamala/containers.conf.d/*
```

I would expect this:

            Overwritten by:
                     /usr/share/ramamala/containers.conf
            Overwritten by:
                     /usr/share/ramamala/containers.conf.d/*
            Overwritten by:
                     /etc/ramamala/containers.conf
            Overwritten by:
                     /etc/ramamala/containers.conf.d/*

rhatdan:
We can leave out .d directories for now, probably overkill for a simple tool.

ericcurtin:
> We can leave out .d directories for now, probably overkill for a simple tool.

I can delete... But it's already implemented, it's just 3 lines of code in the current implementation... Easy to remove though, just delete three lines:

```
        if os.path.exists(path):
            # Load the main config file
            config = parser.parse_file(path)
        if os.path.isdir(path + ".d"):
            # Load all .conf files in ramalama.conf.d directory
            for conf_file in sorted(Path(path + ".d").glob("*.conf")):
                config = parser.parse_file(conf_file)
```

PR #393, opened 2024-10-30
Share these paths if they exist

ericcurtin:
For config


rhatdan:
LGTM
But needs a rebase.

rhatdan:
This PR is going to conflict with my big PR to move away from python inside of the container.

ericcurtin:
Feel free to ignore, if we move to the model where we try to only run llama.cpp/vllm in the container this code becomes irrelevant

rhatdan:
Needs a rebase.

PR #400, opened 2024-11-01
Time for removal of huggingface_hub dependancy

ericcurtin:
We now have our own huggingface client written


ericcurtin:
@swarajpande5 @rhatdan what do you think?

ericcurtin:
I think we can get this list down to:

```
"argcomplete",
"tqdm"
 ```
  
as strict dependancies.

rhatdan:
LGTM

PR #403, opened 2024-11-01
Packit: disable osh diff scan

lsm5:
Often takes too long and results are being ignored for now.

Ref: https://packit.dev/docs/configuration#osh_diff_scan_after_copr_build

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -25,6 +25,7 @@ jobs:
     targets: &fedora_copr_targets
       - fedora-development
       - fedora-latest-stable
+    osh_diff_scan_after_copr_build: false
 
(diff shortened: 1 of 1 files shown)


lsm5:
@rhatdan @ericcurtin @smooge hopefully that should do it. PTAL

ericcurtin (approved):

ericcurtin:
I really need to find out how to make that "Merge on green build" button appear here, we have it in OSTree repo and it's super useful, the button says "Enable auto-merge" and will merge once builds pass.

rhatdan:
I think we should still require one LGTM to pass as well.  Eventually would be nice to get to the point of requiring two.

PR #404, opened 2024-11-01
Make minimal change to allow for ramalama to build on EL9

smooge:
The base version of python shipped with EL9 is python-3.9 and the setuptools shipped with that is too old to understand most of the pyproject.toml input. So the setup.py must contain the name of the project and the version for it to properly build a package in EL9.


rhatdan:
LGTM

ericcurtin:
Nice

PR #407, opened 2024-11-01
reduced the size of the nvidia containerfile

bmahabirbu:
```
podman images
REPOSITORY             TAG                  IMAGE ID      CREATED            SIZE
localhost/rama-cuda    v1                   8d63e6530deb  9 minutes ago      7.72 GB
localhost/rama-small   v1                   23372ba06dba  28 minutes ago     3.02 GB
```
Where small is the reduced container


rhatdan:
LGTM

PR #408, opened 2024-11-04
Consistency changes

ericcurtin:
Ensure llama.cpp version is the same accross containers. Removing some duplicate actions, etc.

### DIFF
--- a/container-images/cuda/Containerfile
+++ b/container-images/cuda/Containerfile
@@ -1,41 +1,22 @@
 # Base image with CUDA for compilation
 FROM docker.io/nvidia/cuda:12.6.2-devel-ubi9 AS builder
 
-# renovate: datasource=github-releases depName=huggingface/huggingface_hub extractVersion=^v(?<version>.*)
-ARG HUGGINGFACE_HUB_VERSION=0.26.2
-# renovate: datasource=github-releases depName=containers/omlmd extractVersion=^v(?<version>.*)
-ARG OMLMD_VERSION=0.1.6
-ARG LLAMA_CPP_SHA=3f1ae2e32cde00c39b96be6d01c2997c29bae555
+ARG LLAMA_CPP_SHA=1329c0a75e6a7defc5c380eaf80d8e0f66d7da78
 # renovate: datasource=git-refs depName=ggerganov/whisper.cpp packageName=https://github.com/ggerganov/whisper.cpp gitRef=master versioning=loose type=digest
 ARG WHISPER_CPP_SHA=4e10afb5a94469c605aae4eceb4021fb0e68c8f5
 
-ARG CUDA_DOCKER_ARCH=default
-
 # Install dependencies only needed for building
-RUN dnf install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm && \
-    dnf install -y git cmake gcc-c++ python3-pip && \
+RUN dnf install -y git cmake gcc-c++ && \
     dnf clean all && rm -rf /var/cache/*dnf*
 
-# Install Python packages
-RUN pip install "huggingface_hub==${HUGGINGFACE_HUB_VERSION}" "omlmd==${OMLMD_VERSION}"
+# Set the temporary installation directory
+ENV INSTALL_PREFIX=/tmp/install
 
-# Build llama.cpp
-RUN git clone https://github.com/ggerganov/llama.cpp && cd llama.cpp && \
-    git reset --hard ${LLAMA_CPP_SHA} && \
-    cmake -B build -DGGML_CUDA=ON -DCUDA_DOCKER_ARCH=${CUDA_DOCKER_ARCH} -DCMAKE_EXE_LINKER_FLAGS=-Wl,--allow-shlib-undefined . && \
-    cmake --build build --config Release -j$(nproc) && \
-    mv build/bin/llama-cli /usr/bin/ && mv build/bin/llama-server /usr/bin/ && \
-    mv build/ggml/src/libggml.so /usr/lib/ && mv build/src/libllama.so /usr/lib/ && \
-    cd / && rm -rf llama.cpp
-
-# Build whisper.cpp
-RUN git clone https://github.com/ggerganov/whisper.cpp && cd whisper.cpp && \
-    git reset --hard ${WHISPER_CPP_SHA} && \
-    cmake -B build -DGGML_CUDA=ON -DCUDA_DOCKER_ARCH=${CUDA_DOCKER_ARCH} -DCMAKE_EXE_LINKER_FLAGS=-Wl,--allow-shlib-undefined . && \
-    cmake --build build --config Release -j$(nproc) && \
-    mv build/bin/main /usr/bin/whisper-main && mv build/bin/server /usr/bin/whisper-server && \
-    if [ -f build/lib/libwhisper.so ]; then mv build/lib/libwhisper.so /usr/lib/; fi && \
-    cd / && rm -rf whisper.cpp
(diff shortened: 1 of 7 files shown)


ericcurtin:
PTAL @bmahabirbu I didn't test the cuda changes FWIW, don't have the hardware

bmahabirbu:
I tested it and it crashed before loading the model. I believe that cmake --install doesn't include 
```
build/ggml/src/libggml.so
build/src/libllama.so
```
files for some reason

container-images/cuda/Containerfile
```diff
@@ -46,16 +45,14 @@ ARG HUGGINGFACE_HUB_VERSION=0.26.2
 ARG OMLMD_VERSION=0.1.6
 
 # Install minimal runtime dependencies
-RUN dnf install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm && \
-    dnf install -y python3 python3-pip && dnf clean all && rm -rf /var/cache/*dnf*
+RUN dnf install -y python3 python3-pip nvidia-driver-cuda-libs && \
```
ericcurtin, line 48: I had to add this for this lib:

/lib64/libcuda.so.1

```
$ ldd /usr/bin/llama-cli
	linux-vdso.so.1 (0x00007f219dc64000)
	libllama.so => /lib64/libllama.so (0x00007f219dae4000)
	libggml.so => /lib64/libggml.so (0x00007f2188000000)
	libstdc++.so.6 => /lib64/libstdc++.so.6 (0x00007f2187c00000)
	libm.so.6 => /lib64/libm.so.6 (0x00007f219da09000)
	libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007f219d9ee000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f2187800000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f219dc66000)
	libcudart.so.12 => /usr/local/cuda/targets/x86_64-linux/lib/libcudart.so.12 (0x00007f2187400000)
	libcublas.so.12 => /usr/local/cuda/targets/x86_64-linux/lib/libcublas.so.12 (0x00007f2180a00000)
	libcublasLt.so.12 => /usr/local/cuda/targets/x86_64-linux/lib/libcublasLt.so.12 (0x00007f215f000000)
	libcuda.so.1 => /lib64/libcuda.so.1 (0x00007f215c000000)
	libgomp.so.1 => /lib64/libgomp.so.1 (0x00007f219d9a5000)
	libdl.so.2 => /lib64/libdl.so.2 (0x00007f219d99e000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f219d999000)
	librt.so.1 => /lib64/librt.so.1 (0x00007f219d994000)
```

which makes me wonder, how was this working before without explicitly installing that.

container-images/cuda/Containerfile
```diff
@@ -1,40 +1,39 @@
...
     git reset --hard ${WHISPER_CPP_SHA} && \
-    cmake -B build -DGGML_CUDA=ON -DCUDA_DOCKER_ARCH=${CUDA_DOCKER_ARCH} -DCMAKE_EXE_LINKER_FLAGS=-Wl,--allow-shlib-undefined . && \
+    cmake -B build -DGGML_CUDA=ON -DCUDA_DOCKER_ARCH=${CUDA_DOCKER_ARCH} \
+      -DCMAKE_INSTALL_PREFIX=${INSTALL_PREFIX} \
+      -DCMAKE_EXE_LINKER_FLAGS=-Wl,--allow-shlib-undefined . && \
     cmake --build build --config Release -j$(nproc) && \
-    mv build/bin/main /usr/bin/whisper-main && mv build/bin/server /usr/bin/whisper-server && \
-    if [ -f build/lib/libwhisper.so ]; then mv build/lib/libwhisper.so /usr/lib/; fi && \
+    cmake --install build && \
+    mv build/bin/main ${INSTALL_PREFIX}/bin/whisper-main && \
```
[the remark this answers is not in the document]
ericcurtin: This line installs the libwhisper stuff:

```
$ ls /usr/lib64/libw*
/usr/lib64/libwhisper.so  /usr/lib64/libwhisper.so.1  /usr/lib64/libwhisper.so.1.7.1
```

container-images/cuda/Containerfile
bmahabirbu: So originally the container would use libllama.so and libggml.so from the build of llama.cpp with cuda. I added those files to /usr/lib and updated the linker to point to those instead (following the example from the official dockerfile). 

Even before that I already had those packages installed in the container. Its most likely due to cuda container toolkit passing the library from the nvidia driver on the host to the container.

ericcurtin:
> I tested it and it crashed before loading the model with a segmentation fault. I originally thought it had something to do with the linker not updating the correct libggml.so file but its just actually do to the second cmakle --install line for whisper. It renames some llama stuff despite the mv commands causing issues

Yes I see what you mean, we have two different versions of /usr/lib64/libggml.so clashing

ericcurtin:
We must somehow align them to the same version...

ericcurtin:
We have two options, split llama.cpp and whisper.cpp into two containerfiles

Or...

Ensure at build-time of the container images that llama.cpp and whisper.cpp versions are ABI compatible...

ericcurtin:
I fixed it by building whisper with -DBUILD_SHARED_LIBS=NO , so using static linking for whisper.cpp seems like an ok option for now.

bmahabirbu:
> I fixed it by building whisper with -DBUILD_SHARED_LIBS=NO , so using static linking for whisper.cpp (but not llama.cpp) seems like an ok option for now.

Nice, I'll give it a test run as well

container-images/cuda/Containerfile
ericcurtin: Should we remove nvidia-driver-cuda-libs? Is it breaking nvidia-container-toolkit?

ericcurtin: I wonder can we CI some of this stuff even without Nvidia hardware in the CI runners

bmahabirbu:
> > I fixed it by building whisper with -DBUILD_SHARED_LIBS=NO , so using static linking for whisper.cpp (but not llama.cpp) seems like an ok option for now.
> 
> Nice, I'll give it a test run as well

Confirmed working! (Had to remove the dnf install nvidia-driver-cuda-libs but that's just due to me setting my environment up with cuda container toolkit)

ericcurtin:
> > > I fixed it by building whisper with -DBUILD_SHARED_LIBS=NO , so using static linking for whisper.cpp (but not llama.cpp) seems like an ok option for now.
> > 
> > 
> > Nice, I'll give it a test run as well
> 
> Confirmed working! (Had to remove the dnf install nvidia-driver-cuda-libs but that's just due to me setting my environment up with cuda container toolkit)

Well lets remove it if that's the case, makes me wonder if somebody could run this without the nvidia-container-toolkit... But anyway don't want to break a working setup

bmahabirbu:
> > > > I fixed it by building whisper with -DBUILD_SHARED_LIBS=NO , so using static linking for whisper.cpp (but not llama.cpp) seems like an ok option for now.
> > > 
> > > 
> > > Nice, I'll give it a test run as well
> > 
> > 
> > Confirmed working! (Had to remove the dnf install nvidia-driver-cuda-libs but that's just due to me setting my environment up with cuda container toolkit)
> 
> Well lets remove it if that's the case, makes me wonder if somebody could run this without the nvidia-container-toolkit... But anyway don't want to break a working setup

Sounds good to me! I'll investigate some more to see if there is a way to run it without the container toolkit

PR #411, opened 2024-11-04
Use subpath for OCI Models

rhatdan:
Adding subpath=/models to the Mount command in quadlet

Currently model-cars store AI Model in /models subdir but standard model-raw are storing them in /.

Changing bot model-cars and model-raw to use the same /models directory allows quadlets to support either model.

The big difference between model cars is that they include executables with the image, where a model-raw is just the model.

Also found a crash when doing ramalama list with images without a tag.

Improved tests to check model type flags.

### DIFF
--- a/ramalama/oci.py
+++ b/ramalama/oci.py
@@ -25,7 +25,7 @@ def list_models(args):
         "--filter",
         f"label={ocilabeltype}",
         "--format",
-        '{"name":"oci://{{ index .Names 0 }}","modified":"{{ .Created }}","size":"{{ .Size }}"},',
+        '{"name":"oci://{{ .Repository }}:{{ .Tag }}","modified":"{{ .Created }}","size":"{{ .Size }}"},',
     ]
     output = run_cmd(conman_args, debug=args.debug).stdout.decode("utf-8").strip()
     if output == "":
@@ -95,11 +95,11 @@ pip install omlmd
         model_name = os.path.basename(source)
         model_raw = f"""\
 FROM {args.image} as builder
-RUN mkdir -p /mnt/models; cd /mnt/models; ln -s {model_name} model.file
+RUN mkdir -p /models; cd /models; ln -s {model_name} model.file
 
 FROM scratch
-COPY --from=builder /mnt/models /
-COPY {model} /{model_name}
+COPY --from=builder /models /models
+COPY {model} /models/{model_name}
 LABEL {ociimage_raw}
 """
         model_car = f"""\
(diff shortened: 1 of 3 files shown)


ramalama/oci.py
```diff
@@ -95,11 +95,11 @@ def _build(self, source, target, args):
...
         model_raw = f"""\
 FROM {args.image} as builder
-RUN mkdir -p /mnt/models; cd /mnt/models; ln -s {model_name} model.file
+RUN mkdir -p /models; cd /models; ln -s {model_name} model.file
 
 FROM scratch
-COPY --from=builder /mnt/models /
-COPY {model} /{model_name}
+COPY --from=builder /models /models
+COPY {model} /models/{model_name}
```
tarilabs, line 102: would it be possible to annotate this layer from the manifest, or the manifest.config ?

maybe this can be used when you want to pull only the "raw" portion, given a [KServe Modelcar format](https://kserve.github.io/website/0.13/modelserving/storage/oci/#prepare-an-oci-image-with-model-data).

i.e. you are pulling `quay.io/my/modelcar` but in this [ramalama] context, you are interested in only pulling the model layer(s).

wdyt?

rhatdan: I have working code to generate a new style k8s yaml file, but in a different local PR.  It is hung up on the support for subPath in k8s.

rhatdan: ./bin/ramalama --runtime=vllm serve --name MyGraniteModel --generate=kube oci://quay.io/rhatdan/tiny-car:latest
# Save the output of this file and use kubectl create -f to import
# it into Kubernetes.
#
# Created with ramalama-0.0.4
apiVersion: v1
kind: Deployment
metadata:
  labels:
    app: MyGraniteModel
  name: MyGraniteModel
spec:
  containers:
  - name: MyGraniteModel
    image: quay.io/ramalama/ramalama:latest
    command: ["vllm"]
    args: ['serve', '--port', '8080', 'quay.io/rhatdan/tiny-car:latest']
    ports:
    - containerPort: 8080
    volumeMounts:
    - mountPath: /mnt/models
      name: model
    - mountPath: /dev/dri
      name: dri
  volumes:
  - name model
    image:
      reference: "quay.io/rhatdan/tiny-car:latest"
      pullPolicy: IfNotPresent
      subPath: "/models"
  - name dri
    hostPath:
      path: /dev/dri

rhatdan: Theoretically this kube.yaml could take your model.car design and allow it to be mounted directly into a container/pod, No need to run a separate sleep process or to search throgh /proc/PID ...  for the rootfs of the image.

tarilabs: right, the KEP-4639 if complemented by subPath will likely benefit similar for KServe deployments, for the time being I think having the models in the `/models` and linking, plus annotation (not in this PR) would benefit as an ad-interim solution; wdyt ?

PR #415, opened 2024-11-06
Bump to v0.0.22

rhatdan:


ericcurtin:
I'm not sure if test/ci.sh is actually running in our CI, the check could be extended to something like this also:

```
local version=$(grep "^version =.*" pyproject.toml | sed s/\"//g | grep -oE '[^ ]+$')
grep version.*$version setup.py
grep version.*$version rpm/python-ramalama.spec
```

doesn't ha

PR #417, opened 2024-11-06
Fix nocontainer mode

ericcurtin:
When using nocontainer we should just reference the file directly.


rhatdan:
I am taking this over.

PR #428, opened 2024-11-07
Remove omlmd as a dependency

ericcurtin:
It's blocking packaging


rhatdan:
So you want to remove it from packaging, but not from the code?

PR #430, opened 2024-11-07
Install llama-cpp-python[server]

ericcurtin:
It implements more of the OpenAI spec for servers than llama.cpp upstream.

## Summary by Sourcery

Build:
- Add installation of llama-cpp-python[server] to the build script for enhanced server capabilities.

### DIFF
--- a/container-images/scripts/build_llama_and_whisper.sh
+++ b/container-images/scripts/build_llama_and_whisper.sh
@@ -1,34 +1,85 @@
 #!/bin/bash
 
+dnf_install() {
+  local rpm_list=("python3" "python3-pip" "gcc-c++" "cmake" "vim" "procps-ng" \
+                  "git")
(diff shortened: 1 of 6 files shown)


container-images/scripts/build_llama_and_whisper.sh
```diff
@@ -29,6 +58,9 @@ main() {
   mv build/bin/server "$install_prefix/bin/whisper-server"
   cd ..
   rm -rf whisper.cpp
+
+  CMAKE_ARGS="${common_flags[*]}" FORCE_CMAKE=1 \
+    pip install --prefix="$install_prefix" 'llama-cpp-python[server]'
```
ericcurtin, line 63: This is the significant change in this PR, the rest is just address maintenance concerns from a previous PR and doing further de-duplication.

rhatdan:
LGTM

PR #433, opened 2024-11-08
When running on macOS via ramalama run we were throwing exceptions

ericcurtin:
Because in the "run" flow, generate isn't even a present argument

## Summary by Sourcery

Bug Fixes:
- Fix exception thrown on macOS when running via ramalama by checking for the presence of the 'generate' argument.

### DIFF
--- a/ramalama/model.py
+++ b/ramalama/model.py
@@ -216,7 +216,7 @@ class Model:
 
     def exec_model_in_container(self, model_path, cmd_args, args):
         if not args.container:
-                return False
+            return False
         conman_args = self.setup_container(args)
         if len(conman_args) == 0:
             return False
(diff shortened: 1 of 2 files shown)


ramalama/model.py
```diff
@@ -237,6 +237,9 @@ def exec_model_in_container(self, model_path, cmd_args, args):
         exec_cmd(conman_args, args.debug, debug=args.debug)
         return True
 
+    def not_args_generate(self, args):
```
rhatdan, line 240: Call this args_generate and then use it with the not flag.

rhatdan:
I think  #434 is a better solution.

We need a way to test run before opening PRs.

PR #435, opened 2024-11-08
Make it easier to test-run manually

rhatdan:
NO need to leak ramalama storage into container any longer.

Since we are disabling SELinux don't add Z.

## Summary by Sourcery

Enhance the test infrastructure by introducing a new helper function to skip tests in non-local environments and updating a specific test to use this function. Add a new Makefile target to simplify manual test execution. Remove unnecessary SELinux options from container setup and adjust volume mounting logic.

Enhancements:
- Improve the test system by adding a new helper function 'skip_if_notlocal' to conditionally skip tests not supported in non-local environments.

Tests:
- Modify the system test 'ramalama run tiny with prompt' to include a conditional skip for non-local environments, ensuring more reliable test execution.

Chores:
- Add a new 'test-run' target to the Makefile to facilitate manual test execution with specific environment variables.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,6 +95,11 @@ lint:
 codespell:
 	codespell --dictionary=- -w
 
+.PHONY: test-run
+test-run:
+	_RAMALAMA_TEST=local RAMALAMA=$(CURDIR)/bin/ramalama bats -T test/system/030-run.bats
+	_RAMALAMA_OPTIONS=--nocontainer _RAMALAMA_TEST=local bats -T test/system/030-run.bats
+
 .PHONY: validate
 validate: codespell lint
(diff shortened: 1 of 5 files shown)


rhatdan:
@ericcurtin @bmahabirbu PTAL

bmahabirbu:
> @ericcurtin @bmahabirbu PTAL

Nice! I tested 030-run.bats and it correctly identified when Ramalama is working vs not. 

I'm wondering if it would be worth capturing the stderr and stdout for a more detailed failure for certain exit codes, I can look into this!

rhatdan:
Yes I wish that we could test this in CI/CD system, but everytime I turn it on, it blows up.

PR #441, opened 2024-11-11
Only do dnf install for cuda images

ericcurtin:
Should fix container build for Vulkan images

## Summary by Sourcery

Enhancements:
- Modify the dnf_install function to conditionally install packages only for CUDA images, avoiding unnecessary installations for Vulkan images.


rhatdan:
LGTM

PR #443, opened 2024-11-12
Set --host option when the server will end up in a container

### LINKED ISSUE #442: `ramalama serve` doesn't work when podman is installed
### What is the problem:

When running ramalama on a computer with podman installed, if you try to setup a server through the `ramalama` serve. The server will be inaccesible from the host computer.

### How to reproduce

* Install ramalama and podman on the same computer
* Start ramalama
```sh
ramalama serve tiny
```
*Run the following script on the host server
```py
import requests

url = "http://localhost:8080/v1/chat/completions"
headers = {
    "Content-Type": "application/json"
}
data = {
    "messages": [
        {"role": "user", "content": "Hello, world!"}
    ]
}
```

### What is happening?

First, here is the difference between running ramalama with and without podman installed.

Without podman installed

```sh
ramalama --dryrun serve tiny
> os.execvp(llama-server, ['llama-server', '--port', '8080', '-m', '/path/to/model'])
```

With podman installed

```sh
ramalama --dryrun serve tiny
> podman run --rm -i --label RAMALAMA ... quay.io/ramalama/ramalama:latest llama-server --port 8080 -m /path/to/mod
```

As it can be seen, both run the `llama-server` command without the --host parameter, this defaults `llama-server` to listen to `127.0.0.1`, which makes sense on the first case, but makes it so that it fails to listen from requests from the host environment on the second case.

I was able to verify this by entering the container through `podman exec` and being successfully able to run the python script from inside the container.

grillo-delmal:
A small patch to extend the use of the --host '0.0.0.0' argument for llama-server to the cases when the app will end up being run in a container.

Fixes #442

## Summary by Sourcery

Bug Fixes:
- Fix the server configuration to correctly set the '--host' option to '0.0.0.0' when the application is run in a container.

### DIFF
--- a/ramalama/model.py
+++ b/ramalama/model.py
@@ -317,7 +317,7 @@ class Model:
         else:
             if args.gpu:
                 exec_args.extend(self.gpu_args())
-            if in_container():
+            if in_container() or (args.container and args.engine):
                 exec_args.extend(["--host", "0.0.0.0"])
 
         if args.generate == "quadlet":


ericcurtin (approved):

rhatdan:
This would only effect when ramalama is running in a container, which does not happen by default any longer.

grillo-delmal:
> This would only effect when ramalama is running in a container, which does not happen by default any longer.

ramalama does run llama-server inside the container by default if a container engine is available.

Since currently `--host '0.0.0.0'` won't be added unless ramalama itself is ran inside a container, you wont be able to access the server while running `ramalama serve` (as described in #442)

This patch fixes that specific case by checking both `args.engine` and `args.container`. Since `args.engine` gets populated [0][1] when a container engine is available [2], and `args.container` [3][4] is true if ramalama is not runing on a container and there is an available container engine [5]. 

[0] https://github.com/containers/ramalama/blob/main/ramalama/cli.py#L183
[1] https://github.com/containers/ramalama/blob/main/ramalama/cli.py#L96
[2] https://github.com/containers/ramalama/blob/main/ramalama/common.py#L29-L54

[3] https://github.com/containers/ramalama/blob/main/ramalama/cli.py#L166
[4] https://github.com/containers/ramalama/blob/main/ramalama/cli.py#L94
[5] https://github.com/containers/ramalama/blob/main/ramalama/cli.py#L36-L40

ericcurtin:
The change makes sense to me, the `in_container()` check only applied if you explicitly run RamaLama python script in a container. Whereas before the check was kinda different as RamaLama python script was implicity re-exec'd in the container.

We could also drastically simplify this by just doing 0.0.0.0 in all cases, that's what I'd do if I opened this PR. But happy with this way also.

rhatdan:
I believe the way to fix this is https://github.com/containers/ramalama/pull/444.

If we do --host 0.0.0.0 always, this means that outside of the container, the service would be listening on the host ports, which means if people outside of the VM wanted to connect to the service, they could (As long as the firewall is open).  Which is probably a better default.

rhatdan:
Bottom line from this discussion, is we should open this up to allow users to specify which host they want to listen on.

ericcurtin:
Thanks for pointing this out @grillo-delmal I merged the other one

PR #446, opened 2024-11-12
Fix RamaLama container image build

ericcurtin:
Error:
(try to add '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)
 Problem: conflicting requests
  - nothing provides libLLVM-17.so()(64bit) needed by mesa-vulkan-drivers-23.3.3-102.el9.x86_64 from copr:copr.fedorainfracloud.org:slp:mesa-krunkit
  - nothing provides libLLVM-17.so(LLVM_17)(64bit) needed by mesa-vulkan-drivers-23.3.3-102.el9.x86_64 from copr:copr.fedorainfracloud.org:slp:mesa-krunkit
Error: building at STEP "RUN chmod +x /scripts/*.sh &&     /scripts/build_llama_and_whisper.sh "ramalama" "$LLAMA_CPP_SHA"       "$WHISPER_CPP_SHA" "/usr" "-DGGML_KOMPUTE=1"": while running runtime: exit status 1
make: *** [Makefile:78: build] Error 1

PR #447, opened 2024-11-12
Document the host flag in ramalama.conf file

rhatdan:


docs/ramalama-serve.1.md
```diff
@@ -39,7 +39,7 @@ Generate specified configuration format for running the AI Model as a service
 show this help message and exit
 
 #### **--host**="0.0.0.0"
-ip address to listen
```
ericcurtin, line 42: Doesn't block merge, but IP address to listen made more sense to me, we will have at least 2 other servers soon llama-cpp-python and vllm

ericcurtin (approved):

ericcurtin:
Looks like https://github.com/containers/ramalama/pull/446 will fix the build

PR #452, opened 2024-11-13
Add llama-run

ericcurtin:
And update llama.cpp and whisper.cpp.

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -215,6 +215,7 @@ The RAMALAMA_IN_CONTAINER environment variable modifies default behaviour.""",
     )
     parser.add_argument("-v", "--version", dest="version", action="store_true", help="show RamaLama version")
 
+
 def configure_subcommands(parser):
     """Add subcommand parsers to the main argument parser."""
     subparsers = parser.add_subparsers(dest="subcommand")
--- a/container-images/asahi/Containerfile
+++ b/container-images/asahi/Containerfile
@@ -1,11 +1,11 @@
 FROM fedora:41
 
-ARG LLAMA_CPP_SHA=1329c0a75e6a7defc5c380eaf80d8e0f66d7da78
+ARG LLAMA_CPP_SHA=0827b2c1da299805288abbd556d869318f2b121e
 # renovate: datasource=git-refs depName=ggerganov/whisper.cpp packageName=https://github.com/ggerganov/whisper.cpp gitRef=master versioning=loose type=digest
-ARG WHISPER_CPP_SHA=6266a9f9e56a5b925e9892acf650f3eb1245814d
+ARG WHISPER_CPP_SHA=3de9deead5759eb038966990e3cb5d83984ae467
 
 COPY ../scripts /scripts
 RUN chmod +x /scripts/*.sh && \
     /scripts/build_llama_and_whisper.sh "asahi" "$LLAMA_CPP_SHA" \
-      "$WHISPER_CPP_SHA" "/usr" "-DGGML_VULKAN=1"
+      "$WHISPER_CPP_SHA"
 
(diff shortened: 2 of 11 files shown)


ericcurtin:
@cooktheryan @lsm5 @mrunalp @slp @rhatdan @tarilabs @umohnani8 @ygalblum PTAL

ygalblum (commented):
LGTM for the code.
But, I think I'm missing something. Is `llama-cpp-python` already a dependency?

ericcurtin:
@ygalblum we probably need to push some container images before merging this, but when we do that, we should be all good.

ramalama/cli.py
```diff
@@ -103,7 +103,7 @@ def load_and_merge_config():
         )
 
     ramalama_config['carimage'] = ramalama_config.get('carimage', "registry.access.redhat.com/ubi9-micro:latest")
-    ramalama_config['runtime'] = ramalama_config.get('runtime', 'llama.cpp')
+    ramalama_config['runtime'] = ramalama_config.get('runtime', 'llama-cpp-python')
```
rhatdan, line 106: Looks like we have a different runtime for ramalama serve versus ramalama run?

ericcurtin: For ramalama run, the only command that are available to do that in a single process are the llama.cpp ones, like llama-cli (and llama-run which I'll be adding in this PR now that the last few bits were merged a couple of hours ago). There are other possibilities for chatbots, but the other require us to set up a server process and a client process. I think llama-run is cleaner than setting up a client and server.

As regards server there are 3 options, vLLM, llama.cpp and llama-cpp-python (which is just a python wrapper around llama.cpp). Although vLLM is the best enterprise option today for a server in my opinion, it lacks the diverse hardware support for local development, which leaves us the other two. llama.cpp server is a very minimal implementation of an OpenAI-compatible server as in it is barely OpenAI-compatible, last I tried it, it didn't work with OpenAI-compatible clients like: https://github.com/open-webui/open-webui

llama.cpp server might catch up as it is actively being contributed to, but it's not there today. One key feature it is missing is multi-model support, so the client normally specifies the model it wants to use.

llama-cpp-python is further ahead in terms of it's OpenAI-compatible server implementation, that's why I'm proposing we switch in this PR.

ericcurtin:
Adding temperature option to llama-run here:

https://github.com/ggerganov/llama.cpp/pull/10899

ramalama/cli.py
rhatdan: My opinion here is choice is bad.  We should only offer llama.cpp and if the server tool llama-cpp-python is better, then default to using that, if a better one shows up in the future, then we could move to it, as long as it implements the Same API as llama-cpp-python.  We are doing the same thing on the run side moving from llama-client to llama-run.

Bottom line is keep the selection to llama.cpp and vLLM (And whisper.cpp) and we can modify the tools used to work with the libraries under the covers.  Eventually we could allow specifying the commands to run within the containers for run and serve, but that is a ways a way.

ramalama/cli.py
```diff
@@ -205,8 +205,8 @@ def configure_arguments(parser):
     parser.add_argument(
         "--runtime",
         default=config.get("runtime"),
-        choices=["llama.cpp", "vllm"],
-        help="specify the runtime to use; valid options are 'llama.cpp' and 'vllm'",
+        choices=["llama-cpp-python", "llama.cpp", "vllm"],
```
rhatdan, line 208: As stated above, do not give this choice, just stick with llama.cpp and use llama-cpp-python server for serve

ericcurtin:
I reduced the scope of this. One of the main drivers for llama-cpp-python server is to attempt to find a server that's compatible with:

https://github.com/open-webui/open-webui

we probably need to explore a few options like llama-server, llama-cpp-python and https://github.com/mostlygeek/llama-swap

ericcurtin:
This is ready to go @rhatdan

rhatdan:
LGTM

PR #454, opened 2024-11-14
Switch to llama-simple-chat

ericcurtin:
This is a new chat program in llama.cpp which is much simpler than the existing one we were using. It doesn't have the debug/verbose output problem and just seems higher quality in general for a simple chatbot, it's a few 100 lines of code.

## Summary by Sourcery

Switch to using llama-simple-chat for a more streamlined and higher quality chat experience, removing verbose options and updating container configurations.

New Features:
- Introduce llama-simple-chat as the new chat program, replacing the previous implementation.

Enhancements:
- Simplify the chat execution process by removing unnecessary arguments and options.

Build:
- Update the LLAMA_CPP_SHA in the container images to a new commit hash.

### DIFF
--- a/container-images/cuda/Containerfile
+++ b/container-images/cuda/Containerfile
@@ -1,7 +1,7 @@
 # Base image with CUDA for compilation
 FROM docker.io/nvidia/cuda:12.6.2-devel-ubi9 AS builder
 
-ARG LLAMA_CPP_SHA=1329c0a75e6a7defc5c380eaf80d8e0f66d7da78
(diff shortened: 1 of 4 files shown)


ericcurtin:
This recently added llama-simple-chat program is more suited to our needs. Maybe we can actively start contributing to it, I already started:

https://github.com/ggerganov/llama.cpp/pull/10291

ericcurtin:
@cooktheryan @lsm5 @mrunalp @slp @rhatdan @tarilabs @umohnani8 @ygalblum PTAL

rhatdan:
LGTM

Issue #471, opened 2024-11-20
Local cuda container build fails with "unsupported instruction `vpdpbusd'"


nzwulfin:
Trying to build on my home system, `./container_build.sh cuda` will fail with the following error
```
/tmp/ccnKypuJ.s: Assembler messages:
/tmp/ccnKypuJ.s:31871: Error: unsupported instruction `vpdpbusd'
/tmp/ccnKypuJ.s:31926: Error: unsupported instruction `vpdpbusd'
/tmp/ccnKypuJ.s:31995: Error: unsupported instruction `vpdpbusd'
/tmp/ccnKypuJ.s:32060: Error: unsupported instruction `vpdpbusd'
/tmp/ccnKypuJ.s:32113: Error: unsupported instruction `vpdpbusd'
gmake[2]: *** [ggml/src/CMakeFiles/ggml.dir/build.make:132: ggml/src/CMakeFiles/ggml.dir/ggml-quants.c.o] Error 1
gmake[2]: *** Waiting for unfinished jobs....
gmake[1]: *** [CMakeFiles/Makefile2:1591: ggml/src/CMakeFiles/ggml.dir/all] Error 2
gmake: *** [Makefile:146: all] Error 2
```

From what I can tell online, this is due to [the binutils in RHEL 9 not being new enough](https://access.redhat.com/solutions/7049696) to support the instruction.

I made some progress by adding the GCC Toolset 12 to the cuda portion of the dnf_install switch statement, but I'm not familiar enough with what needs to really get set to use the toolset correctly. I expect that `scl enable` is doing a lot more with the path than I exported.

```
    dnf install -y gcc-toolset-12 
    export CC=/opt/rh/gcc-toolset-12/root/usr/bin/gcc
    export CCXX=/opt/rh/gcc-toolset-12/root/usr/bin/g++
```

I've hit my limit for testing but thought I'd report the issue anyhow.

nzwulfin:
I examined a UBI 9 container with the toolset and the CUDA dev container and brute forced a few more exports for the build to complete. I don't think this is the right solution, but might serve as a pointer to one.

```
  elif [ "$containerfile" = "cuda" ]; then
    dnf install -y "${rpm_list[@]}"
    dnf install -y gcc-toolset-12 
    export CC=/opt/rh/gcc-toolset-12/root/usr/bin/gcc
    export CCXX=/opt/rh/gcc-toolset-12/root/usr/bin/g++
[...]

PR #473, opened 2024-11-20
Enable GCC Toolet 12 to support AVX VNNI

nzwulfin:
Fixes issue #471 based on a Red Hat KBase article (subscription required), to use AVX VNNI on platforms that support it a later version of binutils is needed.

This adds the GCC 12 Toolset and directly sources the environment vars into the build script.

There are also toolsets available for GCC 13 and 14 available if there's a need for more current versions. They should work the same way.

## Summary by Sourcery

Build:
- Add GCC Toolset 12 to the build script for AVX VNNI support.


nzwulfin:
Thanks @rhatdan, should be signed now. I don't need to update this PR right?

rhatdan:
@bmahabirbu PTAL
LGTM

bmahabirbu:
Looks good!

PR #475, opened 2024-11-20
Fall back to huggingface-cli when pulling via URL fails

rhatdan:
Handle non GGUF files as well.

## Summary by Sourcery

Implement a fallback mechanism to use 'huggingface-cli' for model pulling when URL-based pulling fails, and fix the argument path for the 'vllm serve' command. Update system tests to reflect these changes.

Bug Fixes:
- Fix the argument path for the 'vllm serve' command to use the directory path instead of the model file path.

Enhancements:
- Implement a fallback mechanism to use 'huggingface-cli' for model pulling when URL-based pulling fails.

Tests:
- Update system test to verify the correct arguments for the 'vllm serve' command.

### DIFF
--- a/ramalama/huggingface.py
+++ b/ramalama/huggingface.py
@@ -1,7 +1,7 @@
 import os
 import pathlib
 import urllib.request
-from ramalama.common import available, run_cmd, exec_cmd, download_file, verify_checksum
+from ramalama.common import available, run_cmd, exec_cmd, download_file, verify_checksum, perror
 from ramalama.model import Model
 
 missing_huggingface = """
@@ -45,8 +45,7 @@ class Huggingface(Model):
 
     def login(self, args):
         if not self.hf_cli_available:
-            print("huggingface-cli not available, skipping login.")
-            return
+            raise NotImplementedError("huggingface-cli not available, skipping login.")
         conman_args = ["huggingface-cli", "login"]
         if args.token:
             conman_args.extend(["--token", args.token])
@@ -54,8 +53,7 @@ class Huggingface(Model):
 
     def logout(self, args):
         if not self.hf_cli_available:
-            print("huggingface-cli not available, skipping logout.")
-            return
+            raise NotImplementedError("huggingface-cli not available, skipping logout.")
         conman_args = ["huggingface-cli", "logout"]
         if args.token:
             conman_args.extend(["--token", args.token])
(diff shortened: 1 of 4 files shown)


ramalama/huggingface.py
```diff
@@ -69,6 +67,28 @@ def pull(self, args):
         symlink_dir = os.path.dirname(model_path)
         os.makedirs(symlink_dir, exist_ok=True)
 
+        try:
+            return self.url_pull(args, model_path, directory_path)
+        except (urllib.error.HTTPError, urllib.error.URLError) as e:
```
bentito, line 72:
```suggestion
        except (urllib.error.HTTPError, urllib.error.URLError, KeyError) as e:
```
It's actually a KeyError which is then shown to user as 401 Auth err. Adding KeyError here allows it to fallback (for me anyways)

rhatdan:
@bentito PTAL

bentito:
> @bentito PTAL

When I run latest, I am seeing:

```bash
$ bin/ramalama --debug pull huggingface://ibm-granite/granite-3.0-8b-instruct
URL pull failed and huggingface-cli not available
Error: "Failed to pull model: 'failed to pull https://huggingface.co/ibm-granite/raw/main/granite-3.0-8b-instruct: HTTP Error 401: Unauthorized'"
ramalama fork/rhatdan/huggingface $ huggingface-cli version
usage: huggingface-cli <command> [<args>]
huggingface-cli: error: argument {env,login,whoami,logout,repo,upload,download,lfs-enable-largefiles,lfs-multipart-upload,scan-cache,delete-cache}: invalid choice: 'version' (choose from 'env', 'login', 'whoami', 'logout', 'repo', 'upload', 'download', 'lfs-enable-largefiles', 'lfs-multipart-upload', 'scan-cache', 'delete-cache')
```

Did the usage of hf-cli change from before, because with my suggest mods it was working. See my line comment.

ramalama/huggingface.py
```diff
@@ -69,6 +67,24 @@ def pull(self, args):
         symlink_dir = os.path.dirname(model_path)
         os.makedirs(symlink_dir, exist_ok=True)
 
+        try:
+            return self.url_pull(args, model_path, directory_path)
+        except (urllib.error.HTTPError, urllib.error.URLError, KeyError) as e:
+            if not self.hf_cli_available:
```
bentito, line 73: This isn't working for me, I'm ending up with it showing my hf_cli is not available even though it is.

bentito:
```suggestion
            if self.hf_cli_available:
```
not is reversing the logic, right?

rhatdan: yep

rhatdan:
@bentito PTANL

bentito:
/lgtm

Works for me now!

bentito (approved):
all working for me locally now!

PR #478, opened 2024-11-22
Update model.py to enable CUDA when available

FNGarvin:
Shell out to nvidia-smi for NVidia detection, add device or gpu args to conman as appropriate for docker vs podman.  Import subprocess (to shell out for nvidia-smi) and shutil (to duplicate available() functionality and test whether we're using podman or docker)

## Summary by Sourcery

Enhancements:
- Enable CUDA support in container setup by detecting NVidia GPUs using nvidia-smi and configuring container arguments accordingly.

### DIFF
--- a/ramalama/model.py
+++ b/ramalama/model.py
@@ -152,6 +152,16 @@ class Model:
         gpu_type, gpu_num = get_gpu()
         if gpu_type == "HIP_VISIBLE_DEVICES" or gpu_type == "ASAHI_VISIBLE_DEVICES":
             conman_args += ["-e", f"{gpu_type}={gpu_num}"]
+
+        #podman is not a drop-in replacement for docker... need to "see" which we're in here, ala common.py->available logic
+        #risk of duplicating code here, maint problem
+        if gpu_type == "CUDA_VISIBLE_DEVICES":
+            if shutil.which("podman"):
+                #todo is /all/ appropriate?
+                conman_args += ["--device", "nvidia.com/gpu=all"] #AFAIK, cuda requires this for podman
+            else:
+                conman_args += ["--gpus", "all"] #and this for Docker
+        
         return conman_args
 
     def run_container(self, args, shortnames):
@@ -389,7 +399,15 @@ def get_gpu():
             content = file.read()
             if "asahi" in content.lower():
                 return "ASAHI_VISIBLE_DEVICES", 1
-
+                
+    try:
+        #TODO I don't currently have access to a PC w/ multiple NVidia GPUs nor an NVidia Mac...  but I *think* that
+        #every Linux and Windows machine having modern NVidia will have nvidia-smi and that the number of lines corresponds to the number of zero-indexed gpus
+        check_output = subprocess.run(['nvidia-smi', '-L'], check=True, capture_output=True) #shell to nvidia-smi
+        if not check_output.returncode: #if command EXIT_SUCCESS
+            return "CUDA_VISIBLE_DEVICES", len(check_output.stdout.splitlines()) # ret cuda, #gpus?
+    except Exception: {} #fall through
+    
     return None, None
 
 
--- a/ramalama/model.py
+++ b/ramalama/model.py
@@ -3,6 +3,8 @@ import sys
 import glob
 import atexit
 import shlex
+import subprocess
+import shutil
 
 from ramalama.common import (
     default_image,


ramalama/model.py
```diff
@@ -152,6 +154,16 @@ def setup_container(self, args):
         gpu_type, gpu_num = get_gpu()
         if gpu_type == "HIP_VISIBLE_DEVICES" or gpu_type == "ASAHI_VISIBLE_DEVICES":
             conman_args += ["-e", f"{gpu_type}={gpu_num}"]
+
+        #podman is not a drop-in replacement for docker... need to "see" which we're in here, ala common.py->available logic
+        #risk of duplicating code here, maint problem
+        if gpu_type == "CUDA_VISIBLE_DEVICES":
+            if shutil.which("podman"):
```
rhatdan, line 161: args.engine or conman shouldhave this value.  We also have a function available for discovering executables.

ramalama/model.py
```diff
@@ -389,7 +401,15 @@ def get_gpu():
             content = file.read()
             if "asahi" in content.lower():
                 return "ASAHI_VISIBLE_DEVICES", 1
-
+                
+    try:
+        #TODO I don't currently have access to a PC w/ multiple NVidia GPUs nor an NVidia Mac...  but I *think* that
+        #every Linux and Windows machine having modern NVidia will have nvidia-smi and that the number of lines corresponds to the number of zero-indexed gpus
+        check_output = subprocess.run(['nvidia-smi', '-L'], check=True, capture_output=True) #shell to nvidia-smi
```
rhatdan, line 408: Use run_cmd

rhatdan:
Thanks @FNGarvin.  I found some issues.

ramalama/model.py
FNGarvin: Good change, thank you.

bmahabirbu: I tested between docker and podman both can use `--gpus=all` so there's no need to check which instance is running.

I tested on podman version 3.4.4 and 5.2.5 as well! Speaking of, the ubuntu package for podman is out of date but useful for testing stuff like this!

ramalama/model.py
bmahabirbu: We can go a step further and query the nvidia-smi command itself to get more info! For example doing this command `nvidia-smi --query-gpu=index,memory.total --format=csv,noheader,nounits | sort -t, -k2 -nr | head -n 1` Can get us the largest vram GPU and id formatted as "id, vram-in-mb".

We can do something like this

```
try:
        check_output = subprocess.run(['nvidia-smi', '-L'], check=True, capture_output=True) #shell to nvidia-smi
 if not check_output.returncode: #if command EXIT_SUCCESS
            return "CUDA_VISIBLE_DEVICES", len(check_output.stdout.splitlines()) # ret cuda, #gpus?
    except Exception: {} #fall through

    return None, None
```

ramalama/model.py
FNGarvin: --gpus=all does not work for me on Podman.  I have seen some talk about getting that feature supported for better parity in some other repos, but if changes have been made they certainly aren't yet ubiquitous.  My motivation, selfishly, was to get this working for myself... which is why I chose to use --device for podman and --gpu for Docker (as per all the NVidia support docs I've read).  

Is it a problem to have different conman args according to underlying container, or just not as tidy as one would hope (admittedly, I am not a Python coder and my code is not tidy).

ramalama/model.py
FNGarvin: I think that's a fabulous idea.  My inclination in general is to avoid shelling out, but I don't think we're going to find a better or more lightweight way to test for the presence of NVidia and CUDA.  Probably why all the NVidia Container Toolkit docs seem to use it as a sanity check for installs.

bmahabirbu: Same I'd rather not shell out myself but doing it this way avoids complications with different systems. if we can assume a system has an nvidia GPU then most likely there will be drivers installed with nvidia-smi as well. 

Right now the vulkan backed for llama.cpp doesn't have all the functionality like cuda and hip blas does. But later down the line id like to switch to vulkan and use the vulkan SDK to query GPU data since it supports amd nvidia and intel graphics.

ramalama/model.py
FNGarvin: Thank you for circling back to check with me, mate.  It would NOT have felt great for the change to go through that would needlessly break my ability to use the code.  I think the included session capture should be enough to eliminate user error as the issue, but lmk if you need more data or see something missing.

```bash
Perilous to all of us are the devices of an art deeper than we ourselves
possess.
                -- Gandalf the Grey [J.R.R. Tolkien, "Lord of the Rings"]
fng@CYGNUS:~$ uname -a
Linux CYGNUS 5.15.167.4-microsoft-standard-WSL2 #1 SMP Tue Nov 5 00:21:55 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
fng@CYGNUS:~$ cat /etc/os-release
PRETTY_NAME="Ubuntu 24.04.1 LTS"
NAME="Ubuntu"
VERSION_ID="24.04"
VERSION="24.04.1 LTS (Noble Numbat)"
VERSION_CODENAME=noble
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=noble
LOGO=ubuntu-logo
fng@CYGNUS:~$ podman -v
podman version 4.9.3
fng@CYGNUS:~$ sudo apt-get install --only-upgrade podman
[sudo] password for user:
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
podman is already the newest version (4.9.3+ds1-1ubuntu0.2).
The following packages were automatically installed and are no longer required:
  fonts-lato libgmp-dev libgmpxx4ldbl libnsl2 libruby libruby3.2 python3-argcomplete python3-platformdirs
  python3-psutil python3-userpath rake ruby ruby-net-telnet ruby-rubygems ruby-sdbm ruby-webrick ruby-xmlrpc ruby3.2
  rubygems-integration unzip zip
Use 'sudo apt autoremove' to remove them.
0 upgraded, 0 newly installed, 0 to remove and 1 not upgraded.
fng@CYGNUS:~$ podman run --rm --gpus all ramalama nvidia-smi

==========
== CUDA ==
==========

CUDA Version 12.6.2

Container image Copyright (c) 2016-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.

This container image and its contents are governed by the NVIDIA Deep Learning Container License.
By pulling and using the container, you accept the terms and conditions of this license:
https://developer.nvidia.com/ngc/nvidia-deep-learning-container-license

A copy of this license is made available in this container at /NGC-DL-CONTAINER-LICENSE for your convenience.

WARNING: The NVIDIA Driver was not detected.  GPU functionality will not be available.
   Use the NVIDIA Container Toolkit to start this container with GPU support; see
   https://docs.nvidia.com/datacenter/cloud-native/ .

/opt/nvidia/nvidia_entrypoint.sh: line 67: exec: nvidia-smi: not found
fng@CYGNUS:~$ podman run --rm --device /dev/dri --gpus all ramalama nvidia-smi

==========
== CUDA ==
==========

CUDA Version 12.6.2

Container image Copyright (c) 2016-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.

This container image and its contents are governed by the NVIDIA Deep Learning Container License.
By pulling and using the container, you accept the terms and conditions of this license:
https://developer.nvidia.com/ngc/nvidia-deep-learning-container-license

A copy of this license is made available in this container at /NGC-DL-CONTAINER-LICENSE for your convenience.

WARNING: The NVIDIA Driver was not detected.  GPU functionality will not be available.
   Use the NVIDIA Container Toolkit to start this container with GPU support; see
   https://docs.nvidia.com/datacenter/cloud-native/ .

/opt/nvidia/nvidia_entrypoint.sh: line 67: exec: nvidia-smi: not found
fng@CYGNUS:~$ podman run --rm --device nvidia.com/gpu=all ramalama nvidia-smi

==========
== CUDA ==
==========

CUDA Version 12.6.2

Container image Copyright (c) 2016-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.

This container image and its contents are governed by the NVIDIA Deep Learning Container License.
By pulling and using the container, you accept the terms and conditions of this license:
https://developer.nvidia.com/ngc/nvidia-deep-learning-container-license

A copy of this license is made available in this container at /NGC-DL-CONTAINER-LICENSE for your convenience.

Thu Nov 28 03:59:58 2024
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 560.35.03              Driver Version: 561.09         CUDA Version: 12.6     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA GeForce RTX 4080 ...    On  |   00000000:01:00.0  On |                  N/A |
|  0%   36C    P8             12W /  320W |    1635MiB /  16376MiB |      7%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+

+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI        PID   Type   Process name                              GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
|  No running processes found                                                             |
+-----------------------------------------------------------------------------------------+
fng@CYGNUS:~$
```

tl/dr: I'm on a popular config, I'm using the latest Podman from my distro's repo, --gpu all doesn't work with or without --device /dev/dri, and --device nvidia.com/gpu=all works just fine.   Breaking what works to make Podman appear to be perfectly identical to Docker seems like cutting off your nose to spite your face.

bmahabirbu: Hey no worries! I'm sorry for not getting back to you earlier! I understand now why `--gpu all` doesn't work for you. I totally forgot that in order for podman to utilize nvidia GPU I edited the podman hook to use the nvidia runtime by default that's why for me it works just doing `--gpu all` I didn't know that you could specify "nvidia.com/gpu=all" to force the nvidia runtime during execution.

For reference here's a doc I wrote I'm detailing the steps
https://github.com/containers/ramalama/blob/main/docs/readme/wsl2-podman-cuda.md
Look under Setting Up Podman NVIDIA Hook

Im wondering if docker can use `nvidia.com/gpu=all` instead. Ill give it a whirl and update my PR accordingly. If not I'll update my PR to include separate args depending on the container runtime

bmahabirbu: Moreover the reason why I did this in the first place was because ` podman run --device nvidia.com/gpu=all` didn't work for me in WSL2 since WSL2 doesn't have direct GPU resource mapping

FNGarvin: I'll add your gentle manner to my list of things to be thankful for, lol.  

> I didn't know that you could specify "nvidia.com/gpu=all" to force the nvidia runtime during execution.

It's how NVidia tells you to set Podman up: https://docs.nvidia.com/ai-enterprise/deployment/rhel-with-kvm/latest/podman.html  also https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/cdi-support.html etc.  

>   podman run --device nvidia.com/gpu=all didn't work for me in WSL2 

IDK, I've been using Podman with the NVidia Container Toolkit on WSL for a spell and you would've noticed that I'm running it in the shell log I shared above.  I don't think it's a WSL issue, though (note in my dump above that I'm using WSL2), because I have had the same issue on bare metal Ubuntu/NVidia. More likely to be Podman version, distro choice, rooted vs rootless, CDI vs non-CDI or some such?

I can see the value of Alexandre Teles's gist that you've adopted to ease moving between Docker and Podman, but IDK if it's reasonable to expect users to have the same.  Especially if it's not mentioned in the setup docs.  

OK, goodnight & happy holidays.  I'm sure whatever you do will be fine, so thanks for considering my input.

bmahabirbu: Gotcha,

After doing some research I setup podman the hard way with editing hooks. A consequence of this is that `--device nvidia.com/gpu=all`  doesn't work for me. It turns out there is a much easier way. follow this method  https://podman-desktop.io/docs/podman/gpu and I've tested it with podman desktop (using these instructions which it's based from the nv doc https://podman-desktop.io/docs/podman/gpu) to get both nvidia.com/gpu=all and `gpu --all` working. 

Overall, following a few extra steps can allow podman to use the same args as docker

but I agree, the user shouldn't need to be mingling with configurations for seamless movement between docker and podman. Something I thought was easily possible at first but simply forgot I had podman setup beforehand.

From what I understand there is a way to provide the shared nvidia libraries manually (without nvidia-container-toolkit) to docker and podman. If its possible this would be the best way. Ill look into this some more. 

Happy holidays to you as well!

FNGarvin: > follow this method [...] to get both nvidia.com/gpu=all and gpu --all working.

As you say, those instructions mimic the NVidia ones.  Following them does not make  ```--gpu all``` work, AFAICT, and I don't see anything in any of the documents suggesting otherwise.  They all use the ```--device nvidia.com/gpu=all``` form, at least.

Did you turn up https://github.com/containers/podman/issues/21156 https://github.com/containers/podman/pull/21180  etc in your research?  I'm not deep into any of these tech stacks, so I suppose I could be misunderstanding something or giving too much credence to the wrong people... but I am seeing things like "NVIDIA has actively discouraged the use of --gpus in docker since 2020. It has many inconsistencies and never should have been added in the first place. The long-term plan is to standardize on CDI and eventually deprecate --gpus altogether" mixed in among declarations that the --gpu and --device arguments were never meant to be interchangeable and so forth.  But maybe I'm misunderstanding due to ignorance.

Of course, to some extent the discourse there should be moot because it went on until late Jan and as you can see in my dump the latest Podman in the WSL2/Ubuntu depot was released previously.  This is the WSL configuration someone will get by default at the time of this writing.  It is not safe to assume everyone will have a configuration that allows both options and I suspect that configurations like yours where ```--device nvidia.com/gpu=all``` won't work will be very rare and nonstandard.

Regards,
FNG

bmahabirbu: I did not come across those! From the conversations it's best to stay safe using  --device = nvidia.com/gpu=all for podman. It seems having --gpus all compatibility can be sketchy

As of the latest nv container toolkit doc
https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/sample-workload.html#running-a-sample-workload-with-docker

Nvidia still suggest to use --gpus all for docker. So overall it seems consolidating the args is more complex than I thought. I'll update my PR accordingly.

Docker uses --device /dev/dri for amd support (same as podman) and it works fine. Maybe worth checking out of nvidia runtime can be used with --device.

PR #484, opened 2024-11-22
Allow users to set ctx-size via command line

rhatdan:
## Summary by Sourcery

Introduce a command line option to set the context size for AI models, update related documentation, and adjust tests to ensure proper functionality.

New Features:
- Add command line option to set context size for running and serving AI models.

Enhancements:
- Remove hardcoded common parameters from the Model class.

Documentation:
- Update documentation to include information about the new --ctx-size option for both run and serve commands.

Tests:
- Modify system tests to verify the correct handling of the new --ctx-size option.

### DIFF
--- a/docs/ramalama-run.1.md
+++ b/docs/ramalama-run.1.md
@@ -11,11 +11,24 @@ ramalama\-run - run specified AI Model as a chatbot
 #### **--authfile**=*password*
 path of the authentication file for OCI registries
 
+#### **--ctx-size**, **-c**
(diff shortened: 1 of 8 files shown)


ericcurtin:
If you look at other similar inferencing engines like llama.cpp or Ollama, the default it 2048. 4096 breaks more models, hardware configurations than 4096 from my paying around with it. I think it translates to the number of tokens (aka words) a prompt can have.

ericcurtin:
https://github.com/ollama/ollama/blob/main/docs/faq.md its just 2048 in Ollama

PR #494, opened 2024-11-26
Take in fixes from @churchyard to modernize spec file

smooge:
Pull request was downstream at
https://src.fedoraproject.org/rpms/python-ramalama/pull-request/2 and this will make it match with upstream. At that point we will be able to automate with packit.

## Summary by Sourcery

Enhancements:
- Modernize the spec file by aligning it with upstream changes, including formatting adjustments and updates to build requirements.

### DIFF
--- a/rpm/python-ramalama.spec
+++ b/rpm/python-ramalama.spec
@@ -73,8 +73,10 @@ configure the system for AI themselves. After the initialization, RamaLama
 will run the AI Models within a container based on the OCI image.
 
 
+%if 0%{?fedora} >= 40
 %generate_buildrequires
 %pyproject_buildrequires
+%endif
 
 %prep
 %forgeautosetup -p1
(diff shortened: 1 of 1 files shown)


rhatdan:
Builds are failing on centos-stream because of missing requiements tdqm?

rhatdan:
LGTM

rpm/python-ramalama.spec
```diff
@@ -1,51 +1,46 @@
-%global debug_package %{nil}
-
 %global pypi_name ramalama
-%global forgeurl https://github.com/containers/%{pypi_name}
+%global forgeurl  https://github.com/containers/%{pypi_name}
 # see ramalama/version.py
-%global version0 0.2.0
+%global version0  0.2.0
```
lsm5, line 4: @smooge do you need the version0 stuff at all? Packit will handle the version automatically in Fedora PRs. For regular builds, packit will use `git describe` to fetch the version by default.

PR #495, opened 2024-11-27
Remove %version0 from spec file

smooge:
This was something to allow for the package to build in Fedora koji and to pass review since it needed to be in the upstream code and the Fedora spec file. Now that we are using packit, it is not needed and should be removed from the spec.

## Summary by Sourcery

Build:
- Remove the %version0 macro from the spec file as it is no longer needed with the use of packit.

### DIFF
--- a/rpm/python-ramalama.spec
+++ b/rpm/python-ramalama.spec
@@ -1,7 +1,5 @@
 %global pypi_name ramalama
 %global forgeurl  https://github.com/containers/%{pypi_name}
-# see ramalama/version.py
-%global version0  0.2.0
 %forgemeta
 
 %global summary   RamaLama is a command line tool for working with AI LLM models


smooge:
@lsm5 I am not sure what is causing this :
```
2024-11-27 12:30:16.022 upstream.py       DEBUG  Version: 0.2.0
2024-11-27 12:30:16.022 upstream.py       DEBUG  Sanitized version: 0.2.0
2024-11-27 12:30:16.022 upstream.py       DEBUG  Name + version = ramalama-0.2.0
2024-11-27 12:30:16.022 commands.py       DEBUG  Command: git archive --output rpm/ramalama-0.2.0.tar.gz --prefix ramalama-0.2.0/ HEAD
2024-11-27 12:30:16.045 cmd.py            INFO   git cat-file --batch-check
2024-11-27 12:30:16.045 cmd.py            DEBUG  Popen(['git', 'cat-file', '--batch-check'], cwd=/tmp/tmp2fkdnrc3, stdin=<valid stream>, shell=False, universal_newlines=False)
2024-11-27 12:30:16.059 utils.py          ERROR  Preparation of the repository for creation of an SRPM failed: You need to define Version:, %{commit} or %{tag} before the macro invocation !
  File "/usr/lib/python3.13/site-packages/specfile/spec_parser.py", line 233, in get_rpm_spec
    return rpm.spec(tmp.name, flags)
           ~~~~~~~~^^^^^^^^^^^^^^^^^
ValueError: can't parse specfile


The above exception was the direct cause of the following exception:

```

All I did was remove `version0` from teh spec file in this:

```
%global pypi_name ramalama
%global forgeurl  https://github.com/containers/%{pypi_name}
%forgemeta

%global summary   RamaLama is a command line tool for working with AI LLM models

%global _python_dist_allow_version_zero 1

Name:             python-%{pypi_name}
# DO NOT TOUCH the Version string!
# The TRUE source of this specfile is:
# https://github.com/containers/ramalama/blob/main/rpm/python-ramalama.spec
# If that's what you're reading, Version must be 0, and will be updated by Packit for
# copr and koji builds.
# If you're reading this on dist-git, the version is automatically filled in by Packit.
Version:          0
License:          MIT
Release:          %{autorelease}
Summary:          %{summary}
URL:              %{forgeurl}
# Tarball fetched from upstream
Source:           %{forgesource}
BuildArch:        noarch

```

lsm5 (commented):
The version0 stuff is needed after all for forgemeta. My bad on the earlier suggestion. 

We can either close this PR or maybe set the macro to 0 on line 4 and then also move the comment in the lines above `Version:` to above where we define `version0` .

lsm5:
@smooge UPDATE: I'm told replacing `version0` with `tag` should make things work for `propose-downstream` feature of packit (the job that creates update PRs for dist-git). So, I'd suggest making that change and we can verify if things work with the next upstream release.

PR #505, opened 2024-12-06
Stop checking if command is running in container

rhatdan:
Back when we were embedding the python code into the container image we would check if we were already injected and not run a container, the problem with this, is it is breaking when ramalama is run within a toolbx container or running via podman or docker. If the user of RamaLama wants to run without containers, then they should specify, and the tool should not try to handle this on it's own.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2329826

## Summary by Sourcery

Bug Fixes:
- Remove the check for whether the command is running in a container to prevent issues when running within toolbox, podman, or docker containers.

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -11,7 +11,6 @@ from ramalama.huggingface import Huggingface
 from ramalama.common import (
     container_manager,
     default_image,
-    in_container,
     perror,
     run_cmd,
 )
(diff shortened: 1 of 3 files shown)


rhatdan:
@smooge @ericcurtin @bmahabirbu PTAL

ericcurtin (approved):

ericcurtin:
I have a feeling we will need in container wrapper code again eventually, but maybe its a brand new python script

ericcurtin:
I hope macOS native and nocontainer still works after this

PR #510, opened 2024-12-11
Simplify install

ericcurtin:
So we can just have one version of the installer from the user perspective

## Summary by Sourcery

Simplify the installation process by modifying the install script to conditionally use sudo based on user permissions and update the README to reflect the unified installation method for both Linux and macOS.

Enhancements:
- Simplify the installation script by allowing it to run with or without sudo, depending on user permissions.

Documentation:
- Update the installation instructions in the README to unify the process for Linux and macOS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,15 +73,7 @@ pipx install ramalama
 
 ## Install by script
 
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

PR #511, opened 2024-12-11
Place image name just before command

ericcurtin:
We think the ordering of args are wrong.

## Summary by Sourcery

Enhancements:
- Reorder the arguments in the run_container method to place the image name just before the command execution.


rhatdan:
This change looks fine, but it only effects running ramalama within a container.

Issue #515, opened 2024-12-17
Ramalama rm command should flag missing name or id ?


pghole:
```
pghole@pghole-thinkpadt14sgen2i:~/opensource_contribution/ramalama$ ramalama rm
pghole@pghole-thinkpadt14sgen2i:~/opensource_contribution/ramalama$ 
```

Can we display the error similar to podman here - 
```
pghole@pghole-thinkpadt14sgen2i:~/opensource_contribution/ramalama$ podman rm
Error: you must provide at least one name or id
```

ericcurtin:
Yup, would take a PR for this too

PR #516, opened 2024-12-17
ramalama rm should require at least one argument

rhatdan:
Fixes: https://github.com/containers/ramalama/issues/515

## Summary by Sourcery

Require at least one argument for the 'ramalama rm' command to prevent accidental deletions and add a corresponding test to verify this behavior.

Bug Fixes:
- Ensure the 'ramalama rm' command requires at least one argument to prevent accidental deletions.

Tests:
- Add a test to verify that the 'ramalama rm' command requires at least one argument.

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -244,7 +244,7 @@ def parse_arguments(parser):
 def post_parse_setup(args):
     """Perform additional setup after parsing arguments."""
     mkdirs(args.store)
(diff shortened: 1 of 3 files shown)


ericcurtin (approved):
Will merge on green build

rhatdan:
Strange MacOS is breaking with a weird error where it is not quoting options,

PR #519, opened 2024-12-18
Update shortnames.conf to alias new granite models

ericcurtin:
This updates the shortnames for the "granite" models to point to newer versions.

## Summary by Sourcery

Enhancements:
- Update shortnames.conf to alias new versions of granite models, including granite3.1-dense for 2b and 8b models.


rhatdan:
Should we be going through ollama or point to the same models at huggingface?

shortnames/shortnames.conf
```diff
@@ -1,13 +1,15 @@
 [shortnames]
-  "granite" = "huggingface://ibm-granite/granite-8b-code-instruct-GGUF/granite-8b-code-instruct.Q4_K_M.gguf"
+  "granite" = "granite3.1-dense"
```
rhatdan, line 2: If we want to pull from ollama, this should probably be
ollama://granite3.1-dense
The rest below should be the same.

If user changes default transport these would be attempted to be pulled from OCI or HF.

ericcurtin:
> Should we be going through ollama or point to the same models at huggingface?

The thing is, I don't see granite models being published in .gguf format in other places like huggingface these days.

It's been quite a few months since we've seen a granite gguf published to huggingface, Ollama registry seems to be the place they are publishing them now.

shortnames/shortnames.conf
ericcurtin: Done

ericcurtin:
That's not to say if they got published again to some other place like a hf:// or https:// location that we couldn't change the pull location again

rhatdan:
LGTM

PR #520, opened 2024-12-20
Change pip to pipx in installation instructions

ericcurtin:
It's most common to recommend pip even though it doesn't work on various platforms. Let the users with the tricky platforms figure that out.

## Summary by Sourcery

Documentation:
- Updated installation instructions to use pip instead of pipx.


rhatdan:
LGTM

PR #524, opened 2024-12-22
Add missing files to install.sh script

ericcurtin:
file.py , http_client.py , etc.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -88,9 +88,11 @@ setup_ramalama() {
   $sudo install -m755 -d "$syspath"
   $sudo install -m755 "$to_file" "$ramalama_bin"
 
-  local python_files=("cli.py" "huggingface.py" "model.py" "ollama.py" "common.py" "__init__.py" \
-                      "quadlet.py" "kube.py" "oci.py" "version.py" "shortnames.py" "toml_parser.py")
-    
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -88,9 +88,11 @@ setup_ramalama() {
   $sudo install -m755 -d "$syspath"
   $sudo install -m755 "$to_file" "$ramalama_bin"
 
-  local python_files=("cli.py" "huggingface.py" "model.py" "ollama.py" "common.py" "__init__.py" \
-                      "quadlet.py" "kube.py" "oci.py" "version.py" "shortnames.py" "toml_parser.py")
-    
+  local python_files=("cli.py" "huggingface.py" "model.py" "ollama.py" \
```
rhatdan, line 91: Wouldn't it be easier to just use a GLOB for this?

rhatdan:
Merging but I think we should open a PR to just grab the py files in the ramalama/*py.

install.sh
ericcurtin: The way this installer works is it pulls down install.sh via curl and then the various other files it needs via curl. We could maybe pull down the whole project as a zip/tarball , etc. and then do a glob, but we would pull a lot of unnecessary things also.

But yeah I've also been thinking maybe we can look into somehow making this smarter rather than listing all the files.

Issue #525, opened 2024-12-26
ROCm containers fail on multi-gpu AMD systems


abn:
When attempting to run a model, the command fails.

```console
$ ramalama --debug run granite-code
run_cmd:  podman run --rm -i --label RAMALAMA --security-opt=label=disable --name ramalama_rDnJbnsIEU -t --device /dev/dri --device /dev/kfd -e HIP_VISIBLE_DEVICES=1 --mount=type=bind,src=/home/abn/.local/share/ramalama/models/ollama/granite-code:latest,destination=/mnt/models/model.file,rw=false quay.io/ramalama/rocm:latest /bin/sh -c llama-cli -m /mnt/models/model.file --in-prefix '' --in-suffix '' -c 2048 --temp 0.8 -p 'You are a helpful assistant' -cnv
Error: Command '['podman', 'run', '--rm', '-i', '--label', 'RAMALAMA', '--security-opt=label=disable', '--name', 'ramalama_rDnJbnsIEU', '-t', '--device', '/dev/dri', '--device', '/dev/kfd', '-e', 'HIP_VISIBLE_DEVICES=1', '--mount=type=bind,src=/home/abn/.local/share/ramalama/models/ollama/granite-code:latest,destination=/mnt/models/model.file,rw=false', 'quay.io/ramalama/rocm:latest', '/bin/sh', '-c', "llama-cli -m /mnt/models/model.file --in-prefix '' --in-suffix '' -c 2048 --temp 0.8 -p 'You are a helpful assistant' -cnv"]' returned non-zero exit status 139.
```

Running the `podman` command gives the following.

```console
$ podman run --rm -i --label RAMALAMA --security-opt=label=disable --name ramalama_rDnJbnsIEU -t --device /dev/dri --device /dev/kfd -e HIP_VISIBLE_DEVICES=1 --mount=type=bind,src=/home/abn/.local/share/ramalama/models/ollama/granite-code:latest,destination=/mnt/models/model.file,rw=false quay.io/ramalama/rocm:latest /bin/sh -c llama-cli -m /mnt/models/model.file --in-prefix '' --in-suffix '' -c 2048 --temp 0.8 -p 'You are a helpful assistant' -cnv

rocBLAS error: Cannot read /opt/rocm-6.2.2/lib/rocblas/library/TensileLibrary.dat: Illegal seek for GPU arch : gfx1103
 List of available TensileLibrary Files : 
[...]

PR #526, opened 2024-12-28
Pass ASAHI, CUDA, HIP, HSA prefixed env vars to container

abn:
Relates-to: #525

## Summary by Sourcery

Build:
- Pass environment variables prefixed with ASAHI, CUDA, HIP, and HSA to the container via `get_env_vars`.

### DIFF
--- a/ramalama/common.py
+++ b/ramalama/common.py
@@ -1,5 +1,6 @@
 """ramalama common module."""
 
+import glob
 import hashlib
 import os
 import random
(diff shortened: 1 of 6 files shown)


ericcurtin (approved):

ericcurtin:
I think this is fine, but we have to be mindful of the generators for Quadlet, etc. to ensure the env vars get taken into account there also

abn:
I will push up a change to include those as well.

abn:
@ericcurtin I am curious however, if the better change is to enable explicit `--env | -e` option for the `serve` command instead. But if the generated files are always expected to be used on the local machine, this should be fine I reckon.

abn:
@sourcery-ai review

ericcurtin (approved):

PR #538, opened 2025-01-07
Update llama.cpp to get context size fix

ericcurtin:
Also update whisper.cpp while we are at it.


ericcurtin:
https://github.com/ggerganov/llama.cpp/pull/11094

rhatdan:
LGTM

PR #540, opened 2025-01-07
Remove huggingface warning

ericcurtin:
It prints on all usages of huggingface, it's ugly.

We don't really need huggingface-cli for most usages of hf, this suggests we do.

## Summary by Sourcery

Remove the warning about the huggingface CLI not being available.

Bug Fixes:
- Removed the warning message displayed when the huggingface CLI is not available.

Enhancements:
- Removed the check for huggingface CLI availability.


rhatdan:
LGTM

PR #551, opened 2025-01-08
Add CPU inferencing container build steps to CI

ericcurtin:
Simplified `cmake_steps` function by removing unnecessary `mv` command and adjusted `cmake_flags` handling. Improved `configure_common_flags` function to directly manipulate `common_flags` array. Refactored `clone_and_build_whisper_cpp` and `clone_and_build_llama_cpp` functions to utilize `common_flags` directly. Added `ldconfig` call in the `main` function to ensure library paths are updated.

## Summary by Sourcery

Add CPU inferencing container build steps to CI workflows and refactor build scripts.

Enhancements:
- Simplify and refactor build scripts to use common flags and improve library handling.

CI:
- Add CPU inferencing container build step to CI workflows.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,9 @@ jobs:
            # Install specific podman version
            sudo apt-get upgrade
 
+      - name: Build a container for CPU inferencing
+        run: ./container_build.sh build ramalama
+
       - name: run bats
         run: |
            make validate
@@ -97,6 +100,9 @@ jobs:
            cat /etc/docker/daemon.json
            sudo systemctl restart docker.service
            df -h
+
+      - name: Build a container for CPU inferencing
+        run: ./container_build.sh build ramalama
 
       - name: bats-docker
         run: |
--- a/container-images/scripts/build_llama_and_whisper.sh
+++ b/container-images/scripts/build_llama_and_whisper.sh
@@ -38,12 +38,10 @@ dnf_install() {
 }
 
 cmake_steps() {
-  local cmake_flags=("${!1}")
+  local cmake_flags=("$@")
   cmake -B build "${cmake_flags[@]}"
   cmake --build build --config Release -j"$(nproc)"
-  cmake --install build --prefix=/usr
-  #HACK libggml-cpu.so is being installed in the wrong place should be in /usr/lib64
-  test -f /usr/lib/libggml-cpu.so && mv /usr/lib/libggml-cpu.so /usr/lib64
+  cmake --install build
 }
 
 set_install_prefix() {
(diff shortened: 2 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -28,6 +28,9 @@ jobs:
            # Install specific podman version
            sudo apt-get upgrade
 
+      - name: Build a container for CPU inferencing
+        run: ./container_build.sh build ramlama
```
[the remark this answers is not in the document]
ericcurtin: Fixed and pushing

.github/workflows/ci.yml
```diff
@@ -28,6 +28,9 @@ jobs:
            # Install specific podman version
            sudo apt-get upgrade
 
+      - name: Build a container for CPU inferencing
+        run: ./container_build.sh build ramalama
```
rhatdan, line 32: Should these be build_rm?

container-images/scripts/build_llama_and_whisper.sh
```diff
@@ -38,12 +38,10 @@ dnf_install() {
 }
 
 cmake_steps() {
-  local cmake_flags=("${!1}")
+  local cmake_flags=("$@")
   cmake -B build "${cmake_flags[@]}"
   cmake --build build --config Release -j"$(nproc)"
-  cmake --install build --prefix=/usr
-  #HACK libggml-cpu.so is being installed in the wrong place should be in /usr/lib64
```
ericcurtin, line 45: ldconfig removes this hack basically

.github/workflows/ci.yml
ericcurtin: In this case no, because I want the ci.yml to try and use this image

Otherwise we don't find issues like "llama-run library not found", because it will use the one from quay.io

ericcurtin: This probably isn't complete yet here, but I want to force our test suite to not pull down an image, instead use a locally built one

rhatdan: SGTM

container-images/scripts/build_llama_and_whisper.sh
rhatdan: Why not install in /usr rather then in /usr/local?

ericcurtin: It is installing in /usr (rather than /usr/local) via:

"-DCMAKE_INSTALL_PREFIX=$install_prefix"

it was broken before though you were correct.

I built this all locally and checked.

In the Nvidia build $install_prefix is a temporary directory as we have separate build-time and run-time images

ericcurtin:
We can merge this with a failing build, it's failing anyway. I'll trim ROCm image in another PR

rhatdan:
LGTM

PR #554, opened 2025-01-08
Remove some more gfx9 stuff

ericcurtin:
gfx9 are older GPUs

## Summary by Sourcery

Build:
- Remove gfx9 files from ROCm libraries during Llama and Whisper build.


ericcurtin:
They renamed a directory that we were purging from, that's why the image size exploded

rhatdan:
LGTM

PR #555, opened 2025-01-08
Only use the X/Y parts of the version when pulling images

rhatdan:
## Summary by Sourcery

Enhancements:
- Pull VLLM images using only the major and minor version numbers, instead of the full version string.


rhatdan:
@ericcurtin @bmahabirbu PTAL

ericcurtin (approved):

PR #557, opened 2025-01-09
If podman machine is running use containers

ericcurtin:
Don't check for krunkit hypervisor specifically, it will still work without this, just unaccelerated.

## Summary by Sourcery

Enhancements:
- Remove the check for krunkit hypervisor.


rhatdan:
I think this is fine, but should we put out some warning, that tools would work better with krunkit?

rhatdan:
LGTM

PR #560, opened 2025-01-09
Looks like RHEL9 doesn't support autochangelog

ericcurtin:
Change back to manual changelogs

## Summary by Sourcery

Build:
- Generate changelogs manually instead of automatically.

### DIFF
--- a/rpm/python-ramalama.spec
+++ b/rpm/python-ramalama.spec
@@ -101,4 +101,32 @@ will run the AI Models within a container based on the OCI image.
 %{_mandir}/man5/ramalama*.5*
 
 %changelog
-%autochangelog
+* Thu Jan  9 00:00:00 2025 Daniel J Walsh <dwalsh@redhat.com> - 0.4.0
+- Release 0.4.0
+
+* Tue Nov 26 00:00:00 2024 Stephen Smoogen <ssmoogen@redhat.com> - 0.2.0-7
+- Fix spec file to get CentOS Stream 10 building
+
+* Tue Nov 26 00:00:00 2024 Stephen Smoogen <ssmoogen@redhat.com> - 0.2.0-6
+- Needed to fix to match upstream
+
+* Tue Nov 26 00:00:00 2024 Stephen Smoogen <ssmoogen@redhat.com> - 0.2.0-5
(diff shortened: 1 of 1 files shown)


ericcurtin:
If my reading of the build failures are correct on the main branch, RHEL9 doesn't support autochangelog, EPEL9 seems to though

rhatdan:
@lsm5 @smooge PTAL

ericcurtin:
@lsm5 could we get centos-stream-9 packit builds turned on for PRs so we don't get into the scenario where main is failing but PRs are passing again?

maxamillion:
@ericcurtin it looks like another option would be that we could add the epel repos to the UBI base image and install the `rpmautospec-rpm-macros` package which provides the `%autochangelog` macro.

ericcurtin:
> @ericcurtin it looks like another option would be that we could add the epel repos to the UBI base image and install the `rpmautospec-rpm-macros` package which provides the `%autochangelog` macro.

Yeah I guess I'm more worried if we end up packaging this for RHEL at some point. But we can close for now and re-visit that in future if we do that in future.

@lsm5 or @maxamillion maybe we just turn off the centos-stream-9 builds on the main branch of ramalama for now to get it green again? And only build on EPEL9, not a Packit expert personally.

ericcurtin:
Yeah I guess, closing... @lsm5 or @lbarcziova how do we turn off centos-stream-9 builds on the main branch here in packit? We've been red since forever

lbarcziova:
hi @ericcurtin ! Currently, for the main branch, the builds are implicitly submitted for all the targets in the [rhcontainerbot/podman-next Copr repo](https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/). If you want to override this, you need to define the `targets` list in [here](https://github.com/containers/ramalama/blob/3fdcae16cf8dc3d7a0502b84488944e9c970da2b/.packit.yaml#L38-L45), similarly like you have it defined for pull-requests [here](https://github.com/containers/ramalama/blob/3fdcae16cf8dc3d7a0502b84488944e9c970da2b/.packit.yaml#L25).

ericcurtin:
Thank you @lbarcziova

PR #562, opened 2025-01-09
Fix podman stop --all

rhatdan:
Also make sure all run_cmd with args passes in args.debug

## Summary by Sourcery

Bug Fixes:
- Fixed an issue where `podman stop --all` failed to stop all containers.

### DIFF
--- a/ramalama/common.py
+++ b/ramalama/common.py
@@ -102,7 +102,7 @@ def find_working_directory():
 def run_curl_cmd(args, filename):
     if not verify_checksum(filename):
         try:
-            run_cmd(args)
+            run_cmd(args, debug=args.debug)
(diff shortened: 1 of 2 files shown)


ericcurtin (approved):

ramalama/common.py
```diff
@@ -102,7 +102,7 @@ def find_working_directory():
 def run_curl_cmd(args, filename):
```
ericcurtin, line 102: We can actually delete this whole function, it's not called anywhere anymore, we move to a python based http client, so we could have prettier progress bars, etc. and one less dependancy.

rhatdan: Nice, next PR.

PR #563, opened 2025-01-09
smollm:135m for testing purposes

ericcurtin:
smollm is a model by Hugging Face, it's good for testing and CPU inferencing.

## Summary by Sourcery

Tests:
- Updated system tests to use smollm:135m.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,7 +150,7 @@ You can `list` all models pulled into local storage.
 ```
 $ ramalama list
 NAME                                                                MODIFIED     SIZE
-ollama://tiny-llm:latest                                            16 hours ago 5.5M
+ollama://smollm:135m                                            16 hours ago 5.5M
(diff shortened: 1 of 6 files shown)


rhatdan:
Can you add this to shortnames.conf

ericcurtin:
> Can you add this to shortnames.conf

I will, they publish to ollama though, so it's sorta gonna be a no-op

rhatdan:
Sure, I just want smollm to be the same as smollm:135m, but peraps ollama already handles this.

LGTM

ericcurtin:
> Sure, I just want smollm to be the same as smollm:135m, but peraps ollama already handles this.
> 
> LGTM

Yup we handle all that already, "smollm" on it's own, just like a container image means ":latest", which is ":1.7b" for smollm which is not what we want to use for testing, we want to use the smallest one.

rhatdan:
Just tried it out and it works well, thanks.

PR #564, opened 2025-01-09
run_cmd to exec_cmd

ericcurtin:
podman run with llama-run is only working for me when we use exec_cmd, which should be ok as it's the last process we execute.

## Summary by Sourcery

Enhancements:
- Replace run_cmd with exec_cmd for executing container commands.

### DIFF
--- a/ramalama/model.py
+++ b/ramalama/model.py
@@ -213,7 +213,7 @@ class Model:
 
         atexit.register(cleanup)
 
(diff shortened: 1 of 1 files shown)


ericcurtin:
We should try and consoldate the two functions that execute podman run if possible

rhatdan:
LGTM

PR #569, opened 2025-01-10
Check all files are adding to packaging

ericcurtin:
We've missed this more than once.

## Summary by Sourcery

CI:
- Verify that all Python files are correctly added to the packaging step during the CI process.


rhatdan:
I could not see how it was broken for prior versions of pip, but it did seem to be fixed now.

LGTM

PR #570, opened 2025-01-10
Revert "If podman machine is running use containers"

ericcurtin:
Reverts containers/ramalama#557

## Summary by Sourcery

Bug Fixes:
- Stop using containers if the podman machine is running, even if krunkit or libkrun are used.

### DIFF
--- a/ramalama/common.py
+++ b/ramalama/common.py
@@ -27,9 +27,16 @@ def container_manager():
             return "podman"
 
         podman_machine_list = ["podman", "machine", "list"]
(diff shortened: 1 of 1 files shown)


ericcurtin:
We shouldn't have removed this, it wasn't an issue @smooge has now found an issue if running with "applehv" ramalama complains /dev/dri is not found as we don't have GPU passthrough there.

It was suggested this code was a problem, I think that was a misdiagnosis anyway...

rhatdan:
LGTM
We could not pass the /dev/dri when using applehv.

PR #573, opened 2025-01-10
On ARM by default turn on GPU acceleration

ericcurtin:
On x86 we don't want to do this without a GPU because x86 integrated graphics have very limited access to VRAM and it's normally not worth it. But ARM SoC's share memory between CPU and GPU, meaning it's worth it generally. And we care mostly about Apple Silicon where we want this on in podman machine.

## Summary by Sourcery

Enhancements:
- Increase the number of generations for GPU offload from 99 to 999 on systems where it is enabled.

### DIFF
--- a/ramalama/model.py
+++ b/ramalama/model.py
@@ -1,5 +1,6 @@
 import os
 import sys
+import platform
 
 from ramalama.common import (
     container_manager,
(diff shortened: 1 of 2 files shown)


ericcurtin:
@bmahabirbu PTAL

bmahabirbu:
> @bmahabirbu PTAL

Looks good for running on Mac / arm natively but running inside a podman machine it doesn't apply --ngl 999

ericcurtin:
I think I see the issue, how about now?

bmahabirbu:
needs --ngl instead of -ngl! Should be good to go then!

bmahabirbu:
`Traceback (most recent call last):
  File "/var/home/core/ramalama/ramalama/./bin/ramalama", line 95, in <module>
    main(sys.argv[1:])
  File "/var/home/core/ramalama/ramalama/./bin/ramalama", line 72, in main
    args.func(args)
  File "/var/home/core/ramalama/ramalama/ramalama/cli.py", line 716, in run_cli
    model.run(args)
  File "/var/home/core/ramalama/ramalama/ramalama/model.py", line 295, in run
    exec_args.extend(self.gpu_args(), args.gpu)
TypeError: list.extend() takes exactly one argument (2 given)
[core@localhost ramalama]$

bmahabirbu:
think its just exec_args.extend(self.gpu_args())?

ericcurtin:
I hope it's good now @bmahabirbu really need to head offline

ramalama/model.py
```diff
@@ -174,18 +176,27 @@ def setup_container(self, args):
...
+                    or os.getenv("ASAHI_VISIBLE_DEVICES")
+                    or os.getenv("CUDA_VISIBLE_DEVICES")
+                )
+            )
         ):
-            gpu_args = ["-ngl", "99"]
-        else:
-            print("GPU offload was requested but is not available on this system")
+            if server:
+                gpu_args += ["-ngl"]
```
rhatdan, line 195: The if and else are exactly the same?  So what is the purpose of the server option?

bmahabirbu: One has `-ngl` with one dash and the other has `--ngl` with two dashes. I believe this has to do with extra compatibility for llama run

rhatdan: Yikes, Probably should add a comment, but my eyes would have never seen that.

bmahabirbu:
Tested and working with Podman Desktop with apple hypervisor and libkrun. The apple hypervisor podman machine did not have GPU access and ran with CPU only as expected while libkrun utilized the GPU!

Also did a sanity test on my windows setup with WSL2 to confirm nothing broke functionality for non macs!

ramalama/model.py
ericcurtin: Comment added

rhatdan:
LGTM, merge when tests pass.

PR #574, opened 2025-01-10
This is all dead code which isn't called

ericcurtin:
We should find a static analyzer for this

## Summary by Sourcery

Remove dead code related to running containers.

Tests:
- Update tests to reflect the removal of the container execution functionality.

Chores:
- Remove the `run_container` function and all its usages.


bmahabirbu:
LGTM, Been meaning to do this one

PR #581, opened 2025-01-12
Add flake

jim3692:
I added a `flake.nix` to make it easier to install and/or run on Nix-enabled systems.

To install, add the `packages.ramalama` to the configuration's package list.

I have enabled Vulkan support for `llama-cpp`, since it's not default. ROCm and Cuda support can be enabled for llama.cpp by setting `llamaCppOverrides.rocmSupport` and `llamaCppOverrides.cudaSupport` respectively.

It can also be run directly by `nix run github:containers/ramalama`.

Extra arguments can be passed after a `--`: `nix run github:containers/ramalama -- --nocontainer --gpu run llama3.1:8b`

## Summary by Sourcery

Build:
- Introduce a flake.nix file to define the Nix build process.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -10,6 +10,8 @@
       supportedSystems = [
         "x86_64-linux"
         "aarch64-linux"
+        "x86_64-darwin"
+        "aarch64-darwin"
       ];
 
       forAllSystems = nixpkgs.lib.genAttrs supportedSystems;
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,55 @@
+{
+  inputs = {
+    nixpkgs = {
+      url = "github:NixOS/nixpkgs/nixos-unstable";
+    };
+  };
+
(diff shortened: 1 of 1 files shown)


ericcurtin:
SGTM, you need to "git commit --amend -s" and repush your commit for the DCO build to pass

jim3692:
> SGTM, you need to "git commit --amend -s" and repush your commit for the DCO build to pass

Should I enable Mac support for the flake? I thought that only Linux is supported by the project, but I see a `ci/macos` action running.

ericcurtin:
> > SGTM, you need to "git commit --amend -s" and repush your commit for the DCO build to pass
> 
> Should I enable Mac support for the flake? I thought that only Linux is supported by the project, but I see a `ci/macos` action running.

Yes why not.

We support macOS. We run perfectly on macOS native in fact!

We also plan on running on Windows, but for Windows it will likely be via WSL2

PR #582, opened 2025-01-13
Added vllm cuda support

bmahabirbu:
Some notes, 

Once gpu is detected added to env vars so detection only needs to be called once

Added functionality to pick largest GPU between cuda and rocm by vram

added ```"--max_model_len"= "4096"``` which reduces the token size so models can be loaded. I did this because llama3.2 was not loading on my GPU with 10gb of vram compared to llama.cpp. I believe this has to do with vllm parallelization optimization which increases vram usage


ramalama/model.py
```diff
@@ -171,10 +174,14 @@ def setup_container(self, args):
             conman_args += ["--device", "/dev/kfd"]
 
         for k, v in get_env_vars().items():
-            conman_args += ["-e", f"{k}={v}"]
-
+            # Special case for Cuda
```
ericcurtin, line 177: I think we should remove the else here and the env var in all cases, a use might set CUDA_VISIBLE_DEVICES to run a specific GPU and llama.cpp should be able to process that env var.

ramalama/common.py
```diff
@@ -205,24 +217,44 @@ def get_gpu():
...
+        print(f"error occurred while detecting Nvidia gpu: {ex}")
+
+    # Select GPU
+    if gpu_template:
+        if len(gpu_template) == 1:
+            best_gpu = gpu_template[0]  # Directly use the single entry
+        else:
+            best_gpu = max(gpu_template, key=lambda x: x["vram"])  # Use max for multiple entries
+        # Set env var of picked gpu
+        os.environ[best_gpu["env"]] = best_gpu["index"]
```
ericcurtin, line 250: If CUDA_VISIBLE_DEVICES is set we should skip all this somehow also, can't remember how we did that for ROCm or if we did it correctly. nvidia-smi is probably an expensive technique for nvidia detection, maybe this is a start and we can do somthing cheaper later.

Make sure if nvidia-smi doesn't exists it's super silent, our packagers will complain that ramalama is broken otherwise and needs more dependancies.

ramalama/model.py
```diff
@@ -303,13 +311,16 @@ def build_exec_args_serve(self, args, exec_model_path):
...
             exec_args += ["--seed", args.seed]
+        
         return exec_args
 
     def handle_runtime(self, args, exec_args, exec_model_path):
         if args.runtime == "vllm":
+            get_gpu()
             exec_model_path = os.path.dirname(exec_model_path)
-            exec_args = ["vllm", "serve", "--port", args.port, exec_model_path]
+            exec_args = ["--port", args.port, "--model", MNT_FILE, "--max_model_len", "4096"]
```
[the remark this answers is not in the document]
bmahabirbu: Whats odd is when I had vllm serve as part of the args the container would crash stating those args are recognized. Im guessing when the vllm image is run its entrypoint is vllm serve thats why I removed it.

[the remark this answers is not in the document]
ericcurtin: Ah ok makes sense @bmahabirbu maybe just leave a comment, assumes entrypoint is "vllm serve"

bmahabirbu:
I agree I'll look into doing something simpler like this. No reason to use nvidia-smi if we can see what gpu driver linux is using.

I agree as well lets forget about the comparison its a bit much. If a user has multiple different gpu's we can always have the user set the env var manually to what they want to use.

bmahabirbu:
After testing a bit I think ill stick with calling nvidia-smi for now as I couldn't find any nvidia driver listed inside wsl2. I shorted the code so as long as nvidia-smi is recognized I set the env var.

rhatdan:
@ericcurtin PTAL

ramalama/model.py
ericcurtin: The else should be gone here right?

ericcurtin:
Sorry @bmahabirbu when I asked to remove the else, I meant change:

```
            # Special case for Cuda
            if k == "CUDA_VISIBLE_DEVICES":
                conman_args += ["--device", "nvidia.com/gpu=all"]
            else:
```

to

```
            # Special case for Cuda
            if k == "CUDA_VISIBLE_DEVICES":
                conman_args += ["--device", "nvidia.com/gpu=all"]

```

We were skipping setting the -e in the CUDA case. We want the env vars set inside the container in all cases

bmahabirbu:
Ah gotcha! Will change early today!

ramalama/common.py
```diff
@@ -205,24 +229,17 @@ def get_gpu():
...
 
 
 def get_env_vars():
     prefixes = ("ASAHI_", "CUDA_", "HIP_", "HSA_")
     env_vars = {k: v for k, v in os.environ.items() if k.startswith(prefixes)}
 
-    gpu_type, gpu_num = get_gpu()
-    if gpu_type not in env_vars and gpu_type in {"HIP_VISIBLE_DEVICES", "ASAHI_VISIBLE_DEVICES"}:
-        env_vars[gpu_type] = str(gpu_num)
+    # gpu_type, gpu_num = get_gpu()
```
ericcurtin, line 241: I guess we can delete this

ericcurtin (approved):

PR #588, opened 2025-01-14
We no longer have python dependancies

ericcurtin:
No python3 dependancies outside of the standard python libs.


rhatdan:
LGTM
Merge when tests pass.

PR #589, opened 2025-01-14
Simplify installation spec file

rhatdan:
Fedora 39 is no longer supported, so remove checks in spec file.

More Podman and omlmd to Recommends,  You can run RamaLama with no container engine or with Docker.

## Summary by Sourcery

Remove Fedora 39 checks and move `podman` and `omlmd` to Recommends in the spec file.

Build:
- Move `podman` and `omlmd` to Recommends in the spec file.
- Remove Fedora 39 checks in the spec file.

### DIFF
--- a/rpm/python-ramalama.spec
+++ b/rpm/python-ramalama.spec
@@ -47,13 +47,8 @@ configure the system for AI themselves. After the initialization, RamaLama
 will run the AI Models within a container based on the OCI image.
 
 %package -n python%{python3_pkgversion}-%{pypi_name}
-Requires: podman
-%if 0%{?fedora} >= 40
-# Needed as seen by BZ: 2327515
-Requires: python%{python3_pkgversion}-omlmd
-%else
+Recommends: podman
 Recommends: python%{python3_pkgversion}-omlmd
-%endif
 Summary: %{summary}
 Provides: %{pypi_name} = %{version}-%{release}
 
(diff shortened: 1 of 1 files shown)


rhatdan:
@smooge @lsm5 PTAL

rpm/python-ramalama.spec
```diff
@@ -47,13 +47,8 @@ configure the system for AI themselves. After the initialization, RamaLama
 will run the AI Models within a container based on the OCI image.
 
 %package -n python%{python3_pkgversion}-%{pypi_name}
-Requires: podman
-%if 0%{?fedora} >= 40
-# Needed as seen by BZ: 2327515
-Requires: python%{python3_pkgversion}-omlmd
-%else
+Recommends: podman
 Recommends: python%{python3_pkgversion}-omlmd
```
ericcurtin, line 51: We could actually remove this altogether potentially

rhatdan: If you remove it the first time you pull from Ollama it is going to complain about it not being installed.

ericcurtin: This isn't used for pulling from Ollama, the Ollama puller is written from scratch in python. What omlmd was doing before was a wrapper for oras and tried to pull OCI artefacts using oras.

rhatdan: Ok I see self._pull_omlmd(args) which looks wrong, it is only called in oci.  We really should not call it there, and then we can remove all of the code.

PR #591, opened 2025-01-15
Remove omlmd from OCI calls

rhatdan:
Also Simplify Spec File

Fedora 39 is no longer supported, so remove checks in spec file.

More Podman to Recommends, You can run RamaLama with no container engine or with Docker.

## Summary by Sourcery

Remove the dependency on `omlmd` for pulling OCI models. Use container engines like Podman or Docker for pulling OCI images.

New Features:
- Support pulling models directly using the specified container engine.

Enhancements:
- Simplify OCI model pulling process.

Tests:
- Update tests to reflect the change in the OCI model pulling mechanism.

### DIFF
--- a/test/system/050-pull.bats
+++ b/test/system/050-pull.bats
@@ -54,6 +54,7 @@ load setup_suite
 
 # bats test_tags=distro-integration
 @test "ramalama pull oci" {
+    skip "Waiting for podman artiface support" 
     run_ramalama pull oci://quay.io/mmortari/gguf-py-example:v1
     run_ramalama list
     is "$output" ".*quay.io/mmortari/gguf-py-example" "OCI image was actually pulled locally"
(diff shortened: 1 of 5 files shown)


rhatdan:
@ericcurtin PTAL

ericcurtin:
We still test for it though I see

ramalama pull oci://quay.io/mmortari/gguf-py-example:v1

I guess we are waiting for "podman artefact" or something like that?

rhatdan:
Yes I will remove that test for now, podman artifact PR is in progress.

rhatdan:
@ericcurtin PTAL

test/system/050-pull.bats
```diff
@@ -54,6 +54,7 @@ load setup_suite
 
 # bats test_tags=distro-integration
 @test "ramalama pull oci" {
+    skip "Waiting for podman artiface support"
```
ericcurtin, line 57: Typo here, not worth blocking the merge though, it should be artifact

ericcurtin (approved):

PR #592, opened 2025-01-15
update mountPath in kube yaml

sallyom:
I've found this change is necessary to make a working kube.yaml for `podman kube play ramalama_*.yaml` 

without the fix:
```
$ podman kube play ramalama_MicXluJjP3.yaml 
Pod:
starting container d5f9654b06b3a6fb209a3a9c38986e9c78313417c57a55435050b08d87ca0fab: not a directory
```

to test, this will serve granite-code at localhost:8888:
```
ramalama serve --generate kube granite-code
podman kube play ramalama_*.yaml --publish 8888:8080
```

## Summary by Sourcery

Bug Fixes:
- Fix issue where `podman kube play` would fail with a "not a directory" error when the mount path was set to a directory that didn't exist.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ venv/
 *.container
 *.image
 *.volume
+**ramalama_**
--- a/ramalama/kube.py
+++ b/ramalama/kube.py
@@ -2,7 +2,7 @@ import os
 
 
 from ramalama.version import version
-from ramalama.common import genname, MNT_DIR, get_env_vars
+from ramalama.common import genname, MNT_FILE, get_env_vars
 
 
 class Kube:
@@ -25,8 +25,7 @@ class Kube:
     def gen_volumes(self):
         mounts = f"""\
         volumeMounts:
-        - mountPath: {MNT_DIR}
-          subPath: /models
+        - mountPath: {MNT_FILE}
           name: model"""
 
         volumes = """
--- a/ramalama/quadlet.py
+++ b/ramalama/quadlet.py
@@ -1,6 +1,6 @@
 import os
 
-from ramalama.common import default_image, MNT_DIR, MNT_FILE, get_env_vars
+from ramalama.common import default_image, MNT_FILE, get_env_vars
 
 
 class Quadlet:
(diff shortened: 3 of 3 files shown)


ericcurtin:
Thanks @sallyom I've been meaning to check whether these commands:

```
$ ramalama --dryrun serve llama:3.1
podman run --rm -i --label RAMALAMA --security-opt=label=disable --name ramalama_7mb6H6CGJf --pull=newer -t -p 8080:8080 --device /dev/dri --mount=type=image,src=llama:3.1,destination=/mnt/models,ro,subpath=/models quay.io/ramalama/ramalama:latest llama-server --port 8080 -m /mnt/models/model.file -c 2048 --temp 0.8 -ngl 999 --host 0.0.0.0
$ ramalama --dryrun --runtime vllm serve llama:3.1
podman run --rm -i --label RAMALAMA --security-opt=label=disable --name ramalama_2gEHBiYV0T --pull=newer -t -p 8080:8080 --device /dev/dri --mount=type=image,src=llama:3.1,destination=/mnt/models,ro,subpath=/models quay.io/modh/vllm:rhoai-2.17-cuda --port 8080 --model /mnt/models/model.file --max_model_len 2048
```

still lineup with our generator equivalents. I guess this PR proves they drifted.

ericcurtin:
@rhatdan PTAL. I think you know better than me at this stage what we want the podman run, quadlet, kube file to look like

rhatdan:
I think we have to be careful here.  I want to try this out with Ollama Models, HuggingFace Models, OCI model car models and OCI Model Raw models.  I have a feeling we need to do this two ways.  Depending on the model, we mount at the file path or at the directory path.  Can not try out right now since I am on train.

sallyom:
Also, might be better to use the actual model filename rather than model.file & add ability to load multiple models & switch quickly between them. I figured we'd want to look at this PR closer.

rhatdan:
The problem is without standardizing the OCI Models, we don't know what the path is, which is why we want the models to always have a `model.file` within them. If we had standards, we could have an attribute of the OCI Model to tell us what the path is.  Certain models use additional data within the MODEL DIR to enhance the model.

rhatdan:
We now report an alias, so closing.  Reopen if this is still needed.

PR #595, opened 2025-01-16
Build with curl support

pepijndevos:
This allows the standalone container to download models

As proposed in https://github.com/containers/ramalama/discussions/575 this is useful to use ramalama containers as Home Assistant addons.

## Summary by Sourcery

Build:
- Build the standalone container with cURL.


ericcurtin:
I guess we can add this, it's a little unnecessary as "ramalama run/ramalama pull" has it's own http client for pulling

ericcurtin (approved):

rhatdan:
I am fine with this.

PR #596, opened 2025-01-16
Add ramalama image built on Fedora using Fedora's rocm packages

maxamillion:
Add ramalama image built on Fedora using Fedora's rocm packages, this enables more models of embedded GPUs in the Ryzen APUs for more series of gfx than is available in the official ROCm packages from AMD that are used in the UBI based default images.


rhatdan:
Do we actually need a separate Containerfile or could we just do this with a podman build --from fedora ... or I guess to make it more container engine agnostic use BUILDARGS.

container-images/ramalama-fedora/Containerfile
```diff
@@ -0,0 +1,8 @@
+FROM registry.fedoraproject.org/fedora:latest
+
+COPY ../scripts /scripts
```
ericcurtin, line 3: I know it's not obvious apologies, but "ramalama" means use a generic driver, aka the vulkan version of kompute. I think what we should do here is:

/scripts/build_llama_and_whisper.sh "rocm" "fedora"

with the first parameter being the GPU/AI driver (rocm in this case) and the second parameter could be fedora, ubi, etc.

rhatdan: SGTM

maxamillion:
> Do we actually need a separate Containerfile or could we just do this with a podman build --from fedora ... or I guess to make it more container engine agnostic use BUILDARGS.

That's probably a better option, I'll go that route.

container-images/scripts/build_llama_and_whisper.sh
```diff
@@ -36,7 +36,11 @@ dnf_install() {
     dnf install -y asahi-repos
     dnf install -y mesa-vulkan-drivers "${vulkan_rpms[@]}" "${rpm_list[@]}"
   elif [ "$containerfile" = "rocm" ]; then
-    dnf install -y rocm-dev hipblas-devel rocblas-devel
+    if [ "${ID}" = "fedora" ]; then
```
ericcurtin, line 39: Much nicer :)

container-images/rocm-fedora/Containerfile
```diff
@@ -1,8 +1,10 @@
 FROM registry.fedoraproject.org/fedora:latest
 
+RUN dnf update -y --refresh && dnf clean all
```
ericcurtin, line 3: This step doesn't seem unique to this Containerfile or anything, no big deal though

ericcurtin (approved):
Merging on green build

maxamillion:
@ericcurtin would you prefer I converge the two `rocm` Containerfiles and use `--build-args` as @rhatdan suggested or would that be a separate PR?

ericcurtin:
@maxamillion I don't mind, if it works, you'll probably have to edit container_build.sh for that technique

ericcurtin:
The build failed because it the container image is too large, we probably want to remove some older GPUs from it, for example on the UBI versions we rm *gfx9* files, there are GPUs prior to ~2018

ericcurtin:
@debarshiray and @owtaylor I remember you guys wanted some Fedora based images. Well... It's coming it seems 😄  Thanks to @maxamillion

container-images/scripts/build_llama_and_whisper.sh
```diff
@@ -115,7 +124,8 @@ main() {
   dnf clean all
   rm -rf /var/cache/*dnf* /opt/rocm-*/lib/llvm \
     /opt/rocm-*/lib/rocblas/library/*gfx9* \
-    /opt/rocm-*/lib/hipblaslt/library/*gfx9* llama.cpp whisper.cpp
+    /opt/rocm-*/lib/hipblaslt/library/*gfx9* llama.cpp whisper.cpp \
+    /usr/lib64/rocm/gfx9*
```
ericcurtin, line 128: This might make it small enough :)

It's actually a restriction on the storage space available on our github runner, but it's a welcome restruction, we don't want 20GB container images

One could always contribute a rocm-legacy-fedora type image for gfx9 if they had the motivation to do so.

ericcurtin: We should actually simplify the above line to:

/opt/rocm-*/lib/*/library/*gfx9*

not necessary in this PR or anything.

maxamillion: we'll find out if it's small enough soon! :smile:

maxamillion: @ericcurtin it appears that the build is still angry about it :thinking:

ericcurtin: Yeah I guess you just gotta podman run into it and look around and see what's big, gfx9 was enough to remove for the UBI versions. Trim, test to make sure you didn't break the ROCm support, then add the "rm -fr" here.

ericcurtin: I'm certainly glad GitHub complains about it, don't want humongous images, keeps us on our toes 😄

maxamillion: @ericcurtin question - what's the size goal I'm trying to hit?

ericcurtin: I can't remember exactly I think it's 10G which is pretty damn big. The ROCm images are the only images that suffer from this, I've never had to trim a CUDA one for example.

ericcurtin: It's arguably worth logging an issue in ROCm upstream, I find it hard to believe all those gfx9 files are completely unique, I'm sure there's loads of duplicate data in those files that could be shared using some common files or some other form of de-duplication.

ericcurtin: It leaves a bad first impression for ramalama if users have to pull huge container images just to get going, some people won't wait around for hours for a large image to be pulled.

ericcurtin: We could also do container images like rocm-gfx9-fedora, rocm-gfx10-fedora, etc. We may be able to detect what generation we are pre-pull

ericcurtin:
UBI has nothing lower than gfx9 @maxamillion I see you found gfx8 . If there's gfx7 or less and less, likely worth removing also.

maxamillion:
@ericcurtin nothing older than gfx8, but it's still a lot of space (granted, it was 14G before I pulled both gfx9* and gfx8* out)

```
[root@a854647ecf79 /]# ls /usr/lib64/rocm
gfx10  gfx11  gfx1100  gfx1103
[root@a854647ecf79 /]# du -sh /usr/lib64/rocm/
5.3G    /usr/lib64/rocm/
```

ericcurtin:
Looks like you are almost there, if only we could rank the GPU models by units sold or something and trim based on least units sold.

ericcurtin:
Could also be interesting to see if there is further compression we can do other than whatever podman build does as standard. @rhatdan any compression techniques, etc. we could do off the top of your head?

ericcurtin:
Maybe we will be lucky and this will be small enough to go green

ericcurtin:
Seperate images for gfx8 gfx9 gfx10 gfx11 would be a welcome addition to be honest. We may be able to detect that by querying /proc /sys etc. Pre-pull. Also any GPU with < 1GB VRAM has incredibly low value in supporting, often CPU inferencing is faster (maybe even < 2GB, 4GB, 8GB VRAM). An 8 GB VRAM GPU is considered small these days.

maxamillion:
well bummer ... even splitting them all out the builds are too large :thinking:

ericcurtin:
We could just use the exact same list of GPUs as the UBI9 images and enable additional ones on demand as people request.

maxamillion:
@ericcurtin I don't see a way to trim this down further, I'm open to suggestion though.

This is some sample output from the `rocm-fedora-gfx11` image.
```
[root@167cf24d485c /]# du -sh /usr/* | grep G
8.6G    /usr/lib64

[root@167cf24d485c /]# du -sh /usr/lib64/* | grep G
1.1G    /usr/lib64/librocblas.so.4.2
1.9G    /usr/lib64/librocsparse.so.1.0
3.0G    /usr/lib64/rocm

[root@167cf24d485c /]# du -sh /usr/lib64/rocm/* | grep G
2.3G    /usr/lib64/rocm/gfx11

[root@167cf24d485c /]# du -sh /usr/lib64/rocm/gfx11/* | grep G
2.3G    /usr/lib64/rocm/gfx11/lib

[root@167cf24d485c /]# du -sh /usr/lib64/rocm/gfx11/lib/*
32K     /usr/lib64/rocm/gfx11/lib/cmake
4.0K    /usr/lib64/rocm/gfx11/lib/libhipblas.so
4.0K    /usr/lib64/rocm/gfx11/lib/libhipblas.so.2
996K    /usr/lib64/rocm/gfx11/lib/libhipblas.so.2.2
4.0K    /usr/lib64/rocm/gfx11/lib/librocblas.so
4.0K    /usr/lib64/rocm/gfx11/lib/librocblas.so.4
486M    /usr/lib64/rocm/gfx11/lib/librocblas.so.4.2
4.0K    /usr/lib64/rocm/gfx11/lib/librocsolver.so.0
359M    /usr/lib64/rocm/gfx11/lib/librocsolver.so.0.2
4.0K    /usr/lib64/rocm/gfx11/lib/librocsparse.so.1
830M    /usr/lib64/rocm/gfx11/lib/librocsparse.so.1.0
589M    /usr/lib64/rocm/gfx11/lib/rocblas
```

Based on the output, there's not really anything in there we can rip out based on specific GPU models. It's all just bundled into giant shared object files :confused:

maxamillion:
@ericcurtin in a shocking turn of events, if I set it to `rawhide` as a base image the `rocm-fedora-gf11` image is about 4gb smaller than if I build with Fedora 41 :shrug:  .... any reservations with switching everything to rawhide for now?

maxamillion:
> @ericcurtin in a shocking turn of events, if I set it to `rawhide` as a base image the `rocm-fedora-gf11` image is about 4gb smaller than if I build with Fedora 41 🤷 .... any reservations with switching everything to rawhide for now?

lol nvm .. it's smaller because cmake failed :facepalm:

rhatdan:
I am fine with that, do you know why it is smaller?

maxamillion:
> I am fine with that, do you know why it is smaller?

because cmake failed :facepalm:

ericcurtin:
No issue with rawhide, eventually we can call it Fedora 42

ericcurtin:
This is an isolated image, so that helps

ericcurtin (approved):
LGTM, hopefully we can find a way of figuring out which gfx generation we are in a follow on PR

maxamillion:
> LGTM, hopefully we can find a way of figuring out which gfx generation we are in a follow on PR

I still need to fix one last nagging issue with the rocblas path and then I'll ask for a merge. 🙂

ericcurtin:
rocm-fedora-gfx10 container image is too large from my reading of the build logs

debarshiray:
> @debarshiray and @owtaylor I remember you guys wanted some Fedora based RamaLama images. Well... It's coming it seems 😄 Thanks to @maxamillion

This is fantastic.

While reading through the comments, I saw that you were fighting against the size of the ROCm images because of those different `gfx` files that ROCm has.  I was talking to Kevin Martin about those in the context of Fedora's `llama-cpp` package.  The moment I [turned on](https://src.fedoraproject.org/rpms/llama-cpp/c/329939bab49a33a74adf3aa7d7de636734d833a7) the GPU accelerated HIP code that gets built with the ROCm compiler, the build time and the size of the RPMs exploded, because the HIP sources were being separately compiled for every different GPU.  I was told that it's inherent to how AMD's design works and is a noticeable difference from NVIDIA, and that AMD is aware of it.

The idea of separate ROCm images for different AMD GPUs does sound tempting to me.  I wonder if we should split things like `llama-cpp` similarly.

Finally, if you are looking for the list of GPUs enabled in Fedora's ROCm, then look at [ROCM_GPUS](https://src.fedoraproject.org/rpms/rocm-rpm-macros/blob/rawhide/f/default) in `rocm-rpm-macros`.

ericcurtin:
It does look like poor design, it's hard to believe each of those files are so unique, but that's an AMD problem to solve I guess..

debarshiray:
> LGTM, hopefully we can find a way of figuring out which gfx generation we are in a follow on PR

One can do:
```
$ rocminfo | grep ' gfx' | cut --delimiter ':' --fields 2 | sed 's/ //g'
gfx1031
```

... but I suppose you want to do it in Python with saner APIs.  :)

debarshiray:
> It does look like poor design, it's hard to believe each of those files are so unique, but that's an AMD problem to solve I guess...

I don't know all the details.  Kevin told me that doing it this way, without a deeper redesign, improves performance by a few percentage points and AMD doesn't want to let that go.

ericcurtin:
> > LGTM, hopefully we can find a way of figuring out which gfx generation we are in a follow on PR
> 
> One can do:
> 
> ```
> $ rocminfo | grep ' gfx' | cut --delimiter ':' --fields 2 | sed 's/ //g'
> gfx1031
> ```
> 
> ... but I suppose you want to do it in Python with saner APIs. :)

It would be nice if we could avoid the need for rocminfo to be installed on the host system, feedback from a user:

`
Screenshot 2025-02-04 at 20 02 27
`

he was impressed how he had to install basically nothing and it worked (and he's right, it's how I use it).

ericcurtin:
rocminfo is open source AFAIK, so we may actually be able to figure out the piece of code where it gets that gfx data from

debarshiray:
> > LGTM, hopefully we can find a way of figuring out which gfx generation we are in a follow on PR
> 
> One can do:
> 
> ```
> $ rocminfo | grep ' gfx' | cut --delimiter ':' --fields 2 | sed 's/ //g'
> gfx1031
> ```
> 
> ... but I suppose you want to do it in Python with saner APIs. :)

This is slightly better:
```
$ rocm_agent_enumerator -name
gfx1031
```

... and `rocm_agent_enumerator` is a [one file](https://github.com/ROCm/rocminfo/blob/amd-staging/rocm_agent_enumerator) Python program.  :)

ericcurtin:
Great find @debarshiray ! We also have this problem (which is accounted for in some parts of RamaLama), my machine returns this:

```
gfx1102
gfx90c
```

one of those amd gpus it not suitable for inferencing, it's the integrated graphics, it has tiny VRAM and one does the 1102, it hass 8G VRAM. Just something to keep in mind to take into account

ericcurtin:
readFromKFD() function is basically the important part of that script.

debarshiray:
Reading the `rocm_agent_enumerator` code, it seems that a decent first pass would be to try the approach taken by `readFromKFD()` and fallback to `readFromLSPCI()`.

>We also have this problem (which is accounted for in some parts of RamaLama), my machine returns this:
> 
> ```
> gfx1102
> gfx90c
> ```
> 
> one of those amd gpus it not suitable for inferencing, it's the integrated graphics, it has tiny VRAM and one does the 1102, it hass 8G VRAM. Just something to keep in mind to take into account

What does the `properties` file say inside the sub-directories under `/sys/class/kfd/kfd/topology/nodes`?  That's what `readFromKFD()` looks at.

ericcurtin:
Could probably boil it down to ~10 lines of python if we need to

debarshiray:
> ```
> $ grep gfx /sys/class/kfd/kfd/topology/nodes/*/properties
> /sys/class/kfd/kfd/topology/nodes/0/properties:gfx_target_version 0
> /sys/class/kfd/kfd/topology/nodes/1/properties:gfx_target_version 110002
> /sys/class/kfd/kfd/topology/nodes/2/properties:gfx_target_version 90012
> ```
> 
> there's a little decimal to hex conversion to get "02" and "0c"

Yes, it's:
```
major_ver = int((device_id / 10000) % 100)
minor_ver = int((device_id / 100) % 100)
stepping_ver = int(device_id % 100)
```

... and then:
```
"gfx" + format(major_ver, 'd') + format(minor_ver, 'x') + format(stepping_ver, 'x')
```

Did you figure out how to filter out your smaller GPU?

debarshiray:
> Did you figure out how to filter out your smaller GPU?

Skimming through the [AMDKFD code](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/gpu/drm/amd/amdkfd/) in the Linux kernel gave me some ideas, particularly `kfd_debug_print_topology()`.

Both `cpu_cores_count` and `simd_count` will be set to non-zero values for accelerated processing units (or APUs) in the `/sys/class/kfd/kfd/topology/nodes/*/properties` file.  I have a desktop with an AMD CPU and GPU, not a laptop, so I don't have an AMD APU to verify this, but the rest of the logic does hold for my hardware.

ericcurtin:
> > ```
> > $ grep gfx /sys/class/kfd/kfd/topology/nodes/*/properties
> > /sys/class/kfd/kfd/topology/nodes/0/properties:gfx_target_version 0
> > /sys/class/kfd/kfd/topology/nodes/1/properties:gfx_target_version 110002
> > /sys/class/kfd/kfd/topology/nodes/2/properties:gfx_target_version 90012
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > there's a little decimal to hex conversion to get "02" and "0c"
> 
> Yes, it's:
> 
> ```
> major_ver = int((device_id / 10000) % 100)
> minor_ver = int((device_id / 100) % 100)
> stepping_ver = int(device_id % 100)
> ```
> 
> ... and then:
> 
> ```
> "gfx" + format(major_ver, 'd') + format(minor_ver, 'x') + format(stepping_ver, 'x')
> ```
> 
> Did you figure out how to filter out your smaller GPU?

Yes it's code elsewhere in RamaLama:

```
    # ROCm/AMD CASE
    i = 0
    gpu_num = 0
    gpu_bytes = 0
    for fp in sorted(glob.glob('/sys/bus/pci/devices/*/mem_info_vram_total')):
        with open(fp, 'r') as file:
            content = int(file.read())
            if content > 1073741824 and content > gpu_bytes:
                gpu_bytes = content
                gpu_num = i

        i += 1
```

basically ignore every GPU with less than 1G VRAM and choose the one with the most VRAM.

container-images/scripts/build_llama_and_whisper.sh
```diff
@@ -22,26 +23,36 @@ dnf_install() {
                      "spirv-tools" "glslc" "glslang")
   if [ "$containerfile" = "ramalama" ] || [ "$containerfile" = "rocm" ] || \
     [ "$containerfile" = "vulkan" ]; then # All the UBI-based ones
-    local url="https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm"
-    dnf install -y "$url"
-    crb enable # this is in epel-release, can only install epel-release via url
-    dnf --enablerepo=ubi-9-appstream-rpms install -y "${rpm_list[@]}"
+    if [[ "${ID}" == "rhel" || "${ID}" == "redhat" || "${ID}" == "centos" ]]; then
```
ericcurtin, line 26: I think we'd catch all these with:

PLATFORM_ID

ericcurtin: Or even ID is not fedora

container-images/scripts/build_llama_and_whisper.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+source /etc/os-release
```
ericcurtin, line 2: Normally like to keep everything in main() function

ericcurtin:
I'd merge this only you still have "WIP:" in it. I didn't check the CI logs yet to see if all these were actually built

ericcurtin:
But LGTM in general

ericcurtin:
My suspicions were right, the builds for the container images were turned off, we should turn them on again.

container-images/rocm-fedora-gfx10/Containerfile
```diff
@@ -0,0 +1,11 @@
+FROM registry.fedoraproject.org/fedora:latest
```
rhatdan, line 1: Rather then have 4 Containerfiles, that we need to manage can we just consolidate this into one ROCM containerfile and then manage this with container_build.sh and build_llama_whisper.sh

These containerfiles have very little difference that I can see then which files are being removed for size.

If users say make `build IMAGE=rocm-fedora-gfx10` Then this should be enough to change the `rm -fr ...` line and potentially the FROM line.

ericcurtin: @rhatdan the problem is the resulting image is like 40GB , do we want people to have to pull a container image of that size?

There's little getting around it other than splitting.

The reason the UBI9 one doesn't suffer from this is it only includes support for a small subset of AMD GPUs

ericcurtin:
PR to build all container image builds in CI:

https://github.com/containers/ramalama/pull/831

container-images/rocm-fedora-gfx10/Containerfile
ericcurtin: We could just include the UBI9 subset in the Fedora image and add other individual GPUs on demand (like @maxamillion 's GPU). That's another option.

ericcurtin: I see what you mean not @rhatdan I'd just add another argument to the build_llama_and_whisper.sh script if it was me, would remove these lines:

```
RUN dnf update -y --refresh && dnf clean all
...
ENV WHISPER_CPP_SHA=${WHISPER_CPP_SHA}
ENV LLAMA_CPP_SHA=${LLAMA_CPP_SHA}
```

container-images/rocm-fedora-gfx8/Containerfile
```diff
@@ -0,0 +1,11 @@
+FROM registry.fedoraproject.org/fedora:latest
```
ericcurtin, line 1: Would prefer to tag :latest to a fixed fedora version and let renovate update it.

maxamillion:
@ericcurtin @rhatdan sorry for the delay, I finally had time to refactor this based on feedback and it should be ready for review now

rhatdan:
@maxamillion Are you thinking we will have 4 or more ROCM images?

container-images/scripts/build_llama_and_whisper.sh
```diff
@@ -133,7 +146,29 @@ clone_and_build_ramalama() {
...
+      ;;
+    gfx9)
+      rm -fr /usr/lib64/rocm/gfx8* /usr/lib64/rocm/gfx10* /usr/lib64/rocm/gfx11* 
+      ;;
+    gfx10)
+      rm -fr /usr/lib64/rocm/gfx8* /usr/lib64/rocm/gfx9* /usr/lib64/rocm/gfx11* 
+      ;;
+    gfx11)
+      rm -fr /usr/lib64/rocm/gfx8* /usr/lib64/rocm/gfx9* /usr/lib64/rocm/gfx10* && \
+      ln -s /usr/lib64/rocm/gfx1103/lib/rocblas/library/TensileLibrary_lazy_gfx1103.dat \
```
ericcurtin, line 162: Interesting you figured out we need to do this, worth alerting Tom Rix to this, it might be better to fix this in the rpm package. Worth checking if this is needed for UBI9 also

maxamillion: @ericcurtin that isn't needed in the Rawhide and Fedora 42 nightlies of ROCm but support for gfx1103 is considered experimental in Fedora 41 so I'm not sure where the fix is appropriate.

maxamillion:
> @maxamillion Are you thinking we will have 4 or more ROCM images?

Unfortunately I'm not sure. I wanted it to be all-in-one but the made the container image massive and it was suggested that we break it out so I did. I don't know how many Fedora ROCm images we'll need over time, but I suspect at least one per `gfx` major release series unless we can find an alternative method of making the ROCm stack smaller in the Fedora builds.

rhatdan:
Ok we can start by building multiple images, but we need to get a naming standard and make the GPU check pull the right image.  Also if we can make sure we have as big of a shared image as possible,

container-images/scripts/build_llama_and_whisper.sh
trixirt: Yes.  gfx1103 was experimental in F41.
gfx1035,gfx1103,gfx1150,gfx1151,gfx1152,gfx1200,gfx1201 were added in F42.

ericcurtin:
I dunno by how much or which one, but it appears one of the images is still too big.

ericcurtin:
If you want @maxamillion to get this merged you can skip this image in the build here:

```
    # todo, trim and get building in CI again
    if $ci && [[ "$i" =~ ^container-images/rocm$ ]]; then
      continue
    fi
```

but we should then try and trim in another follow on PR to get hook this up to the CI build.

rhatdan:
I am fine with not building these images in the test so we can get this in.  It would be good to get these released for next weeks release.

maxamillion:
@rhatdan I added `rocm-fedora` to be ignored in the github workflow, if I did that correctly this should be good to merge if you're happy with it

PR #501, opened 2024-12-03
[WIP] Add ramalama rag command

rhatdan:

PR #597, opened 2025-01-16
Begin process of packaging PRAGmatic

rhatdan:
Building Pragmatic into a container image is fairly easy.

podman build --build-arg IMAGE=quay.io/ramalama/rocm -t quay.io/ramalama/rocm-rag container-images/pragmatic

## Summary by Sourcery

Build:
- Install Python and the PRAGmatic package.

### DIFF
--- a/container-images/pragmatic/build_pragmatic.sh
+++ b/container-images/pragmatic/build_pragmatic.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+
+set -exu -o pipefail
+
+export  PYTHON_VERSION="python3 -m"
+if [ "$(python3 --version)" \< "Python 3.11" ]; then
+    dnf install -y python3.11 python3.11-pip git
+    export PYTHON_VERSION="/usr/bin/python3.11 -m"
+else
+    dnf install -y python3-pip git
+fi
+
+cuda="cu124"
+
+rocm="rocm6.2"
+
+cpu="cpu"
+
+vulkan=$cpu
+
+asahi=$cpu
+
+install_pytorch() {
+    version=${!1}
+    echo ${PYTHON_VERSION} pip install torch==${version} -f https://download.pytorch.org/whl/torch_stable.html
+    ${PYTHON_VERSION} pip install torch==${version} -f https://download.pytorch.org/whl/torch_stable.html
+}
+
+clone_and_build_pragmatic() {
+if [ "$2" == "docling" ]; then
+   ${PYTHON_VERSION} pip install docling --extra-index-url https://download.pytorch.org/whl/$1
+fi
+  git clone https://github.com/redhat-et/PRAGmatic
+  cd PRAGmatic
+  git submodule update --init --recursive
+
+  ${PYTHON_VERSION} pip install -r requirements.txt --prefix=/usr
+  ${PYTHON_VERSION} pip install --prefix=/usr .
+  cd ..
+}
+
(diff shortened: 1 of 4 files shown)


rhatdan:
This could get painful:
$ podman images
REPOSITORY                           TAG                   IMAGE ID      CREATED         SIZE
quay.io/ramalama/ramalama            latest                4c2e061e2975  7 minutes ago   10.5 GB

container-images/scripts/build_llama_and_whisper.sh
```diff
@@ -94,6 +95,18 @@ clone_and_build_llama_cpp() {
   cd ..
 }
 
+clone_and_build_pragmatic() {
+  local llama_cpp_sha="924518e2e5726e81f3aeb2518fb85963a500e93a"
```
ericcurtin, line 99: Unused stuff could be cleaned up here

ericcurtin (approved):

container-images/pragmatic/build_pragmatic.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+clone_and_build_pragmatic() {
```
ericcurtin, line 3: We should add

set -exu -o pipefail

We want the build to fail if these commands fail.

rhatdan: Ok, I am still working on making the install smaller.  Gotten CPU cut in 1/3 but little change in ROCM and CUDA.

rhatdan: Should we add the same flag to build_llama_and_whisper.sh?

ericcurtin: `set -ex` is there, it's just in the main function (we could add, `-o pipefail`, etc. if it suits). I tend to create main functions in shell scripts to encourage using scope and less global variable.

`set -e` alone covers a lot to be fair sometimes it's sufficient.

I do like to add `-x` at times also, makes CI easy to debug.

But `set -e` at least is a good idea.

container-images/pragmatic/build_pragmatic.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
+set -exu -o pipefail
+
+export  PYTHON_VERSION="python3 -m"
+if [ "$(python3 --version)" \< "Python 3.11" ]; then
+    dnf install -y python3.11 python3.11-pip
```
dougsland, line 7: Question, python3.12 and python3.13 are not supported? Still learning these bits.

rhatdan: They are supported, this is just checking for RHEL 9 systems with python 3.9.  Basically ramalama requires 3.11 or later.

rhatdan:
@ericcurtin @bmahabirbu this is ready to go in.  It does not build anything yet, just allows people to build RAG Models layers on top of our images.  Once this is in, we can start experimenting with using RAG with RamaLama.

ericcurtin (approved):

bmahabirbu:
@rhatdan thanks for merging! Will play around with this this weekend!

rhatdan:
Here is the lastest stuff I am working on for creation of the vector.db

https://github.com/containers/ramalama/pull/501

rhatdan:
I also just got this merged: https://github.com/redhat-et/PRAGmatic/pull/39

PR #600, opened 2025-01-19
Various README.md updates

ericcurtin:
Want to show our new progress bar. Update the diagram. Use newer model granite3-moe as an example.

## Summary by Sourcery

Documentation:
- Update README.md to reflect the new progress bar and use of the granite3-moe model.


ericcurtin:
One of the main drivers for this is we encountered a bug that reproduces using:

granite-code

model recently, @engelmi is looking into it. While we are fixing it's probably better we don't use it as a reference model. Anyway we have newer models such as the granite3 family.

rhatdan:
LGTM

PR #603, opened 2025-01-21
Update to version that has command history

ericcurtin:
We added a feature to llama-run to traverse prompt history via up/down arrows. Kind of like a shell.

## Summary by Sourcery

New Features:
- Add command history to llama-run.


rhatdan:
LGTM

PR #605, opened 2025-01-21
ROCm build broken

ericcurtin:
The build flags changed in upstream llama.cpp so we were no longer building with ROCm acceleration.

## Summary by Sourcery

Fix ROCm builds.

Build:
- Use `-DGGML_HIP=ON` build flag instead of `-DGGML_HIPBLAS=1`.
- Remove unnecessary ROCm libraries after build.

### DIFF
--- a/container-images/scripts/build_llama_and_whisper.sh
+++ b/container-images/scripts/build_llama_and_whisper.sh
@@ -39,11 +39,15 @@ dnf_install() {
   fi
 }
 
+cmake_check_warnings() {
+  awk -v rc=0 '/CMake Warning:/ { rc=1 } 1; END {exit rc}'
+}
+
 cmake_steps() {
   local cmake_flags=("$@")
-  cmake -B build "${cmake_flags[@]}"
-  cmake --build build --config Release -j"$(nproc)"
-  cmake --install build
+  cmake -B build "${cmake_flags[@]}" 2>&1 | cmake_check_warnings
+  cmake --build build --config Release -j"$(nproc)" 2>&1 | cmake_check_warnings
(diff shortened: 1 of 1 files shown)


ericcurtin:
@maxamillion @bmahabirbu upstream llama.cpp changed their build flags upstream for ROCm, as a result we've been building without ROCm here for a little while as the code would still build, just without ROCm.

ericcurtin:
This was happening basically:

CMake Warning:
  Manually-specified variables were not used by the project:

    GGML_HIPBLAS

ericcurtin:
@rhatdan we probably need new container images, once we get the green PRs in here. ROCm is broken

bmahabirbu:
Really good catch! I wonder if there is a way to catch these sort of errors? I can look into editing the make file to look for mismatched arguments

ericcurtin:
> Really good catch! I wonder if there is a way to catch these sort of errors? I can look into editing the make file to look for mismatched arguments

If you find out great, it would be appreciated, I tried briefly to make cmake treat this as an error, but didn't figure it out.

ericcurtin:
I thought of something to check for cmake warnings @bmahabirbu , I'd prefer something that was built-into cmake, as it would be more elegant than this, but this should work.

ericcurtin:
This build may actually fail, it might find other hidden warnings around the place

ericcurtin:
> LGTM!
> 
> This is great, thank you for catching this! This might explain why I wasn't seeing any GPU utilization on my Ryzen APU when watching `nvtop` ... I had tossed it up to `nvtop` oddities.

Could have been! If you want to test this before it gets pushed, there is a flag to manually pass container images to RamaLama (you could build this locally).

PR #606, opened 2025-01-21
Remove these lines they are unused

ericcurtin:
We now store the sha in the shell script

## Summary by Sourcery

Build:
- Remove the `WHISPER_CPP_SHA` and `LLAMA_CPP_SHA` environment variables from the Containerfile.


rhatdan:
LGTM

PR #609, opened 2025-01-21
Had to make this change for my laptop to suppor nvidia

rhatdan:
## Summary by Sourcery

Chores:
- Use the first available GPU instead of the second one.

### DIFF
--- a/ramalama/common.py
+++ b/ramalama/common.py
@@ -211,7 +211,7 @@ def get_gpu():
     try:
         command = ['nvidia-smi']
         run_cmd(command).stdout.decode("utf-8")
-        os.environ["CUDA_VISIBLE_DEVICES"] = "1"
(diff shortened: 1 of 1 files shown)


rhatdan:
@bmahabirbu PTAL

ericcurtin (approved):
This will have to be smarter than a hardcoding eventually. ROCm does it properly

bmahabirbu:
LGTM, I had the functionality before @ericcurtin but it called nvidia-smi which was an expensive process. I believe for Cuda we don't even need to specify the gpu and it should work fine gotta test that later and make a pr

bmahabirbu:
Tested my Nvidia gpu with the new change and it worked! Not sure why setting the env var to 1 initially worked for me but I'm all set with this change and can be on the new version

PR #613, opened 2025-01-22
We need the rocm libraries in here

ericcurtin:
We didn't before, now we do, or else we crash.

## Summary by Sourcery

Build:
- Remove the LLVM libraries and ROCm gfx9 libraries from the Llama and Whisper build.


rhatdan:
LGTM

rhatdan:
LGTM

Issue #617, opened 2025-01-22
Provide stats on token/s


vpavlin:
When trying various models on resource restricted devices (e.g. Raspberry Pi 5), it is useful to see how each model performs. Ollama has `--verbose` for `run` command to print stats on how long a given prompt took to process and what was the token/s etc.

It would be great if ramalama could also print stats like that based on some argument

ericcurtin:
If we implemented "ramalama bench" and which effectively did this, would that be sufficient for this @vpavlin ? :

```
$ llama-bench -m /Users/ecurtin/.local/share/ramalama/models/ollama/smollm:135m
| model                          |       size |     params | backend    | threads |          test |                  t/s |
| ------------------------------ | ---------: | ---------: | ---------- | ------: | ------------: | -------------------: |
| llama ?B Q4_0                  |  85.77 MiB |   134.52 M | Metal,BLAS |       6 |         pp512 |     11720.35 ± 83.53 |
| llama ?B Q4_0                  |  85.77 MiB |   134.52 M | Metal,BLAS |       6 |         tg128 |       311.96 ± 11.47 |

build: f8feb4b0 (4453)
```

vpavlin:
Yeah, I guess that would be sufficient :)

ericcurtin:
Wanna open a PR to implement that? It's an accessible first contribution

vpavlin:
I would love to open a PR for this, but my time is limited - I am playing with ramalama in my spare time, completely unrelated to my job:) And I am mostly trying ramalama to see if it will work better than Ollama for me because of my RH history and because I ordered a mini pc with AMD 780m and stumbled upon your comments here https://github.com/ollama/ollama/pull/5059#issuecomment-2601881722 :) 

Hopefully I'll be able to submit some PR(s) at some point, but unlikely in the near future:D Hope it's fine if I just dump all my issues on you though:)

ericcurtin:
Merged

PR #620, opened 2025-01-23
Introduce ramalama bench

ericcurtin:
Allows benchmarking of models, GPU stacks, etc.

### DIFF
--- a/ramalama/model.py
+++ b/ramalama/model.py
@@ -188,7 +188,7 @@ class Model:
             conman_args += ["-e", f"{k}={v}"]
         return conman_args
 
-    def gpu_args(self, force=False, server=False):
+    def gpu_args(self, force=False, runner=False):
         gpu_args = []
         if (
             force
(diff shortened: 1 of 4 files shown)


ericcurtin:
https://github.com/containers/ramalama/issues/617

ramalama/model.py
```diff
@@ -197,10 +197,10 @@ def gpu_args(self, force=False, server=False):
             or os.getenv("CUDA_VISIBLE_DEVICES")
             or platform.machine() in {"aarch64", "arm64"}  # linux and macOS report aarch64 differently
         ):
-            if server:
-                gpu_args += ["-ngl"]  # single dash
+            if runner:
+                gpu_args += ["--ngl"]  # single dash
```
rhatdan, line 201: Comments are backwards.

rhatdan:
I take it bench is a command available for llama.cpp?  We'll have to do something for vllm or just use llama.cpp bench command.

ramalama/model.py
ericcurtin: It's deliberate run is "--ngl" and everything else is "-ngl"

ericcurtin:
> I take it bench is a command available for llama.cpp? We'll have to do something for vllm or just use llama.cpp bench command.

A benchmarking functionality was requested numerous times. I don't know how to do it vllm yet. But llama.cpp but have a simple tool in. I agree we will need a vllm equivalent eventually.

ramalama/model.py
ericcurtin: Oh yes comments, fixing

PR #622, opened 2025-01-23
Avoid dnf install on OSTree system

ericcurtin:
It's not mutable

## Summary by Sourcery

Enhancements:
- Prevent accidental modification attempts on immutable OSTree systems.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -21,8 +21,27 @@ download() {
       --remote-time --retry 10 --retry-max-time 10 "$1"
 }
 
+dnf_install() {
+  if ! available podman; then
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -47,7 +47,7 @@ check_platform() {
       sudo="sudo"
     fi
 
-    if available dnf; then
+    if available dnf && ! grep -q ostree= /proc/cmdline; then
```
rhatdan, line 50: ALso add
! available podman

ericcurtin: Added

rhatdan:
LGTM

PR #623, opened 2025-01-23
Add man page for cuda support

rhatdan:
## Summary by Sourcery

Documentation:
- Add a man page detailing CUDA setup and usage with ramalama on Linux systems.

### DIFF
--- a/docs/ramalama-cuda.7.md
+++ b/docs/ramalama-cuda.7.md
@@ -1,53 +1,84 @@
-# Setting Up RamaLama with CUDA Support in WSL2 Using Docker
+% ramalama 7
 
-This guide will walk you through the steps required to set up RamaLama in WSL2 with CUDA support using Docker.
+# Setting Up RamaLama with CUDA Support on Linux systems
 
-## Prerequisites
-
-1. **NVIDIA Game-Ready Drivers**
-   Ensure that you have the appropriate NVIDIA game-ready drivers installed on your Windows system. This is necessary for CUDA support in WSL2.
-
-2. **Docker Desktop Installation**
-   Install Docker Desktop and in settings under **General**, ensure the option to **Use the WSL 2 based engine** is checked.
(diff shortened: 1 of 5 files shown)


rhatdan:
@bmahabirbu PTAL

docs/ramalama-cuda.7.md
```diff
@@ -0,0 +1,110 @@
...
+   sudo gpg --dearmor -o /usr/share/keyrings/nvidia-container-toolkit-keyring.gpg
+   
+   curl -s -L https://nvidia.github.io/libnvidia-container/stable/deb/nvidia-container-toolkit.list
+   sed 's#deb https://#deb [signed-by=/usr/share/keyrings/nvidia-container-toolkit-keyring.gpg] https://#g' | \
+   sudo tee /etc/apt/sources.list.d/nvidia-container-toolkit.list
+   ```
+
+* Update the packages list from the repository
+
+   ```bash
```
ericcurtin, line 34: Ok I guess but it seems very Debian/Ubuntu-specific

ericcurtin (approved):

docs/ramalama-cuda.7.md
rhatdan: This is the Debian section, above there is a dnf/fedora section.

ericcurtin: Ah yes, we are missing the "add a repo" step in the dnf parts, like we have for the apt-get parts.

We could add these steps to install.sh potentially, but I am wary of over automating the process (plus you gotta detect that we want to use an Nvidia GPU which isn't always as straightforward as it sounds.)

bmahabirbu: I added the debian section as an example originally! The link has the rest of the installation instructions

bmahabirbu:
LGTM!

PR #625, opened 2025-01-25
Added --jinja to llama-run command

engelmi:
In https://github.com/ggerganov/llama.cpp/pull/11016/ `llama.cpp` the support for Jinja templates was added, incl. a [default template](https://github.com/ggerganov/llama.cpp/blob/master/common/common.cpp#L1889-L1894). Simply appending the new `--jinja` to the llama-run command in ramalama gets the granite-code running again. However, this only works for the ollama model, the model pulled from huggingface still seems broken.

```bash
# model registry: ollama, without --jinja
$ ramalama run ollama://granite-code
failed to apply the chat template

# model registry: ollama, with --jinja
$ ramalama run ollama://granite-code
> Write a hello world application in C                                                                     
\```c
#include <stdio.h>

int main() {
    printf("hello world\n");
    return 0;
}
\```
> 

# model registry: huggingface
$ ramalama run granite-code
> Write a hello world application in C                                                                     
Hello,
How are you today?
<|im_end|>
<|im_start|>user
Good. How about you?
<|im_end|>
<|im_start|>assistant
...
```

Trying the `--jinja` also for models with no jinja templates seems to work as well, therefore it seems fine to use `--jinja` in general:
```bash
$ ramalama run ollama://smollm:135m
> tell me a joke
I'd be happy to tell you a joke. Here's one:
```

## Summary by Sourcery

New Features:
- Enable Jinja template support for the ollama model registry.

### DIFF
--- a/ramalama/model.py
+++ b/ramalama/model.py
@@ -282,7 +282,7 @@ class Model:
 
     def build_exec_args_run(self, args, model_path, prompt):
         exec_model_path = model_path if not args.container else MNT_FILE
-        exec_args = ["llama-run", "-c", f"{args.context}", "--temp", f"{args.temp}"]
+        exec_args = ["llama-run", "-c", f"{args.context}", "--temp", f"{args.temp}", "--jinja"]
 
         if args.seed:
             exec_args += ["--seed", args.seed]


rhatdan:
Is there a way to check the model and know that it needs --jinja? So we couldset it automatically?  How would a non expert know that they need that flag?

ericcurtin:
Tagging @ochafik who wrote minja, might have some advice when using the flag is most appropriate

ericcurtin (approved):
Happy to merge this FWIW, sorry for the slow review for a one-liner change, had only looked on phone before :)

rhatdan:
Merging, but I really would like to see if we could get this enabled by default in the cases where it is needed.

ochafik:
> Tagging @ochafik who wrote minja, might have some advice when using the flag is most appropriate

Sorry for slow reply, it’s been flu time here.

The flag should be useful whenever the chat template (in the gguf or overridden by command line flag) is a valid Jinja template. Without it, llama.cpp relies either on pattern detection on said jinja template, or on a set of known template names. If a model had such a short-form template name instead of an actual template, then ‘—jinja’ would fail (no clue if this ever happens tho, hopefully convert.py only stores actual templates in ggufs; otherwise could maybe detect if the string looks like a template).

ericcurtin:
> > Tagging @ochafik who wrote minja, might have some advice when using the flag is most appropriate
> 
> Sorry for slow reply, it’s been flu time here.

Get well soon, and thanks very much for letting us pick your brain.

> 
> The flag should be useful whenever the chat template (in the gguf or overridden by command line flag) is a valid Jinja template. Without it, llama.cpp relies either on pattern detection on said jinja template, or on a set of known template names. If a model had such a short-form template name instead of an actual template, then ‘—jinja’ would fail (no clue if this ever happens tho, hopefully convert.py only stores actual templates in ggufs; otherwise could maybe detect if the string looks like a template).

JadedBlueEyes:
Just chiming in here that when this is re-added, it should also be added to `build_exec_args_serve`. I'm using this with Goose, for example.

PR #628, opened 2025-01-26
added mac cpu only support

bmahabirbu:
inside a podman machine without krun support still allow mac to use ramalama

## Summary by Sourcery

New Features:
- Enable running on CPU for macOS systems inside a podman machine without krun support.


rhatdan:
LGTM

PR #632, opened 2025-01-27
fixed rocm detection by adding gfx targets in containerfile

bmahabirbu:
`REPOSITORY                                 TAG         IMAGE ID      CREATED            SIZE
localhost/rocm                             latest      f901b99e3216  37 minutes ago     6.34 GB
quay.io/ramalama/rocm                      latest      a50e205ddb8f  43 hours ago       6.21 GB
`

Adding the targets increased the container marginally but now it works on my system. Wondering how I can update the quay.io file

### DIFF
--- a/container-images/scripts/build_llama_and_whisper.sh
+++ b/container-images/scripts/build_llama_and_whisper.sh
@@ -62,7 +62,7 @@ configure_common_flags() {
   common_flags=("-DGGML_NATIVE=OFF")
   case "$containerfile" in
     rocm)
-      common_flags+=("-DGGML_HIP=ON")
+      common_flags+=("-DGGML_HIP=ON" "-DAMDGPU_TARGETS=gfx1010,gfx1030,gfx1032,gfx1100,gfx1101,gfx1102")
       ;;
     cuda)
       common_flags+=("-DGGML_CUDA=ON" "-DCMAKE_EXE_LINKER_FLAGS=-Wl,--allow-shlib-undefined")


ericcurtin:
I am curious what's the difference between:

-DAMDGPU_TARGETS=gfx1010,gfx1030,gfx1032,gfx1100,gfx1101,gfx1102

and not using this build flag at all? Just hope we don't lose compatibility with some AMD cards because of manually specifying models. Might be worth reading llama.cll CMakeFiles to be sure

bmahabirbu:
What I found is that when not specifying the targets it doesn't work on any of my and cards and im not sure why. Id assume if you don't specify a target it should work for all but that's not the case.

I also tried specifying the architecture during the runtime of llama-run with an env car to override the default room arch which did override it but still ran into a seg fault issue

Good call I checked out the llama.cpp docker file and they had the arch specified so I went off of that

ericcurtin:
> What I found is that when not specifying the targets it doesn't work on any of my and cards and im not sure why. Id assume if you don't specify a target it should work for all but that's not the case.
> 

There is likely a default set. No targets works with my GPU. I'm just hoping this list enables more GPUs than we had before.

I can't test this though because I'm travelling and don't have access to that GPU.

> I also tried specifying the architecture during the runtime of llama-run with an env car to override the default room arch which did override it but still ran into a seg fault issue
> 
> Good call I checked out the llama.cpp docker file and they had the arch specified so I went off of that

PR #634, opened 2025-01-28
feat: add argument to define amd gpu targets

jobcespedes:
This commit add a container buid arg to set custom the amd gpu targets. It also keeps the default amd gpu target values.

## Summary by Sourcery

New Features:
- Allow users to specify custom AMD GPU targets when building the ROCm container image.

### DIFF
--- a/container-images/rocm/Containerfile
+++ b/container-images/rocm/Containerfile
@@ -3,6 +3,8 @@ FROM registry.access.redhat.com/ubi9/ubi:9.5
 COPY rocm/amdgpu.repo /etc/yum.repos.d/
 COPY rocm/rocm.repo /etc/yum.repos.d/
 COPY scripts /scripts
+
+ARG AMDGPU_TARGETS=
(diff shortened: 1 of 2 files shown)


ericcurtin:
Whats the difference between CMAKE_HIP_ARCHITECTURES and AMDGPU_TARGETS?

Could we de-deduplicate this list somehow so we only have to edit it in one place?

gfx1010,gfx1030,gfx1032,gfx1100,gfx1101,gfx1102

jobcespedes:
Hi,

I am using only `AMDGPU_TARGETS` in the commit I just force pushed. I learned that using `CMAKE_HIP_ARCHITECTURES` and `AMDGPU_TARGETS` is [redundant](https://github.com/ggerganov/llama.cpp/blob/2b8525d5c89b124c4578a2621cbeb64354ff3d9c/ggml/src/ggml-hip/CMakeLists.txt#L33). Consequently, the list is only in one place now.

ericcurtin (approved):

rhatdan:
lgtm

PR #637, opened 2025-01-28
Add perplexity subcommand to RamaLama CLI

ericcurtin:
- Added a new subcommand `perplexity` to the RamaLama CLI in `cli.py`.
- Implemented the `perplexity` method in the `Model` class in `model.py`.
- Updated the documentation in `ramalama.1.md` to include the new `perplexity` command.

## Summary by Sourcery

Add a `perplexity` subcommand to calculate the perplexity of a specified AI model.

New Features:
- Added a new `perplexity` subcommand to the RamaLama CLI.

Tests:
- Added tests for the new perplexity feature.

### DIFF
--- a/docs/ramalama-perplexity.1.md
+++ b/docs/ramalama-perplexity.1.md
@@ -1,0 +1,44 @@
+% ramalama-perplexity 1
+
+## NAME
+ramalama\-perplexity - calculate the perplexity value of an AI Model
+
+## SYNOPSIS
+**ramalama perplexity** [*options*] *model* [arg ...]
+
+## MODEL TRANSPORTS
+
+| Transports    | Prefix | Web Site                                            |
+| ------------- | ------ | --------------------------------------------------- |
+| URL based    | https://, http://, file:// | `https://web.site/ai.model`, `file://tmp/ai.model`|
+| HuggingFace   | huggingface://, hf://, hf.co/ | [`huggingface.co`](https://www.huggingface.co)      |
+| Ollama        | ollama:// | [`ollama.com`](https://www.ollama.com)              |
+| OCI Container Registries | oci:// | [`opencontainers.org`](https://opencontainers.org)|
+|||Examples: [`quay.io`](https://quay.io),  [`Docker Hub`](https://docker.io),[`Artifactory`](https://artifactory.com)|
+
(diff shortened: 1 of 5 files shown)


ramalama/model.py
```diff
@@ -244,6 +245,25 @@ def run(self, args):
...
 
+    def perplexity(self, args):
+        self.check_name_and_container(args)
+        model_path = self.get_model_path(args)
+        exec_args = self.build_exec_args_perplexity(args, model_path)
+        self.execute_model(model_path, exec_args, args)
+
+    def build_exec_args_perplexity(self, args, model_path):
+        exec_model_path = MNT_FILE if args.container else model_path
+        exec_args = ["llama-perplexity"]
```
rhatdan, line 256: Can we do something similar to this for vLLM?

docs/ramalama-perplexity.1.md
```diff
@@ -0,0 +1,44 @@
+% ramalama-perplexity 1
+
+## NAME
+ramalama\-perplexity - measure perplexity of specified AI Model
```
rhatdan, line 4: I have no idea what perplexity is, can you describe it in this command?  When would a user want to run this command?

rhatdan:
I think we should be very careful about adding new commands to RamaLama.  We might want to make it easy to just execute random commands on a model.  Perhaps something like `ramalama exec model llama-perplexity` Rather then add lots of commands specify to certain runtimes.

docs/ramalama-perplexity.1.md
rhatdan: Can you add information from: https://github.com/ggerganov/llama.cpp/blob/master/examples/perplexity/README.md

ramalama/model.py
ericcurtin: Not sure, has to be investigated

docs/ramalama-perplexity.1.md
ericcurtin: Added more info

ramalama/model.py
rhatdan: Well if the perplexity command works for non GGUF models, then we could use the llama.cpp one for both vllm as well as llama.cpp

PR #644, opened 2025-01-28
Add container image to support Intel ARC GPU

cgruver:
This PR adds a container image, `intel-gpu`.

This image enables GPU support for Intel GPUs.  It has been tested on a Meteor Lake chipset with Intel ARC GPU.  I have additional hardware with 11, 12, and 13th gen Intel chipsets which I will also test.

This image includes a build of llama.cpp, built from release `b4523`. 

This image does not include whisper.cpp.

This image is built from Fedora 41.

## Summary by Sourcery

New Features:
- Add `intel-gpu` container image with oneAPI support, enabling GPU acceleration for Intel ARC GPUs.

### DIFF
--- a/container-images/intel-gpu/Containerfile
+++ b/container-images/intel-gpu/Containerfile
@@ -1,0 +1,32 @@
+FROM quay.io/fedora/fedora:41 as builder
+
+COPY intel-gpu/oneAPI.repo /etc/yum.repos.d/
+
+RUN dnf install -y lspci clinfo intel-opencl g++ cmake git libcurl-devel intel-oneapi-base-toolkit ; \
+    git clone https://github.com/ggerganov/llama.cpp.git -b b4523 ; \
+    cd llama.cpp ; \
+    mkdir -p build ; \
+    cd build ; \
+    source /opt/intel/oneapi/setvars.sh ; \
+    cmake .. -DGGML_SYCL=ON -DCMAKE_C_COMPILER=icx -DCMAKE_CXX_COMPILER=icpx -DLLAMA_CURL=ON -DGGML_CCACHE=OFF -DGGML_NATIVE=ON ; \
+    cmake --build . --config Release -j -v ; \
+    cmake --install . --prefix /llama-cpp
+
+FROM quay.io/fedora/fedora:41
+
+ENV LD_LIBRARY_PATH="/usr/local/lib64:/usr/local/lib/:/opt/intel/oneapi/mkl/2025.0/lib:/opt/intel/oneapi/compiler/2025.0/opt/compiler/lib:/opt/intel/oneapi/compiler/2025.0/lib/clang/19/lib:/opt/intel/oneapi/compiler/2025.0/lib:/opt/intel/oneapi/umf/0.9/lib:/opt/intel/oneapi/tbb/2022.0/lib:/opt/intel/oneapi/tcm/1.2/lib:/opt/intel/oneapi/redist/opt/compiler/lib:/opt/intel/oneapi/redist/lib/clang/19/lib:/opt/intel/oneapi/redist/lib:/opt/intel/oneapi/mkl/2025.0/lib:/opt/intel/oneapi/compiler/2025.0/opt/compiler/lib:/opt/intel/oneapi/compiler/2025.0/lib/clang/19/lib:/opt/intel/oneapi/compiler/2025.0/lib:/opt/intel/oneapi/umf/0.9/lib:/opt/intel/oneapi/tbb/2022.0/lib:/opt/intel/oneapi/tcm/1.2/lib:/opt/intel/oneapi/redist/opt/compiler/lib:/opt/intel/oneapi/redist/lib/clang/19/lib:/opt/intel/oneapi/redist/lib"
+
+COPY --from=builder /llama-cpp/bin/ /usr/local/bin/
+COPY --from=builder /llama-cpp/lib/ /usr/local/lib/
+COPY --from=builder /llama-cpp/lib64/ /usr/local/lib64/
+COPY --from=builder /llama-cpp/include/ /usr/local/include/
+COPY intel-gpu/oneAPI.repo /etc/yum.repos.d/
+
+RUN dnf install -y intel-opencl libcurl lspci clinfo intel-oneapi-runtime-compilers intel-oneapi-mkl-core intel-oneapi-mkl-sycl-blas intel-oneapi-runtime-dnnl ; \
+    chown 0:0 /etc/passwd ; \
+    chown 0:0 /etc/group ; \
+    chmod g=u /etc/passwd /etc/group ; \
+    useradd -u 1000 -g render -G video -s /bin/bash -d /home/llama-user llama-user
+
+USER 1000
+WORKDIR /home/llama-user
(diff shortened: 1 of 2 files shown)


ericcurtin:
Approving to start CI but we should make sure this calls:

https://github.com/containers/ramalama/blob/main/container-images/scripts/build_llama_and_whisper.sh

to at least build llama.cpp from that script, so we are consistent but more importantly you'll want the version of llama.cpp to get updated like the other containers for new features, new model support, bug fixes, to make sure this stays compatible with upstream llama.cpp, etc.

cgruver:
> Approving to start CI but we should make sure this calls:
> 
> https://github.com/containers/ramalama/blob/main/container-images/scripts/build_llama_and_whisper.sh
> 
> to at least build llama.cpp from that script, so we are consistent but more importantly you'll want the version of llama.cpp to get updated like the other containers for new features, new model support, bug fixes, to make sure this stays compatible with upstream llama.cpp, etc.

There's a couple of reasons that I didn't use the shell script for this image -

1. There's a note in the llama.cpp docs that refers to some releases breaking the SYCL backend - [https://github.com/ggerganov/llama.cpp/blob/master/docs/backend/SYCL.md#recommended-release](https://github.com/ggerganov/llama.cpp/blob/master/docs/backend/SYCL.md#recommended-release)
2. I discovered that a change to the chat template support in release `b4524` broke the usage of the granite-code models that I was testing with.  So, I need to open an issue on that.  Every release after `b4523` seems to be broken.

ericcurtin:
> > Approving to start CI but we should make sure this calls:
> > https://github.com/containers/ramalama/blob/main/container-images/scripts/build_llama_and_whisper.sh
> > to at least build llama.cpp from that script, so we are consistent but more importantly you'll want the version of llama.cpp to get updated like the other containers for new features, new model support, bug fixes, to make sure this stays compatible with upstream llama.cpp, etc.
> 
> There's a couple of reasons that I didn't use the shell script for this image -
> 
> 1. There's a note in the llama.cpp docs that refers to some releases breaking the SYCL backend - https://github.com/ggerganov/llama.cpp/blob/master/docs/backend/SYCL.md#recommended-release
> 2. I discovered that a change to the chat template support in release `b4524` broke the usage of the granite-code models that I was testing with.  So, I need to open an issue on that.  Every release after `b4523` seems to be broken.

I still think we should continue updating with upstream and the other images.

We have versioning of the RamaLama container images now, so one could always manually peg back to an old RamaLama container image if the really want to. We need to find such bugs and submit issues upstream to llama.cpp

cgruver:
> > > Approving to start CI but we should make sure this calls:
> > > https://github.com/containers/ramalama/blob/main/container-images/scripts/build_llama_and_whisper.sh
> > > to at least build llama.cpp from that script, so we are consistent but more importantly you'll want the version of llama.cpp to get updated like the other containers for new features, new model support, bug fixes, to make sure this stays compatible with upstream llama.cpp, etc.
> > 
> > 
> > There's a couple of reasons that I didn't use the shell script for this image -
> > 
> > 1. There's a note in the llama.cpp docs that refers to some releases breaking the SYCL backend - https://github.com/ggerganov/llama.cpp/blob/master/docs/backend/SYCL.md#recommended-release
> > 2. I discovered that a change to the chat template support in release `b4524` broke the usage of the granite-code models that I was testing with.  So, I need to open an issue on that.  Every release after `b4523` seems to be broken.
> 
> I still think we should continue updating llama.cpp upstream and the other images.
> 
> We have versioning of the RamaLama container images now, so one could always manually peg back to an old RamaLama container image if they really wanted to. We need to find such bugs and submit issues upstream to llama.cpp
> 
> I won't block merging of the PR on this but I think it would be a better idea.

No, I totally agree with you.  I'll get the upstream issue opened and work on a followup PR that moves the appropriate build logic into the script.

cgruver:
@ericcurtin The build stage for this image is running out of space.  Unfortunately, the Intel OneAPI Toolkit results in a build stage image of around 14GB...  The final image is a little under 3GB.

rhatdan:
I will merge this, and attempt to build an image.  My concern is that we share the image RHEL AI Intel image.

container-images/intel-gpu/Containerfile
```diff
@@ -0,0 +1,32 @@
...
+
+COPY intel-gpu/oneAPI.repo /etc/yum.repos.d/
+
+RUN dnf install -y lspci clinfo intel-opencl g++ cmake git libcurl-devel intel-oneapi-base-toolkit ; \
+    git clone https://github.com/ggerganov/llama.cpp.git -b b4523 ; \
+    cd llama.cpp ; \
+    mkdir -p build ; \
+    cd build ; \
+    source /opt/intel/oneapi/setvars.sh ; \
+    cmake .. -DGGML_SYCL=ON -DCMAKE_C_COMPILER=icx -DCMAKE_CXX_COMPILER=icpx -DLLAMA_CURL=ON -DGGML_CCACHE=OFF -DGGML_NATIVE=ON ; \
```
ericcurtin, line 11: We actually want DGGML_NATIVE=OFF

=ON creates not so portable binaries (and you get different results based on what generation CPU one has.

rhatdan: Do we need another PR?

container-images/intel-gpu/Containerfile
```diff
@@ -0,0 +1,32 @@
...
+
+COPY --from=builder /llama-cpp/bin/ /usr/local/bin/
+COPY --from=builder /llama-cpp/lib/ /usr/local/lib/
+COPY --from=builder /llama-cpp/lib64/ /usr/local/lib64/
+COPY --from=builder /llama-cpp/include/ /usr/local/include/
+COPY intel-gpu/oneAPI.repo /etc/yum.repos.d/
+
+RUN dnf install -y intel-opencl libcurl lspci clinfo intel-oneapi-runtime-compilers intel-oneapi-mkl-core intel-oneapi-mkl-sycl-blas intel-oneapi-runtime-dnnl ; \
```
eero-t, line 25: To improve build time, I would suggest using additional "base" stage that install these libraries, and which is used both by build, and final container stages.

That way packages common to final and build stages would be installed only once.

container-images/intel-gpu/Containerfile
cgruver: Good point...  AFAIK, intermediate images are not automatically cleaned up.  At least not with local workstation builds.  Not sure how GitHub actions handles it though...

eero-t: AFAIK that's not an image, just a named layer, and whether it's cleaned afterwards, depends on whether image build has (layer) caching enabled or not...

eero-t: If one would want to avoid _both_ caching OneAPI build layer due to its disk usage, and re-downloads of layer content (when it's not cached), I would suggest:
* Using OneAPI base images instead of their package versions: https://hub.docker.com/r/intel/oneapi
* Doing `docker pull` to make sure latest version of the image(s) are present, and
* Running `docker build` with caching disabled.

eero-t: Another option for reducing layer size would be installing just suitable subset of `intel-oneapi-base-toolkit` metapackage dependencies directly.  Finding out which of them are not needed will need some iteration though...

(To test that, I would recommend running container with full `intel-oneapi-base-toolkit`, where subset of its packages is  uninstalled before build.  That way one avoid large deps being re-downloaded.)

PR #657, opened 2025-01-29
Update intel-gpu Containerfile to reduce the size of the builder image

cgruver:
Modified the `as builder` stage of the Containerfile to minimize the build dependency packages installed for building llama.cpp.

Reduced the build stage image from 14 to 7GB

## Summary by Sourcery

Build:
- Reduce the size of the builder image by removing unnecessary packages.

### DIFF
--- a/container-images/intel-gpu/Containerfile
+++ b/container-images/intel-gpu/Containerfile
@@ -2,13 +2,13 @@ FROM quay.io/fedora/fedora:41 as builder
 
 COPY intel-gpu/oneAPI.repo /etc/yum.repos.d/
 
-RUN dnf install -y lspci clinfo intel-opencl g++ cmake git libcurl-devel intel-oneapi-base-toolkit ; \
+RUN dnf install -y intel-opencl g++ cmake git tar libcurl-devel intel-oneapi-mkl-sycl-devel intel-oneapi-dnnl-devel intel-oneapi-compiler-dpcpp-cpp ; \
     git clone https://github.com/ggerganov/llama.cpp.git -b b4523 ; \
     cd llama.cpp ; \
     mkdir -p build ; \
     cd build ; \
     source /opt/intel/oneapi/setvars.sh ; \
-    cmake .. -DGGML_SYCL=ON -DCMAKE_C_COMPILER=icx -DCMAKE_CXX_COMPILER=icpx -DLLAMA_CURL=ON -DGGML_CCACHE=OFF -DGGML_NATIVE=ON ; \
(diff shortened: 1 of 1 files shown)


container-images/intel-gpu/Containerfile
```diff
@@ -2,13 +2,13 @@ FROM quay.io/fedora/fedora:41 as builder
...
 
-RUN dnf install -y lspci clinfo intel-opencl g++ cmake git libcurl-devel intel-oneapi-base-toolkit ; \
+RUN dnf install -y intel-opencl g++ cmake git tar libcurl-devel intel-oneapi-mkl-sycl-devel intel-oneapi-dnnl-devel intel-oneapi-compiler-dpcpp-cpp ; \
     git clone https://github.com/ggerganov/llama.cpp.git -b b4523 ; \
     cd llama.cpp ; \
     mkdir -p build ; \
     cd build ; \
     source /opt/intel/oneapi/setvars.sh ; \
-    cmake .. -DGGML_SYCL=ON -DCMAKE_C_COMPILER=icx -DCMAKE_CXX_COMPILER=icpx -DLLAMA_CURL=ON -DGGML_CCACHE=OFF -DGGML_NATIVE=ON ; \
+    cmake .. -DGGML_SYCL=ON -DCMAKE_C_COMPILER=icx -DCMAKE_CXX_COMPILER=icpx -DLLAMA_CURL=ON -DGGML_CCACHE=OFF -DGGML_NATIVE=OFF ; \
```
dougsland, line 11: I believe the change is nice to have as DGGML_NATIVE=off generates more generic arch binary. However, I would keep intel-oneapi-base-toolkit to avoid breakages in case intel decide to add more sub-packages that in the future might break us. 

Additionally, I would make DGGML_NATIVE (it seems possible) as ARG in the container file so it's easy for users or devs to understand and change a constant.

```
# Set a build argument for GGML_NATIVE with a default value of OFF
# - ON  = Enables CPU-specific optimizations (faster but machine-dependent)
# - OFF = Disables optimizations (slightly slower but portable across CPUs)
ARG GGML_NATIVE=OFF

<SNIP>

    cmake .. -DGGML_SYCL=ON -DCMAKE_C_COMPILER=icx -DCMAKE_CXX_COMPILER=icpx \
             -DLLAMA_CURL=ON -DGGML_CCACHE=OFF -DGGML_NATIVE=${GGML_NATIVE}
```

dougsland:
See-Also: https://github.com/containers/ramalama/pull/657

ericcurtin (approved):

container-images/intel-gpu/Containerfile
cgruver: I had to remove the whole of `intel-oneapi-base-toolkit` because it generates a 14GB build image.  It was blowing up the GitHub actions.  We'll just have to keep track of build dependencies as Intel releases new generations of chips.

I'm already waiting with $$$ in hand for ASUS to release the 15th GEN NUCs...  ;-)

Issue #655, opened 2025-01-29
podman for nvidia may need `--group-add keep-groups`


khumarahn:
Hi! Ramalama is a very nice project.

I have some trouble making it use an nvidia gpu: on my system `/dev/dri/card*` are protected by the group `video`, and this privilege is given up by podman.
```
$ ls -alh /dev/dri
total 0
drwxr-xr-x   3 root root        140 Jan 28 11:15 .
drwxr-xr-x  21 root root       4.5K Jan 28 13:16 ..
drwxr-xr-x   2 root root        120 Jan 28 11:15 by-path
crw-rw----+  1 root video  226,   0 Jan 28 11:15 card0
crw-rw----+  1 root video  226,   1 Jan 28 11:15 card1
crw-rw-rw-   1 root render 226, 128 Jan 28 11:15 renderD128
crw-rw-rw-   1 root render 226, 129 Jan 28 11:15 renderD129
```

This is similar to https://github.com/containers/ramalama/pull/376 and related to https://github.com/containers/podman/issues/10166

ericcurtin:
Lets just add it like @bmahabirbu 's #376 since this is the second reported case.

Please open a PR and ensure it's called in all cases kublet, quadlet, run, serve, etc.

khumarahn:
```

**Comment by khumarahn** [comment]
I think, there needs to be a way to pass `--gpu-layers` to llama.cpp too. Ollama seems to compute that automatically somehow

**Comment by ericcurtin** [comment]
--gpu-layers is ngl in llama.cpp cli

**Comment by khumarahn** [comment]
Ah, right! This also works:
```
diff --git a/ramalama/model.py b/ramalama/model.py
index 710a27a..ad87b85 100644
--- a/ramalama/model.py
+++ b/ramalama/model.py
@@ -185,6 +185,7 @@ class Model:
             # Special case for Cuda
             if k == "CUDA_VISIBLE_DEVICES":
                 conman_args += ["--device", "nvidia.com/gpu=all"]
+                conman_args += ["--group-add", "keep-groups"]
             conman_args += ["-e", f"{k}={v}"]
         return conman_args
 
@@ -206,7 +207,7 @@ class Model:
             else:
                 gpu_args += ["-ngl"]  # single dash
 
-            gpu_args += ["999"]
+            gpu_args += ["16"]
 
         return gpu_args
 
```

ericcurtin:
I think we might need:

"--gpu-layers", "16"

[...]

PR #659, opened 2025-01-29
add --ngl to specify the number of gpu layers, and --keep-groups so podman has access to gpu

khumarahn:
Hi. This follows an invitation to submit a PR in https://github.com/containers/ramalama/issues/655

I do not feel qualified to submit PRs to ramalama, but I tried to keep it as simple as possible, and to not change any defaults.

With these changes the following works on my system (gentoo linux, an ancient 1080ti with just 11gb of GPU memory):
```
ramalama --keep-groups --ngl 34 serve --name deepseek deepseek-r1:32b
```

I didn't touch the documentation yet, hope to get feedback first.

## Summary by Sourcery

Add two new command-line options: `--ngl` to specify the number of GPU layers to use and `--keep-groups` to give podman access to the GPU.

New Features:
- Added `--ngl` option to control the number of GPU layers used by the model.
- Added `--keep-groups` option for podman to enable GPU access.

## Summary by Sourcery

Add support for specifying the number of GPU layers to offload using the --ngl option, and allow keeping user groups when using podman with the --keep-groups flag.

New Features:
- Added `--ngl` option to specify the number of GPU layers to use.
- Added `--keep-groups` option to pass `--group-add keep-groups` to podman when launching the container. This is useful for enabling GPU access in some systems, but has security implications.

### DIFF
--- a/docs/ramalama.1.md
+++ b/docs/ramalama.1.md
@@ -109,6 +109,13 @@ The default can be overridden in the ramalama.conf file or via the the
 RAMALAMA_IMAGE environment variable. `export RAMALAMA_TRANSPORT=quay.io/ramalama/aiimage:latest` tells
 RamaLama to use the `quay.io/ramalama/aiimage:latest` image.
 
+#### **--keep-groups**
+pass --group-add keep-groups to podman (default: False)
+Needed to access the gpu on some systems, but has an impact on security, use with caution.
+
+#### **--ngl**
+number of gpu layers (default: 999)
+
 #### **--nocontainer**
 do not run RamaLama in the default container (default: False)
 The default can be overridden in the ramalama.conf file.
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -190,6 +190,21 @@ The RAMALAMA_CONTAINER_ENGINE environment variable modifies default behaviour.""
         help="offload the workload to the GPU",
     )
     parser.add_argument(
+        "--ngl",
+        dest="ngl",
+        type=int,
+        default=config.get("ngl", 999),
+        help="Number of layers to offload to the gpu, if available"
+    )
+    parser.add_argument(
+        "--keep-groups",
+        dest="podman_keep_groups",
+        default=config.get("keep_groups", False),
+        action="store_true",
+        help="""pass `--group-add keep-groups` to podman, if using podman.
+Needed to access gpu on some systems, but has security implications.""",
+    )
+    parser.add_argument(
         "--image",
         default=config.get("image"),
         help="OCI container image to run with the specified AI model",
--- a/ramalama/model.py
+++ b/ramalama/model.py
@@ -160,6 +160,8 @@ class Model:
 
         if os.path.basename(args.engine) == "podman":
             conman_args += ["--pull=newer"]
+            if args.podman_keep_groups:
+                conman_args += ["--group-add", "keep-groups"]
         elif os.path.basename(args.engine) == "docker":
             try:
                 run_cmd([args.engine, "pull", "-q", args.image], ignore_all=True)
@@ -188,10 +190,10 @@ class Model:
             conman_args += ["-e", f"{k}={v}"]
         return conman_args
 
-    def gpu_args(self, force=False, runner=False):
+    def gpu_args(self, args, runner=False):
         gpu_args = []
         if (
-            force
+            args.gpu
             or os.getenv("HIP_VISIBLE_DEVICES")
             or os.getenv("ASAHI_VISIBLE_DEVICES")
             or os.getenv("CUDA_VISIBLE_DEVICES")
@@ -206,7 +208,7 @@ class Model:
             else:
                 gpu_args += ["-ngl"]  # single dash
 
-            gpu_args += ["999"]
+            gpu_args += [ f'{args.ngl}' ]
 
         return gpu_args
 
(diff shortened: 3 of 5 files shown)


ramalama/model.py
```diff
@@ -188,10 +190,10 @@ def setup_container(self, args):
             conman_args += ["-e", f"{k}={v}"]
         return conman_args
 
-    def gpu_args(self, force=False, runner=False):
+    def gpu_args(self, args, runner=False):
```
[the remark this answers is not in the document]
khumarahn: fixed

khumarahn:
@sourcery-ai review

ericcurtin (approved):

khumarahn:
I added the docs, I think this is ready for a review

@sourcery-ai review

ericcurtin:
@khumarahn changes look fine, CI failed around documentation man pages. You basically need to keep running this:

hack/xref-helpmsgs-manpages

and change things until it agrees you've done it right.

docs/ramalama.1.md
```diff
@@ -109,6 +109,13 @@ The default can be overridden in the ramalama.conf file or via the the
 RAMALAMA_IMAGE environment variable. `export RAMALAMA_TRANSPORT=quay.io/ramalama/aiimage:latest` tells
 RamaLama to use the `quay.io/ramalama/aiimage:latest` image.
 
+#### **--keep-groups**
```
rhatdan, line 112: I would prefer to have RamaLama figure this out. or always set it when Podman is used in rootless mode.
Users are not going to understand when and when not to use this.

rhatdan: The downside of using it all of the time is a leak of GIDs into the container when not necessary.  Add option to allow users to disable it for security purposes,  But I think we should enable it by default for rootless containers so that users will not stumble upon it.

ramalama/cli.py
```diff
@@ -189,6 +189,15 @@ def configure_arguments(parser):
         action="store_true",
         help="offload the workload to the GPU",
     )
+    parser.add_argument("--ngl", dest="ngl", type=int, default=999, help="Number of layers to offload to the gpu, if available")
+    parser.add_argument(
+        "--keep-groups",
+        dest="podman_keep_groups",
+        default=False,
```
rhatdan, line 196: Add function to detect, also option should be configurable in ramalama.conf file.

docs/ramalama.1.md
```diff
@@ -109,6 +109,13 @@ The default can be overridden in the ramalama.conf file or via the the
...
+pass --group-add keep-groups to podman (default: False)
+Needed to access the gpu on some systems, but has an impact on security, use with caution.
+
+#### **--ngl**
+number of gpu layers (default: 999)
```
rhatdan, line 117: Should be configurable in ramalama.conf file.

ramalama/cli.py
khumarahn: What is the right way to detect a rootless podman? I found how to detect podman by `os.path.basename(args.engine) == "podman"`, but I'm not sure about rootless

rhatdan:
There is something wrong with ubi9 images.

rhatdan:
Please squash your commits.

khumarahn:
@rhatdan I added the options to ramalama.conf.

I don't know how to detect if podman is running rootless to turn on keep-groups. Also, I'm not sure this is needed: it goes against the least privilege principle. And turning on this option would be the easiest thing to do for people running gpu in containers (e.g. with nvidia one has to first set up their container toolkit, and keep it updated with the drivers update)

khumarahn:
> Please squash your commits.

done

khumarahn:
@sourcery-ai review

rhatdan:
LGTM

ramalama/model.py
```diff
@@ -188,10 +190,10 @@ def setup_container(self, args):
...
         gpu_args = []
         if (
-            force
+            args.gpu
```
ericcurtin, line 196: I don't think this PR is correct, on systems with integrated graphics we don't want to use GPU acceleration, it's typically much slower than CPU inferencing, practically unusable.

With the default being a static 999 this will always be the case now, which is not good. We should look at this again before we spin up a new release.

ericcurtin: We can have this option but the default being 999 is not good

khumarahn: I don't think I changed any existing logic here, I just added two options that allow ramalama to work on my system. The default of 999 layers is of course questionable

ericcurtin: It's the lack of VRAM on integrated graphics also, most of the time one will crash on one of these systems because of running out of VRAM

khumarahn: Sure, but somehow ramalama works on my laptop with no gpu without any configuration. This PR did not change the default behaviour of ramalama, unless I'm missing something

docs/ramalama.1.md
```diff
@@ -109,6 +109,13 @@ The default can be overridden in the ramalama.conf file or via the the
 RAMALAMA_IMAGE environment variable. `export RAMALAMA_TRANSPORT=quay.io/ramalama/aiimage:latest` tells
 RamaLama to use the `quay.io/ramalama/aiimage:latest` image.
 
+#### **--keep-groups**
+pass --group-add keep-groups to podman (default: False)
+Needed to access the gpu on some systems, but has an impact on security, use with caution.
```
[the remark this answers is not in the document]
ericcurtin: I think these are great suggestions for documentation at least, dunno if you saw this bot comment @rhatdan @khumarahn

ramalama/model.py
ericcurtin: If this is try, we should try and figure out why 999 was or wasn't included on your system

PR #673, opened 2025-01-30
Add generated man pages for section 7 into gitignore

jistr:
Sections 1 and 5 are already ignored, but there is also `ramalama-cuda.7.md` file now which gets rendered into `ramalama-cuda.7`, so the latter should be in gitignore.

## Summary by Sourcery

Chores:
- Ignore generated man pages for section 7.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 ramalama/__pycache__
 docs/*.1
 docs/*.5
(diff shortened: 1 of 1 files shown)


dougsland:
LGTM

rhatdan:
LGTM

Issue #665, opened 2025-01-30
Feature request: Add option to disable network access


joshua-stone:
There doesn't appear to be a way to pass `--network=none` for limiting network access:

https://github.com/containers/ramalama/blob/4851d600d7e3d09d21bb73bb07270ef7c3a06c89/ramalama/model.py#L140

I believe that having a straightforward way to confine locally-running AI models would be useful for guaranteeing privacy and security in general.

rhatdan:
I think it makes sense to default to --network=none, to help alleviate fears from people using random AI Models (deepseek for example). 

Care to open a PR?

rhjostone:
@rhatdan Should I make this a global option for subcommands like `serve`, or should it be made exclusive for `run` for now since `serve` implicitly requires network access anyway?

https://github.com/containers/ramalama/pull/674

rhatdan:
Anything that runs a container should have this.

run, serve, convert, and push off the top of my head.

rhatdan:
--network=mode is merged.

PR #674, opened 2025-01-30
Add --network-mode option

rhjostone:
Addresses https://github.com/containers/ramalama/issues/665

## Summary by Sourcery

New Features:
- Allow users to specify the network mode for containers.

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -379,6 +379,12 @@ def bench_cli(args):
 
 def bench_parser(subparsers):
     parser = subparsers.add_parser("bench", aliases=["benchmark"], help="benchmark specified AI Model")
+    parser.add_argument(
+        "--network-mode",
+        type=str,
+        default="none",
+        help="set the network mode for the container",
+    )
     parser.add_argument("MODEL")  # positional argument
     parser.set_defaults(func=bench_cli)
 
@@ -600,6 +606,13 @@ type of OCI Model Image to push.
 Model "car" includes base image with the model stored in a /models subdir.
 Model "raw" contains the model and a link file model.file to it stored at /.""",
     )
+    # https://docs.podman.io/en/latest/markdown/podman-build.1.html#network-mode-net
+    parser.add_argument(
+        "--network-mode",
+        type=str,
+        default="none",
+        help="sets the configuration for network namespaces when handling RUN instructions",
+    )
     parser.add_argument("SOURCE")  # positional argument
     parser.add_argument("TARGET")  # positional argument
     parser.set_defaults(func=convert_cli)
@@ -717,6 +730,15 @@ def _run(parser):
 def run_parser(subparsers):
     parser = subparsers.add_parser("run", help="run specified AI Model as a chatbot")
     _run(parser)
+    # Disable network access by default, and give the option to pass any supported network mode into
+    # podman if needed:
+    # https://docs.podman.io/en/latest/markdown/podman-run.1.html#network-mode-net
+    parser.add_argument(
+        "--network-mode",
+        type=str,
+        default="none",
+        help="set the network mode for the container",
+    )
     parser.add_argument("MODEL")  # positional argument
     parser.add_argument(
         "ARGS", nargs="*", help="Overrides the default prompt, and the output is returned without entering the chatbot"
@@ -741,6 +763,17 @@ def serve_parser(subparsers):
     )
     parser.add_argument(
         "-p", "--port", default=config.get('port', "8080"), help="port for AI Model server to listen on"
+    )
+    # --network-mode=default lets the container listen on localhost, and is an option that's compatible
+    # with podman and docker. It should use the bridge driver for rootful podman, the pasta driver for
+    # rootless podman, and the bridge driver for docker:
+    # https://docs.podman.io/en/latest/markdown/podman-run.1.html#network-mode-net
+    # https://docs.docker.com/engine/network/#drivers
+    parser.add_argument(
+        "--network-mode",
+        type=str,
+        default="default",
+        help="set the network mode for the container",
     )
     parser.add_argument("MODEL")  # positional argument
     parser.set_defaults(func=serve_cli)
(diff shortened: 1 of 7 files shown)


rhatdan:
This option should be available for run and serve.  Basically for any container operation.

ramalama/cli.py
```diff
@@ -721,6 +721,15 @@ def run_parser(subparsers):
...
         "ARGS", nargs="*", help="Overrides the default prompt, and the output is returned without entering the chatbot"
     )
+    # Disable network access by default, and give the option to pass any supported network mode into
+    # podman if needed:
+    # https://docs.podman.io/en/latest/markdown/podman-run.1.html#network-mode-net
+    parser.add_argument(
+        "--network-mode",
+        type=str,
+        default="none",
+        help="Set the network mode for the container.",
```
rhatdan, line 731: This needs to be documented in man pages.  I do not know why the man page checker allowed this through.

dougsland:
I am assuming this is related to --network=host and --network=private?

ramalama/cli.py
rhjostone: I believe the new options should be documented now.

rhjostone:
> I am assuming this is related to --network=host and --network=private?

Yeah, this should let you set any network driver that's supported in podman/docker:

https://docs.podman.io/en/latest/markdown/podman-run.1.html#network-mode-net
https://docs.docker.com/engine/network/#drivers

ramalama/cli.py
```diff
@@ -721,6 +721,15 @@ def run_parser(subparsers):
     parser.add_argument(
         "ARGS", nargs="*", help="Overrides the default prompt, and the output is returned without entering the chatbot"
     )
+    # Disable network access by default, and give the option to pass any supported network mode into
+    # podman if needed:
+    # https://docs.podman.io/en/latest/markdown/podman-run.1.html#network-mode-net
+    parser.add_argument(
+        "--network-mode",
+        type=str,
+        default="none",
```
dougsland, line 739: @rhatdan should we use private by default?

ramalama/cli.py
```diff
@@ -751,6 +764,12 @@ def serve_parser(subparsers):
     parser.add_argument(
         "-p", "--port", default=config.get('port', "8080"), help="port for AI Model server to listen on"
     )
+    parser.add_argument(
+        "--network-mode",
+        type=str,
+        default="bridge",
```
dougsland, line 774: @rhatdan @rhjostone I am wondering if this should be private instead of bridge. The private creates the bridge.

ramalama/cli.py
rhatdan: No none is a better default. I don't see a container needing access to the network for `run`, might need `private` for serve.

ramalama/cli.py
rhatdan: Let's make this `private`, since I believe `bridge` will work differently in `rootless` containers. `private` allows the container engine to make the choice.  If the network is set to private, on Docker then we would need to to force bridge, since `private` is not understood on Docker.

@Luap99 Correct me if I am wrong?

rhatdan: Looks like bridge works for rootless podman ok, so will merge.

rhatdan:
Thanks @rhjostone this is very good.

Please squash your commits.

git rebase -i origin
git push --force

ramalama/cli.py
rhjostone: `private` would be my preferred choice since it works for both rootful and rooltess podman:

```
private: Create a new namespace for the container. This uses the bridge mode for rootful containers and slirp4netns for rootless ones.
```

Docker however doesn't appear to have a `private` mode:

https://docs.docker.com/engine/network/#drivers

Since ramalama supports both podman and docker, I wanted to use whatever option would work between all engines. Users can at least still explicitly set `private` for more confined networking.

rhjostone:
> Thanks @rhjostone this is very good.
> 
> Please squash your commits.
> 
> git rebase -i origin git push --force

Done!

rhatdan:
LGTM

dougsland:
alright, lets do it. LGTM

rhjostone:
Thanks @dougsland @rhatdan!

ramalama/cli.py
Luap99: You should be able to request `default` instead, I think that works on docker too. I don't think bridge is a great default for rootless podman. Using the direct pasta driver has more advantages, keeps source ip and is a less complex network setup. The only advantage of bridge is that other containers can connect directly to it without going through host ports.

rhjostone: That's a good idea. I've changed the default value to `default`.

rhatdan:
Test wants the options to be alphabetically sorted in the man pages.

rhjostone:
@rhatdan Does the ordering in the man pages look correct now?

dougsland:
CI/CD not related to this change. Personally, I would love to have a documentation explaining the options and showing what happens if you change from one option to another. However, can be in a different patch.

dougsland:
As he fixed the last milestone as @rhatdan  requested and the patch looks good, moving forward. Thanks @rhjostone

ericcurtin:
This PR should have not been merged @dougsland it didn't pass a single bats build

ericcurtin:
Opening a revert PR.

@rhjostone after the revert is complete, could you push this PR again? We need to then fix the tests so we can merge it when it's green.

A lot of people are opening PRs right now and they are gonna fail unless we revert.

dougsland:
> Opening a revert PR.
> 
> @rhjostone after the revert is complete, could you push this PR again? We need to then fix the tests so we can merge it when it's green.
> 
> A lot of people are opening PRs right now and they are gonna fail unless we revert.

Could you please share the error @ericcurtin ? To me the CI/CD is already broken for ages.

rhatdan:
This PR is reverted.

rhjostone:
@ericcurtin It looks like the PR was reverted, then re-merged thanks to @rhatdan:

https://github.com/containers/ramalama/commit/760b6fd34176b87e0983fdca0a9e5f505b66feaa

Do I still have to re-push the PR?

rhjostone:
Good to know, thanks!

Issue #675, opened 2025-01-30
Making the ollama/huggingface model to OCI process more seamless


kush-gupt:
There's probably opportunity for improvement when pushing and running OCI images with models on ramalama 0.5.3. 

I'm running this on a Mac M3 and installed ramalama with the install.sh.

I would have assumed ramalama push would check if the source model exists locally, else download from the specified protocol. Instead,I get the following:
```
$ ramalama --debug push huggingface://lmstudio-community/granite-3.1-2b-instruct-GGUF/granite-3.1-2b-instruct-Q4_K_M.gguf oci://quay.io/kugupta/granite-3.1-2b:latest
Pushing quay.io/kugupta/granite-3.1-2b:latest...
Converting /Users/kugupta/.local/share/ramalama/models/huggingface/lmstudio-community/granite-3.1-2b-instruct-GGUF/granite-3.1-2b-instruct-Q4_K_M.gguf to quay.io/kugupta/granite-3.1-2b:latest...
run_cmd:  podman manifest rm quay.io/kugupta/granite-3.1-2b:latest
Working directory: None
Ignore stderr: True
Ignore all: False
Building quay.io/kugupta/granite-3.1-2b:latest...
run_cmd:  podman build --no-cache -q -f /var/folders/zd/vny5_y1j19q40tb71dz4ftd40000gn/T/RamaLama_Containerfile_b6xg2zwr /Users/kugupta/.local/share/ramalama/models/huggingface/lmstudio-community/granite-3.1-2b-instruct-GGUF
Working directory: None
Ignore stderr: False
Ignore all: False
Error: context must be a directory: "/Users/kugupta/.local/share/ramalama/models/huggingface/lmstudio-community/granite-3.1-2b-instruct-GGUF"
Error: Command '['podman', 'build', '--no-cache', '-q', '-f', '/var/folders/zd/vny5_y1j19q40tb71dz4ftd40000gn/T/RamaLama_Containerfile_b6xg2zwr', '/Users/kugupta/.local/share/ramalama/models/huggingface/lmstudio-community/granite-3.1-2b-instruct-GGUF']' returned non-zero exit status 125.
```

```
$ ramalama --debug push ollama://granite3.1-dense:2b oci://quay.io/kugupta/granite-3.1-2b:latest 
Error: granite3.1-dense:2b does not exist
```

The clear workaround is to pull and then convert, but not sure if you want to have a prompt that says `model doesn't exist, would you like to pull?` or something similar to knock this out in one command:
[...]

PR #676, opened 2025-01-30
remove ro as an option when mounting images

kush-gupt:
From #675 , ro is an invalid mount option for `type=image`.

From the podman [--mount option docs](https://docs.podman.io/en/v5.1.1/markdown/podman-run.1.html#mount-type-type-type-specific-option):
```
Options specific to type=image:

- rw, readwrite: true or false (default if unspecified: false).

- subpath: Mount only a specific path within the image, instead of the whole image.
```

## Summary by Sourcery

Bug Fixes:
- Fixed an issue where the `ro` mount option was incorrectly used with `type=image`, which is invalid according to the podman documentation.

### DIFF
--- a/ramalama/model.py
+++ b/ramalama/model.py
@@ -220,7 +220,7 @@ class Model:
         if model_path and os.path.exists(model_path):
             conman_args += [f"--mount=type=bind,src={model_path},destination={MNT_FILE},ro"]
         else:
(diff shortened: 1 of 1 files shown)


ericcurtin:
Please sign your commit with:

```
git commit -a --amend -s
```

and repush.

kush-gupt:
Hey Eric, I believe I got the commit signed now!

rhatdan:
LGTM

PR #683, opened 2025-01-31
Revert "Added --jinja to llama-run command"

ericcurtin:
Reverts containers/ramalama#625


ericcurtin:
I'm gonna propose we remove this just for a few weeks @engelmi and @rhatdan and re-introduce again when the newer container images have propagated to peoples machines a little more.

I'm a little worried about giving the FOSDEM talk and things not working.

So it's just a temporary revert, for a few weeks.

ericcurtin:
This also makes me question how well our container image versioning is actually working, I had an older container image on my machine and I was trying to run a newer version of ramalama and it didn't pull the newer image (and hence it didn't work because of lack of --jinja argument)

engelmi (approved):
LGTM

codefromthecrypt:
Hi, I have a public demo in about 10 days and currently I will have to jump around lack of jinja support as it uses tool calls. I'm trying to show the ease/devex of ramalama, so this gets in the way. Any updates or how I can avoid failures like below when running against: `ramalama serve qwen2.5:3b`?

```
  File "/Users/adriancole/oss/genai-instrumentation/examples/python/pydantic-ai/.venv/lib/python3.12/site-packages/pydantic_ai/models/openai.py", line 290, in _completions_create
    raise ModelHTTPError(status_code=status_code, model_name=self.model_name, body=e.body) from e
pydantic_ai.exceptions.ModelHTTPError: status_code: 500, model_name: qwen2.5:3b, body: {'code': 500, 'message': 'tools param requires --jinja flag', 'type': 'server_error'}
```

codefromthecrypt:
following up here https://github.com/containers/ramalama/pull/917#issuecomment-2708697504

Issue #684, opened 2025-01-31
ramalama pull progress bar causes high CPU usage in Linux terminal


mkesper:
During a `ramalama pull` command, the progress bar causes high CPU load in the process responsible for the Linux terminal (e.g. tmux or the terminal itself).

Possible solutions:
- Reduce progress update rate
- Add a -q/--quiet option for the `ramalama pull` subcommand.

ericcurtin:
I think this is fixed but if not maybe you could reopen? @mkesper

PR #685, opened 2025-01-31
Unify CLI options (verbosity, version)

mkesper:
Feel free to adjust block size. But 1k blocks are ridiculously low for downloading GiB of LLM data.

Related: #684

## Summary by Sourcery

New Features:
- Added a `--quiet` flag to suppress the progress bar during downloads.

### DIFF
--- a/ramalama/common.py
+++ b/ramalama/common.py
@@ -198,7 +198,7 @@ def download_file(url, dest_path, headers=None, show_progress=True):
     while retries < max_retries:
         try:
             # Initialize HTTP client for the request
-            http_client.init(url=url, headers=headers, output_file=dest_path, progress=show_progress)
+            http_client.init(url=url, headers=headers, output_file=dest_path, show_progress=show_progress)
             return  # Exit function if successful
 
         except urllib.error.HTTPError as e:
--- a/ramalama/http_client.py
+++ b/ramalama/http_client.py
@@ -13,7 +13,7 @@ class HttpClient:
     def __init__(self):
         pass
 
-    def init(self, url, headers, output_file, progress, response_str=None):
+    def init(self, url, headers, output_file, show_progress, response_str=None):
         output_file_partial = None
         if output_file:
             output_file_partial = output_file + ".partial"
@@ -33,7 +33,7 @@ class HttpClient:
 
             self.now_downloaded = 0
             self.start_time = time.time()
-            self.perform_download(out.file, progress)
+            self.perform_download(out.file, show_progress)
 
         if output_file:
             os.rename(output_file_partial, output_file)
@@ -46,7 +46,7 @@ class HttpClient:
         if self.response.status not in (200, 206):
             raise IOError(f"Request failed: {self.response.status}")
 
-    def perform_download(self, file, progress):
+    def perform_download(self, file, show_progress):
         self.total_to_download += self.file_size
         self.now_downloaded = 0
         self.start_time = time.time()
@@ -58,7 +58,7 @@ class HttpClient:
                 return
 
             size = file.write(data)
-            if progress:
+            if show_progress:
                 accumulated_size += size
                 if time.time() - last_update_time >= 0.1:
                     self.update_progress(accumulated_size)
@@ -68,7 +68,7 @@ class HttpClient:
         if accumulated_size > 0:
             self.update_progress(accumulated_size)
 
-        if progress:
+        if show_progress:
             print("\033[K", end="\r")
 
     def human_readable_time(self, seconds):
--- a/ramalama/ollama.py
+++ b/ramalama/ollama.py
@@ -16,7 +16,7 @@ def fetch_manifest_data(registry_head, model_tag, accept):
     return manifest_data
 
 
-def pull_config_blob(repos, accept, registry_head, manifest_data):
+def pull_config_blob(repos, accept, registry_head, manifest_data, show_progress):
     cfg_hash = manifest_data["config"]["digest"]
     config_blob_path = os.path.join(repos, "blobs", cfg_hash)
 
@@ -27,7 +27,7 @@ def pull_config_blob(repos, accept, registry_head, manifest_data):
     download_file(url, config_blob_path, headers=headers, show_progress=False)
 
 
-def pull_blob(repos, layer_digest, accept, registry_head, models, model_name, model_tag, model_path):
+def pull_blob(repos, layer_digest, accept, registry_head, models, model_name, model_tag, model_path, show_progress):
     layer_blob_path = os.path.join(repos, "blobs", layer_digest)
     url = f"{registry_head}/blobs/{layer_digest}"
     headers = {"Accept": accept}
@@ -35,7 +35,7 @@ def pull_blob(repos, layer_digest, accept, registry_head, models, model_name, mo
     if local_blob is not None:
         run_cmd(["ln", "-sf", local_blob, layer_blob_path])
     else:
-        download_file(url, layer_blob_path, headers=headers, show_progress=True)
+        download_file(url, layer_blob_path, headers=headers, show_progress=show_progress)
         # Verify checksum after downloading the blob
         if not verify_checksum(layer_blob_path):
             print(f"Checksum mismatch for blob {layer_blob_path}, retrying download...")
@@ -49,15 +49,15 @@ def pull_blob(repos, layer_digest, accept, registry_head, models, model_name, mo
     run_cmd(["ln", "-sf", relative_target_path, model_path])
 
 
-def init_pull(repos, accept, registry_head, model_name, model_tag, models, model_path, model):
+def init_pull(repos, accept, registry_head, model_name, model_tag, models, model_path, model, show_progress):
     manifest_data = fetch_manifest_data(registry_head, model_tag, accept)
-    pull_config_blob(repos, accept, registry_head, manifest_data)
+    pull_config_blob(repos, accept, registry_head, manifest_data, show_progress)
     for layer in manifest_data["layers"]:
         layer_digest = layer["digest"]
         if layer["mediaType"] != "application/vnd.ollama.image.model":
             continue
 
-        pull_blob(repos, layer_digest, accept, registry_head, models, model_name, model_tag, model_path)
+        pull_blob(repos, layer_digest, accept, registry_head, models, model_name, model_tag, model_path, show_progress)
 
     return model_path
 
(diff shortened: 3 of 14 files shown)


ericcurtin (approved):

ericcurtin:
Change looks fine to me @mkesper you just need to sign your commits with something like:

```
git commit -a -s --amend
```

to satisfy the DCO bot.

ericcurtin:
Lots of tests failed to, we need to get this build green somehow, at least some of the failures are flakes, not sure how many.

This will all be rebuilt anyway when you sign the commits.

ramalama/common.py
```diff
@@ -177,7 +177,7 @@ def download_file(url, dest_path, headers=None, show_progress=True):
         show_progress = False
 
     try:
-        http_client.init(url=url, headers=headers, output_file=dest_path, progress=show_progress)
+        http_client.init(url=url, headers=headers, output_file=dest_path, show_progress=show_progress)
```
jhjaggars, line 180: This rename needs to be made to HttpClient.init as well.

ericcurtin:
Anybody know what "podman pull" or "podman artifact pull" block size is? @baude @rhatdan 

We could just standardise on that, it's likely a sane value.

I probably care about reliability over performance. I know in practice at the TCP level of the stack a max packet size is 1500 bytes, but that's probably not so important at the http level.

ericcurtin:
It could be the progress bar causing the CPU usage issue also, maybe we could change it so that it doesn't update every single block iteration.

ericcurtin:
Does quiet alone solve the CPU usage @mkesper ?

jhjaggars:
@ericcurtin Best I can tell, the image copier used in podman updates once per second by default:
https://github.com/containers/podman/blob/main/vendor/github.com/containers/common/libimage/copier.go#L315

ericcurtin:
If you guys could check if this makes a difference I'd appreciate it, not much CPU usage issues on my machine:

https://github.com/containers/ramalama/pull/717

ramalama/ollama.py
```diff
@@ -26,11 +26,11 @@ def pull_config_blob(repos, accept, registry_head, manifest_data):
     download_file(url, config_blob_path, headers=headers, show_progress=False)
```
swarajpande5, line 26: https://github.com/containers/ramalama/blob/94bd95ee88a9e9690d7056438ab0b1544923be5f/ramalama/ollama.py#L18-L26
Please add the `show_progress` parameter in this function and remove the harded coded `False` in the `download_file()` call.

swarajpande5 (changes requested):

ramalama/ollama.py
mkesper: I can do this to make it more consistent although this will change behaviour to before the patch.

mkesper:
To add: Interrupting the download is still perfectly possible with Ctrl-C.

rhatdan:
Thanks @mkesper 
LGTM

ramalama/http_client.py
```diff
@@ -60,7 +60,7 @@ def perform_download(self, file, progress):
         self.now_downloaded = 0
         self.start_time = time.time()
         while True:
-            data = self.response.read(1024)
+            data = self.response.read(100 * 1024)
```
dougsland, line 63: The patch looks good to me. How about we add a constant 100KB=100*1024 and use it here? It's faster reading for all people.

dougsland:
LGTM I add minor comment.

dougsland:
CI/CD is complaning about:

```
xref-helpmsgs-manpages: 'ramalama pull --help' lists '--quiet', which is not in docs/ramalama-pull.1.md
make: *** [Makefile:133: validate] Error 1
```

dougsland:
The another issue in CI/CD don't see related to the patch:
```
# tags: distro-integration
# (from function `bail-now' in file test/system/helpers.podman.bash, line 122,
#  from function `die' in file test/system/helpers.podman.bash, line 848,
#  from function `run_ramalama' in file test/system/helpers.bash, line 187,
#  in test file test/system/055-convert.bats, line 12)
#   `run_ramalama 1 convert bogus foobar' failed
#
# [10:29:52.5[68](https://github.com/containers/ramalama/actions/runs/13183889748/job/36843129951?pr=685#step:5:69)712822] $ /home/runner/work/ramalama/ramalama/bin/ramalama convert
# [10:29:52.667041482] usage: ramalama convert [-h] [--type {car,raw}] [--network-mode NETWORK_MODE]
#                         SOURCE TARGET
# ramalama convert: error: the following arguments are required: SOURCE, TARGET
# [10:29:52.6[70](https://github.com/containers/ramalama/actions/runs/13183889748/job/36843129951?pr=685#step:5:71)168914] [ rc=2 (expected) ]
#
# [10:29:52.681213373] $ /home/runner/work/ramalama/ramalama/bin/ramalama convert tiny
# [10:29:52.777658054] usage: ramalama convert [-h] [--type {car,raw}] [--network-mode NETWORK_MODE]
#                         SOURCE TARGET
# ramalama convert: error: the following arguments are required: TARGET
# [10:29:52.7809198[75](https://github.com/containers/ramalama/actions/runs/13183889748/job/36843129951?pr=685#step:5:76)] [ rc=2 (expected) ]
#
# [10:29:52.792616049] $ /home/runner/work/ramalama/ramalama/bin/ramalama convert bogus foobar
# [10:29:52.892675855] usage: ramalama [-h] [--container] [--debug] [--dryrun] [--engine ENGINE]
#                 [--gpu] [--ngl NGL] [--keep-groups] [--image IMAGE]
#                 [--nocontainer] [--runtime {llama.cpp,vllm}] [--store STORE]
#                 [-v]
#                 {help,bench,benchmark,containers,ps,convert,info,list,ls,login,logout,perplexity,pull,push,rm,run,serve,stop,version}
#                 ...
# ramalama: requires a subcommand
# #/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
# #| FAIL: exit code is 0; expected 1
# #\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
# # [teardown]
ok 34 [055] ramalama convert file to image in 3116ms
ok 35 [055] ramalama convert tiny to image in 16685ms
ok 36 [060] ramalama info in 16[83](https://github.com/containers/ramalama/actions/runs/13183889748/job/36843129951?pr=685#step:5:84)ms
make: *** [Makefile:149: bats-nocontainer] Error 1e
```

ramalama/http_client.py
ericcurtin: I'm not sure this is the best fix, I think this works because the read is so big it delays the time that the next iteration of the loop is caused. In practice you cannot depend on TCP packet sizes being larger than 1400 bytes (because of middleboxes, etc.) and this encourages fragmenting the bytes, which isn't spectacular for reliability, performance, etc.

I think it's like the whole printing the progress bar thing is spiking the CPU usage, so if we just avoid updating that I think it could be a better fix to try (just updated this PR): 

https://github.com/containers/ramalama/pull/717

But I don't have CPU spike issues personally so cannot test the above PR.

ericcurtin: But reading python http library, they seem to default to 8kb, that seems more reasonable than 100kb:

https://github.com/python/cpython/blob/main/Lib/http/client.py

ericcurtin: We could try both PRs together

ericcurtin: I'm happy to drop #717 if it doesn't make much difference to CPU usage, although it does seem excessive to update the progress bar on each iteration.

mkesper: > I'm not sure this is the best fix, I think this works because the read is so big it delays the time that the next iteration of the loop is called. In practice you cannot depend on TCP packet sizes being larger than 1400 bytes (because of middleboxes, legacy, etc.) and this encourages fragmenting the bytes, which isn't spectacular for reliability, performance (from the transport perspective), etc.

As far as I understand the Python http client library code, read() does not operate at TCP level but at HTTP stream level, right? That would render this concern moot.

ericcurtin: True, but it's still a consideration as HTTP is on top of TCP

dougsland: IMHO, Splitting in two patches make sense, one for the progress bar and another for buffer size to be discussed. The progress bar skipping should already help as @ericcurtin mentioned.

rhatdan:
Was this a duplicate of the other PR?

ericcurtin:
They are related, I can thing we can take this PR in too if the tests are fixed so it's passing CI. I think 100kb is an excessively high block size though, we'd have to reduce that

ericcurtin:
High block sizes can break some network connections outright...

mkesper:
I have soloed out the --quiet flag and added it to documentation.

ramalama/ollama.py
ericcurtin: I think False is always correct here, it's not a bulky file, the progress means very little, we only show progress for the gguf file

ericcurtin (approved):

swarajpande5 (approved):

ramalama/ollama.py
swarajpande5: sounds good !

ericcurtin:
Need to get green CI builds

rhatdan:
If you run the command that is blowing up with the subcommand missing error with --debug, it will show you the stack dump.

ramalama --debug stop --all

ericcurtin (approved):

mkesper:
It's the convert command that's blowing up as it does not use the pull parser but, naturally needs to download a source. As also other commands use a --quiet switch I'm thinking maybe this should be a global option instead like in many other programs? @ericcurtin @swarajpande5

mkesper:
Syntax would then be `ramalama --quiet <command> [argument1] [argument2]...`

ericcurtin:
--quiet as a global seems fine to me, I'm gonna add you as a collaborator now @mkesper means you won't have to wait for us to click the magic button for CI

rhatdan:
The add subcommand means most likely that --quiet option was not added to one of the parsers.

rhatdan:
Squash and sign your commits.

mkesper:
Seems I'm stuck now, can't see how to fix this (seems to be connected to autocomplete):
```
./hack/xref-helpmsgs-manpages
xref-helpmsgs-manpages: 'ramalama  --help' lists '--quiet,', which is not in docs/ramalama.1.md
xref-helpmsgs-manpages: 'ramalama ': '--quiet' in docs/ramalama.1.md, but not in --help```

rhatdan:
https://github.com/containers/ramalama/pull/835 Should fix the --quiet issue.

rhatdan:
@mkesper Rebase and repush and we should get this in before the release.

ericcurtin:
Can we change --dry-run back to --dryrun please

ericcurtin:
If there is inconsistence "--dry-run" -> "--dryrun"

rhatdan:
We should alias the two, since I can never remember to use --dryrun or --dry-run.

Podman actually uses both spellings. podman auto update uses --dry-run and quadlet uses --dryrun.

ericcurtin:
Yeah I actually copied quadlet at the time

mkesper:
I added a clarifying comment to the second dryrun argument. I honestly thought that it was mistakenly added twice.

ericcurtin (approved):

PR #687, opened 2025-01-31
fix error on macOS for M1 pro

### LINKED ISSUE #686: can't run on Docker on macOS due to "/dev/dri": no such file or directory error
Attempting to run any model on macOS yields the following error:

```
docker: Error response from daemon: error gathering device information while adding custom device "/dev/dri": no such file or directory.
```

CLI command plus output:
```
% ramalama --debug serve smollm:135m
run_cmd:  docker inspect quay.io/ramalama/ramalama:0.5
Working directory: None
Ignore stderr: False
Ignore all: True
exec_cmd:  docker run --rm -i --label RAMALAMA --security-opt=label=disable --name ramalama_yAjVCHPh3U -t -p 8080:8080 --device /dev/dri --mount=type=bind,src=/Users/mmccurdy/.local/share/ramalama/models/ollama/smollm:135m,destination=/mnt/models/model.file,ro quay.io/ramalama/ramalama:latest llama-server --port 8080 -m /mnt/models/model.file -c 2048 --temp 0.8 --host 0.0.0.0
docker: Error response from daemon: error gathering device information while adding custom device "/dev/dri": no such file or directory.
```

Should ramallama be falling back to CPU support in this case?

If I manually alter the docker command to remove the `--device /dev/dri` it works as expected.

volker48:
Fixes #686 

## Summary by Sourcery

Bug Fixes:
- Fix a bug where the container setup on macOS was failing when a `/dev/dri` device file does not exist.

### DIFF
--- a/ramalama/model.py
+++ b/ramalama/model.py
@@ -175,7 +175,7 @@ class Model:
         if hasattr(args, "port"):
             conman_args += ["-p", f"{args.port}:{args.port}"]
 
-        if sys.platform == "darwin" or os.path.exists("/dev/dri"):
+        if (sys.platform == "darwin" and os.path.basename(args.engine) != "docker") or os.path.exists("/dev/dri"):
             conman_args += ["--device", "/dev/dri"]
 
         if os.path.exists("/dev/kfd"):


ramalama/model.py
```diff
@@ -175,7 +175,7 @@ def setup_container(self, args):
         if hasattr(args, "port"):
             conman_args += ["-p", f"{args.port}:{args.port}"]
 
-        if sys.platform == "darwin" or os.path.exists("/dev/dri"):
```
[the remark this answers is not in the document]
ericcurtin: This is gonna break accelerated support with podman machine, what we could do is:

```
(sys.platform == "darwin" and whatever_variable != "docker") or os.path.exists("/dev/dri"):
```

ramalama/model.py
```diff
@@ -175,7 +175,7 @@ def setup_container(self, args):
         if hasattr(args, "port"):
             conman_args += ["-p", f"{args.port}:{args.port}"]
 
-        if sys.platform == "darwin" and os.path.exists("/dev/dri"):
```
dougsland: @volker48 could you please use: 
```
if not sys.platform == "darwin"
```

Why? It make sure nobody in the future adds it back trying to solve something. BTW, how to reproduce this one?

ramalama/model.py
```diff
@@ -175,7 +175,7 @@ def setup_container(self, args):
         if hasattr(args, "port"):
             conman_args += ["-p", f"{args.port}:{args.port}"]
 
-        if sys.platform == "darwin" or os.path.exists("/dev/dri"):
+        if sys.platform == "darwin" and os.path.exists("/dev/dri"):
```
dougsland: I don't think /dev/dri exists in macOS. Do you have it? 

```
 sw_vers 
ProductName:		macOS
ProductVersion:		15.2
BuildVersion:		24C101
$ ls /dev/dri
ls: /dev/dri: No such file or directory
```

ramalama/model.py
volker48: Sure I'll update. To reproduce I just tried to run any image on an M1 Pro Mac using Docker.

ramalama/model.py
volker48: Correct, the path does not exist on Mac with Docker. I was trying to fix the issue by changing from or to and because the or was short circuiting since the platform check is first. I realized that isn't the right fix. I was also not entirely sure why this check for platform was here and combined with the existence check for the path since this path did not exist on Mac I felt like I was missing something.

volker48: Correct, the path does not exist on Mac with Docker. I was trying to fix the issue by changing from or to and because the or was short circuiting since the platform check is first. I realized that isn't the right fix. I was also not entirely sure why this check for platform was here and combined with the existence check for the path since this path did not exist on Mac I felt like I was missing something.

ramalama/model.py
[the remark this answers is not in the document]
volker48: Maybe I'm missing something, but why is the platform check needed? Is it not sufficient to only check for the existence of /dec/dri?

ramalama/model.py
dougsland: @volker48 all good, just update the patch as soon you test the new logic, thanks for reaching out and contributing.

ericcurtin: /dev/dri exists if you use the podman-machine hypervisor with GPU acceleration krunkit.

It's a funny problem because you can't really check if /dev/dri exists on macOS (where this python code runs). But we want GPU acceleration with podman at least.

I believe docker has this GPU acceleration feature also, they used krunkit written by @slp and called it something else.

So another possible fix is to detect if docker on macOS is using krunkit. But the main thing is, we don't want to break accelerated GPU support with podman on macOS here.

volker48: Thanks @ericcurtin I installed Podman and can see the code is working fine there. Let me update my approach so I don't break anything.

volker48:
Tested on both podman and docker and its working now.

rhatdan:
This works with Podman machine running?

dougsland:
LGTM please confirm the info for @rhatdan

ericcurtin (approved):

ericcurtin:
Checking for docker acceleration would be a nice follow on PR for this, so that if it is enabled, that we do actually expose /dev/dri. They call it "Docker VMM" or something like this (but it's actually krunkit made by the containers team at Red Hat). It was BETA in macOS Docker Desktop don't know if it's still beta or not.

volker48:
> This works with Podman machine running?

Yes, I tested with both Podman and Docker. I double checked when using the `RAMALAMA_CONTAINER_ENGINE=podman` env var that the podman command still contained `--device=/dev/dri`.

PR #692, opened 2025-01-31
bump llama.cpp to latest release hash aa6fb13

maxamillion:
llama.cpp dropped a new release today, this updates the build script to point to the latest release

## Summary by Sourcery

Build:
- Update the llama.cpp build script to use the latest release hash.


dougsland:
LGTM

PR #704, opened 2025-02-01
Update install.sh to include "gpu_detector.py"

graystevens:
Added `gpu_detector.py` to the list of Python files to be installed, else the CLI fails to run due to the missing module.

## Summary by Sourcery

Bug Fixes:
- Fix the CLI failing to run due to a missing module.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -129,7 +129,8 @@ setup_ramalama() {
   local python_files=("cli.py" "huggingface.py" "model.py" "ollama.py" \
                       "common.py" "__init__.py" "quadlet.py" "kube.py" \
                       "oci.py" "version.py" "shortnames.py" "toml_parser.py" \
-                      "file.py" "http_client.py" "url.py" "annotations.py")
+                      "file.py" "http_client.py" "url.py" "annotations.py" \
+                      "gpu_detector.py" )
   for i in "${python_files[@]}"; do
     if $local_install; then
       url="ramalama/${i}"


rhatdan:
Thanks @graystevens 
LGTM

Please sign. your commits.
git commit -a --amend -s
git push --force

rhatdan:
We need to figure out a way to test this so these issues do. not arise.

dougsland:
> We need to figure out a way to test this so these issues do. not arise.

Agreed, should be automatic.

ericcurtin:
We added this an automatic check a while ago:

https://github.com/containers/ramalama/blob/900eca6ecbdd7ce7cb596c63f2ddc3ef42d9c635/test/ci.sh#L41

There's a couple of things in CI that don't seem to be running anymore. "make lint" is another one that doesn't seem to be running.

dougsland:
> We added this an automatic check a while ago:
> 
> https://github.com/containers/ramalama/blob/900eca6ecbdd7ce7cb596c63f2ddc3ef42d9c635/test/ci.sh#L41
> 
> There's a couple of things in CI that don't seem to be running anymore. "make lint" is another one that doesn't seem to be running.

@ericcurtin regarding to be careful with maintenance. We do what we can, bugs will exist or bad designs. I am working in some right now, relax a bit, lets enjoy as group building the community around Ramalama.

Issue #693, opened 2025-02-01
(Timeout download model): python3.13/ssl.py", line 1138, in read  return self._sslobj.read(len, buffer)


dougsland:
During download of deepseek I got timeout and ramalama just "exploded" with `TimeoutError: The read operation timed out`. We should be more user friendly.

```
ramalama run deepseek-r1:8b          
 23% |████████████████████████████████████                                                                                                                          |    1.08 GB/   4.58 GB   9.32 MB/s    8m 18sTraceback (most recent call last):
  File "/opt/homebrew/bin/ramalama", line 95, in <module>
    main(sys.argv[1:])
    ~~~~^^^^^^^^^^^^^^
  File "/opt/homebrew/bin/ramalama", line 72, in main
    args.func(args)
    ~~~~~~~~~^^^^^^
  File "/Users/douglaslandgraf/ramalama/ramalama/cli.py", line 729, in run_cli
    model.run(args)
    ~~~~~~~~~^^^^^^
  File "/Users/douglaslandgraf/ramalama/ramalama/model.py", line 244, in run
    model_path = self.get_model_path(args)
  File "/Users/douglaslandgraf/ramalama/ramalama/model.py", line 289, in get_model_path
    model_path = self.pull(args)
  File "/Users/douglaslandgraf/ramalama/ramalama/ollama.py", line 108, in pull
    return init_pull(repos, accept, registry_head, model_name, model_tag, models, model_path, self.model)
  File "/Users/douglaslandgraf/ramalama/ramalama/ollama.py", line 56, in init_pull
    pull_blob(repos, layer_digest, accept, registry_head, models, model_name, model_tag, model_path)
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/douglaslandgraf/ramalama/ramalama/ollama.py", line 33, in pull_blob
    download_file(url, layer_blob_path, headers=headers, show_progress=True)
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/douglaslandgraf/ramalama/ramalama/common.py", line 180, in download_file
    http_client.init(url=url, headers=headers, output_file=dest_path, progress=show_progress)
[...]

PR #713, opened 2025-02-03
common: general improvements

dougsland:
- if timeout happens we try 5 times before sending Timeout error to users.

- improve user experience when timeout occurs

- Added console source tree for handling messages

Resolves: https://github.com/containers/ramalama/issues/693

## Summary by Sourcery

Improve the reliability of file downloads and introduce logging with user-friendly messages.

Bug Fixes:
- Retry file downloads up to five times if a timeout or network error occurs.

Enhancements:
- Added logging for network errors, timeouts, and other download-related issues.
- Improved user experience by providing more informative messages during file downloads.

### DIFF
--- a/ramalama/console.py
+++ b/ramalama/console.py
@@ -1,0 +1,39 @@
+import os
+import logging
+
+
+def is_locale_utf8():
+    """Check if the system locale is UTF-8."""
+    return 'UTF-8' in os.getenv('LC_CTYPE', '') or 'UTF-8' in os.getenv('LANG', '')
+
+
+def supports_emoji():
+    """Detect if the terminal supports emoji output."""
+    term = os.getenv("TERM")
+    if not term or term in ("dumb", "linux"):
+        return False
+
+    return is_locale_utf8()
+
+
+# Allow users to override emoji support via an environment variable
+# If RAMALAMA_FORCE_EMOJI is not set, it defaults to checking supports_emoji()
+RAMALAMA_FORCE_EMOJI = os.getenv("RAMALAMA_FORCE_EMOJI")
+FORCE_EMOJI = RAMALAMA_FORCE_EMOJI.lower() == "true" if RAMALAMA_FORCE_EMOJI else None
+EMOJI = FORCE_EMOJI if FORCE_EMOJI is not None else supports_emoji()
+
+
+# Define emoji-aware logging messages
+def error(msg):
+    formatted_msg = f"❌ {msg}" if EMOJI else f"[ERROR] {msg}"
+    logging.error(formatted_msg)
+
+
+def warning(msg):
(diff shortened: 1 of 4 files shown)


dougsland:
I believe the code is ready for review this time. Please let me know if I missed something.

ramalama/console.py
```diff
@@ -0,0 +1,39 @@
+import os
+import logging
+
+
+def is_locale_utf8():
```
ericcurtin, line 5: I tested this on macOS with alacritty it needs one more check to work, because this is the only of the 3 env vars populated on this platform:

```
 $ echo $LC_ALL
 en_IE.UTF-8
 ```
 
 So we need to add one more or:
 
 ```
 return 'UTF-8' in os.getenv('LC_CTYPE', '') or 'UTF-8' in os.getenv('LANG', '') or 'UTF-8' in os.getenv('LC_ALL', '')
 ```

File: ramalama/console.py [hunk 2]
```diff
@@ -0,0 +1,39 @@
...
+        return False
+
+    return is_locale_utf8()
+
+
+# Allow users to override emoji support via an environment variable
+# If RAMALAMA_FORCE_EMOJI is not set, it defaults to checking supports_emoji()
+RAMALAMA_FORCE_EMOJI = os.getenv("RAMALAMA_FORCE_EMOJI")
+FORCE_EMOJI = RAMALAMA_FORCE_EMOJI.lower() == "true" if RAMALAMA_FORCE_EMOJI else None
+EMOJI = FORCE_EMOJI if FORCE_EMOJI is not None else supports_emoji()
```
ericcurtin, line 23: These lines are long-winded it took me a while to read this, lets simplify:

```
EMOJI = os.getenv("RAMALAMA_FORCE_EMOJI") or supports_emoji()
```

ericcurtin:
This PR is much easier to review, a much more focused PR. I think we are close.

ramalama/console.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+# Define emoji-aware logging messages
+def error(msg):
```
ericcurtin, line 27: The spacing isn't consistent in the printed output in these three functions, some are double-space some are single-space.

If you are pushing new commits you might as well catch, we can also do in a follow on PR.

ramalama/console.py
ericcurtin: Could also do:

```
return any('UTF-8' in os.getenv(var, '') for var in ['LC_CTYPE', 'LANG', 'LC_ALL'])
```

as long as it works I don't mind too much. And it just checks these three env vars for UTF-8, I don't care after that point.

ericcurtin:
I'll catch the changes in another PR I have the time

PR #716, opened 2025-02-03
Fix macOS emoji compatibility with Alacritty

ericcurtin:
The main fix here is to check LC_ALL also. It's required for this
platform combination. The rest is code simplifications and
refactorings. Fix to have consitent spacing (there was a mix of
double and single spacings in output based on log-level).

## Summary by Sourcery

Bug Fixes:
- Fix UTF-8 locale detection on macOS when using Alacritty.

### DIFF
--- a/ramalama/console.py
+++ b/ramalama/console.py
@@ -4,7 +4,7 @@ import logging
 
 def is_locale_utf8():
     """Check if the system locale is UTF-8."""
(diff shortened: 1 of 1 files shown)


ericcurtin:
@dougsland PTAL

maxamillion (commented):
Code LGTM but I don't have a Mac so I can't test.

dougsland:
> @dougsland PTAL

sure, little late (after merge) but all good, important is getting the wheel moving and improvements coming. :)

PR #720, opened 2025-02-03
Refactor image builds to put all of the hardware and platform selection logic into one place.

cgruver:
This is a proposed refactor of the image build process that I believe will make it easier to maintain as we add more hardware and platform types.

I moved all of the selection logic into a `case` statement in the `main` function.

Each image has its own selection in the `case`.

## Summary by Sourcery

Consolidate hardware and platform selection logic for image builds into a single script.

Enhancements:
- Standardize the installation directory to `/usr` for CUDA and Intel GPU builds.
- Update the base image for Asahi builds to `quay.io/fedora/fedora:41`.

Build:
- Introduce a unified `build_llama_and_whisper.sh` script to manage builds across different hardware and platforms (ramalama, ROCm, CUDA, Vulkan, Asahi, and Intel GPU).
- Remove per-platform build scripts.


cgruver:
__Note:__ Please review, but don't merge yet until we are sure I did not break any of the container images.

:-)

container-images/build_llama_and_whisper.sh
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/env bash
+
+# Bash does not easily pass arrays as a single arg to a function.  So, make this a global var in the script.
+CMAKE_FLAGS=""
+LLAMA_CPP_CMAKE_FLAGS=""
```
ericcurtin, line 5: If we need a global variables, could we put local variables in main instead? Trying to keep everything scoped.

ericcurtin: The styling changes aren't a blocker for merge, but why change it?

We don't put "function" in from of the functions in any of the other shell scripts in the project, we don't use camelCase in any other shell scripts either.

Now this is the odd shell script in terms of consistency 😄

container-images/build_llama_and_whisper.sh
```diff
@@ -0,0 +1,113 @@
...
+  local container_image=${1}
+  local install_prefix=/usr
+  local package_list
+  local whisper_cpp_sha=${WHISPER_CPP_SHA:-8a9ad7844d6e2a10cddf4b92de4089d7ac2b14a9}
+  local llama_cpp_sha=${LLAMA_CPP_SHA:-aa6fb1321333fae8853d0cdc26bcb5d438e650a1}
+  local common_rpms=("python3" "python3-pip" "python3-argcomplete" "python3-dnf-plugin-versionlock" "gcc-c++" "cmake" "vim" "procps-ng" "git" "dnf-plugins-core" "libcurl-devel")
+  local vulkan_rpms=("vulkan-headers" "vulkan-loader-devel" "vulkan-tools" "spirv-tools" "glslc" "glslang")
+  local intel_rpms=("intel-oneapi-mkl-sycl-devel" "intel-oneapi-dnnl-devel" "intel-oneapi-compiler-dpcpp-cpp" "intel-level-zero" "oneapi-level-zero" "oneapi-level-zero-devel" "intel-compute-runtime")
+
+  LLAMA_CPP_CMAKE_FLAGS+=("-DGGML_CCACHE=OFF" "-DGGML_NATIVE=OFF" "-DBUILD_SHARED_LIBS=NO" "-DLLAMA_CURL=ON")
```
ericcurtin, line 56: This isn't right, we want BUILD_SHARE_LIBS for one but not the other

container-images/build_llama_and_whisper.sh
```diff
@@ -0,0 +1,113 @@
...
+      source /opt/intel/oneapi/setvars.sh
+    ;;
+  esac
+
+  CMAKE_FLAGS+=("${WHISPER_CPP_CMAKE_FLAGS[@]}")
+  cloneAndBuild https://github.com/ggerganov/whisper.cpp ${whisper_cpp_sha} ${install_prefix}
+  CMAKE_FLAGS+=(${LLAMA_CPP_CMAKE_FLAGS[@]})
+  cloneAndBuild https://github.com/ggerganov/llama.cpp ${llama_cpp_sha} ${install_prefix}
+  dnf -y clean all
+  rm -rf /var/cache/*dnf* /opt/rocm-*/lib/*/library/*gfx9*
```
ericcurtin, line 108: This was where the bulk of the trimming was done for rocm in the past. Not sure why it's ballooned again.

I know this sounds a little hard, but I'd revert this shell script back to it's original state and introduce just enough changes to get "intel-gpu" working with this script. There's no need to rewrite the whole script, it's asking for breakages.

ericcurtin:
Besides the build_llama_and_whisper.sh file, the changes look fine.

container-images/build_llama_and_whisper.sh
cgruver: Understood.  It was a wild hair 🤓

container-images/build_llama_and_whisper.sh
cgruver: Yeah, I don't want to be the oddball...

container-images/build_llama_and_whisper.sh
cgruver: I think -DBUILD_SHARED_LIBS breaks for the intel-gpu build.  I'll double check that.

cgruver: I bet that's why the rocm image got really big...  `-DBUILD_SHARED_LIBS=NO`.  That resulted in some huge executables in /usr/bin...

ericcurtin: Yeah, I think you got it, -DBUILD_SHARED_LIBS=NO end up in the libraries being duplicated in every single binary produced.

The reason we do it for whisper.cpp is it uses some of the same libraries, with the same names, but with different versions sometimes, so one install will replace the library of the other, breaking whichever was installed first.

whisper.cpp is a small project so we statically link that one to fix the issue.

PR #721, opened 2025-02-03
gh actions: add install.sh workflow

dougsland:
## Summary by Sourcery

CI:
- Add a workflow to install RamaLama using a shell script. The workflow runs on pushes and pull requests to the main branch.

### DIFF
--- a/.github/workflows/install_ramalama.yml
+++ b/.github/workflows/install_ramalama.yml
@@ -1,0 +1,50 @@
+name: Install RamaLama
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  install_ramalama:
+    strategy:
+      matrix:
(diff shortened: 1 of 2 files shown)


dougsland:
without removing `-e` github actions just exit without explanion, failing with timeout in linux and error on macos.

.github/workflows/install_ramalama.yml
```diff
@@ -0,0 +1,41 @@
...
+          if ! command -v brew &> /dev/null; then
+            echo "Homebrew is required but not found. Please install it."
+            exit 1
+          fi
+          brew install lshw curl || true
+
+      - name: Run RamaLama Installer
+        run: |
+          chmod +x install.sh
+          ./install.sh
```
ericcurtin, line 41: We probably want to do:

```
./install.sh -l
```

here.


```
./install.sh
```

alone doesn't make a while pile of sense because it's pull from "s" branch, we want to test the current branch rather than the "s" branch.

PR #722, opened 2025-02-03
ramalama container: Make it possible to build basic container on all RHEL architectures

jcajka:
Make it possible to build ramalama container on all RHEL arches.

## Summary by Sourcery

Build:
- Add support for building the ramalama container on multiple architectures by conditionally installing dependencies based on the architecture.

### DIFF
--- a/container-images/scripts/build_llama_and_whisper.sh
+++ b/container-images/scripts/build_llama_and_whisper.sh
@@ -6,6 +6,7 @@ dnf_install() {
                   "procps-ng" "git" "dnf-plugins-core" "libcurl-devel")
   local vulkan_rpms=("vulkan-headers" "vulkan-loader-devel" "vulkan-tools" \
                      "spirv-tools" "glslc" "glslang")
+  local blas_rpms=("openblas-devel")
   local intel_rpms=("intel-oneapi-mkl-sycl-devel" "intel-oneapi-dnnl-devel" \
                   "intel-oneapi-compiler-dpcpp-cpp" "intel-level-zero" \
                   "oneapi-level-zero" "oneapi-level-zero-devel" "intel-compute-runtime")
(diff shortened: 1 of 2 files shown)


container_build.sh
```diff
@@ -70,7 +70,14 @@ build() {
 }
 
 determine_platform() {
-  local platform="linux/amd64"
+  case $conman_bin in
+    podman)
+      local platform="$(podman 2>/dev/null info | grep OsArch | cut -d' ' -f4)"
```
rhatdan, line 75: podman info --format '{{ .Version.OsArch }}'

ericcurtin:
Could you describe what OpenBlas give us that it's only applicable to ppc and s390? Worth putting in the commit message also

container_build.sh
```diff
@@ -70,7 +70,14 @@ build() {
 }
 
 determine_platform() {
-  local platform="linux/amd64"
+  case $conman_bin in
+    podman)
+      local platform="$(podman info --format '{{ .Version.OsArch }}' 2>/dev/null)"
+      ;;
```
ericcurtin, line 76: This is fine, but why not continue to use "uname -m" like all the other CPU architectures, they you don't have to worry about podman/docker differences at all, it's easier to test and maintain.

container-images/scripts/build_llama_and_whisper.sh
```diff
@@ -14,16 +15,21 @@ dnf_install() {
...
-    dnf config-manager --add-repo "$url"
-    url="http://mirror.centos.org/centos/RPM-GPG-KEY-CentOS-Official"
-    curl --retry 8 --retry-all-errors -o \
-      /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-Official "$url"
-    rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-Official
-    dnf install -y mesa-vulkan-drivers "${vulkan_rpms[@]}"
+    # x86_64 and aarch64 means kompute
+    if [ "$(uname -m)" = "x86_64" ] || [ "$(uname -m)" = "aarch64" ] ;then
+      local uname_m
+      uname_m="$(uname -m)"
```
ericcurtin, line 21: Just do:

local uname_m
uname_m="$(uname -m)"
    
before this if block and then use the $uname_m variable after this point everywhere.

container_build.sh
jcajka: Unfortunately container world is using the Go arch strings that are not compatible with uname -m.

rhatdan:
This PR needs a rebase.

jcajka:
> Could you describe what OpenBlas give us that it's only applicable to ppc and s390? Worth putting in the commit message also

Feels kind of nitpickym, arbitrary. Why are you using kompute instead of naive CPU backends?

ericcurtin:
> > Could you describe what OpenBlas give us that it's only applicable to ppc and s390? Worth putting in the commit message also
> 
> Feels kind of nitpickym, arbitrary. Why are you using kompute instead of naive CPU backends?

It's ok, don't let it block you, just curious. Know nothing about OpenBlas personally.

For the generic RamaLama container, the intent is that it is used for CPU-inferencing or use a generic GPU backend. The most generic API around is probably Vulkan, there are two Vulkan backends available for llama.cpp. One enabled via GGML_VULKAN=ON (this is probably more popular), the other enabled via GGML_KOMPUTE=ON

One of the gpu's we use this for is the one exposed via krunkit on podman-machine on macOS. @slp has had great joy using kompute there.

If it felt nitpicky I apologise. I am curious what are the advantages of this backend over the other. It's good that we give people choice over backends.

ericcurtin (approved):

PR #728, opened 2025-02-04
Packit: downstream jobs for EPEL 9,10

lsm5:
This commit enables downstream jobs for EPEL 9 and 10.

Since we're only updating epel branches downstream and not running tests on EPEL targets, we can include the epel targets, we can keep them as part of `&fedora_targets` yaml anchor and not include them separately.

## Summary by Sourcery

CI:
- Enable downstream jobs for EPEL 9 and 10 using the existing `fedora_targets` YAML anchor.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -61,6 +61,8 @@ jobs:
     packages: [ramalama-fedora]
     dist_git_branches: &fedora_targets
       - fedora-all
+      - epel10
+      - epel9
 
   - job: koji_build
     trigger: commit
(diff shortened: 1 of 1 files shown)


ericcurtin (approved):

lsm5:
The PR is good to go. Effects will only be noticeable on downstream update jobs.

Packit jobs are [slow today](https://status.packit.dev/), so not seeing them triggered or successful isn't a problem for this PR AFAICT.

rhatdan:
LGTM

Issue #727, opened 2025-02-04
Detect Ubuntu Asahi during container setup


chrootchad:
https://github.com/containers/ramalama/pull/526 is failing to detect and use the ashi image on Ubuntu Asahi - ref:

> Thanks @ericcurtin I ran 
> ```
> ramalama --gpu --debug --image quay.io/ramalama/asahi:0.5.4 run ollama://deepseek-r1:7b
> ```
> and forced the latest asahi image and it worked (including the gpu support I couldn't get working previously with other models when the quay.io/ramalama/ramalama:latest image was being used)
> 
> Seems to be an ubuntu asahi quirk, but it looks like this:
> https://github.com/containers/ramalama/blob/3032571edf208a300d7f6f2ee09cf7a0dc10152c/ramalama/common.py#L203
> 
> ```python
>     # ASAHI CASE
>     if os.path.exists('/etc/os-release'):
>         with open('/etc/os-release', 'r') as file:
>             if "asahi" in file.read().lower():
>                 # Set Env Var and break
>                 os.environ["ASAHI_VISIBLE_DEVICES"] = "1"
>                 return
> ```
> isn't setting ASAHI_VISIBLE_DEVICES=1 
> because on ubuntu asahi
> 
> ```bash
> cat /etc/os-release
> PRETTY_NAME="Ubuntu 24.04.1 LTS"
> NAME="Ubuntu"
> VERSION_ID="24.04"
> VERSION="24.04.1 LTS (Noble Numbat)"
> VERSION_CODENAME=noble
> ID=ubuntu
> ID_LIKE=debian
> HOME_URL="https://www.ubuntu.com/"
> SUPPORT_URL="https://help.ubuntu.com/"
> BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
> PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
> UBUNTU_CODENAME=noble
> LOGO=ubuntu-logo
> ```
> 
> Doesn't contain "asahi" anywhere in the output, this seems to though
> 
> ```bash
> hostnamectl | grep asahi
>           Kernel: Linux 6.12.0-1004-asahi-arm
> ```
> 
> Thanks again for your help.
>  

 _Originally posted by @chrootchad in [#616](https://github.com/containers/ramalama/issues/616#issuecomment-2629417564)_

ericcurtin:
[...]

PR #730, opened 2025-02-04
Check for apple,arm-platform in /proc

ericcurtin:
In /proc/device-tree/compatible specifically, thanks Asahi Lina!

It's more portable accross Fedora Asahi Remix and Ubuntu Asahi
Remix.

Also added env var to container image.

Co-authored-by: Asahi Lina <lina@asahilina.net>
Co-authored-by: Daniel J Walsh <dwalsh@redhat.com>

### DIFF
--- a/container-images/asahi/Containerfile
+++ b/container-images/asahi/Containerfile
@@ -1,5 +1,6 @@
 FROM fedora:41
 
+ENV ASAHI_VISIBLE_DEVICES 1
 COPY ../scripts /scripts
 RUN chmod +x /scripts/*.sh && \
     /scripts/build_llama_and_whisper.sh "asahi"
--- a/ramalama/common.py
+++ b/ramalama/common.py
@@ -244,12 +244,16 @@ def get_gpu():
         return
 
     # ASAHI CASE
-    if os.path.exists('/etc/os-release'):
-        with open('/etc/os-release', 'r') as file:
-            if "asahi" in file.read().lower():
-                # Set Env Var and break
-                os.environ["ASAHI_VISIBLE_DEVICES"] = "1"
-                return
+    if os.path.exists('/proc/device-tree/compatible'):
+        try:
+            with open('/proc/device-tree/compatible', 'rb') as f:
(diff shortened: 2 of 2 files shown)


ericcurtin:
Does this work for:

http://github.com/containers/ramalama/issues/727

@chrootchad

rhatdan:
A better way might be to set an environment variable within the image to identify the image type. One issue with doing it based on command line, is that someone could launch service on a different container based on the asahi container.

Lets change the containerfiles to set the GPU type as an environment variable.

ENV ASAHI_VISIBLE_DEVICES 1

During the build, then during the run, there is no issue.

ericcurtin:
> A better way might be to set an environment variable within the image to identify the image type. One issue with doing it based on command line, is that someone could launch service on a different container based on the asahi container.
> 
> Lets change the containerfiles to set the GPU type as an environment variable.
> 
> ENV ASAHI_VISIBLE_DEVICES 1
> 
> During the build, then during the run, there is no issue.

Added

chrootchad:
https://github.com/containers/ramalama/pull/526#issuecomment-2564146148 
The fantastic tooling you have here & specifically the first diagram helped me confirm my theory as to what was going wrong  (having not had any prior knowledge of the codebase [or python])

ericcurtin:
> [#526 (comment)](https://github.com/containers/ramalama/pull/526#issuecomment-2564146148) The fantastic tooling you have here & specifically the first diagram helped me confirm my theory as to what was going wrong (having not had any prior knowledge of the codebase [or python])

Yeah sometimes the diagrams are quite cool, I don't always take them for granted though. They are AI generated at the end of the day.

ericcurtin:
> I don't fully understand what is going on here, but this does look like an improvement over what we had before. I trust the asahi people will test.

if we ignore the env var for a second. It's because we need a way of detecting we are on Asahi to download the Asahi container image. I initially put in a way that was only compatible with Fedora Asahi Remix. The new way is more portable and can detect we need to pull the Asahi container image on Ubuntu Asahi Remix.

PR #734, opened 2025-02-04
Add --network-mode option

rhatdan:
## Summary by Sourcery

New Features:
- Added a `--network-mode` option to the `bench`, `convert`, `push`, `run`, and `serve` commands. This option allows users to specify the network mode for the container, such as `host`, `bridge`, or `none`.

### DIFF
--- a/ramalama/model.py
+++ b/ramalama/model.py
@@ -188,6 +188,8 @@ class Model:
             if k == "CUDA_VISIBLE_DEVICES":
                 conman_args += ["--device", "nvidia.com/gpu=all"]
             conman_args += ["-e", f"{k}={v}"]
+        if args.network_mode != "":
+            conman_args += ["--network", args.network_mode]
         return conman_args
 
     def gpu_args(self, args, runner=False):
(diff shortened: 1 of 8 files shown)


ramalama/model.py
```diff
@@ -188,6 +188,8 @@ def setup_container(self, args):
             if k == "CUDA_VISIBLE_DEVICES":
                 conman_args += ["--device", "nvidia.com/gpu=all"]
             conman_args += ["-e", f"{k}={v}"]
+        if args.network_mode != "":
```
ericcurtin, line 191: Nice and simple fix :)

ericcurtin (approved):

bmahabirbu:
Shoulnt we have --network=bridge as default for llama-serve or else we cant connect to it by default

ericcurtin:
> Shoulnt we have --network=bridge as default for llama-serve or else we cant connect to it by default

Yes, we want to be able to connect to the IP address and port in the server case somehow, but we don't pass any --network by default, so the default behaviour remains unchanged to what was there prior to this PR.

ericcurtin:
Basically there's already a different default in the code here for serve than for any other usage.

ericcurtin:
I do wish we called the CLI option --network rather than reinventing the name for the CLI option as --network-mode

PR #736, opened 2025-02-04
modify container_build.sh to add capability to use podman farm for multi-arch images

cgruver:
Modified container_build.sh -

1. Add `multi-arch` as a command
2. Add `multi-arch-targets.list` to be processed if `$target` == `all`

Usage: 

`./container_build.sh multi-arch all` - Uses `podman farm` to build all of the images in `multi-arch-targets.list`

`./container_build.sh multi-arch vulkan` - Uses`podman farm` to build quay.io/ramalama/vulkan.

## Summary by Sourcery

Add support for building multi-arch images using `podman farm`.

New Features:
- Building multi-architecture container images is now supported through the `multi-arch` command in `container_build.sh` script.

Build:
- Add `multi-arch` command to `container_build.sh` to build multi-arch images using `podman farm`.
- Add support for building all multi-arch images defined in `multi-arch-targets.list`.
- Set default registry path to `quay.io/ramalama`.

### DIFF
--- a/container_build.sh
+++ b/container_build.sh
@@ -93,7 +93,7 @@ parse_arguments() {
         rm_after_build="true"
         shift
         ;;
-      build|push)
+      build|push|multi-arch)
         command="$1"
         shift
         ;;
@@ -116,12 +116,24 @@ process_all_targets() {
   done
 }
 
+process_multi_arch_targets() {
+  for i in $(cat ./multi-arch-targets.list); do
+    build_multi_arch "$i"
+  done
+}
+
+build_multi_arch() {
+  local target="$1"
+  podman farm build -t quay.io/ramalama/"$target" -f container-images/"$target"/Containerfile ./container-images
+}
+
 print_usage() {
   echo "Usage: $(basename "$0") [-h|--help] [-d] <command> [target]"
   echo
   echo "Commands:"
   echo "  build        Build the container images"
   echo "  push         Push the container images"
+  echo "  multi-arch   Build and Push multi-arch images with podman farm"
   echo
   echo "Options:"
   echo "  -d           Some option description"
@@ -149,12 +161,25 @@ main() {
     print_usage
     exit 1
   fi
+  if [ "$command" = "multi-arch" ] && [ "$conman_bin" != "podman" ]; then
+    echo "Error: command 'multi-arch' only works with podman farm"
+    print_usage
+    exit 1
+  fi
 
   target="${target:-all}"
-  if [ "$target" = "all" ]; then
-    process_all_targets "$command" "$option"
+  if [ "$command" = "multi-arch" ]; then
+    if [ "$target" = "all" ]; then
+      process_multi_arch_targets
+    else
+      build_multi_arch "$target"
+    fi
   else
-    build "container-images/$target" "$command" "$option"
+    if [ "$target" = "all" ]; then
+      process_all_targets "$command" "$option"
+    else
+      build "container-images/$target" "$command" "$option"
+    fi
   fi
 }
 
--- a/container_build.sh
+++ b/container_build.sh
@@ -115,8 +115,13 @@ process_all_targets() {
     if [ "$i" == "container-images/scripts" ]; then
       continue
     fi
-    if [ "$command" = "multi-arch" ] && [ ! -f "$i"/.multi-arch ]; then
-      continue
+    # skip images that don't make sense for multi-arch builds
+    if [ "$command" = "multi-arch" ]; then
+      case "${i//container-images\//}" in
+        rocm|intel-gpu)
+          continue
+          ;;
+      esac
     fi
     build "$i" "$command" "$option"
   done
--- a/container_build.sh
+++ b/container_build.sh
@@ -115,7 +115,7 @@ process_all_targets() {
     if [ "$i" == "container-images/scripts" ]; then
       continue
     fi
-    if [ "$command" = "multi-arch" ] && [ !-f $"i"/.multi-arch ]; then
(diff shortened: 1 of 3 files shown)


cgruver:
@rhatdan @ericcurtin 

WDYT?

ericcurtin (approved):

container_build.sh
```diff
@@ -116,12 +116,24 @@ process_all_targets() {
   done
 }
 
+process_multi_arch_targets() {
+  for i in $(cat ./multi-arch-targets.list); do
+    build_multi_arch "$i"
+  done
+}
+
+build_multi_arch() {
```
rhatdan, line 125: Can this be simplified by passing in "farm" to build_arch()
and then doing `podman $2 build ...`

rhatdan: BTW This is a good idea, and potentially useful for automated builds.

ericcurtin:
It's fine, but maybe the list should exclude certain targets rather than include certain targets. Ideally, Everything should be built multi-arch except for "rocm" and "asahi" if they cause issues. Interestingly, "asahi" might build fine for x86_64, because they do some GPU x86_64 stuff via muvm, so there's actually some rare use-cases for that.

So maybe even just try and exclude "rocm". Generally multi-arch should always be built except in exceptional circumstances. (not taking s390 and ppc into account for a moment)

A different list file maybe isn't necessary either, but that's up to you.

container_build.sh
cgruver: Let me see.

`podman farm` does not like a tag on the image.  So, this might be more complex.

cgruver:
> It's fine, but maybe the list should exclude certain targets rather than include certain targets. Ideally, Everything should be built multi-arch except for "rocm" and "asahi" if they cause issues. Interestingly, "asahi" might build fine for x86_64, because they do some GPU x86_64 stuff via muvm, so there's actually some rare use-cases for that.
> 
> So maybe even just try and exclude "rocm". Generally multi-arch should always be built except in exceptional circumstances. (not taking s390 and ppc into account for a moment)
> 
> A separately list file maybe isn't necessary either, but happy to leave that up to you.

Intel probably doesn't make sense for multi-arch either.

But, I like the idea of an exclude list too.  I'll tinker with it.

ericcurtin:
Yeah sorry intel too :)

container_build.sh
cgruver: Ah, but the script does not explicitly add the `latest` tag either.  So this might work.

I'll tinker with it.

cgruver: @rhatdan I think I succeeded in simplifying based on your request.

cgruver:
@ericcurtin WDYT?

I trigger the podman farm build based on the presence of a `.multi-arch` file in the container image folder.

ericcurtin:
This is mergeable @cgruver but again, I'd flip it to be exclusive rather than inclusive, it's an exception when we don't do multi-arch with the exceptions being rocm, intel (and maybe asahi, but asahi actually use the x86_64 GPU drivers on muvm as I said before).

container_build.sh
```diff
@@ -158,5 +173,6 @@ main() {
   fi
 }
 
+REGISTRY_PATH=${REGISTRY_PATH:-quay.io/ramalama}
```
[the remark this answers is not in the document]
ericcurtin: I think consider this feedback optional, but I am impressed at how well the bot wrote the regex here.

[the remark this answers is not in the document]
ericcurtin: Much better than I would!

cgruver:
> This is mergeable @cgruver but again, I'd flip it to be exclusive rather than inclusive, it's an exception when we don't do multi-arch with the exceptions being rocm, intel (and maybe asahi, but asahi actually use the x86_64 GPU drivers on muvm as I said before).

Do you prefer a `.skip-multi-arch` file in the image folder, or just code into the script which ones to skip?

ericcurtin:
> > This is mergeable @cgruver but again, I'd flip it to be exclusive rather than inclusive, it's an exception when we don't do multi-arch with the exceptions being rocm, intel (and maybe asahi, but asahi actually use the x86_64 GPU drivers on muvm as I said before).
> 
> @ericcurtin Do you prefer a `.skip-multi-arch` file in the image folder, or just code into the script which ones to skip?

If it was me I'd code it in, people are gonna wonder what the mystery files are about (and they are hidden), but I'll leave it upto you.

I just want to get people into the habit of building multi-arch. Like it's very easy to think Nvidia is x86-only, as one example, but it's actually not. And an eGPU can connect to multiple arches also.

ericcurtin:
Plus there's less platforms to skip than include, at present it's max 3 we should skip for multi-arch

cgruver:
> > > This is mergeable @cgruver but again, I'd flip it to be exclusive rather than inclusive, it's an exception when we don't do multi-arch with the exceptions being rocm, intel (and maybe asahi, but asahi actually use the x86_64 GPU drivers on muvm as I said before).
> > 
> > 
> > @ericcurtin Do you prefer a `.skip-multi-arch` file in the image folder, or just code into the script which ones to skip?
> 
> If it was me I'd code it in, people are gonna wonder what the mystery files are about (and they are hidden), but I'll leave it upto you.
> 
> I just want to get people into the habit of building multi-arch. Like it's very easy to think Nvidia is x86-only, as one example, but it's actually not. And an eGPU can connect to multiple arches also.

Done and done.  :-)  I added a `case` to exclude `intel-gpu` and `rocm`

ericcurtin (approved):

rhatdan:
LGTM

PR #738, opened 2025-02-04
github actions: ramalama install

dougsland:
## Summary by Sourcery

CI:
- Set up a GitHub Actions workflow to install `ramalama` and pull the `tinyllama` model on push to `main` and pull requests targeting `main`. The workflow runs on Ubuntu and macOS and installs required dependencies like `lshw` and `curl`.

### DIFF
--- a/.github/workflows/install_ramalama.yml
+++ b/.github/workflows/install_ramalama.yml
@@ -1,0 +1,48 @@
+name: Install RamaLama
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  install_ramalama:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]  # Runs on Ubuntu, macOS, and Fedora
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v3
+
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@ericcurtin PTAL

dougsland:
@rhatdan there is a comma (,) in the install that is fixed in this PR as well.

rhatdan:
Yes I saw that.

rhatdan:
@dougsland does `install -l` blow up on errors?

dougsland:
> @dougsland does `install -l` blow up on errors?

yes but we had no CI and I tested with make install, that's why nobody catch that early too.

dougsland:
> > @dougsland does `install -l` blow up on errors?
> 
> yes but we had no CI and I tested with make install, that's why nobody catch that early too.

I need to look the code to see the differences but maybe in another patch, we could do make install calling install.sh

dougsland:
> > @dougsland does `install -l` blow up on errors?
> 
> yes but we had no CI and I tested with make install, that's why nobody catch that early too.

Another example that we could handle that in our global timezone and merge it. install.sh is broken right now due comma before console.py. @rhatdan

rhatdan:
This PR needs to be refreshed then we can merge.
LGTM

dougsland:
@rhatdan ready

dougsland:
CI/CD failure not related to the patch.

PR #739, opened 2025-02-04
There's a comma in the list of files in install.sh

ericcurtin:
It shouldn't be there. Also remove the -qq, it makes it feel like the script isn't making progress as podman takes a while to install.

## Summary by Sourcery

Remove the -qq flag from apt-get to show installation progress and fix a syntax error in a list of files.

Bug Fixes:
- Removed the unnecessary `-qq` flag from `apt-get` to make the installation progress visible.

Chores:
- Fixed a syntax error in a Python file list by removing an extra comma.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -35,12 +35,12 @@ dnf_install() {
 }
 
 apt_get_install() {
-  apt-get -qq -y install "$1"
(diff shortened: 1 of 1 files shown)


rhatdan:
:+1:

PR #741, opened 2025-02-05
On native macOS this was returning an incorrect path

ericcurtin:
$ ramalama --dryrun run granite-code
llama-run -c 2048 --temp 0.8 /path/to/model


ramalama/model.py
```diff
@@ -285,12 +285,14 @@ def build_prompt(self, args):
         return prompt
 
     def get_model_path(self, args):
+        model_path = self.exists(args)
+        if model_path:
+            return model_path
+
         if args.dryrun:
             return "/path/to/model"
```
jhjaggars, line 293: In this case, should we return something that indicates the model doesn't exist in the store already and we'd try to pull it?

ericcurtin: I don't think we should pull for "--dryrun", this is kinda just for, please print an approximate "podman run" command

rhatdan:
LGTM

ramalama/model.py
rhatdan: I am not sure if we should print an error in the case that the model does not exists. But I could be swayed either way.

PR #746, opened 2025-02-06
Update vLLM containers

ericcurtin:
From rhoai-2.17 -> rhoai-2.18

## Summary by Sourcery

Build:
- Update the vLLM base container image from `rhoai-2.17` to `rhoai-2.18` for both CUDA and ROCm runtimes.

### DIFF
--- a/ramalama/model.py
+++ b/ramalama/model.py
@@ -118,9 +118,9 @@ class Model:
 
         if args.runtime == "vllm":
             if gpu_type == "HIP_VISIBLE_DEVICES":
-                return "quay.io/modh/vllm:rhoai-2.17-rocm"
+                return "quay.io/modh/vllm:rhoai-2.18-rocm"
 
-            return "quay.io/modh/vllm:rhoai-2.17-cuda"
+            return "quay.io/modh/vllm:rhoai-2.18-cuda"
 
         split = version().split(".")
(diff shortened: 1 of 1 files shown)


ramalama/model.py
```diff
@@ -118,9 +118,9 @@ def _image(self, args):
 
         if args.runtime == "vllm":
             if gpu_type == "HIP_VISIBLE_DEVICES":
-                return "quay.io/modh/vllm:rhoai-2.17-rocm"
+                return "quay.io/modh/vllm:rhoai-2.18-rocm"
```
engelmi, line 121: Would it make sense to have such values configurable in a dedicated .conf file for ramalama? For example, having 
`VLLM_CONTAINER_CPU=quay.io/modh/vllm:rhoai-2.18-rocm`
`VLLM_CONTAINER_GPU=quay.io/modh/vllm:rhoai-2.18-cuda`

engelmi (approved):
LGTM

ramalama/model.py
ericcurtin: We started something like that, if you grep around the codebase you'll see it:

ramalama.conf

to keep it simple we map things directly to the CLI args (but a CLI arg would take precedence over a conf setting). But I haven't used it in a long time, feel free to test, edit, enhance that config file.

PR #749, opened 2025-02-06
Detect Intel ARC GPU in Meteor Lake chipset

cgruver:
Add logic to detect an Intel GPU and set `INTEL_VISIBLE_DEVICES` if found.

__Note:__ This is preliminary logic that narrowly detects the Intel ARC GPU in the Meteor Lake Core Ultra 7 chipset.

## Summary by Sourcery

Detect Intel ARC GPUs and use them if available.

New Features:
- Enable the use of Intel ARC GPUs on systems with Meteor Lake Core Ultra 7 chipsets.

Tests:
- Update tests to support Intel ARC GPUs.

### DIFF
--- a/ramalama/common.py
+++ b/ramalama/common.py
@@ -283,15 +283,15 @@ def get_gpu():
         return
 
     # INTEL iGPU CASE (Look for ARC GPU)
-    if os.path.exists('/sys/bus/pci/drivers/i915/0000:00:02.0/device'):
-        try:
-            with open('/sys/bus/pci/drivers/i915/0000:00:02.0/device', 'rb') as f:
-                content = f.read()
-                if b"0x7d55" in content:
-                    os.environ["INTEL_VISIBLE_DEVICES"] = "1"
-        except OSError:
-            # Handle the case where the file does not exist
-            pass
+    igpu_num = 0
+    for fp in sorted(glob.glob('/sys/bus/pci/drivers/i915/*/device')):
+        with open(fp, 'rb') as file:
+            content = file.read()
+            if b"0x7d55" in content:
+                igpu_num += 1
+
+    if igpu_num:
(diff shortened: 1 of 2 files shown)


cgruver:
@ericcurtin @rhatdan 

WDYT?

This is an attempt at Intel GPU detection.  It's very narrow at this point, but we can iterate on it to make it capable of broader detection.

ramalama/common.py
```diff
@@ -282,9 +282,19 @@ def get_gpu():
         os.environ["HIP_VISIBLE_DEVICES"] = str(gpu_num)
         return
 
+    # INTEL iGPU CASE (Look for ARC GPU)
+    if os.path.exists('/sys/bus/pci/drivers/i915/0000:00:02.0/device'):
+        try:
+            with open('/sys/bus/pci/drivers/i915/0000:00:02.0/device', 'rb') as f:
```
rhatdan, line 288: Would it make sense to use a glob for this?

glob.glob(/sys/bus/pci/drivers/i915/*/device)

cgruver: I'll try that.

Learning Python on the Go as it were...  ;-)

cgruver: Modified to use glob.

WDYT?

cgruver:
There are other iGPU types from Intel which will work in theory.  Right now the test is for a very specific chip.

As we identify other iGPUs that provide a boost over CPU inference, we'll need to decide how to handle a table of supported iGPUs.

cgruver:
I've got some 13th gen and 12 gen NUCs which I plan to test.  They are currently part of OCP clusters in my lab, but I can tear them down this weekend.  Anything older than that is probably not worth it.  CPU will likely be better.

rhatdan:
LGTM
But you need to take draft flag off.

cgruver:
> LGTM But you need to take draft flag off.

Done.

Issue #754, opened 2025-02-06
ramalama version do not show git hash and date


dougsland:
It's hard to catch which change the ramalama tool is based.

--version show `0.5.5` but which change is based ?

rhatdan:
I don't see a reason to dirty the display when this information is fairly easy to find.

https://github.com/containers/ramalama/releases/tag/v0.5.5

dougsland:
Use git log, closing this one.

PR #755, opened 2025-02-06
--version points to git hash and date

dougsland:
Today ramalama just share version 0.5.5 (as example) but don't tell users and developer which commit it's using.

Resolves: https://github.com/containers/ramalama/issues/754

## Summary by Sourcery

Update the versioning scheme to use setuptools_scm, dynamically deriving the version from Git.

Build:
- Use setuptools_scm for version management.

Chores:
- Remove the hardcoded version and replace it with dynamic versioning.


dougsland:
https://github.com/containers/ramalama/pull/756

PR #756, opened 2025-02-06
--version points to git hash

dougsland:
Today ramalama just share version 0.5.5 (as example) but don't tell users and developer which commit it's using.

Resolves: https://github.com/containers/ramalama/issues/754

## Summary by Sourcery

Update the version reporting to use the Git commit hash if available, falling back to a file-based version if not in a Git repository.

New Features:
- Add the ability to determine the version from the git hash.

Enhancements:
- Improve version reporting to provide more context.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -100,6 +100,29 @@ check_platform() {
   return 0
 }
 
+setup_version_file() {
+  local config_dir="$HOME/.config/ramalama"
+  local version_file="$config_dir/version"
+
+  # Ensure ~/.config/ramalama/ exists
+  mkdir -p "$config_dir"
+
+  version="0.0.0"
+  # Try to get version from Git
+  if git rev-parse --is-inside-work-tree >/dev/null 2>&1; then
+    version=$(git describe --tags --long --always)
+    # Format version properly (strip 'g' from commit hash)
+    if [[ "$version" =~ ([0-9]+\.[0-9]+\.[0-9]+)-([0-9]+)-g([a-f0-9]+) ]]; then
+      version="${BASH_REMATCH[1]}.dev${BASH_REMATCH[2]}+${BASH_REMATCH[3]}"
+    fi
+  fi
+
+  # Save version to file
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -100,6 +100,29 @@ check_platform() {
   return 0
 }
 
+setup_version_file() {
```
rhatdan, line 103: This would not work in the pypi install, rpm install, or any install that does not use the install.sh.

rhatdan:
In order to get PRs merged, you need to modify the tests to make sure they pass.

rhatdan:
Why do you care about a git hash?  We should be only cutting releases, and then you can go to github.com and look up the release which will point at the github hash?

rhatdan:
https://github.com/containers/ramalama/releases/tag/v0.5.5

install.sh
engelmi: Could we simply use a `setup.py.in` with a placeholder for the version, i.e. `version=@VERSION@`, which we can replace at build-time? This could then be used by all means to install ramalama by simply adding this prepare step. Additionally, the version would be hard-coded. We use that approach for example in [BlueChi](https://github.com/eclipse-bluechi/bluechi/blob/main/src/bindings/python/setup.py.in#L15)
WDYT? @dougsland @rhatdan

dougsland: @engelmi I like the idea but maybe we will close this one. WDYT @rhatdan @engelmi  ?

dougsland:
> Why do you care about a git hash? We should be only cutting releases, and then you can go to github.com and look up the release which will point at the github hash?

It's generally useful to include version details in the --version flag so users when visit the release page and can match the commit or just locally via git. Additionally, during development, having the version information helps track which patch is being used, especially when working with multiple patches. After all, when developing it always says v0.5.5 which is not really correct it's release v0.5.5 + current git commit. However, I am okay to close this, if you and @ericcurtin decide, it's just something I found during my tests+development.

PR #762, opened 2025-02-07
Check if krunkit process is running

ericcurtin:
The podman machine list technique isn't consisent across podman machine versions.


ramalama/common.py
```diff
@@ -32,33 +32,37 @@ def container_manager():
...
-        return "podman"
+        return None
 
     if available("docker"):
         return "docker"
 
     return None
 
 
+def is_krunkit_running():
```
rhatdan, line 46: Try first with       --all-providers 

This should be in newer podman's

ericcurtin:
Re-opened here https://github.com/containers/ramalama/pull/763

PR #763, opened 2025-02-07
Check if krunkit process is running with --all-providers

ericcurtin:
Required for newer podman machine.

### DIFF
--- a/ramalama/common.py
+++ b/ramalama/common.py
@@ -3,6 +3,7 @@
 import glob
 import hashlib
 import os
+import re
 import random
 import logging
 import shutil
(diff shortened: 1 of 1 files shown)


ericcurtin:
This won't work with older podman machine... Which I think is fine... Our krunkit technique isn't mature yet, so it's fine it just works in newer podman machine

ericcurtin:
@rhatdan @benoitf I'm heading offline, but a quick test would be appreciated...

ericcurtin:
Drastically simplified, we fork one process now instead of two in this block

rhatdan:
LGTM
Merge once tests pass.

PR #768, opened 2025-02-08
There would be one case where this wouldn't work

ericcurtin:
When accumulated_size has just been refreshed to zero

## Summary by Sourcery

Bug Fixes:
- Fix an edge case where the download progress was not being updated correctly when the total downloaded size was a multiple of the read buffer size.

### DIFF
--- a/ramalama/http_client.py
+++ b/ramalama/http_client.py
@@ -59,7 +59,7 @@ class HttpClient:
         while True:
             data = self.response.read(1024)
             if not data:
-                break
+                return
 
             size = file.write(data)
             if progress:
@@ -69,8 +69,10 @@ class HttpClient:
                     accumulated_size = 0
                     last_update_time = time.time()
 
-        if accumulated_size:
+        if accumulated_size > 0:
             self.update_progress(accumulated_size)
+
+        if progress:
             print("\033[K", end="\r")
 
     def human_readable_time(self, seconds):


ramalama/http_client.py
```diff
@@ -66,11 +67,14 @@ def perform_download(self, file, progress):
...
                 if time.time() - last_update_time >= 0.1:
                     self.update_progress(accumulated_size)
+                    do_final_progress = True
                     accumulated_size = 0
                     last_update_time = time.time()
 
-        if accumulated_size:
-            self.update_progress(accumulated_size)
+        if do_final_progress:
+            if accumulated_size:
```
rhatdan, line 75: This makes no sense, to add the option.

Why not do 

if accumulated_size > 0:

You reset the accumalated_size option right after the do_final_progress.

ericcurtin: Ah it's the "\033[K" outside the if block that's the main fix...

I dislike python's indenting as '{' '}' sometimes, makes things like this not so obvious :)

ericcurtin: Wait sorry this does make sense... It's just not clear, I'll do it a different way that's more clear

rhatdan: So you can break out with `if not data:` when you have accumalated size and do not want to update_progress?

rhatdan:
LGTM

Much better.

Merge when tests pass.

PR #782, opened 2025-02-10
Reuse Ollama cached image when available

kush-gupt:
Created a new function called in_existing_cache to take in a model name and tag, search a default Ollama cache and see if there's a matching model layer to symlink in the ramalama store.

Allows the process of someone running ollama to use ramalama with no model pulls!
```
$ ollama ls
NAME                                   ID              SIZE      MODIFIED      
starcoder2:3b                          9f4ae0aff61e    1.7 GB    9 minutes ago    
sroecker/nuextract-tiny-v1.5:latest    be09317bfbbb    994 MB    7 hours ago      
deepseek-r1:32b                        38056bbcbb2d    19 GB     3 weeks ago      
qwen2.5-coder:32b                      4bd6cbf2d094    19 GB     2 months ago     
nomic-embed-text:latest                0a109f422b47    274 MB    8 months ago     

$ ramalama ls
NAME MODIFIED SIZE

$ ramalama run starcoder2:3b 
> We have a cache and didn't have to pull the model down
...

$ ramalama ls
NAME                   MODIFIED       SIZE   
ollama://starcoder2:3b 15 seconds ago 1.59 GB
```

PR #784, opened 2025-02-11
Check if file exists before sorting them into a list

kush-gupt:
A dangling symlink can cause `ramalama ls` to fail with:
```
$ ramalama ls            
Error: [Errno 2] No such file or directory: 'ollama/qwen:1.8b
```

So checking to see if the file exists before appending it into a list to be sorted fixes this:
```
$ ramalama ls
NAME                       MODIFIED     SIZE    
ollama://starcoder2:3b     1 hour ago   1.59 GB 
ollama://deepseek-r1:32b   12 hours ago 18.49 GB
ollama://qwen2.5-coder:32b 12 hours ago 18.49 GB
```

Root cause could be solved by creating a hard link instead of symbolic from to ollama model blob to the ramalama blob store

## Summary by Sourcery

Bug Fixes:
- Fix an error where a dangling symlink could cause `ramalama ls` to fail.

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -429,7 +429,7 @@ def human_duration(d):
         return f"{d // 31536000} years"
 
 
-def list_files_by_modification():
+def list_files_by_modification(args):
     paths = Path().rglob("*")
     models = []
     for path in paths:
@@ -440,6 +440,9 @@ def list_files_by_modification():
                 continue
         if os.path.exists(path):
             models.append(path)
+        else:
+            print(f"Broken symlink found in: {args.store}/models/{path} \nAttempting removal")
+            New(str(path).replace("/", "://", 1), args).remove(args)
 
     return sorted(models, key=lambda p: os.path.getmtime(p), reverse=True)
 
@@ -537,7 +540,7 @@ def _list_models(args):
     models = []
 
     # Collect model data
-    for path in list_files_by_modification():
+    for path in list_files_by_modification(args):
         if path.is_symlink():
             if str(path).startswith("file/"):
                 name = str(path).replace("/", ":///", 1)
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -438,7 +438,8 @@ def list_files_by_modification():
                 path = str(path).replace("file/", "file:///")
                 perror(f"{path} does not exist")
                 continue
-        models.append(path)
+        if os.path.exists(path):
+            models.append(path)
 
     return sorted(models, key=lambda p: os.path.getmtime(p), reverse=True)
 


kush-gupt:
Fixes the issue reproduced here: https://github.com/containers/ramalama/pull/782#issuecomment-2650930131

ericcurtin (approved):

kush-gupt:
In the context of where the other inodes are located, it _should_ be just the ollama and hf caches right? So the user would have to ollama/hf rm them separately to completely wipe the model from their filesystem but that's also what we'd expect right? 

Scenario I see is: User pulls some model with ramalama, it's nice to load from cache but it's still on the source cache/tooling to manage the original model?

rhatdan:
I was thinking of creating a hard link to a file in the same destination to where ramalama pull ollaman:foobar puts it now.

So you would still have a symlink, but instead of a symlink to the file under OLLAMA it would be a symlink to a hardlink file under $HOME/.local/share/ramalama

rhatdan:
With this change does `ramalama rm MODEL` with broken symlink report MODEL does not exists?  I want ramalama rm MODEL to remove broken symlink.

kush-gupt:
ramalama rm works to remove the broken symlink:
```
$ ramalama rm qwen:1.8b
Untagged: qwen:1.8b
Deleted: sha256:1296b084ed6bc4c6eaee99255d73e9c715d38e0087b6467fd1c498b908180614
Deleted: sha256:9ece4a97bfb61bdb539531db5584fa119ad55684281d8a2d864339ae3fdd6c15

$ ls $HOME/.local/share/ramalama/models/ollama/
deepseek-r1:32b		qwen2.5-coder:32b	starcoder2:3b
```

ericcurtin:
> With this change does `ramalama rm MODEL` with broken symlink report MODEL does not exists? I want ramalama rm MODEL to remove broken symlink.

I'm not sure there is a case where broken symlink happens. I suspect @kush-gupt just created a dangling symlink already. The problem with hardlinks is the Ollama structure:

```
                  ----> modela:latest
sha256:somechecksum 
                  ----> modela:8b
```

8b and latest are the same file, both just symlinks for de-duplication purposes.

If we make things hardlinks, we have to do a lot of searching for the other linked hardnames, it will be expensive. We don't really gain the automatic reference counting cleanup capabilities of hardlinks because we don't refer to models by their checksum (and we have to find that checksum somehow, with symlink it's easy because you are pointed to that SHA, when it's a hardlink we aren't pointed to any sha)

ericcurtin:
IIRC the "ramalama rm" does a decent job of trying to delete all orphaned things, which is why here we see two sha256's deleted in @kush-gupt 's output

kush-gupt:
Indeed, I had that broken symlink created after I did `ollama rm qwen:1.8b` after I had already done a `ramalama pull qwen:1.8b`

rhatdan:
Ok I think I am ok with just symlink as long as RamaLama handles it.  I think we also create a symlink to a model stored as a file.

file:///tmp/model.gguf for example.

rhatdan:
I would like to see a test for this though.

Create a model with file://$TEMPDIR/model.gguf  Show it in list.  Remove $TEMPDIR/model.gguf
Now see if the list sees it.  Then attempt to remove it and make sure the rm succeeds.

kush-gupt:
Added a call to remove if a broken symlink is detected when listing models:
```
$ ollama ls
NAME                                   ID              SIZE      MODIFIED     
nomic-embed-text:latest                0a109f422b47    274 MB    8 months ago  
  
$ ramalama ls 
NAME                       MODIFIED     SIZE    
ollama://starcoder2:3b     7 hours ago  1.59 GB 
ollama://deepseek-r1:32b   17 hours ago 18.49 GB
ollama://qwen2.5-coder:32b 17 hours ago 18.49 GB

$ ramalama pull nomic-embed-text:latest 

$ ollama rm nomic-embed-text:latest 
deleted 'nomic-embed-text:latest'

kugupta@kugupta-mac ramalama % ramalama ls                          
Broken symlink found in: /Users/kugupta/.local/share/ramalama/models/ollama/nomic-embed-text:latest 
Attempting removal
Untagged: nomic-embed-text:latest
Deleted: sha256:970aa74c0a90ef7482477cf803618e776e173c007bf957f635f1015bfcfef0e6
NAME                       MODIFIED     SIZE    
ollama://starcoder2:3b     7 hours ago  1.59 GB 
ollama://deepseek-r1:32b   17 hours ago 18.49 GB
ollama://qwen2.5-coder:32b 18 hours ago 18.49 GB
```

rhatdan:
LGTM

maxamillion (approved):
LGTM

Issue #788, opened 2025-02-11
Optimization: check if container capable only once


ericcurtin:
On execution of a ramalama command we check if the environment is suitable to run via podman/docker before executing some command.

The way we do it at present is suboptimal we fork a podman machine command 4 times as an example on macOS.

PR #790, opened 2025-02-11
Stash output from container_manager

rhatdan:
Fixes: https://github.com/containers/ramalama/issues/788

## Summary by Sourcery

Bug Fixes:
- Fixes an issue where the container engine was not being cached, leading to unnecessary checks and potential inconsistencies across multiple calls.

### DIFF
--- a/ramalama/common.py
+++ b/ramalama/common.py
@@ -26,19 +26,30 @@ HTTP_RANGE_NOT_SATISFIABLE = 416
 DEFAULT_IMAGE = "quay.io/ramalama/ramalama"
 
 
+_engine = ""
+
+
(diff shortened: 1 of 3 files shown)


ramalama/common.py
```diff
@@ -26,19 +26,30 @@
...
 
 
+_engine = ""
+
+
 def container_manager():
+    global _engine
+    if _engine != "":
+        if _engine == "None":
+            return None
```
ericcurtin, line 36: Empty string, "None" and None... Python is frustrating with all the various kinds of null :'(

ericcurtin (approved):

PR #792, opened 2025-02-12
Install llama.cpp for mac and nocontainer tests

rhatdan:


.github/workflows/ci.yml
```diff
@@ -60,8 +60,14 @@ jobs:
         shell: bash
         run: |
            sudo apt-get update
-           sudo apt-get install podman bats bash codespell python3-argcomplete pipx
+           sudo apt-get install podman bats bash codespell python3-argcomplete pipx git cmake
            make install-requirements
+           git clone https://github.com/ggerganov/llama.cpp
+           cd llama.cpp
```
ericcurtin, line 66: I recommend doing something like:

container-images/scripts/build_llama_and_whisper.sh build

and introduce this "build" or similar argument to build_llama_and_whisper.sh .

Keeping everything on the same version of llama.cpp will help avoid flakey builds.

rhatdan: I looked at that but the script was pretty focused on dnf install commands, which will not work on Ubuntu.

Lets get this in and then we ca work towards that.

ericcurtin: Sure, as a follow-on:

It's pretty easy to avoid the dnf stuff in that script, if dnf doesn't exist, don't execute that part

ericcurtin: @rhatdan https://github.com/containers/ramalama/pull/794

PR #793, opened 2025-02-12
_engine is set to None or has a value

ericcurtin:
Simplification, less states of this variable

## Summary by Sourcery

Enhancements:
- Simplify the logic for determining the container engine by reducing the number of possible states for the `_engine` variable.

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -821,6 +821,7 @@ def _run(parser):
         help="require HTTPS and verify certificates when contacting registries",
     )
 
+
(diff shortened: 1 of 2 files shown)


rhatdan:
The problem with this, is if there is no engine it will continuously go through all of the steps including launching the podman machine command to figure out that the machine is not running with libkrun.

My version only checks once in the None case.

ericcurtin:
How about now?

rhatdan:
LGTM

PR #800, opened 2025-02-13
Change --network-mode to --network

ericcurtin:
So users can easily understand what podman/docker CLI this maps to.

## Summary by Sourcery

Documentation:
- Update CLI documentation to reflect the argument name change from `--network-mode` to `--network`.


rhatdan:
Sorry I missed this, I would definitely not want an option with a "-" in it, and keeping to the Podman/Docker standard is best.

PR #802, opened 2025-02-13
If ngl is not specified

ericcurtin:
ngl should default to 999 in all these cases.

## Summary by Sourcery

Bug Fixes:
- Set ngl to 999 if it is less than 0 to resolve issues caused by unspecified ngl values.


rhatdan:
LGTM

PR #803, opened 2025-02-13
Prepare containers to run with ai-lab-recipes

rhatdan:
Add two new scripts llama-server.sh and whisper-server.sh which can handle environment variables from the ai-lab-recipes.

## Summary by Sourcery

New Features:
- Added `llama-server.sh` and `whisper-server.sh` to support running the containers with ai-lab-recipes.

### DIFF
--- a/container-images/cuda/Containerfile
+++ b/container-images/cuda/Containerfile
@@ -14,3 +14,6 @@ RUN dnf install -y python3 && \
 # Copy the entire installation directory from the builder
 COPY --from=builder /tmp/install /usr
 
+ENV MODEL_PATH=/mnt/models/model.file
+
+COPY --chmod=755 ../scripts /usr/bin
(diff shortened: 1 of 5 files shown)


rhatdan:
@benoitf @bmahabirbu @ericcurtin PTAL

benoitf (commented):
these scripts look ok but my question is:
who will call these scripts ?

because it does not seems that there is an entrypoint calling these scripts automatically

rhatdan:
I considered doing the entrypoint, but I don't want to have separate scripts for lama-server and whisper-server, so the idea would be to change the quadlets used to launch the model server.

benoitf:
I would expect to have a default mode as I'm using ramalama

```
$ podman run --rm -it quay.io/ramalama/ramalama
[root@f6ae4f8862d6 /]#
```
today it's defaulting to bash

so at elast I would expect the list of scripts I could call ?

but for example if there is a `MODEL_CHAT_FORMAT` it could just run `llama-server` script
if there is nothing it then just run `/bin/bash` as the parent image


or if I provide env variable RAMALA_something="serve" it's running the script to serve a model

Issue #642, opened 2025-01-28
Add support for safetensors (non-GGUF) model format


sallyom:
Instead of a single model file, ramalama will need to mount and pull a directory.  It would be really nice to have `ramalama pull the/model --format  safetensors` or something like that

bmahabirbu:
Great idea, if we're using vllm (or there that support safe tensor) as the runtime we can mount the folder as is. If we are using llama.cpp we can convert the safetensors to a gguf then mount the file as usual!

ericcurtin:
I agree we should implement this. I would say even drop the "--format safetensors", I think it would not be too hard to automatically detect a certain model is safetensors.

rhatdan:
SGTM

rhatdan:
@ericcurtin could you update this issue, with the current state where we are with safetensor support.

rhatdan:
@olliewalsh @engelmi Where are we on this support?  Do we currently support mounting safetensors into containers now?

engelmi:
> [@olliewalsh](https://github.com/olliewalsh) [@engelmi](https://github.com/engelmi) Where are we on this support? Do we currently support mounting safetensors into containers now?

The model store allows to manage arbitrary number of files - which files are pulled is defined in the model registry integration layer, i.e. the `pull` function in our model implementations like huggingface, ollama, etc. 

AFAIK we are able to pull safetensor files from the huggingface registry and mount them into the ramalama container. However, I think only `--runtime mlx` (on apple) will be able to run them. I described [in this comment](https://github.com/containers/ramalama/issues/1816#issuecomment-3223012115) how to convert the safetensor model to gguf and run it via llama.cpp.

rhatdan:
@engelmi what is the current state of this issue?

engelmi:
> [@engelmi](https://github.com/engelmi) what is the current state of this issue?

@rhatdan I think these PRs:
- https://github.com/containers/ramalama/pull/1976
- https://github.com/containers/ramalama/pull/2009

(re-)added support to run safetensors with ramalama, provided the runtime supports it. 
llama.cpp only supports gguf. I think vllm only uses safetensors, so `--runtime vllm` in combination with the proper container image (or local vllm installation for `--nocontainer`) would be needed. Another runtime supporting safetensors would be mlx. But I haven't tried both, unfortunately. 
But overall, I think we could close this issue.

RobertKrawitz:
The existing message is not clear:

```
[...]

PR #805, opened 2025-02-13
Introduced a model store

engelmi:
Added a model store to standardize pulling, storing and using models across the different repositories.

A key goal of this store is to support downloading multiple files, e.g. the chat template from ollama models or additional metadata from non-GGUF models. In addition, this is probably a first step towards enabling the usage of safetensors (https://github.com/containers/ramalama/issues/642) where a model consists of multiple files. 

The proposed structure is inspired by how the huggingface-cli stores its files and extends it for the multi-source usage in ramalama. 

The proposed storage structure looks like this after running 
- `ramalama pull ollama://tinyllama`
- `ramalama pull hf://ibm-granite/granite-3b-code-base-2k-GGUF/granite-3b-code-base.Q4_K_M.gguf`
```
~/.local/share/ramalama/store
   |-- ollama
   |   |-- tinyllama
   |   |   |-- blobs
   |   |   |   |-- sha256:2af3b81862c6be03c769683af18efdadb2c33f60ff32ab6f83e42c043d6c7816
   |   |   |   |-- sha256:6331358be52a6ebc2fd0755a51ad1175734fd17a628ab5ea6897109396245362
   |   |   |   |-- sha256:af0ddbdaaa26f30d54d727f9dd944b76bdb926fdaf9a58f63f78c532f57c191f
   |   |   |-- refs
   |   |   |   |-- latest
   |   |   |-- snapshots
   |   |   |   |-- sha256:6331358be52a6ebc2fd0755a51ad1175734fd17a628ab5ea6897109396245362
   |   |   |   |   |-- chat_template -> ../../blobs/sha256:af0ddbdaaa26f30d54d727f9dd944b76bdb926fdaf9a58f63f78c532f57c191f
   |   |   |   |   |-- config.json -> ../../blobs/sha256:6331358be52a6ebc2fd0755a51ad1175734fd17a628ab5ea6897109396245362
   |   |   |   |   |-- tinyllama -> ../../blobs/sha256:2af3b81862c6be03c769683af18efdadb2c33f60ff32ab6f83e42c043d6c7816
   |-- huggingface
   |   |-- ibm-granite
   |   |   |-- granite-3b-code-base-2k-GGUF
   |   |   |   |-- blobs
   |   |   |   |   |-- sha256:c803a9bb910be0699501319140329ba0d2850aeb8827389e38f24e8370f04293
   |   |   |   |-- refs
   |   |   |   |   |-- latest
   |   |   |   |-- snapshots
   |   |   |   |   |-- sha256:c803a9bb910be0699501319140329ba0d2850aeb8827389e38f24e8370f04293
   |   |   |   |   |   |-- granite-3b-code-base.Q4_K_M.gguf -> ../../blobs/sha256:c803a9bb910be0699501319140329ba0d2850aeb8827389e38f24e8370f04293
```

## Summary by Sourcery

New Features:
- Add a model store to handle pulling, storing, and accessing models across repositories.

### DIFF
--- a/ramalama/common.py
+++ b/ramalama/common.py
@@ -21,6 +21,7 @@ logging.basicConfig(level=logging.WARNING, format="%(asctime)s - %(levelname)s -
 
 MNT_DIR = "/mnt/models"
 MNT_FILE = f"{MNT_DIR}/model.file"
+HTTP_NOT_FOUND = 404
 HTTP_RANGE_NOT_SATISFIABLE = 416
 
 DEFAULT_IMAGE = "quay.io/ramalama/ramalama"
@@ -154,6 +155,19 @@ def verify_checksum(filename):
     # Compare the checksums
     return sha256_hash.hexdigest() == expected_checksum
 
+def generate_sha256(to_hash: str) -> str:
+    """
+    Generates a sha256 for a string.
+
+    Args:
+    to_hash (str): The string to generate the sha256 hash for.
+
+    Returns:
+    str: Hex digest of the input appended to the prefix sha256:
+    """
+    h = hashlib.new("sha256")
+    h.update(to_hash.encode("utf-8"))
+    return f"sha256:{h.hexdigest()}"
 
 # default_image function should figure out which GPU the system uses t
 # then running appropriate container image.
@@ -199,8 +213,9 @@ def download_file(url, dest_path, headers=None, show_progress=True):
             return  # Exit function if successful
 
         except urllib.error.HTTPError as e:
-            if e.code == HTTP_RANGE_NOT_SATISFIABLE:  # "Range Not Satisfiable" error (file already downloaded)
-                return  # No need to retry
+            # "Range Not Satisfiable" error (file already downloaded)
+            if e.code in [HTTP_RANGE_NOT_SATISFIABLE, HTTP_NOT_FOUND]:
+                raise e
 
         except urllib.error.URLError as e:
             console.error(f"Network Error: {e.reason}")
--- a/ramalama/huggingface.py
+++ b/ramalama/huggingface.py
@@ -1,8 +1,9 @@
 import os
 import pathlib
 import urllib.request
-from ramalama.common import available, run_cmd, exec_cmd, download_file, verify_checksum, perror
+from ramalama.common import available, run_cmd, exec_cmd, download_file, verify_checksum, perror, generate_sha256
 from ramalama.model import Model, rm_until_substring
+from ramalama.model_store import ModelRegistry, SnapshotFile
 
 missing_huggingface = """
 Optional: Huggingface models require the huggingface-cli module.
@@ -14,29 +15,30 @@ pip install huggingface_hub
 
 def is_huggingface_cli_available():
     """Check if huggingface-cli is available on the system."""
-    if available("huggingface-cli"):
-        return True
-    else:
-        return False
+    return available("huggingface-cli")
 
-
-def fetch_checksum_from_api(url):
+def fetch_checksum_from_api(organization, file):
     """Fetch the SHA-256 checksum from the model's metadata API."""
-    with urllib.request.urlopen(url) as response:
-        data = response.read().decode()
-    # Extract the SHA-256 checksum from the `oid sha256` line
-    for line in data.splitlines():
-        if line.startswith("oid sha256:"):
-            return line.split(":", 1)[1].strip()
-    raise ValueError("SHA-256 checksum not found in the API response.")
-
+    checksum_api_url = f"https://huggingface.co/{organization}/raw/main/{file}"
+    try:
+        with urllib.request.urlopen(checksum_api_url) as response:
+            data = response.read().decode()
+        # Extract the SHA-256 checksum from the `oid sha256` line
+        for line in data.splitlines():
+            if line.startswith("oid sha256:"):
+                return line.replace("oid", "").strip()
+        raise ValueError("SHA-256 checksum not found in the API response.")
+    except urllib.error.HTTPError as e:
+        raise KeyError(f"failed to pull {checksum_api_url}: " + str(e).strip("'"))
+    except urllib.error.URLError as e:
+        raise KeyError(f"failed to pull {checksum_api_url}: " + str(e).strip("'"))
 
 class Huggingface(Model):
-    def __init__(self, model):
+    def __init__(self, model, store_path=""):
         model = rm_until_substring(model, "hf.co/")
         model = rm_until_substring(model, "://")
-        super().__init__(model)
-        self.type = "huggingface"
+        super().__init__(model, store_path, ModelRegistry.HUGGINGFACE)
+
         self.hf_cli_available = is_huggingface_cli_available()
 
     def login(self, args):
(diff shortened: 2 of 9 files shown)


engelmi:
@rhatdan @ericcurtin Could you have a brief look at this? Its still in draft, but I wanted to get your opinion if this is going in the right direction or not.

ramalama/common.py
```diff
@@ -199,8 +213,9 @@ def download_file(url, dest_path, headers=None, show_progress=True):
             return  # Exit function if successful
 
         except urllib.error.HTTPError as e:
-            if e.code == HTTP_RANGE_NOT_SATISFIABLE:  # "Range Not Satisfiable" error (file already downloaded)
-                return  # No need to retry
+            # "Range Not Satisfiable" error (file already downloaded)
+            if e.code in [HTTP_RANGE_NOT_SATISFIABLE, HTTP_NOT_FOUND]:
```
ericcurtin, line 217: +1

ramalama/huggingface.py
```diff
@@ -14,29 +15,30 @@
 
 def is_huggingface_cli_available():
     """Check if huggingface-cli is available on the system."""
-    if available("huggingface-cli"):
-        return True
-    else:
-        return False
+    return available("huggingface-cli")
```
ericcurtin, line 18: +1

ramalama/model_store.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+from ramalama.common import verify_checksum, download_file
+
+
+class ModelRegistry(StrEnum):
+    HUGGINGFACE = "huggingface"
+    OLLAMA = "ollama"
+    OCI = "oci"
+    URL = "url"
```
ericcurtin, line 17: Could see the community adding even more here, such as s3:// , just something to keep in mind, I wonder should we continue with https, etc. We do have to code up any protocol we support.

ericcurtin:
The code looks great to me so far, don't see any significant issues at the moment.

ramalama/common.py
```diff
@@ -154,6 +155,19 @@ def verify_checksum(filename):
...
+
+    Args:
+    to_hash (str): The string to generate the sha256 hash for.
+
+    Returns:
+    str: Hex digest of the input appended to the prefix sha256:
+    """
+    h = hashlib.new("sha256")
+    h.update(to_hash.encode("utf-8"))
+    return f"sha256:{h.hexdigest()}"
```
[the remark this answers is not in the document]
engelmi: Switching `:` with `-` makes much sense. 
Since this PR would introduce a breaking regarding the file storage anyway, I can include this as well.

ramalama/model_store.py
engelmi: Good point! Its already different with the default https download and the huggingface-/ollama-cli - although they use https as well, the "library" is different. Maybe I can inject a class/callable - which specifies HOW to download - to the model stores `new_snapshot` function. Then this would be easily extensible.

ericcurtin:
Since this is a breaking change I guess that's the main question, do we want to break all users now on upgrade? I just want to be sure the benefits are worth it. If we gain some functional benefits I think it's a good idea. What I'm less of a fan of is non-functional breaking changes like renaming "repos/models" -> "store" . But if we need to break everything for functional reasons anyway, now is certainly the time to do renames and the rename is fine.

I also think it's possible to do this in a non-breaking way in the existing ~/.local/share/ramalama/repos directory.

~/.local/share/ramalama/models/ is just a directory intended to be a more presentable way for humans and for "ramalama ls" to parse, etc. The metadata, messy implementation details were intended to go to ~/.local/share/ramalama/repos/ directory.

It's worth passing this by @swarajpande5 also, he wrote a fair bit of this.

If we decide we are doing breaking changes, another one we should do is add .gguf extension to the Ollama .gguf files, some tools refused to load them in the past without that file extension. Example that we ended up fixing vllm-side: https://github.com/vllm-project/vllm/issues/7993

But just fair warning @engelmi there's many tests, assumptions built up around the existing technique. This could be tough to get through CI. You will likely spend a lot of time massaging the tests here, etc. If you do this in a non-breaking way it will be less effort to code and less impact on the users when multi-file models such as the safetensors one are enabled.

engelmi:
> Since this is a breaking change I guess that's the main question, do we want to break all users now on upgrade? I just want to be sure the benefits are worth it. If we gain some functional benefits I think it's a good idea. What I'm less of a fan of is non-functional breaking changes like renaming "repos/models" -> "store" . But if we need to break everything for functional reasons anyway, now is certainly the time to do renames and the rename is fine.

The "store" directory is just an intermediate for local development at the moment. It would probably be best to switch to the model directory, i.e. having `~/.local/share/ramalama/models/ollama/tinyllama/<blobs|refs|snapshots>`. The repos directory would go into the respective blobs directory. It would be a breaking change nonetheless. 
And I agree, there should be functional benefits for this. These are the benefits I can think of:
- clear distinction between `name`, `tag` and `organization` of a model - also on a file/directory level (the repository, i.e. ollama or hf, is already used in the current implementation). This also helps during development, I think. 
- using `model tags` in the "refs" directory enables the user to easily switch between different versions. Ollama already provides tags and for huggingface one can use the commit, e.g. https://huggingface.co/ibm-granite/granite-3.0-8b-instruct/raw/257d6976020e06daa75f9b19d056a5e7590bf7fc/model-00001-of-00004.safetensors
I think it would be possible possible to use tags with the current approach, but this would either require updating the symlinks in the models directory to point to the correct sha256 or encode it into the path in the models directory. 
- all files for one model and version are bundled together - so its harder to break things
- easier cleanup, repairing and symlink checking for models
- support for safetensors on a file storage level "out of the box" (not for running them, only to pull, store and provide these files)

> I also think it's possible to do this in a non-breaking way in the existing ~/.local/share/ramalama/repos directory.
> 
> ~/.local/share/ramalama/models/ is just a directory intended to be a more presentable way for humans and for "ramalama ls" to parse, etc. The metadata, messy implementation details were intended to go to ~/.local/share/ramalama/repos/ directory.
> 
> It's worth passing this by @swarajpande5 also, he wrote a fair bit of this.

Yes, it should be possible to support multiple files with the current approach. I am wondering, though, if it makes things harder to understand and maintain - code-wise as well as from the directories and files. 
For example: 
When having multiple files (incl. metadata). Those should be stored in "repos" and linking only the model from "models" to the "repos" directory, right? As a human inspecting the "models" directory, wouldn't I want to know which files are used? And if I add symlinks to the other files, how can I ensure the versions of those files stay the same? Introducing model tags/versions and encoding it into the path would also be a (smaller) breaking change. Adding a "refs" directory, we'd have to keep track of all files for a version there. 

The messy implementation details should be only in code where we can encapsulate them (i.e. in the Ollama/HF/URL/etc. Model), the file storage should be as standardized as possible, I think. 

> If we decide we are doing breaking changes, another one we should do is add .gguf extension to the Ollama .gguf files, some tools refused to load them in the past without that file extension. Example that we ended up fixing vllm-side: [vllm-project/vllm#7993](https://github.com/vllm-project/vllm/issues/7993)

Sounds good to me!
In the model store class this should be simple - basically  a check with `is_model_gguf` on the downloaded blob with a check for the file extension on the model name. 
Same for replacing `:` with `-`, which can also be done in a central location of the store class.

> But just fair warning @engelmi there's many tests, assumptions built up around the existing technique. This could be tough to get through CI. You will likely spend a lot of time massaging the tests here, etc. If you do this in a non-breaking way it will be less effort to code and less impact on the users when multi-file models such as the safetensors one are enabled.

Yes, that is true and will be quite cumbersome to change.

rhatdan:
I really like this PR, but it needs a rebase.  We need to get this in ASAP, and probably need a mechanism to upgrade people, rather then break their stores.

ddiddi:
Happy to help rebase and update

engelmi:
> I really like this PR, but it needs a rebase. We need to get this in ASAP, and probably need a mechanism to upgrade people, rather then break their stores.

I'll rebase and rework this so we can hide the new store behind a CLI flag, e.g. `--use-model-store`. This way, users can rely on the current installation without the need to change anything and still be able to switch. And in the future we simply mark not using that CLI flag as deprecated and enforce it after some time.

rhatdan:
Other PR is merged, this needs to wait until I am back from PTO (two weeks) in case anything goes wrong.

engelmi:
> Other PR is merged, this needs to wait until I am back from PTO (two weeks) in case anything goes wrong.

Have a great PTO!
I'll ensure the PR is ready in two weeks.

engelmi:
Closing in favor of #905

PR #572, opened 2025-01-10
Add privileged flag

ericcurtin:
Split container-related argument configurations into separate functions for better modularity and readability. Improved help descriptions for argument parser. Added new functions to handle GPU, debug, store, and version-related arguments. Refactored `setup_container` method in `Model` class by splitting into smaller functions to enhance maintainability. Added functions to handle specific container setup arguments (e.g., podman, tty, detach, port, privileged, device, environment variables).

## Summary by Sourcery

New Features:
- Added a "--privileged" flag to allow running containers with extended privileges.

PR #809, opened 2025-02-13
Add run and serve arguments for --device and --privileged

cgruver:
Removed `RAMALAMA_DEVICE` for --device override in favor of a `--device` argument.

Added `--privileged` argument to satisfy the desired outcome of #572 

Added the ability to pass multiple instances of `--device /dev/some/device` to allow users to explicitly define which devices to leak into the container.

`--device` and `--privileged` are arguments of `ramalama run|serve`

## Summary by Sourcery

Add the `--device` and `--privileged` options to the run and serve commands. Allow multiple devices to be passed to the `--device` argument. Replace the `RAMALAMA_DEVICE` environment variable with the new `--device` argument.

New Features:
- Add `--device` and `--privileged` arguments to the `ramalama run` and `ramalama serve` commands. The `--device` argument allows for passing multiple device paths to expose to the container.  The `--privileged` argument grants extended privileges to the container.
- Replace the `RAMALAMA_DEVICE` environment variable with the `--device` argument for specifying devices to make available within the container during model serving or running. Multiple `--device` arguments can be provided to specify multiple devices

### DIFF
--- a/docs/ramalama-run.1.md
+++ b/docs/ramalama-run.1.md
@@ -31,6 +31,9 @@ path of the authentication file for OCI registries
 #### **--ctx-size**, **-c**
 size of the prompt context (default: 2048, 0 = loaded from model)
 
+#### **--device**
+declare host device to leak into the container 
+
 #### **--help**, **-h**
 show this help message and exit
 
@@ -42,6 +45,9 @@ name of the container to run the Model in
 
 #### **--network**=*none*
 set the network mode for the container
+
+#### **--privileged**
+give extended privileges to container
 
 #### **--seed**=
 Specify seed rather than using random seed model interaction
--- a/docs/ramalama-run.1.md
+++ b/docs/ramalama-run.1.md
@@ -32,7 +32,13 @@ path of the authentication file for OCI registries
 size of the prompt context (default: 2048, 0 = loaded from model)
 
 #### **--device**
-declare host device to leak into the container 
+Add a host device to the container. Optional permissions parameter  can
+be  used  to  specify device permissions by combining r for read, w for
+write, and m for mknod(2).
+
+Example: --device=/dev/dri/renderD128:/dev/xvdc:rwm
+
+The device specifiaction is passed directly to the underlying container engine.  See documentation of the supported container engine for more information.
 
 #### **--help**, **-h**
 show this help message and exit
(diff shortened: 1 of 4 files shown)


cgruver:
@ericcurtin  Here's my attempt to complete #572 and add `--device`

@rhatdan  I satisfied your ask to make the arguments part of the subcommands `run` and `serve` rather than global.

I will add docs for these if we like this approach.

cgruver:
Looks like I broke the tests.  Need to update the appropriate `.md` docs.

docs/ramalama-run.1.md
```diff
@@ -43,6 +46,9 @@ name of the container to run the Model in
 #### **--network**=*none*
 set the network mode for the container
 
+#### **--privileged**
+give extended privileges to container
```
rhatdan, line 50:
```
       By  default, RamaLama containers are unprivileged (=false) and cannot, for
       example, modify parts of the operating system. This is  because  by  de‐
       fault  a  container is only allowed limited access to devices. A "privi‐
       leged" container is given the same access to devices as the user launch‐
       ing the container, with the exception of virtual consoles  (/dev/tty\d+)
       when running in systemd mode (--systemd=always).

       A  privileged container turns off the security features that isolate the
       container from the host. Dropped Capabilities,  limited  devices,  read-
       only  mount points, Apparmor/SELinux separation, and Seccomp filters are
       all disabled.  Due to the disabled  security  features,  the  privileged
       field  should  almost never be set as containers can easily break out of
       confinement.

       Containers running in a user namespace (e.g., rootless containers)  can‐
       not have more privileges than the user that launched them.
```

rhatdan: Please add this, from Podman man pages.

docs/ramalama-serve.1.md
```diff
@@ -46,6 +46,9 @@ The default is TRUE. The --nocontainer option forces this option to False.
 
 Use the `ramalama stop` command to stop the container running the served ramalama Model.
 
+#### **--device**
+declare host device to leak into the container
```
rhatdan, line 50: Add something like.

```
       Add a host device to the container. Optional permissions parameter  can
       be  used  to  specify device permissions by combining r for read, w for
       write, and m for mknod(2).

       Example: --device=/dev/sdc:/dev/xvdc:rwm.

The device specifiaction is passed directly to the underlying container engine.  See documentation of the supported container engine for more information.
```

docs/ramalama-serve.1.md
```diff
@@ -70,6 +73,9 @@ set the network mode for the container
 #### **--port**, **-p**
 port for AI Model server to listen on
 
+#### **--privileged**
+give extended privileges to container
```
rhatdan, line 77: ditto

ramalama/model.py
```diff
@@ -187,13 +187,11 @@ def setup_container(self, args):
...
-        # Check for env var RAMALAMA_DEVICE to explicitly declare the GPU device path
-        device_override = 0
-        gpu_device = os.environ.get("RAMALAMA_DEVICE")
-        if gpu_device:
-            conman_args += ["--device", gpu_device]
-            device_override = 1
-        if device_override != 1:
+        if args.device:
+            for device_arg in args.device:
+                print(device_arg)
```
rhatdan, line 192: This looks like debug line.

ramalama/cli.py
```diff
@@ -233,6 +233,12 @@ def configure_arguments(parser):
         help="store AI Models in the specified directory",
     )
     parser.add_argument("-v", "--version", dest="version", action="store_true", help="show RamaLama version")
+    # parser.add_argument("--device",
```
rhatdan, line 236: Either turn this code on or delete it.

rhatdan:
LGTM

PR #817, opened 2025-02-14
Encountered a bug where this function was returning -1

ericcurtin:
Now the function will never return -1

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -801,11 +801,9 @@ def _run(parser):
         default=config.get('ctx_size', 2048),
         help="size of the prompt context (0 = loaded from model)",
     )
-    parser.add_argument("--device",
-        dest="device",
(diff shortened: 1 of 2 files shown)


rhatdan:
How about:
```
_engine = -1
def container_manager():
    global _engine
    def get_engine():
        engine = os.getenv("RAMALAMA_CONTAINER_ENGINE")
        if engine:
            return engine
        if available("podman") and (sys.platform != "darwin" or is_podman_machine_running_with_krunkit()):
            return "podman"
        if available("docker") and sys.platform != "darwin":
            return docker
        return None

    if _engine != -1:
        return _engine
    _engine=get_engine()
    return _engine
```

ericcurtin:
I must learn to push my commit with no elif!

ericcurtin:
re-pushed without nested functions

PR #819, opened 2025-02-14
Add entrypoint container images

rhatdan:
Install podman-remote and ramalama so we can use ramalama
from within a container.

$ podman run --env XDG_RUNTIME_DIR -v $HOME:$HOME -v /run/user:/run/user --userns=keep-id -ti --privileged quay.io/ramalama/ramalama ramalama run granite

### DIFF
--- a/container-images/scripts/build_llama_and_whisper.sh
+++ b/container-images/scripts/build_llama_and_whisper.sh
@@ -15,7 +15,7 @@ dnf_install_intel_gpu() {
 }
 
 dnf_install() {
-  local rpm_list=("python3" "python3-pip" "python3-argcomplete" \
+  local rpm_list=("podman-remote" "python3" "python3-pip" "python3-argcomplete" \
                   "python3-dnf-plugin-versionlock" "gcc-c++" "cmake" "vim" \
                   "procps-ng" "git" "dnf-plugins-core" "libcurl-devel")
   local vulkan_rpms=("vulkan-headers" "vulkan-loader-devel" "vulkan-tools" \
@@ -122,6 +122,17 @@ clone_and_build_llama_cpp() {
   rm -rf llama.cpp
 }
 
+clone_and_build_ramalama() {
+  # link podman-remote to podman for use by RamaLama
+  ln -sf /usr/bin/podman-remote /usr/bin/podman
+  git clone https://github.com/containers/ramalama
+  cd ramalama
+  git submodule update --init --recursive
+  pip install . --prefix=/usr
+  cd ..
+  rm -rf ramalama
+}
+
 main() {
   set -ex
 
@@ -134,6 +145,9 @@ main() {
   configure_common_flags
   common_flags+=("-DGGML_CCACHE=OFF" "-DCMAKE_INSTALL_PREFIX=$install_prefix")
   available dnf && dnf_install
+  if [ -n "$containerfile" ]; then 
+      clone_and_build_ramalama
+  fi
   clone_and_build_whisper_cpp
   common_flags+=("-DLLAMA_CURL=ON")
   case "$containerfile" in
--- a/container_build.sh
+++ b/container_build.sh
@@ -37,6 +37,22 @@ rm_container_image() {
   fi
 }
 
+add_entrypoint() {
+    containerfile=$(mktemp)
+    cat > ${containerfile} <<EOF
+FROM $2
+ENTRYPOINT [ "/usr/bin/$3.sh" ]
+EOF
+    echo $1 build --no-cache -t "$2-$3" -f ${containerfile} .
+    eval $1 build --no-cache -t "$2-$3" -f ${containerfile} .
+    rm ${containerfile}
+}
+
+add_entrypoints() {
+    add_entrypoint "$1" "$2" "whisper-server"
+    add_entrypoint "$1" "$2" "llama-server"
+}
+
 build() {
   cd "container-images"
   local image_name="${1//container-images\//}"
@@ -55,6 +71,7 @@ build() {
       echo "${conman_build[@]}"
       "${conman_build[@]}"
       "${conman_show_size[@]}"
+      add_entrypoints "${conman[@]}" "$REGISTRY_PATH"/"$image_name"
       rm_container_image
       ;;
     push)
@@ -62,6 +79,7 @@ build() {
       ;;
     multi-arch)
       podman farm build -t "$REGISTRY_PATH"/"$image_name" -f "$image_name"/Containerfile .
+      add_entrypoints "podman farm" "$REGISTRY_PATH"/"$image_name"
       ;;
     *)
       echo "Invalid command: ${2:-}. Use 'build', 'push' or 'multi-arch'."


rhatdan:
@benoitf PTAL

container_build.sh
```diff
@@ -37,6 +37,21 @@ rm_container_image() {
   fi
 }
 
+add_entrypoint() {
+    containerfile=$(mktemp)
+    cat > ${containerfile} <<EOF
+FROM $2
+ENTRYPOINT [ "/usr/bin/$3.sh" ]
+EOF
+    eval $1 --no-cache -t "$2-$3" -f ${containerfile} .
```
benoitf, line 46: on my end it's failing on `--no-cache` parameter not found in podman CLI

benoitf: it seems $1 $2 $3 are not the expected parameters
1 = podman
2 = build
3 = whisper-server

benoitf:
so changing the order of the arguments in the script, I was able to build an image and start it using

```
podman run --rm -it --env=HOME=/tmp --cap-drop=all --security-opt=no-new-privileges --init -v $HOME/.local/share/containers/podman-desktop/extensions-storage/redhat.ai-lab/models:/tmp/models -p 9000:9000 --device /dev/dri -e PORT=9000 -e MODEL_PATH=/tmp/models/hf.bartowski.granite-3.1-8b-instruct-GGUF/granite-3.1-8b-instruct-Q4_K_M.gguf ramalama-llama-server:latest
```

and was able to use curl to query the model and it worked 👍 


```
 curl --location 'http://localhost:9000/v1/chat/completions' \
--header 'Content-Type: application/json' \
--data '{
  "messages": [
    {
      "content": "You are a helpful assistant.",
      "role": "system"
    },
    {
      "content": "What is the capital of France?",
      "role": "user"
    }
  ]
}'
```

```
{"choices":[{"finish_reason":"stop","index":0,"message":{"content":"The capital of France is Paris.\n\....
```

rhatdan:
@benoitf This ramalama image now contains the upstream version of RamaLama along with podman-remote, so we should be able to run RamaLama in this container.

With a command like:

$ podman run --env XDG_RUNTIME_DIR -v $HOME:$HOME -v /run/user:/run/user --userns=keep-id -ti --privileged quay.io/ramalama/ramalama ramalama run granite

benoitf (approved):
tried the dedicated images 

I think there is an issue in the script (not related to this PR)

I provided `-e MODEL_CHAT_FORMAT="llama-2"` and got a core dumped about a missing --jinja

```
podman run --rm -it --env=HOME=/tmp --cap-drop=all --security-opt=no-new-privileges --init --pull=newer -v $HOME/.local/share/containers/podman-desktop/extensions-storage/redhat.ai-lab/models:/tmp/models -p 9000:9000 --device /dev/dri -e PORT=9000 -e HOST=0.0.0.0 -e MODEL_PATH=/tmp/models/hf.bartowski.granite-3.1-8b-instruct-GGUF/granite-3.1-8b-instruct-Q4_K_M.gguf -e MODEL_CHAT_FORMAT="llama-2" quay.io/ramalama/ramalama-llama-server:latest
terminate called after throwing an instance of 'std::runtime_error'
  what():  error: the supplied chat template is not supported: llama-2
note: llama.cpp was started without --jinja, we only support commonly used templates

/usr/bin/llama-server.sh: line 14:     3 Aborted                 (core dumped) llama-server --model ${MODEL_PATH} --host ${HOST:=0.0.0.0} --port ${PORT:=8001} --gpu_layers ${GPU_LAYERS:=0} ${CHAT_FORMAT}
```

it seems that llama.cpp server accept "llama2" and crash with "llama-2"

but llamacpp python is using `llama-2`
https://github.com/abetlen/llama-cpp-python?tab=readme-ov-file#chat-completion

I'm wondering if we could for example replace "llama-2" by "llama2" in the llama-sever script to have a better backward compatibility

container-images/scripts/build_llama_and_whisper.sh
```diff
@@ -122,6 +122,17 @@ clone_and_build_llama_cpp() {
   rm -rf llama.cpp
 }
 
+clone_and_build_ramalama() {
+  # link podman-remote to podman for use by RamaLama
+  ln -sf /usr/bin/podman-remote /usr/bin/podman
+  git clone https://github.com/containers/ramalama
+  cd ramalama
```
ericcurtin, line 129: How about a pip install? At least it wouldn't be installing from main branch, a more unstable target...

benoitf: could it just copy the files from where we are.
Like if I change some scripts I would assume it's embedding the files from my current working directory (can be my changes, the tag, etc) rather than the latest stable or main branch

Also if I checkout 2 months after it would use the same files (the one stored in the current working copy)

rhatdan: We have a chicken and egg situation:

Currently we are building the images at the same time as the release.  Juggling the two at the same time is going to be difficult.

When I build the 0.7.0 image, I usually do this before cutting the release.  Installing the older version say 0.6.0 into the 0.7.0 image seems wrong.  I don't think that is the proper solution.

Building a users home checked out version is also very risky, because that version may never be released to github period.  We could change the scripts to allow it to install the current version if this is really necessary, but I don't think it is, and users could hack their scripts in themselves.

benoitf: I've thought about it over the week-end and my conclusion would be that there should be a dedicated image with only the CLI (no any inference implementation like llama.cpp).

ramalama-cli image would be the image for the python code of RamaLama that will the run containers (so with podman-remote packaged inside this image)

And I would not expect to have podman-remote or any python code/dependency in any of the ramalama image that provides llama.cpp

And then the RamaLama cli should also be a smaller image (and the other as well)


ramalama-cli (python code + podman-remote)--> start containers like llama.cpp or whisper.cpp or vLLM images (that only contain the runner)



> Building a users home checked out version is also very risky, because that version may never be released to github period

if I do 'make images' I expect that it builds a specific tag like `:next`  or `:dev` or another one from my local copy.
Then I can test the images locally.

The CI job would publish only for tags or from main branch code so I don't see any issues

ericcurtin: > I've thought about it over the week-end and my conclusion would be that there should be a dedicated image with only the CLI (no any inference implementation like llama.cpp).
> 
> ramalama-cli image would be the image for the python code of RamaLama that will the run containers (so with podman-remote packaged inside this image)
> 
> And I would not expect to have podman-remote or any python code/dependency in any of the ramalama image that provides llama.cpp
> 
> And then the RamaLama cli should also be a smaller image (and the other as well)
> 
> ramalama-cli (python code + podman-remote)--> start containers like llama.cpp or whisper.cpp or vLLM images (that only contain the runner)

This almost works, except for on Windows.

> 
> > Building a users home checked out version is also very risky, because that version may never be released to github period
> 
> if I do 'make images' I expect that it builds a specific tag like `:next` or `:dev` or another one from my local copy. Then I can test the images locally.
> 
> The CI job would publish only for tags or from main branch code so I don't see any issues

ericcurtin: > We have a chicken and egg situation:
> 

I think we solve the chicken and egg situation, if we do something like:

COPY ../../* /ramalama/

(maybe even a create an .rpm/tarball during the build script, copy it, install it).

git clone main branch, could be in any sort of random state.

> Currently we are building the images at the same time as the release. Juggling the two at the same time is going to be difficult.
> 
> When I build the 0.7.0 image, I usually do this before cutting the release. Installing the older version say 0.6.0 into the 0.7.0 image seems wrong. I don't think that is the proper solution.
> 
> Building a users home checked out version is also very risky, because that version may never be released to github period. We could change the scripts to allow it to install the current version if this is really necessary, but I don't think it is, and users could hack their scripts in themselves.

rhatdan:
I don't think it is worth managing another Image for adding RamaLama to the base image.  Their is a maintainence issue and a pain for every user who wants to use RamaLama within a container.  Lets get this merged and then we can debate ways to get the current RamaLama into the container at Runtime.

Adding a -v .:/run/ramalama might be enough.  But I would like to get this in for this weeks release.

PR #829, opened 2025-02-16
Changes to "apt" command syntax in install.sh

christopherpaquin:
Original install.sh had incorrect syntax as shown below

`apt_get_install`

Changed to 
`apt install`

Had a `-y `in my changes orginally, but removed as that would actually install the package, and the offending line was only a "test if true"

NOTE: `apt-get install `would also be correct, but `apt install` is newer and not many still use `apt-get`

## Summary by Sourcery

Bug Fixes:
- Fix the syntax of the `apt` command used for installing packages in `install.sh` script. The incorrect `apt_get_install` and `apt-get install` was replaced with `apt install`

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -43,9 +43,9 @@ apt_install() {
     $sudo apt-get update || true
 
     # only install docker if podman can't be
-    if ! $sudo apt install -y podman; then
+    if ! $sudo apt install podman; then
       if ! available docker; then
-        $sudo apt install -y docker || true
+        $sudo apt install docker || true
       fi
     fi
   fi
--- a/install.sh
+++ b/install.sh
@@ -43,9 +43,9 @@ apt_install() {
     $sudo apt-get update -qq || true
 
     # only install docker if podman can't be
-    if ! $sudo apt_get_install podman; then
+    if ! $sudo apt install -y podman; then
       if ! available docker; then
-        $sudo apt_get_install docker || true
+        $sudo apt install -y docker || true
       fi
     fi
   fi


install.sh
```diff
@@ -43,9 +43,9 @@ apt_install() {
     $sudo apt-get update || true
 
     # only install docker if podman can't be
-    if ! $sudo apt_get_install podman; then
+    if ! $sudo apt install podman; then
```
engelmi, line 46: `apt_get_install` is calling the function in line 37 to 39, so it should be changed there (and probably renamed).

engelmi:
Please sign your commit via `git commit -s` for the DCO

install.sh
ericcurtin: apt-get update should probably be changed to it's apt equivalent to be consistent also

ericcurtin: Also this should change:

available apt-get

to

available apt

ericcurtin: We do want -y 's in this script btw, the check to see if these things are installed in the "available" check

christopherpaquin:
From what I am reading I should make the following changes. Please confirm
This includes
1. Renaming the apt install check available function to differ from the apt-get install function
2. Renaming the apt-get install function to apt_install to track with the changes to deprecate usage of apt-get

```
37 apt_install() {
 38   apt_install_available"$1"
 39 }
 40  
 41 apt_install_available() {
 42   if ! available podman; then
 43     $sudo apt update || true
 44  
 45     # only install docker if podman can't be
 46     if ! $sudo apt install podman; then
 47       if ! available docker; then
 48         $sudo apt install docker || true
 49       fi
 50     fi
 51   fi
 52 }

```
Should I make my changes locally and then submit another pull request?

I can sign with `git commit -s.` What is DCO?

ericcurtin:
> From what I am reading I should make the following changes. Please confirm This includes
> 
> 1. Renaming the apt install check available function to differ from the apt-get install function
> 2. Renaming the apt-get install function to apt_install to track with the changes to deprecate usage of apt-get
> 
> ```
> 37 apt_install() {
>  38   apt_install_available"$1"
>  39 }
>  40  
>  41 apt_install_available() {
>  42   if ! available podman; then
>  43     $sudo apt update || true
>  44  
>  45     # only install docker if podman can't be
>  46     if ! $sudo apt install podman; then
>  47       if ! available docker; then
>  48         $sudo apt install docker || true
>  49       fi
>  50     fi
>  51   fi
>  52 }
> ```

Almost, it should be `apt install -y` rather than `apt install`

> 
> Should I make my changes locally and then submit another pull request?
> 
> I can sign with `git commit -s.` What is DCO?

DCO is the build that checks someone has signed their commits. It's a requirement for all projects in the:

https://github.com/containers

namespace.

christopherpaquin:
Closing as there is a dupe. https://github.com/containers/ramalama/pull/832

Issue #823, opened 2025-02-15
port forwarding doesn't work in latest commit b8e779


tuananh:
i have problem with shutting down ramalama with 0.5.0 and 0.6.0 but works fine in this latest commit b8e779c

however, i have a new issue where i can't access port 8080 from host. the logs looks fine to me

`ramalama serve ollama://deepseek-r1:32b`

```
llama_kv_cache_init:        CPU KV buffer size =   512.00 MiB
llama_init_from_model: KV self size  =  512.00 MiB, K (f16):  256.00 MiB, V (f16):  256.00 MiB
llama_init_from_model:        CPU  output buffer size =     0.58 MiB
llama_init_from_model:        CPU compute buffer size =   307.00 MiB                          llama_init_from_model: graph nodes  = 2246
llama_init_from_model: graph splits = 1
common_init_from_params: setting dry_penalty_last_n to ctx_size = 2048
common_init_from_params: warming up the model with an empty run - please wait ... (--no-warmup to disable)
srv          init: initializing slots, n_slots = 1                                            slot         init: id  0 | task -1 | new slot n_ctx_slot = 2048
[...]

PR #830, opened 2025-02-16
Make serve by default expose network

ericcurtin:
But the default for the rest should be to encapsulate the network in the container. Also de-deduplicate code.

## Summary by Sourcery

Introduce a helper function to manage network arguments and change the default network setting for the serve command to expose the network.

Enhancements:
- Decouple network settings by introducing a shared function to handle network arguments across different subcommands.
- Expose the network by default for the `serve` subcommand, while other subcommands default to isolating the container network.

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -435,14 +435,23 @@ def bench_cli(args):
     model.bench(args)
 
 
-def bench_parser(subparsers):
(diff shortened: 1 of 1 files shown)


rhatdan:
Please add tests for this.  Simply do --dryrun and make sure --default=none or --default="" for most commands is on command line for serve.

tuananh:
Confirm this PR fix #823 for me.

PR #831, opened 2025-02-16
Make CI build all images

ericcurtin:
To ensure they all continue to build and remain of reasonable size.

## Summary by Sourcery

CI:
- Rebuild all container images in the CI workflow using a new script, `container_build.sh -r build`

### DIFF
--- a/.github/workflows/ci-images.yml
+++ b/.github/workflows/ci-images.yml
@@ -7,11 +7,13 @@ on:
   pull_request:
     paths:
       - "**/container-images/**"
+      - ".github/workflows/ci-images.yml"
   push:
     branches:
       - main
     paths:
       - "**/container-images/**"
+      - ".github/workflows/ci-images.yml"
 
 jobs:
   build:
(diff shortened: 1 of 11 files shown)


ericcurtin:
@rhatdan pragmatic doesn't seem to be building... But we can just leave it untouched and skip it in the build if needs be?

rhatdan:
Lets skip pragmatic for now, since this is in major flux at the moment.

ericcurtin:
> Lets skip pragmatic for now, since this is in major flux at the moment.

I actually got it green in the end, but no problem skipping for now. Everything has a maintenance cost.

ericcurtin:
@rhatdan this is good to go. It turns on various container image builds again. Skipping ROCm I need to split UBI9 ROCm like @maxamillion 's PR. Trimming is a pain and the trimming is just removing GPU support anyway which people will complain about.

There's also a lot of shellcheck fixes in here, hope they are all fine. Turning on shellcheck again for all shell scripts.

PR #840, opened 2025-02-17
Some tests around --network, --net options

ericcurtin:
For run and serve commands.

## Summary by Sourcery

Tests:
- Added system tests to verify the behavior of the `--network` and `--net` options with the `run` and `serve` commands.


rhatdan (approved):
LGTM

PR #847, opened 2025-02-18
Just one add_argument call for --dryrun/--dry-run

ericcurtin:
Just alias a single function call, rather than calling the function twice

## Summary by Sourcery

Enhancements:
- The `--dry-run` argument is now an alias for `--dryrun`, instead of being a separate argument.


rhatdan:
LGTM

PR #856, opened 2025-02-19
Fix up handling of image selection on generate

rhatdan:
Also fall back to trying OCI images on ramalama run and serve.

## Summary by Sourcery

This pull request modifies the application to improve image selection and handling, particularly when using `run` and `serve` commands. It introduces a fallback mechanism to attempt using OCI images if the initial image type fails. Additionally, it refactors the `Quadlet` and `Kube` classes to receive the image as a parameter, ensuring correct image usage in container configurations.

Bug Fixes:
- Fixes an issue where the application would not attempt to use OCI images when the initially specified image type failed, specifically in `run` and `serve` commands.

Enhancements:
- Modifies the `Quadlet` and `Kube` classes to receive the image as a parameter during initialization, ensuring the correct image is used for container configuration.
- Improves image handling by ensuring the correct image is used when generating container configurations.

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -838,8 +838,17 @@ def run_parser(subparsers):
 
 
 def run_cli(args):
-    model = New(args.MODEL, args)
-    model.run(args)
+    try:
+        model = New(args.MODEL, args)
+        model.run(args)
+
+    except KeyError as e:
+        try:
+            args.quiet = True
+            model = OCI(args.MODEL, args.engine, ignore_stderr=True)
+            model.run(args)
+        except Exception:
+            raise e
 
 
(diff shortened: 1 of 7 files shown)


ericcurtin (approved):
Haven't looked at why the tests are failing or anything but the code looks fine

rhatdan:
The issue is an older version of podman, I believe in Ubuntu.

ericcurtin:
> The issue is an older version of podman, I believe in Ubuntu.

There's a super hacky "Upgrade to podman 5" stage in at least some of our builds. I think it was installing Ubuntu 24.10 podman on Ubuntu 24.04

PR #863, opened 2025-02-20
Fix macOS GPU acceleration via podman

ericcurtin:
We should always have acceleration on when running on macOS. There
is a possible case where one may not want to use acceleration on
macOS. If someone is hell-bent on using podman without krunkit on
macOS. But for now, just turn it on regardless.


benoitf (approved):
image

now it works (it consumes my GPU) 🚀

PR #866, opened 2025-02-21
Added support for apple hypervisor support on macos

bmahabirbu:
Slightly reworked the way krunkit is being detected so ramalama can use apple hyper v (the default apple vm for podman desktop) as well. 

Even though this is a niche use case I think its useful if the default VM can spin up ramalama too!

## Summary by Sourcery

This pull request enables ramalama to run on macOS using the Apple hypervisor (the default VM for Podman Desktop) by adjusting krunkit detection logic. It also ensures that GPU acceleration is enabled on Apple Silicon machines.

New Features:
- Adds support for the Apple hypervisor in macOS environments when using Podman Desktop.

Enhancements:
- Refactors the logic for detecting krunkit to enable its usage with the Apple hypervisor.

### DIFF
--- a/ramalama/common.py
+++ b/ramalama/common.py
@@ -36,7 +36,7 @@ def get_engine():
     if engine is not None:
         return engine
 
-    if available("podman") and (sys.platform != "darwin" or is_podman_machine_running_with_krunkit()):
+    if available("podman"):
         return "podman"
 
     if available("docker") and sys.platform != "darwin":
@@ -55,7 +55,7 @@ def container_manager():
     return _engine
 
 
-def is_podman_machine_running_with_krunkit():
+def krunkit():
     podman_machine_list = ["podman", "machine", "list", "--all-providers"]
     try:
         output = run_cmd(podman_machine_list, ignore_stderr=True).stdout.decode("utf-8").strip()
--- a/ramalama/model.py
+++ b/ramalama/model.py
@@ -11,6 +11,7 @@ from ramalama.common import (
     genname,
     get_env_vars,
     get_gpu,
+    krunkit,
     run_cmd,
 )
 from ramalama.gguf_parser import GGUFInfoParser
@@ -211,7 +212,7 @@ class Model:
             for device_arg in args.device:
                 conman_args += ["--device", device_arg]
         else:
-            if (sys.platform == "darwin" and os.path.basename(args.engine) != "docker") or os.path.exists("/dev/dri"):
+            if os.path.exists("/dev/dri") or krunkit():
                 conman_args += ["--device", "/dev/dri"]
 
             if os.path.exists("/dev/kfd"):
@@ -237,8 +238,8 @@ class Model:
             or (
                 # linux and macOS report aarch64 differently, on Apple Silicon
                 # (arm64), we should have acceleration on regardless.
-                machine == "arm64"
-                or (machine == "aarch64" and os.path.exists("/dev/dri"))
+                machine
+                in {"arm64", "aarch64"}
             )
         ):
             if args.ngl < 0:


ramalama/common.py
```diff
@@ -36,7 +36,7 @@ def get_engine():
     if engine is not None:
         return engine
 
-    if available("podman") and (sys.platform != "darwin" or is_podman_machine_running_with_krunkit()):
+    if available("podman"):
```
rhatdan, line 39: This will only check if podman is available and not whether the podman machine is running.  The basic idea of this check was the GPU passtrhough is not happening unless krunkit is running in a podman machine.  If not then we default to running the AI model locally.

rhatdan:
The apple hypervisor currently does not support GPU passthrough.

ramalama/model.py
```diff
@@ -237,8 +238,8 @@ def gpu_args(self, args, runner=False):
             or (
                 # linux and macOS report aarch64 differently, on Apple Silicon
                 # (arm64), we should have acceleration on regardless.
-                machine == "arm64"
-                or (machine == "aarch64" and os.path.exists("/dev/dri"))
+                machine
+                in {"arm64", "aarch64"}
```
ericcurtin, line 242: The separate check for "arm64" is deliberate, it means we are on Apple Silicon and macOS. The "/dev/dri" check is also deliberate, some ARM machines on Linux (a VM is one example) don't have access to graphics.

ericcurtin: If -ngl 999 works on Linux machines without /dev/dri for "ramalama run" and "ramalama serve" we could do this, but we'd have to check that to be sure.

ramalama/common.py
bmahabirbu: I added the krunkit check later in model.py in around line 215 when checking for the existance of dev/dri. My thought process was that if macos has podman installed (via brew or podman desktop etc) it should move through the process as usual

Apple hyperv doest have dev/dri but it still can spin up ramalama without gpu.

ramalama/model.py
bmahabirbu: We also do the check for dev/dri at around line 215 in model.py so I was a bit confused at the check here but your right in this case it will executive with ngl 999 without explicit gpu support.

 In the past ngl is ignored if you don't have gpu but this shouldn't be the way we do it

ramalama/common.py
ericcurtin: It's tricky, presence of /dev/dri doesn't guarantee GPU acceleration. It means something capable of DRI exists. It could be software graphics like simpledrm as an example. In a simpledrm case as an example, you don't want to go via GPU, as it's software graphics which would even be slower than CPU.

One of the reasons we check for that, is if you pass --device /dev/dri and it doesn't exist, podman will fail. It's a hint that we might have a GPU capable of running an AI workload, but only a hint. Another example, integrated graphics on many machines perform worse than CPU and only have 512M VRAM which isn't large enough to fit a decently sized model.

bmahabirbu:
Makes total sense! My original thought process was If we want the user to have gpu on macos they should explicitly turn on krunkit support on podman desktop and have that vm be the default. But if using the default hyper v vm, instead of defaulting to running without a container it would at least spin up without gpu support.

I can see how defaulting to no container can better influence someone to turn on krunkit and make it the default in Podman desktop. This brings into discussion that podman desktop should default to use krunkit for the end user and theres no extra steps.

ericcurtin:
> Makes total sense! My original thought process was If we want the user to have gpu on macos they should explicitly turn on krunkit support on podman desktop and have that vm be the default. But if using the default hyper v vm, instead of defaulting to running without a container it would at least spin up without gpu support.
> 
> I can see how defaulting to no container can better influence someone to turn on krunkit and make it the default in Podman desktop. This brings into discussion that podman desktop should default to use krunkit for the end user and theres no extra steps.

I think it's planned FWIW, it's just a waiting game...

bmahabirbu:
Awesome! Sounds good I'll close this PR then! If we end up getting some request for use with mac cpu only I can try this again!

Issue #869, opened 2025-02-23
ramalama list with json returns formatted value


benoitf:
when I use `ramama list` it displays the size with MB/GB formatting, etc (human size)

but using json formatting, I expect to see the raw value to have tools parsing the value

but today `--json` is returning the human size display

PR #870, opened 2025-02-23
chore: do not format size for --json export in list command

benoitf:
fixes https://github.com/containers/ramalama/issues/869

## Summary by Sourcery

This pull request fixes an issue where the size of models was being formatted when exporting in JSON format. The size is now only formatted for human readability when printing to the console.

Bug Fixes:
- Fixes an issue where the size of models was being formatted when exporting in JSON format, which is not desired.

Enhancements:
- The size of the model is now formatted for human readability only when printing to the console, and not when exporting to JSON.

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -533,7 +533,7 @@ def human_readable_size(size):
 
 
 def get_size(file):
-    return human_readable_size(os.path.getsize(file))
+    return os.path.getsize(file)
 
 
 def _list_models(args):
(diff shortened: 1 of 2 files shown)


benoitf:
discovered that the OCI data were formatted by podman CLI and not by ramalama CLI
so, ensure we send the raw numbers from OCI

test/system/010-list.bats
```diff
@@ -23,22 +23,20 @@ load helpers
...
-    #FIXME jq version on mac does not like regex handling
-    skip_if_darwin
     # 'created': ramalama includes fractional seconds, ramalama-remote does not
     tests="
-name              | [a-z0-9A-Z:/]\\\+
-modified          | [0-9]\\\+
-size              | [0-9]\\\+
+name              | test(\"^[a-z0-9A-Z:/]\\+\")
+modified          | tostring | test(\"^[a-z0-9A-Z:/]\\+\")
+size              | tostring | test(\"^[0-9]\\+\$\")
```
benoitf, line 30: it checks also now with a $ at the end to ensure the size is only a digit and not '123 MB' or '4 GB'
else the test was passing

rhatdan:
LGTM

PR #877, opened 2025-02-24
[WIP] Add support for kserve

rhatdan:


ramalama/kserve.py
```diff
@@ -0,0 +1,105 @@
...
+                f"""\
+apiVersion: serving.kserve.io/v1alpha1
+kind: ServingRuntime
+metadata:
+  name: {self.runtime}-runtime
+  annotations:
+    openshift.io/display-name: KServe ServingRuntime for {self.model}
+    opendatahub.io/recommended-accelerators: '["{_gpu}"]'
+  labels:
+    opendatahub.io/dashboard: 'true'
```
danielezonca, line 50:
```suggestion
```
Let's remove them for now, they are openshift/openshift ai specific.
Sorry I should have excluded them in the example I have shared

ramalama/kserve.py
```diff
@@ -0,0 +1,105 @@
...
+  annotations:
+    prometheus.io/port: '{self.args.port}'
+    prometheus.io/path: '/metrics'
+  multiModel: false
+  supportedModelFormats:
+    - autoSelect: true
+      name: vLLM
+  containers:
+    - name: kserve-container
+      image: {self.image}
```
danielezonca, line 61: This code looks wrong, as far as I understand checking the example code, it will return ramalama as image and not vLLM.
Moreover vLLM is specific per accelerator so the same if/else condition to produce the gpu requirements should be applied here to select the right image

ramalama/kserve.py
```diff
@@ -0,0 +1,105 @@
...
+        outfile = self.name + "-kserve-runtime.yaml"
+        outfile = outfile.replace(":", "-")
+        print(f"Generating kserve runtime file: {outfile}")
+        with open(outfile, 'w') as c:
+            c.write(
+                f"""\
+apiVersion: serving.kserve.io/v1alpha1
+kind: ServingRuntime
+metadata:
+  name: {self.runtime}-runtime
```
danielezonca, line 45: I think this is misleading, it produces a `llama.cpp-runtime` value but this is supposed to be vLLM

docs/ramalama-serve.1.md
```diff
@@ -60,8 +60,9 @@ Generate specified configuration format for running the AI Model as a service
 
 | Key          | Description                                                              |
 | ------------ | -------------------------------------------------------------------------|
-| quadlet      | Podman supported container definition for running AI Model under systemd |
+| kserve       | Kserve YAML definition for running the AI Model as a kserve service in Kubernetes        |
```
danielezonca, line 63:
```suggestion
| kserve       | Kserve YAML definition for running the AI Model as a kserve service in Kubernetes using vLLM        |
```

ramalama/kserve.py
rhatdan: You would need to specify --runtime vllm and then you get the correct output.  Is kserve totally tied to vllm?  Or is there ability to support multiple runtimes?

danielezonca: KServe is multi runtime, there is a list of runtime included OOTB but it is also dynamic, you can bring your own runtime

rhatdan: Link to the list? vllm and llama.cpp included?

rhatdan: https://kserve.github.io/website/0.10/modelserving/v1beta1/serving_runtime/

This looks like something different and we would need to hard code  the most common one in the tool?  Which is that, does not look like kserve gets down to the llama.cpp or vllm level?

ramalama/kserve.py
```diff
@@ -0,0 +1,105 @@
...
+            _gpu = 'nvidia.com/gpu'
+        elif os.getenv("HIP_VISIBLE_DEVICES") != "":
+            _gpu = 'amd.com/gpu'
+        if _gpu != "":
+            gpu = f'\n          {_gpu}: "1"'
+
+        outfile = self.name + "-kserve-runtime.yaml"
+        outfile = outfile.replace(":", "-")
+        print(f"Generating kserve runtime file: {outfile}")
+        with open(outfile, 'w') as c:
```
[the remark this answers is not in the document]
afazekas: IMHO: jinja2 is not fast, maybe a more readable `f"" string template`way.

docs/ramalama-serve.1.md
```diff
@@ -187,6 +188,64 @@ CONTAINER ID  IMAGE                             COMMAND               CREATED
...
+```
+$ ramalama serve --pull=never --threads 10 --port 8081 --generate kserve oci://quay.io/rhatdan/granite
+Generating kserve runtime file: granite-cuda-kserve-runtime.yaml
+Generating kserve file: granite-cuda-kserve.yaml
+
+$  cat granite-cuda-kserve-runtime.yaml
+apiVersion: serving.kserve.io/v1alpha1
+kind: ServingRuntime
+metadata:
+  name: llama.cpp-cuda-runtime
```
terrytangyuan, line 201: Should this be granite instead of llama?

docs/ramalama-serve.1.md
```diff
@@ -187,6 +188,64 @@ CONTAINER ID  IMAGE                             COMMAND               CREATED
...
+  annotations:
+    prometheus.io/port: '8081'
+    prometheus.io/path: '/metrics'
+  multiModel: false
+  supportedModelFormats:
+    - autoSelect: true
+      name: vLLM
+  containers:
+    - name: kserve-container
+      image: quay.io/ramalama/cuda:0.8
```
terrytangyuan, line 216: Is this the correct image? I see that CPUs are used in the generated InferenceService

docs/ramalama-serve.1.md
```diff
@@ -187,6 +188,64 @@ CONTAINER ID  IMAGE                             COMMAND               CREATED
 3f64927f11a5  quay.io/ramalama/ramalama:latest  /usr/bin/ramalama...  17 seconds ago  Up 17 seconds  0.0.0.0:8082->8082/tcp  ramalama_YMPQvJxN97
 ```
 
+### Generate kserve service off of OCI Model car quay.io/ramalama/granite:1.0
+```
+$ ramalama serve --pull=never --threads 10 --port 8081 --generate kserve oci://quay.io/rhatdan/granite
```
terrytangyuan, line 193: Is Ramalama only designed for single-node serving?

docs/ramalama-serve.1.md
rhatdan: That is the runtime. But maybe granite makes more sense.

docs/ramalama-serve.1.md
rhatdan: What should the InferenceService look like if it was using nvidia/cuda?

docs/ramalama-serve.1.md
rhatdan: RamaLama is just a tool to launch AI models in containers, the idea would be to generate kubernetes content, to allow models to run cross multi-nodes, but not something ramalama would do from command line.

danielezonca: I think we can consider this for another iteration: given that ramalama is mainly designed for local I don't expect a user will run a model big enough to require multi node GPU setup

docs/ramalama-serve.1.md
danielezonca: The official community image for vLLM is on DockerHub and the cuda image is 
`vllm/vllm-openai:${version}`
like 
`vllm/vllm-openai:v0.8.5`
and I think we should use them as starting point (it is CUDA only)

I suggest to consider this as additional parameter like `--vllm-version` 

Note 1:
We have a [quay.io](https://quay.io/repository/vllm/vllm?tab=tags) build that is like `quay.io/vllm/vllm:0.8.4.20250429` but it is still in progress

Note 2:
RHOAI product builds use a different org and the images are like
`quay.io/modh/vllm:rhoai-2.19-cuda` / `quay.io/modh/vllm:rhoai-2.19-rocm`
but I don't think we want to use product images here

docs/ramalama-serve.1.md
```diff
@@ -187,6 +188,64 @@ CONTAINER ID  IMAGE                             COMMAND               CREATED
...
+      modelFormat:
+        name: vLLM
+      storageUri: "oci://quay.io/rhatdan/granite"
+      resources:
+        limits:
+          cpu: "10"
+          memory: 24Gi
+        requests:
+          cpu: "10"
+          memory: 24Gi
```
danielezonca, line 246: We need at least the GPU required too
```suggestion
        limits:
          cpu: "10"
          memory: 24Gi
          nvidia.com/gpu: '1'
        requests:
          cpu: "10"
          memory: 24Gi
          nvidia.com/gpu: '1'
```

PR #878, opened 2025-02-24
Check if terminal is compatible with emojis before using them

ericcurtin:
Just in case it doesn't.

## Summary by Sourcery

Enhancements:
- The application now checks if the terminal supports emojis before using them in the prompt prefix.


benoitf (approved):
thanks @ericcurtin I was unaware of that constant

I checked on macOS and it's working as expected (with or without containers)

PR #879, opened 2025-02-24
The package available via dnf is in a good place

ericcurtin:
Defaulting to that on platforms that have dnf, if it fails for whatever reason, fall back to this script.

## Summary by Sourcery

This pull request modifies the installation script to first attempt to install `ramalama` using `dnf` if it is available. If `dnf` is not available or the installation fails, the script falls back to the existing installation method. The installation process is also refactored into separate functions for installing the binary and libraries.

New Features:
- Adds support for installing ramalama via the `dnf` package manager on systems where it is available.

Enhancements:
- Refactors the installation process into separate functions for installing the binary and libraries.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -29,8 +29,12 @@ download() {
 }
 
 dnf_install() {
+  $sudo dnf install -y "$1"
+}
+
+dnf_install_podman() {
   if ! available podman; then
-    $sudo dnf install -y --best podman || true
+    dnf_install podman || true
   fi
 }
 
@@ -87,7 +91,7 @@ check_platform() {
     fi
 
     if available dnf && ! grep -q ostree= /proc/cmdline; then
-      dnf_install
+      dnf_install_podman
     elif available apt; then
       apt_update_install
     fi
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -147,30 +155,44 @@ setup_ramalama() {
...
+    get_installation_dir
+    return 0
+  fi
 
   local os
   os="$(uname -s)"
   local sudo=""
   check_platform
 
+  if available dnf; then
```
engelmi, line 190: Via dnf we'd get a versioned `ramalama` installation and with `setup_ramalama` we'd get the latest from main. So would it make sense to enable switching between these two options, e.g. with `--use-dnf`?

ericcurtin: This installs from the "s" branch by default which I push roughly whenever we do a release. I guess it would be nice to have alternate install techniques work even on systems that can install via dnf.

So I added checks to see if someone explicitly tried to specify a BRANCH or local install.

But we do want to install by dnf here by default. AFAIK this is the only install technique going to appear on the ramalama.ai website. And we want people on platforms like Fedora and RHEL to use the rpm version.

ericcurtin: This was always kinda meant as just a thrown together thing until the various distros package RamaLama properly. It will probably stay around forever because yeah it's nice to be able to install direct from upstream also like you say.

PR #880, opened 2025-02-25
Use vllm-openai upstream image

ericcurtin:
The one we are currently using is old and doesn't have .gguf compatibility.

## Summary by Sourcery

Enhancements:
- Update the vLLM image to the upstream vllm-openai image, providing compatibility with the .gguf format.


engelmi (approved):
LGTM

PR #881, opened 2025-02-25
Add Ollama to CI and system tests for its caching

kush-gupt:
This adds improvements to the logic for detecting existing Ollama caches and adds a pull system test to verify the cache functionality.

This also updates the GitHub CI to install Ollama.

## Summary by Sourcery

This pull request enhances the project by integrating Ollama into the CI pipeline and adding a system test to validate its caching mechanism. It also fixes a bug in the logic for detecting existing Ollama caches.

Bug Fixes:
- Improves the logic for detecting existing Ollama caches by checking if the manifest path exists and is readable.

CI:
- Configures CI to install Ollama, ensuring it's available for testing and integration.

Tests:
- Adds a system test to verify the functionality of pulling images from the Ollama cache.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,11 +56,7 @@ jobs:
 
       - name: install ollama
         shell: bash
-        run: |
-           curl -L https://ollama.com/download/ollama-linux-amd64.tgz -o ollama-linux-amd64.tgz
-           sudo tar -C /usr -xzf ollama-linux-amd64.tgz
-           sudo useradd -r -s /bin/false -U -m -d /usr/share/ollama ollama
-           sudo usermod -a -G ollama runner
+        run: ./.github/scripts/install-ollama.sh
 
       - name: Upgrade to podman 5
         run: |
@@ -95,11 +91,7 @@ jobs:
 
       - name: install ollama
         shell: bash
-        run: |
-           curl -L https://ollama.com/download/ollama-linux-amd64.tgz -o ollama-linux-amd64.tgz
-           sudo tar -C /usr -xzf ollama-linux-amd64.tgz
-           sudo useradd -r -s /bin/false -U -m -d /usr/share/ollama ollama
-           sudo usermod -a -G ollama runner
+        run: ./.github/scripts/install-ollama.sh
 
       - name: Upgrade to podman 5
         run: |
@@ -129,11 +121,7 @@ jobs:
 
       - name: install ollama
         shell: bash
-        run: |
-           curl -L https://ollama.com/download/ollama-linux-amd64.tgz -o ollama-linux-amd64.tgz
-           sudo tar -C /usr -xzf ollama-linux-amd64.tgz
-           sudo useradd -r -s /bin/false -U -m -d /usr/share/ollama ollama
-           sudo usermod -a -G ollama runner
+        run: ./.github/scripts/install-ollama.sh
 
       - name: Upgrade to podman 5
         run: |
@@ -185,9 +173,7 @@ jobs:
 
       - name: install ollama
         shell: bash
-        run: |
-           brew install ollama
-           brew services start ollama
+        run: ./.github/scripts/install-ollama.sh
 
       - name: Run a one-line script
         shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,6 +54,14 @@ jobs:
            sudo apt-get install podman bats bash codespell python3-argcomplete pipx
            make install-requirements
 
+      - name: install ollama
+        shell: bash
+        run: |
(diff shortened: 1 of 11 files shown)


swarajpande5 (approved):
The changes LGTM !

ramalama/ollama.py
```diff
@@ -68,12 +68,12 @@ def in_existing_cache(model_name, model_tag):
...
         os.path.join(os.environ['HOME'], '.ollama/models'),
         '/usr/share/ollama/.ollama/models',
-        f'C:\\Users\\{os.getlogin()}\\.ollama\\models',
+        'C:\\Users\\%username%\\.ollama\\models',
     ]
 
     for cache_dir in default_ollama_caches:
         manifest_path = os.path.join(cache_dir, 'manifests', 'registry.ollama.ai', model_name, model_tag)
-        if os.path.exists(manifest_path):
+        if os.path.exists(manifest_path) and os.access(manifest_path, os.R_OK):
```
engelmi, line 76: NIT: 
```suggestion
        if os.access(manifest_path, os.R_OK):
```

.github/workflows/ci.yml
```diff
@@ -54,6 +54,14 @@ jobs:
            sudo apt-get install podman bats bash codespell python3-argcomplete pipx
            make install-requirements
 
+      - name: install ollama
+        shell: bash
+        run: |
+           curl -L https://ollama.com/download/ollama-linux-amd64.tgz -o ollama-linux-amd64.tgz
+           sudo tar -C /usr -xzf ollama-linux-amd64.tgz
+           sudo useradd -r -s /bin/false -U -m -d /usr/share/ollama ollama
+           sudo usermod -a -G ollama runner
```
engelmi, line 63: I think we can move the installation to a install script and/or as make target. This way, we reduce the duplication here and we enable/simplify local execution of the ollama cache test.

test/system/050-pull.bats
```diff
@@ -29,6 +29,37 @@ load setup_suite
     is "$output" "Error: ${random_image_name} was not found in the Ollama registry"
 }
 
+# bats test_tags=distro-integration
+@test "ramalama pull ollama cache" {
+    skip_if_no_ollama
```
engelmi, line 34: Instead of skipping if the ollama cli doesn't exist, could we only install the cli here and uninstall at the end of this test? This way we don't affect other test runs on the same host. WDYT?

ericcurtin: If we do add this, I think this is the right way to do it. Install requires root access, we don't want our tests operating as root.

Also, we want to avoid overwriting someones ollama install and mucking around with someone's system like this, say the local developer that runs this test suite.

engelmi: I agree that requiring root access for the install is bad. However, when the ollama cli is present, it alters the behavior of ramalama also for all other tests - checking for cached files - and this might lead to really tricky test failures (or worse passing tests when it shouldn't). 

> Also, we want to avoid overwriting someones ollama install and mucking around with someone's system like this, say the local developer that runs this test suite.
I would assume the automatic install would first check for an already setup ollama cli installation and if not detected install it. This way it should not mess with someones system.

kush-gupt: Yeah I was having the same mentality as Eric, didn't want to force an ollama install for a regular user running the tests locally so that's why we skip if ollama commands don't exist.

If there's a better place for the Ollama install to happen, I can move it! 

Installing at CI time was just my first idea, and could emulate a good amount of the users running Ramalama (since they're into local LLMs, the majority may have tried/installed Ollama already)

engelmi: Let's keep the skip in the bat :)

@kush-gupt What do you think about creating `.github/scripts/install-ollama.sh` which reduces the 4 lines in the CI to a single one? It would need to detect the host (macos or ubuntu). Its kind of a compromise between using it only in the CI, but still having a runnable script.

kush-gupt: Can knock that out in a bit!

kush-gupt: I'll just add the `$(logname)` for easy reproducibility locally

engelmi: Nice, looks great! 

> Was thinking of replacing the hardcoded runner user with $(logname) if we expect folks to run this script outside of CI, but is that something we'd expect/want @engelmi?

Probably not, considering that the majority has already installed Ollama as you wrote.

.github/workflows/ci.yml
kush-gupt: Script at `.github/scripts/install-ollama.sh` should resolve this!

ericcurtin (approved):

rhatdan:
Nice work @kush-gupt

PR #882, opened 2025-02-25
Moved pruning protocol from model to factory

engelmi:
By moving the pruning of the protocol from the model input to the model_factory and encapsulating it in a dedicated function, unit tests can be written more easily.

## Summary by Sourcery

Moves the protocol parsing logic from the model classes to the model factory to improve testability and code organization. It introduces a `prune_model_input` method in the `ModelFactory` class to remove the protocol from the model input string. It also adds a `validate_oci_model_input` method to validate OCI model inputs. Additionally, it includes new unit tests to validate the new logic.

Enhancements:
- The protocol parsing logic has been moved from the model classes to the model factory, and is now encapsulated in a dedicated function.

Tests:
- Added unit tests for the OCI model input validation.
- Added unit tests for the protocol pruning logic.

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -13,7 +13,6 @@ from ramalama.common import container_manager, default_image, perror, run_cmd
 from ramalama.gpu_detector import GPUDetector
 from ramalama.model import MODEL_TYPES
 from ramalama.model_factory import ModelFactory
-from ramalama.oci import OCI
 from ramalama.shortnames import Shortnames
 from ramalama.toml_parser import TOMLParser
 from ramalama.version import print_version, version
@@ -699,7 +698,7 @@ def convert_cli(args):
     if not tgt:
         tgt = target
 
-    model = OCI(tgt, args.engine)
+    model = ModelFactory(tgt, engine=args.engine).create_oci()
     model.convert(source, args)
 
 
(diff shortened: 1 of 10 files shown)


engelmi:
The `ramalama convert file to image` test fails, but I am not sure this is related to this change:
```bash
[07:52:37.127614592] $ /home/runner/work/ramalama/ramalama/bin/ramalama convert /tmp/ramalama_bats.yyeHxU/aimodel foobar
# [07:52:37.962614926] Error: invalid repository name (1b572995d33b4edde32e4b0112342f32bbd43dcbe5df317d452d962e3b15c691), cannot specify 64-byte hexadecimal strings
# Failed to create manifest for OCI foobar : Command '['podman', 'manifest', 'create', 'foobar', '1b572995d33b4edde32e4b0112342f32bbd43dcbe5df317d452d962e3b15c691']' returned non-zero exit status 125.
# Converting /tmp/ramalama_bats.yyeHxU/aimodel to foobar...
# Building foobar...
```
@ericcurtin WDYT?

ramalama/cli.py
```diff
@@ -775,7 +774,7 @@ def push_cli(args):
                 raise e
         try:
             # attempt to push as a container image
-            m = OCI(tgt, config.get('engine', container_manager()))
+            m = ModelFactory(tgt, engine=config.get('engine', container_manager())).create_oci()
```
ericcurtin, line 777: It's probably not for this PR, but it would be better if we resolved all environment variables, configuration files, CLI args, once early in execution of ramalama. I started that here at one point, but it drifted:

load_and_merge_config

we tend to do things like, use the "engine" value or container_manager() over and over again in this codebase which isn't great.

ericcurtin (approved):

ramalama/cli.py
engelmi: I agree and good to know. I'll keep that in mind and will try to refactor occurrences like this.

PR #888, opened 2025-02-26
benchmark failing because of lack of flag

ericcurtin:
Specifically priviledged because it's not present in the args object.

## Summary by Sourcery

Fixes a bug where the `bench` command was missing some arguments. This change introduces a new function `bench_run_serve_perplexity_args` to define arguments shared between `bench`, `run` and `serve` commands. The `ngl` argument was removed from the `bench` command and is now loaded from the model.


rhatdan:
Can you add a test to run bench without checking for output, to at least catch crashes.

rhatdan:
LGTM

PR #891, opened 2025-02-26
Switch from tiny to smollm:135m

ericcurtin:
This is probably a consequence of my slow network, but I switched to smollm:135m, it's easier for demos. tiny was taking too long to download.

## Summary by Sourcery

Switch the default model in the demo script from `tiny` to `smollm:135m` to improve the out-of-the-box experience for users with slower networks.


rhatdan:
LGTM

PR #898, opened 2025-02-28
Try to choose a free port on serve if default one is not available

### LINKED ISSUE #797: ramalama serve is not checking if the port is available before running the command
Being a pretty dumb user, I don't care the port where the model will be served. It can be a random port as it's being displayed

1. using 8080 as a default port is quite challenging because as a developer, I have tons of app that I start that are already using this port.
 (so picking up a port different than 8080 can increase the "first attempt succeed"

1. most of the cli launchers detect that the port is in use and is serving the app using a basic '+1' on the default port

so If I launch 2 serve commands, it won't fail (unless I specify a port number and in that case yes I expect it fails if it's already taken but without specifying a port number I will assume the tool is picking one for me that works)

here is the current error message:
```
Error: preparing container ee6f2c8cd4377db38836b25c7523147a2383bcd9d86f22db0446b15212bd5bfb for attach: cannot listen on the TCP port: listen tcp4 :8080: bind: address already in use
```

andreadecorte:
This change tries first to find if the default port 8080 is available.
If not, it tries to find an available free port in the range 8081-8090 in random order.
An error is raised if no free port is found.

In case of success, the chosen port is printed out for the user.

This does not apply if the user chooses a port different from 8080.

Note that this check could be still not be enough if the chosen port is taken
by another process after our check, in that case we will still fail at a later phase as today.

Includes unit testing.

## Summary by Sourcery

Tests:
- Add a unit test for the port validation logic.

### DIFF
--- a/ramalama/config.py
+++ b/ramalama/config.py
@@ -5,6 +5,8 @@ from typing import Any, Dict
 from ramalama.common import container_manager, default_image
 from ramalama.toml_parser import TOMLParser
 
+DEFAULT_PORT_RANGE = (8080, 8090)
+
 
 def get_store():
     if os.geteuid() == 0:
@@ -60,6 +62,10 @@ def load_config_from_env(config: Dict[str, Any], env: Dict):
     config['transport'] = env.get('RAMALAMA_TRANSPORT', config.get('transport', "ollama"))
 
 
+def int_tuple_as_str(input: tuple) -> str:
+    return '-'.join(map(str, input))
+
+
 def load_config_defaults(config: Dict[str, Any]):
     """Set configuration defaults if these are not yet set."""
     config['nocontainer'] = config.get('nocontainer', False)
@@ -71,7 +77,8 @@ def load_config_defaults(config: Dict[str, Any]):
     config['pull'] = config.get('pull', "newer")
     config['temp'] = config.get('temp', "0.8")
     config['host'] = config.get('host', "0.0.0.0")
-    config['port'] = config.get('port', "8080")
+    # print tuple as a range to avoid confusion
+    config['port'] = config.get('port', int_tuple_as_str(DEFAULT_PORT_RANGE))
     config['use_model_store'] = config.get('use_model_store', False)
 
 
--- a/ramalama/model.py
+++ b/ramalama/model.py
@@ -1,5 +1,7 @@
 import os
 import platform
+import random
+import socket
 import sys
 
 from ramalama.common import (
@@ -14,6 +16,7 @@ from ramalama.common import (
     podman_machine_accel,
     run_cmd,
 )
+from ramalama.config import DEFAULT_PORT_RANGE, int_tuple_as_str
 from ramalama.console import EMOJI
 from ramalama.gguf_parser import GGUFInfoParser
 from ramalama.kube import Kube
@@ -630,6 +633,7 @@ class Model(ModelBase):
 
     def serve(self, args):
         self.validate_args(args)
+        args.port = compute_serving_port(args.port, args.debug)
         model_path = self.get_model_path(args)
         exec_model_path = MNT_FILE
         if not args.container and not args.generate:
@@ -690,3 +694,45 @@ def distinfo_volume():
         return ""
 
     return f"-v{path}:/usr/share/ramalama/{dist_info}:ro"
+
+
+def compute_ports() -> list:
+    first_port = DEFAULT_PORT_RANGE[0]
+    last_port = DEFAULT_PORT_RANGE[1]
+    ports = list(range(first_port + 1, last_port + 1))
+    random.shuffle(ports)
+    # try always the first port before the randomized others
+    return [first_port] + ports
+
+
+def get_available_port_if_any(debug: bool) -> int:
+    ports = compute_ports()
+    print(ports)
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+        chosen_port = 0
+        for target_port in ports:
+            if debug:
+                print(f"Checking if {target_port} is available")
+            try:
+                s.bind(('localhost', target_port))
+            except OSError:
+                continue
+            else:
+                chosen_port = target_port
+                break
+        return chosen_port
+
+
+def compute_serving_port(port: str, debug: bool) -> str:
+    if not port:
+        raise IOError("serving port can't be empty.")
+    if port != int_tuple_as_str(DEFAULT_PORT_RANGE):
+        # user specified a custom port, don't override the choice
+        return port
+    # otherwise compute a random serving port in the range
+    target_port = get_available_port_if_any(debug)
+
+    if target_port == 0:
+        raise IOError("no available port could be detected. Please ensure you have enough free ports.")
+    print(f"serving on port {target_port}")
+    return str(target_port)
--- a/README.md
+++ b/README.md
@@ -221,7 +221,9 @@ $ ramalama stop mylama
 
 To use a UI, run a `ramalama serve` command, then connect via your browser at:
 
-127.0.0.1:8080
+127.0.0.1:<port>
+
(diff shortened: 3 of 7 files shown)


rhatdan:
You need to modify the man pages to describe what is happening here.

rhatdan:
$ ramalama containers --format '{{ .Ports }}'
0.0.0.0:8080->8080/tcp

Shows ports, but I have not figured out how to get the right number easily.

$ ramalama containers --format '{{ json }}' | jq 'map(.Ports)[0][0].host_port' 
8080

ramalama/model.py
```diff
@@ -633,3 +636,29 @@ def distinfo_volume():
...
+    if port != DEFAULT_PORT:
+        # user custom port, don't override the choice
+        return port
+    port = int(port)
+    found = False
+    # loop until we find an open port
+    while port <= MAX_PORT:
+        if not is_port_in_use(port):
+            found = True
+            break
```
ericcurtin, line 656: Rather than validating port, can we just try and bind to it. And increment if it fails? This solution is less prone to race conditions. I'd also only attempt to bind to 100 ports (less than max port), if they all fail, likely something is drastically wrong, failing ~65k times  could take a while.

It's also worse for performance to check before attempting to connect, we should assume we can connect and try the next port if we fail.

ericcurtin:
On many OSes (likely all the OSes we are compatible with) if you try to bind to port zero it will bind to a random open port, not saying we should or shouldn't do this, but it's an option to consider.

andreadecorte:
> I am not sure if incrementing the port is better or grabbing a random port. Users are comfortable with low number ports,  but you are much more likely to hit conflicts.

Yes it could be worth trying a random port (limiting the number of attempts as suggested by Eric)

> $ ramalama containers --format '{{ .Ports }}' 0.0.0.0:8080->8080/tcp
> 
> Shows ports, but I have not figured out how to get the right number easily.
> 
> $ ramalama containers --format '{{ json }}' | jq 'map(.Ports)[0][0].host_port' 8080
If you mean to show the users the output from there, we would need to be able to print this after launching the command, do we have something similar? And in any case we would not cover the --nocontainer case

ramalama/model.py
andreadecorte: makes sense to limit the number of attempts, thanks
> It's also worse for performance to do a test connect before attempting to connect, we should assume we can connect and try the next port if we fail.

Can you expand this? afaik the real connection is done inside the runtime, so are you saying to try to parse the error there and try again to launch the command?

ericcurtin:
I am gonna critique myself for a moment though... 😄  Lets take another tool Ollama, it uses a specific defined port, port 11434 , third party tools that interact with Ollama know where to find the Ollama service on a host, at port 11434, so I would say first try the default port, then fallback to random open port, because it will make some third part tools really tricky to work with, one would have to reconfigure the server port to connect to everytime...

rhatdan:
Should RamaLama bind to 11434 by default then, and not 8080, which eliminates a common port used for web serving?

One issue with using `--port 0` would be that other then the output line of llama.cpp, there would be know what to know which port it bound to, unless llama.cpp had a way to query it.

ericcurtin:
> Should RamaLama bind to 11434 by default then, and not 8080, which eliminates a common port used for web serving?

I'm open minded around the port. Some users install both Ollama and RamaLama, both applications would be competing for the port. RamaLama would probably end up fighting the Ollama daemon to compete for the port. Do we want to compete for the port?

We could by default do something like:

-p 11534-11634 (long alias --port)

This is one of those problems that will go away for the most part once we implement the ollama-like proxy server thing in RamaLama.

> 
> One issue with using `--port 0` would be that other then the output line of llama.cpp, there would be know what to know which port it bound to, unless llama.cpp had a way to query it.

rhatdan:
I think it is best to take the race condition for now, especially for containerized, since we need to expose the port on the host from within the container. So lets stick with this patch and deal with the race.  I prefer to pick a random Port within a range to limit the chance of the race.

andreadecorte:
ack, thanks. To summarize, I noted 3 things to work on:
- chose random port within a limited range (I can go for 11580-11680 as suggested, or we can refine that further) instead of increment by 1
- reuse socket for better perfs
- update manpages/docs
I will work on these points

andreadecorte:
updated based on the above comments, including docs.
I chose for now:
- 10 attempts
- range 8080-65536
but it can be changed if needed

ramalama/config.py
```diff
@@ -5,6 +5,8 @@
 from ramalama.common import container_manager, default_image
 from ramalama.toml_parser import TOMLParser
 
+DEFAULT_PORT_RANGE = (8080,65536)
```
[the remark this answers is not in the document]
andreadecorte: actually, I also added a limit of 10 attempts here, so it was ten ports out of the 8080-65536 range.
https://github.com/containers/ramalama/pull/898/files#diff-df19838abdc928ad5378587b7b23b29b5623d368daf8b19d59f74863049aebb5R648

This was to maximize the chances to find a good one.
Is this ok for you, as I think it addresses the concerns over the 60k ports?

rhatdan:
Needs a rebase.

ramalama/config.py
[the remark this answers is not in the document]
ericcurtin: I think we want to do something like this:

```
import socket
import random

def find_unused_port(start=8080, end=8090):
    # First try port 8080
    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
        try:
            s.bind(('localhost', start))
            return start
        except OSError:
            pass

    # If port 8080 is in use, try random ports in the range
    ports = list(range(start + 1, end))
    random.shuffle(ports)

    for port in ports:
        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
            try:
                s.bind(('localhost', port))
                return port
            except OSError:
                continue

    raise RuntimeError("No available ports in the specified range")

if __name__ == "__main__":
    unused_port = find_unused_port()
    print(f"Found unused port: {unused_port}")
```

we want to try 8080 first, because some external tools will just hardcode to that, then just keep trying random ports in the range until you find an unused one.

ericcurtin:
"make format" needs to be run to pass the build.

andreadecorte:
> "make format" needs to be run to pass the build.

aah thanks, I didn't have time to investigate, https://github.com/containers/ramalama/blob/main/CONTRIBUTING.md#python-format-and-lint we suggest here `make validate` but that doesn't include the format target, it should then

ericcurtin:
Now it should be more complete:

https://github.com/containers/ramalama/pull/916

ramalama/model.py
```diff
@@ -685,3 +689,42 @@ def distinfo_volume():
...
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+        max_attempts = 10
+
+        chosen_port = 0
+        current_attempt = 0
+        to_continue = True
+        while to_continue and current_attempt < max_attempts:
+            target_port = compute_random_port()
+            if debug:
+                print(f"Checking if {target_port} is available")
```
engelmi, line 708: General question:
Couldn't we use a logger instead of `if debug`, e.g.:
```python
logger.debug(f"Checking if {target_port} is available")
```

andreadecorte: yes that would avoid passing around debug flags. I didn't introduce it here not to diverge from the main goal as it was not used before in the code, but +1 as a follow up

ramalama/config.py
```diff
@@ -5,6 +5,8 @@
 from ramalama.common import container_manager, default_image
 from ramalama.toml_parser import TOMLParser
 
+DEFAULT_PORT = 8080
```
ericcurtin, line 8: We want the default to be 8080-8090

The point is not that we hardcode the tries to 10. Moreso that we hardcode the range to 8080-8090

andreadecorte: updated with a default range instead, hopefully we are converging to the expected solution

rhatdan:
LGTM

ericcurtin (approved):

Issue #895, opened 2025-02-28
[MacOS] Forcing container mode without a libkrun VM should fail with a clear error


andreadecorte:
On MacOS, ramalama only supports podman machines based on libkrun.
This is autodetected, but if user forces it, there is no clear error reported to the user.

## How to reproduce:
- podman machine list only has applehv machines
- ramalama --container run <something>

## Expectations:
An error saying that no podman machines with libkrun has been found

## Actual result:
a bad command line is build
```
llama-run -c 2048 --temp 0.8 /mnt/models/model.file
```
An model not found error is raised
```
Fetching resource 'https://registry.ollama.ai/v2//mnt/models/model.file/manifests/latest' failed: Failure when receiving data from the peer
Loading modelgguf_init_from_file: failed to open GGUF file 'model.file'
llama_model_load: error loading model: llama_model_loader: failed to load model from model.file

llama_model_load_from_file_impl: failed to load model
initialize_model: error: unable to load model from file: model.file
```

ericcurtin:
We should use podman-machine in this case, when --container is explicitly specified

@andreadecorte just so you know even if we fix this, using applehv will only have the ability to do CPU-based inferencing.

@bmahabirbu interested in looking at this one?

rhatdan:
Forcing a --container could have multiple issues. 

1. No Podman or Docker being installed
2. No Podman Machine or Docker Machine running.
3. Only CPU inferencing if the machine does not support GPU passthrough
4. Successful running of a container within a machine.

andreadecorte:
@ericcurtin this is not necessarily fixed by using applehv. What I had in mind was to just error out clearly if only applehv machine is available. Right now, this is confusing for the casual user as the error message is not really linked.

ericcurtin:
@andreadecorte how does this look for you in the latest release, closing, please re-open if unsure

andreadecorte:
looking good with latest release
```
Warning! Your VM podman-machine-default is using applehv, which does not support GPU. Only the provider libkrun has GPU support. Do you want to proceed without GPU? (yes/no):
```
Thanks @bmahabirbu for the change

PR #902, opened 2025-03-03
Added support for mac cpu and clear warning message

bmahabirbu:
Address issue https://github.com/containers/ramalama/issues/895

My thought process is if we're going to give a error sign to the user for not having libkrun we can give them the option to use cpu as well given they know it doesnt have gpu support!

## Summary by Sourcery

Adds support for running on macOS with the CPU by prompting the user to proceed without GPU support if the 'applehv' provider is being used. Modifies GPU arguments to account for the absence of /dev/dri when not using libkrun.

### DIFF
--- a/ramalama/common.py
+++ b/ramalama/common.py
@@ -2,10 +2,12 @@
 
 import glob
 import hashlib
+import json
 import logging
 import os
 import random
-import re
+
+# import re
 import shutil
 import string
 import subprocess
@@ -36,7 +38,7 @@ def get_engine():
     if engine is not None:
         return engine
 
-    if available("podman") and (sys.platform != "darwin" or is_podman_machine_running_with_krunkit()):
+    if available("podman") and (sys.platform != "darwin" or apple_vm()):
         return "podman"
 
     if available("docker") and sys.platform != "darwin":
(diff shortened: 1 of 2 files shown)


ramalama/common.py
```diff
@@ -55,14 +57,45 @@ def container_manager():
...
+                            f"Warning! Your VM {name} is using {provider}, which does not support GPU. "
+                            "Only the provider libkrun has GPU support. "
+                            "Do you want to proceed without GPU? (yes/no): "
+                        )
+                        .strip()
+                        .lower()
+                    )  # Make sure input is stripped and lowercase
+
+                    if user_input in ["yes", "y"]:
+                        os.environ["LIBKRUN"] = "False"
```
ericcurtin, line 88: We might as well delete these:

os.environ["LIBKRUN"]

sets, they aren't used anywhere.

FWIW, I do try to avoid using environment variables. Being global variables, it's preferable to avoid. I did add usage of environment variables like HIP_VISIBLE_DEVICES, CUDA_VISIBLE_DEVICES but these environment variables pre-date RamaLama, that's trying to take advantage of existing functionality and not re-invent the wheel.

Can be caught in a follow on PR though.

ramalama/common.py
```diff
@@ -55,14 +57,45 @@ def container_manager():
     return _engine
 
 
-def is_podman_machine_running_with_krunkit():
-    podman_machine_list = ["podman", "machine", "list", "--all-providers"]
+def apple_vm():
+    podman_machine_list = ["podman", "machine", "list", "--format", "json", "--all-providers"]
```
ericcurtin, line 61: json is a great idea, the existing parsing was more hacky. Although I would have kept the loose *krun* check, the name there has changed before between libkrun and krunkit.

ramalama/common.py
ericcurtin: Wait I see what you did now, still not ideal to use an env var. Anyway this is good enough for merge at least.

ericcurtin (approved):

PR #907, opened 2025-03-03
Use python variable instead of environment variable

ericcurtin:
environment variables are more global than we need.


ericcurtin:
@bmahabirbu PTAL

ramalama/common.py
```diff
@@ -57,43 +58,47 @@ def container_manager():
     return _engine
 
 
+def confirm_no_gpu(name, provider):
+    while True:
+        user_input = (
+            input(
```
ericcurtin, line 64: One disadvantage of this technique @bmahabirbu is it would be sort of annoying to answer this question over and over again interactively, but... Maybe it's annoying enough that it will poke people towards krunkit.

ericcurtin: I hope long-term:

podman-machine + applehv

is just completely migrated to:

podman-bootc + krunkit

but that could take a little bit.

bmahabirbu: My thoughts as well! I don't like using potentially blocking code  but needed a method for the user to be informed

bmahabirbu:
Looks good! Waiting on the last test

PR #910, opened 2025-03-03
Build a non-kompute Vulkan container image

ericcurtin:
There's no image to pull to play around with this backend right now.

## Summary by Sourcery

Chores:
- Build a container image for the Vulkan backend.

### DIFF
--- a/container-images/vulkan/Containerfile
+++ b/container-images/vulkan/Containerfile
@@ -4,5 +4,5 @@ ENV MODEL_PATH=/mnt/models/model.file
 
 COPY --chmod=755 ../scripts /usr/bin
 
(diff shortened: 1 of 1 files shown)


rhatdan:
Yikes that is a big bug. @lsm5 how close are we getting to removing kompute altogether or at least switching the defaults.  RamaLama builds with Vulkan and we build a Kompute image?

LGTM

lsm5:
> Yikes that is a big bug. @lsm5 how close are we getting to removing kompute altogether or at least switching the defaults. RamaLama builds with Vulkan and we build a Kompute image?
> 
> LGTM

@rhatdan I guess you meant someone else? I have no idea about Kompute.

Issue #885, opened 2025-02-26
Feature Request: Add support for unsupported Ascend NPU devices


leo-pony:
I want to integrates Ascend NPU hardware capabilities into the ramalama, and enables users to leverage the NPUs for inference and serving of LLMs.

 [Ascend NPU](https://www.hiascend.com/en/)  is an AI processor that has supports LLM inference engines such as [vllm](https://github.com/vllm-project/vllm-ascend), [llama.cpp](https://github.com/ggml-org/llama.cpp), [onnxruntime](https://github.com/microsoft/onnxruntime).

 As mentioned above,  Ascend NPU also supports AI frameworks like PyTorch, TensorFlow, which has already integrated by [huggingface](https://github.com/huggingface/accelerate/commit/c33adecc9f92808cbaeb0d34928e52e34180f964), [deepspeed](https://github.com/microsoft/DeepSpeed/commit/23a11a39510e2aefb48236e3d2672a7dcbfc42a3), and libraries such as Transformers/Accelerate for traning and fine-tune.

This patch address two aspects:
Add code in ramalama to build llama.cpp-cann binary and docker image.
Add code in ramalama to select vllm ascend docker image.

leo-pony:
@ericcurtin The coding of "Add code in ramalama to build llama.cpp-cann binary and docker image." has complete, currently in debuging pahse.

ericcurtin:
SGTM happy to look at PRs around this

geoman2:
Not sure what where to post so posting here to ask about NPU support (specially ANE NPU engine from M Apple series), but also the NPU from ARM smartphones like the Dimencity 9400/9500 and Snapdragon 8 Elite.

Can anyone reply?

ericcurtin:
I recommend talking to llama.cpp community first, once we are confident it's working there, then the RamaLama bits can be done

PR #911, opened 2025-03-04
Add support for llama.cpp engine to use ascend NPU device

leo-pony:
Add support for llama.cpp engine to use ascend NPU device.
[Ascend NPU](https://www.hiascend.com/en/) is an AI processor that has supports LLM inference engines such as [vllm](https://github.com/vllm-project/vllm-ascend), [llama.cpp](https://github.com/ggml-org/llama.cpp), [onnxruntime](https://github.com/microsoft/onnxruntime).

Corresponding Feature Request is #885.

## Summary by Sourcery

Adds support for Ascend NPU devices to the llama.cpp engine, enabling the use of Ascend NPUs for accelerated inference. This includes changes to detect Ascend NPUs, configure the build environment, and set necessary environment variables.

New Features:
- Adds support for Ascend NPU devices to the llama.cpp engine.
- Detects Ascend NPUs using the `npu-smi` command.
- Adds a new containerfile for building images with CANN support.
- Adds documentation for setting up RamaLama with Ascend NPU support on Linux systems

### DIFF
--- a/container-images/cann/Containerfile
+++ b/container-images/cann/Containerfile
@@ -1,0 +1,19 @@
+# Base image with CANN for compilation
+ARG ASCEND_VERSION=cann:8.0.0-910b-openeuler22.03-py3.10
+
+FROM quay.io/ascend/${ASCEND_VERSION} AS builder
+ARG GOLANG_VERSION
+COPY ../scripts /scripts
+RUN chmod +x /scripts/*.sh && \
+    sh -x /scripts/build_llama_and_whisper.sh "cann"
+
+FROM quay.io/ascend/${ASCEND_VERSION}
+# Copy the entire installation directory from the builder
+COPY --from=builder /tmp/install /usr
+ENV MODEL_PATH=/mnt/models/model.file
+COPY --chmod=755 ../scripts /usr/bin
+ENTRYPOINT [ \
+    "/bin/bash", \
+    "-c", \
+    "export LD_LIBRARY_PATH=/usr/lib:${LD_LIBRARY_PATH} && source /usr/local/Ascend/ascend-toolkit/set_env.sh && exec \"$@\"", "--" \
+]
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -187,6 +187,14 @@ def show_gpus_available_cli(args):
 
     else:  # Linux/Other OS GPU detection
         try:
+            ascend_gpus = gpu_detector.get_ascend_npu()
+            # Since Ascend devices are not required, keep quiet if no Ascend devices are detected
+            if ascend_gpus:
+                gpu_info.extend(ascend_gpus)
+        except Exception as e:
+            errors.append({"Vendor": "Ascend", "INFO": str(e)})
+
+        try:
             nvidia_gpus = gpu_detector.get_nvidia_gpu()
             if nvidia_gpus:
                 gpu_info.extend(nvidia_gpus)
--- a/ramalama/gpu_detector.py
+++ b/ramalama/gpu_detector.py
@@ -1,6 +1,8 @@
 import glob
 import logging
 import platform
+import re
+import shutil
 import subprocess
 
 logging.basicConfig(level=logging.WARNING, format="%(asctime)s - %(levelname)s - %(message)s")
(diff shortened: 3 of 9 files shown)


container-images/cann/Containerfile
```diff
@@ -0,0 +1,35 @@
+# Base image with CANN for compilation
+ARG ASCEND_VERSION=8.0.rc2.beta1-910b-openeuler22.03-py3.10
+
+FROM quay.io/ascend/cann:${ASCEND_VERSION} AS builder
+ARG GOLANG_VERSION
+COPY ../scripts /scripts
+RUN export http_proxy=http://172.17.0.1:7890 && https_proxy=http://172.17.0.1:7890 && chmod +x /scripts/*.sh && \
```
ericcurtin, line 7: Will need to remove the hardcoded proxy strings before merge. I would recommend putting this whole section in:

```
```

as "cann" specific scripting.

ericcurtin:
This LGTM for the most part so far

leo-pony:
> This LGTM for the most part so far

Thank you very much for your quick review! I will revise all the questions you raised and the issues found by CI, and add corresponding contents to the guidance document. 
Let you check it later.

container-images/cann/entrypoint.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+export LD_LIBRARY_PATH=/usr/lib:${LD_LIBRARY_PATH}
+
+cann_in_sys_path=/usr/local/Ascend/ascend-toolkit
```
rhatdan, line 5: Since we control what is put in the containerfile, why do we need this entrypontcode outside of the Containerfile?  The environemtn variable and entrypoint should be there, or am I missing something?

ramalama/cli.py
```diff
@@ -186,6 +186,15 @@ def show_gpus_available_cli(args):
             errors.append({"Vendor": "Apple", "INFO": str(e)})
 
     else:  # Linux/Other OS GPU detection
+        try:
+            ascend_gpus = gpu_detector.get_ascend_npu()
+            if ascend_gpus:
+                gpu_info.extend(ascend_gpus)
+            else:
+                errors.append({"Vendor": "Ascend", "INFO": "No GPU detected or drivers missing."})
```
rhatdan, line 194: Is this actually an error, or the expected behaviour?  IE Should this be removed?

ramalama/gpu_detector.py
```diff
@@ -147,6 +149,47 @@ def get_macos_gpu(self):
...
 
+    def get_ascend_npu(self):
+        """Detects Ascend NPUs using npu-smi (Linux only)."""
+        if platform.system() != "Linux":
+            return  # Skip on macOS and other platforms
+
+        try:
+            gpus = []
+            # get total npu number
+            total_count_process = subprocess.run(["npu-smi", "info", "-l"], capture_output=True, text=True, check=True)
```
rhatdan, line 160: Can you check if npu-smi exists, and if not return the skip.

ramalama/gpu_detector.py
```diff
@@ -147,6 +149,47 @@ def get_macos_gpu(self):
...
+                hbm_capacity = int(hbm_capacity_match.group(1))
+
+                self._update_best_gpu(hbm_capacity, npu_id, "CANN_VISIBLE_DEVICES")
+                gpu_info["VRAM"] = hbm_capacity
+                gpus.append(gpu_info)
+            
+            return gpus
+        except subprocess.CalledProcessError as e:
+            logging.error(f"Failed to detect ascend npu on linux: {e}")
+            raise RuntimeError("`npu-smi` not found. No ASCEND NPU detected or drivers missing.")
```
rhatdan, line 188: This will cause the error to be reported twice.  The command line tool will display the RuntimeError to the user.

ramalama/gpu_detector.py
```diff
@@ -147,6 +149,47 @@ def get_macos_gpu(self):
...
+        except Exception as e:
+            logging.error(f"Unexpected error while detecting linux ascend npu: {e}")
```
rhatdan, line 190: This error should be raised as well.

ramalama/gpu_detector.py
```diff
@@ -159,6 +202,17 @@ def detect_best_gpu(self, gpu_template):
...
+            try:
+                ascend_gpus = self.get_ascend_npu()
+                for gpu in ascend_gpus:
+                    vram = int(gpu.get("VRAM", "0 MiB").split()[0])
+                    if vram > best_vram:
+                        best_gpu = gpu
+                        best_vram = vram
+                        best_env = "CANN"
+            except RuntimeError as e:
+                logging.warning(f"Warning: Ascend detection failed: {e}")
```
rhatdan, line 214: Can you try this on a machine without Ascend and make sure it stays quiet.

This should either raise the error all the way up, or not say anything.  Would there be a case where a user could have a machine with say an NVIDIA GPU and a CANN GPU and this would be noice for the NVIDIA GPU?

container-images/cann/Containerfile
```diff
@@ -0,0 +1,35 @@
...
+    cann_in_sys_path=/usr/local/Ascend/ascend-toolkit; \
+    cann_in_user_path=$HOME/Ascend/ascend-toolkit; \
+    if [ -f "${cann_in_sys_path}/set_env.sh" ]; then \
+        source ${cann_in_sys_path}/set_env.sh; \
+        export LD_LIBRARY_PATH=${cann_in_sys_path}/latest/lib64:${cann_in_sys_path}/latest/aarch64-linux/devlib:${LD_LIBRARY_PATH} ; \
```
shink, line 12: We should consider supporting both x86 and arm

container-images/cann/Containerfile
```diff
@@ -0,0 +1,35 @@
+# Base image with CANN for compilation
+ARG ASCEND_VERSION=8.0.rc2.beta1-910b-openeuler22.03-py3.10
+
```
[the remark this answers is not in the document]
ericcurtin: Note RamaLama does most of it's work on RHEL-like platforms. For this container you may use Ubuntu 22.04, the community has the option to do that, there's advantages to aligning on RHEL-like platforms.

We will probably create a federated network of sorts for container images, so the RamaLama project may not build and publish every container image to https://quay.io/organization/ramalama namespace, but for this container image it's a good starting point to collaborate here at the moment.

ericcurtin:
Commits need to be signed via `git commit -s` to satisfy the DCO bot.

container-images/cann/Containerfile
ericcurtin: If it's easy this is a good idea.

ramalama/gpu_detector.py
ericcurtin: A dumb check is to just skip all this if `npu-smi` doesn't exist. There's potential to align this technique with `nvidia-smi` tool.

ericcurtin (approved):

ramalama/model.py
```diff
@@ -158,6 +158,7 @@ def _image(self, args):
             "CUDA_VISIBLE_DEVICES": "quay.io/ramalama/cuda",
             "ASAHI_VISIBLE_DEVICES": "quay.io/ramalama/asahi",
             "INTEL_VISIBLE_DEVICES": "quay.io/ramalama/intel-gpu",
+            "CANN_VISIBLE_DEVICES": "quay.io/ramalama/cann",
```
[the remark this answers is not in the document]
rhatdan: Yes we should have a table in ramallama.conf which is easily customizable.

PR #916, opened 2025-03-07
Extend make validate check to do more

ericcurtin:
It also does check-format now.

## Summary by Sourcery

Extends the `make validate` command to include format checking using black and isort. Also, modifies the linting process to include more files and updates test cases to use a constant for the Hugging Face blob URL.

Build:
- Adds `check-format` to the `validate` target in the Makefile, enforcing code formatting standards.
- Extends linting to include more files in subdirectories in the Makefile.
- Updates the Makefile to use black and isort for format checking.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -114,18 +114,18 @@ docs:
 
 .PHONY: lint
 lint:
-	flake8 *.py */*.py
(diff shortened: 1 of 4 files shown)


engelmi (approved):
LGTM

PR #917, opened 2025-03-07
Add chat template support

engelmi:
This PR enables the (automatic) use of the chat template file via `ramalama run` by passing the chat template file to `llama-run`. The chat template can be either provided/downloaded directly or is extracted from the GGUF model and stored - preference is given to the provided chat template file. 

TODO:
- [ ] Include https://pypi.org/project/go2jinja/ to convert Go Templates (used by Ollama) to Jinja templates


engelmi:
@curtine Could we rebuild a new version of the ramalama container image? 
It seems that it does not yet include the changes from https://github.com/ggml-org/llama.cpp/pull/11961/ since the `--chat-template-file` CLI option for `llama-run` is missing.

ericcurtin:
@rhatdan has the build infrastructure set up to do it, so it would be less effort to wait for him for a new release.

But if you build a container image locally, you can just pass that to llama.cpp for development purposes.

ramalama/cli.py
```diff
@@ -817,6 +819,9 @@ def run_cli(args):
             model.run(args)
         except Exception:
             raise e
+    except Exception as ex:
+        print(ex)
```
rhatdan, line 823: Debugging info?

rhatdan: If you run ramalama --debug, you should see tracebacks when they happen.

ramalama/model_store.py
```diff
@@ -47,24 +57,78 @@ def download(self, blob_file_path: str, snapshot_dir: str) -> str:
...
+                parts = filename.split(RefFile.SEP)
+                if len(parts) != 2:
+                    ref_file.filenames.append(filename)
+                    filename = file.readline().strip()
+                    continue
+
+                ref_file.filenames.append(parts[0])
+                if parts[1] == RefFile.MODEL_SUFFIX:
+                    ref_file.model_name = parts[0]
+                elif parts[1] == RefFile.CHAT_TEMPLATE_SUFFIX:
```
rhatdan, line 116: Switch elif to if, no need for else.

ramalama/model_store.py
```diff
@@ -267,25 +338,27 @@ def get_cached_files(self, model_tag: str) -> Tuple[str, list[str], bool]:
...
 
         ref_file_path = self.get_ref_file_path(model_tag)
         if not self.file_exists(ref_file_path):
             ref_file = RefFile()
             ref_file.hash = snapshot_hash
             ref_file.filenames = [file.name for file in snapshot_files]
+            for file in snapshot_files:
+                if file.type == SnapshotFileType.Model:
+                    ref_file.model_name = file.name
+                elif file.type == SnapshotFileType.ChatTemplate:
```
rhatdan, line 352: Switch elif  to if.  No need for else.

ramalama/model_store.py
```diff
@@ -306,6 +379,52 @@ def new_snapshot(self, model_tag: str, snapshot_hash: str, snapshot_files: list[
...
             os.symlink(blob_relative_path, self.get_snapshot_file_path(snapshot_hash, file.name))
 
+    def _ensure_chat_template(self, model_tag: str, snapshot_hash: str, snapshot_files: list[SnapshotFile]):
+        model_file: SnapshotFile = None
+        for file in snapshot_files:
+            # Give preference to a chat template that has been specified in the file list
+            if file.type == SnapshotFileType.ChatTemplate:
+                return
+            if file.type == SnapshotFileType.Model:
+                model_file = file
```
rhatdan, line 389: Should break here?  Can there be multiple SnapshotfileType.Model, if yes then we only see the last.

rhatdan (approved):

rhatdan:
We will do a release on Monday. or Sunday,

codefromthecrypt:
So, when I run the following, and access the openai endpoint I still get jinja errors on tool calls as this is foundational to jinja, but not yet jinja, right?
```bash
$ gh pr checkout 917
$ python3 -m venv .venv && source .venv/bin/activate && pip install -e . && python bin/ramalama serve qwen2.5:3b
```

codefromthecrypt:
ps if I make this change, and do `ramalama serve qwen2.5:3b`, my tool calls examples work, except one (semantic-kernel dotnet)

```diff
--- a/ramalama/model.py
+++ b/ramalama/model.py
@@ -586,6 +586,7 @@ class Model(ModelBase):
         else:
             exec_args = [
                 "llama-server",
+                "--jinja",
                 "--port",
                 args.port,
                 "-m",
```

https://github.com/ggml-org/llama.cpp/issues/12279 has details on failures in general with hf qwen2.5, and the error of semantic-kernel dotnet, which also applies here.

ramalama/model_store.py
engelmi: The ModelStore would allow multiple models at the moment and huggingface allows that as well (e.g. [mradermacher/SmolLM-135M-GGUF](https://huggingface.co/mradermacher/SmolLM-135M-GGUF/tree/main)). However, this would be invalid for ramalama as input. When the `refs` file is serialized, the current approach is to use the last seen model file - the same applies to the chat template file. So currently, I think its probably worth to add some kind of validaton for this (i.e. only one model and chat template in the list of files) and raise an Exception. WDYT?

engelmi:
@codefromthecrypt By using the `--jinja` option for `llama-run` or `llama-server`, implicitly the built-in chat templates are being used. I assume that one of the llama.cpp built-in templates produces quite similar output to one of the one required by your model. When you run
```bash
$ ramalama inspect qwen2.5:3b | grep chat_template
```
You get basically the jinja template required by that model. 
(If the model has been pulled from ollama, then its probably not there since ollama uses Go Templates instead of Jinja - we are working on that as well)

This PR is part of enabling to detect and use the chat templates provided by the platformas such as ollama as well as extract this information from .gguf models. However, this will take a bit longer. So I think its best to use ramalama with your changes for the presentation.

ericcurtin:
> @codefromthecrypt By using the `--jinja` option for `llama-run` or `llama-server`, implicitly the built-in chat templates are being used. I assume that one of the llama.cpp built-in templates produces quite similar output to one of the one required by your model. When you run
> 
> ```shell
> $ ramalama inspect qwen2.5:3b | grep chat_template
> ```
> 
> You get basically the jinja template required by that model. (If the model has been pulled from ollama, then its probably not there since ollama uses Go Templates instead of Jinja - we are working on that as well)
> 
> This PR is part of enabling to detect and use the chat templates provided by the platformas such as ollama as well as extract this information from .gguf models. However, this will take a bit longer. So I think its best to use ramalama with your changes for the presentation.

Long-term we want llama-server to just default to jinja without any manual intervention and fallback to other techniques, needs upstream llama.cpp work.

codefromthecrypt:
thanks for the advice folks! ps please applaud @ochafik for the work upstream on llama-cpp! https://github.com/ggml-org/llama.cpp/issues/12279

engelmi:
Added also the feature to convert ollama's Go Templates to Jinja ones so those should work as well. For example, when running:
```bash
$ ramalama --use-model-store run --pull=never ollama://granite-code
🦭 > write a hello world program in python
print("hello world")

🦭 > 
```
I had to use `--pull=never` since the quay.io container image doesn't have the changes for `--chat-template-file` in llama.cpp (yet). 

It downloads the go template from ollama and then converts it to jinja, passing the latter one to llama-run. 
```bash
$ cat chat_template
{{ if .Suffix }}<fim_prefix> {{ .Prompt }}<fim_suffix> {{ .Suffix }}<fim_middle>
{{- else if .Messages }}
{{- range $i, $_ := .Messages }}
{{- $last := eq (len (slice $.Messages $i)) 1 }}
{{- if eq .Role "user" }}Question:
{{ .Content }}
...
$ cat chat_template_converted
{% if suffix %}<fim_prefix> {{ prompt }}<fim_suffix> {{ suffix }}<fim_middle>
{%- elif messages %}
{%- for m in messages %}
{%- set last = ((messages)[loop.index0:])|length==1 %}
{%- if m["role"]=="user" %}Question:
{{ m["content"] }}
...
```

@rhatdan @ericcurtin PTAL, I think the PR is ready for review.

ericcurtin:
Tested this branch on macOS, problem:

```
./install.sh -l
cp: /var/folders/q3/5msn8s8j02d62vcdg_y2hh2w0000gn/T/tmp.UMDH5HbFl3/share/go2jinja/__init__.py: No such file or directory
cp: /var/folders/q3/5msn8s8j02d62vcdg_y2hh2w0000gn/T/tmp.UMDH5HbFl3/share/go2jinja/go2jinja.py: No such file or directory
```

ramalama/model_store.py
rhatdan: Yes I think we should throw an error, since RamaLama has no way to know if it chose the right one.

engelmi: Added a validation function, incl. a few unit test cases.

ericcurtin:
> > Tested this branch on macOS, problem:
> > ```
> > ./install.sh -l
> > cp: /var/folders/q3/5msn8s8j02d62vcdg_y2hh2w0000gn/T/tmp.UMDH5HbFl3/share/go2jinja/__init__.py: No such file or directory
> > cp: /var/folders/q3/5msn8s8j02d62vcdg_y2hh2w0000gn/T/tmp.UMDH5HbFl3/share/go2jinja/go2jinja.py: No such file or directory
> > ```
> 
> Did you use the current version? I pushed some updates recently since the CI was failing due to this.

Just refreshed branch, still occuring

ericcurtin:
We have this problem again on macOS /usr/bin/python3 . It would be nice to set up CI to check this. It's a very easy fix everytime, just populate a variable first. If you remember to only use variables in your f-string's it will stop happening.

```
  File "/Users/ecurtin/git/ramalama/bin/ramalama", line 98, in <module>
    main(sys.argv[1:])
  File "/Users/ecurtin/git/ramalama/bin/ramalama", line 52, in main
    import ramalama
  File "/Users/ecurtin/git/ramalama/./ramalama/__init__.py", line 5, in <module>
    from ramalama.cli import HelpException, init_cli, print_version
  File "/Users/ecurtin/git/ramalama/./ramalama/cli.py", line 10, in <module>
    import ramalama.oci
  File "/Users/ecurtin/git/ramalama/./ramalama/oci.py", line 9, in <module>
    from ramalama.model import Model
  File "/Users/ecurtin/git/ramalama/./ramalama/model.py", line 24, in <module>
    from ramalama.model_store import ModelStore
  File "/Users/ecurtin/git/ramalama/./ramalama/model_store.py", line 14, in <module>
    from ramalama.go2jinja import go2jinja
  File "/Users/ecurtin/git/ramalama/./ramalama/go2jinja/go2jinja.py", line 117
    return f"{self.operands[0].to_jinja()}.format({", ".join([op.to_jinja() for op in self.operands[1:]])})"
                                                    ^
SyntaxError: f-string: expecting '}'
```

engelmi:
> We have this problem again on macOS /usr/bin/python3 . It would be nice to set up CI to check this. It's an easy fix everytime, just populate a variable first. If we remember to only use variables in f-string's it will stop happening.
> 
> ```
> Traceback (most recent call last):
>   File "/Users/ecurtin/git/ramalama/bin/ramalama", line 98, in <module>
>     main(sys.argv[1:])
>   File "/Users/ecurtin/git/ramalama/bin/ramalama", line 52, in main
>     import ramalama
>   File "/Users/ecurtin/git/ramalama/./ramalama/__init__.py", line 5, in <module>
>     from ramalama.cli import HelpException, init_cli, print_version
>   File "/Users/ecurtin/git/ramalama/./ramalama/cli.py", line 10, in <module>
>     import ramalama.oci
>   File "/Users/ecurtin/git/ramalama/./ramalama/oci.py", line 9, in <module>
>     from ramalama.model import Model
>   File "/Users/ecurtin/git/ramalama/./ramalama/model.py", line 24, in <module>
>     from ramalama.model_store import ModelStore
>   File "/Users/ecurtin/git/ramalama/./ramalama/model_store.py", line 14, in <module>
>     from ramalama.go2jinja import go2jinja
>   File "/Users/ecurtin/git/ramalama/./ramalama/go2jinja/go2jinja.py", line 117
>     return f"{self.operands[0].to_jinja()}.format({", ".join([op.to_jinja() for op in self.operands[1:]])})"
>                                                     ^
> SyntaxError: f-string: expecting '}'
> ```

This error is new to me: Why does this cause an issue on mac? And yes, would be great to have the CI properly checking for this on mac - same for the install.sh.

engelmi:
> > > Tested this branch on macOS, problem:
> > > ```
> > > ./install.sh -l
> > > cp: /var/folders/q3/5msn8s8j02d62vcdg_y2hh2w0000gn/T/tmp.UMDH5HbFl3/share/go2jinja/__init__.py: No such file or directory
> > > cp: /var/folders/q3/5msn8s8j02d62vcdg_y2hh2w0000gn/T/tmp.UMDH5HbFl3/share/go2jinja/go2jinja.py: No such file or directory
> > > ```
> > 
> > 
> > Did you use the current version? I pushed some updates recently since the CI was failing due to this.
> 
> Just refreshed branch, still occuring

I can't test this on my system or via the CI, so its quite hard to fix this. Could you help me here? @ericcurtin

ericcurtin:
> > We have this problem again on macOS /usr/bin/python3 . It would be nice to set up CI to check this. It's an easy fix everytime, just populate a variable first. If we remember to only use variables in f-string's it will stop happening.
> > ```
> > Traceback (most recent call last):
> >   File "/Users/ecurtin/git/ramalama/bin/ramalama", line 98, in <module>
> >     main(sys.argv[1:])
> >   File "/Users/ecurtin/git/ramalama/bin/ramalama", line 52, in main
> >     import ramalama
> >   File "/Users/ecurtin/git/ramalama/./ramalama/__init__.py", line 5, in <module>
> >     from ramalama.cli import HelpException, init_cli, print_version
> >   File "/Users/ecurtin/git/ramalama/./ramalama/cli.py", line 10, in <module>
> >     import ramalama.oci
> >   File "/Users/ecurtin/git/ramalama/./ramalama/oci.py", line 9, in <module>
> >     from ramalama.model import Model
> >   File "/Users/ecurtin/git/ramalama/./ramalama/model.py", line 24, in <module>
> >     from ramalama.model_store import ModelStore
> >   File "/Users/ecurtin/git/ramalama/./ramalama/model_store.py", line 14, in <module>
> >     from ramalama.go2jinja import go2jinja
> >   File "/Users/ecurtin/git/ramalama/./ramalama/go2jinja/go2jinja.py", line 117
> >     return f"{self.operands[0].to_jinja()}.format({", ".join([op.to_jinja() for op in self.operands[1:]])})"
> >                                                     ^
> > SyntaxError: f-string: expecting '}'
> > ```
> 
> This error is new to me: Why does this cause an issue on mac? And yes, would be great to have the CI properly checking for this on mac - same for the install.sh. Edit: Ah, its the use of `"` inside the `{...}`.

Older version on python3 doesn't have PEP 701

ericcurtin:
It's working now

ericcurtin (approved):

PR #919, opened 2025-03-07
Modify GPU detection to match against env var value instead of prefix

cgruver:
This PR fixes and issue where ramalama will fail to properly detect a GPU if there is an environment variable that has the same prefix as the env vars that ramalama looks for.

i.e. the Intel OneApi dev environment sets `INTEL_PYTHONHOME`

The previous logic will incorrectly return true because it matched agains the prefix of `INTEL_`

## Summary by Sourcery

Bug Fixes:
- Fixes incorrect GPU detection when environment variables share prefixes with GPU-related variables.

### DIFF
--- a/ramalama/common.py
+++ b/ramalama/common.py
@@ -371,14 +371,13 @@ def get_gpu():
 def get_env_vars():
     gpu_vars = (
         "ASAHI_VISIBLE_DEVICES",
+        "CANN_VISIBLE_DEVICES",
         "CUDA_VISIBLE_DEVICES",
+        "CUDA_LAUNCH_BLOCKING",
         "HIP_VISIBLE_DEVICES",
         "HSA_VISIBLE_DEVICES",
-        "INTEL_VISIBLE_DEVICES",
-        "CANN_VISIBLE_DEVICES",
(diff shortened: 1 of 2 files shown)


ramalama/common.py
```diff
@@ -365,15 +365,22 @@ def get_gpu():
...
-    #     env_vars[gpu_type] = str(gpu_num)
+    gpu_vars = (
+        "ASAHI_VISIBLE_DEVICES",
+        "CUDA_VISIBLE_DEVICES",
+        "HIP_VISIBLE_DEVICES",
+        "HSA_VISIBLE_DEVICES",
+        "INTEL_VISIBLE_DEVICES",
+        "CANN_VISIBLE_DEVICES",
+        "CUDA_LAUNCH_BLOCKING",
+        "HIP_SOMETHING",
```
ericcurtin, line 380: Can we remove fake ones, like HIP_SOMETHING ?

cgruver: Yeah, I wondered about that too...  It looks like it is only for testing the presence of `HIP_*` env vars.

I can modify the test to look for a legit var.

PR #920, opened 2025-03-08
Add Intel ARC 155H to list of supported hardware

cgruver:
Small change to add Intel Meteor Lake with ARC GPU to the list of hardware in readme.md.

## Summary by Sourcery

Documentation:
- Adds Intel ARC GPU (Meteor Lake 155H) to the list of supported hardware in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,8 @@ curl -fsSL https://raw.githubusercontent.com/containers/ramalama/s/install.sh | 
 | Nvidia GPU (cuda)                  | :white_check_mark: |
 | AMD GPU (rocm)                     | :white_check_mark: |
 | Ascend NPU (Linux)                 | :white_check_mark: |
+| Intel ARC GPU (Meteor Lake 155H)   | :white_check_mark: |
+
 ## COMMANDS
 
 | Command                                                | Description                                                |


ericcurtin (approved):

hanthor:
@cgruver I can ask around and collect more people to test Intel Arc GPU support so we can add to a list? I'm pretty sure i915 GPUs will work. But I'll as some A and B dGPU users to test. Were just looking for SYCL in the `ramalama bench -ngl 99` command?

ericcurtin:
> @cgruver I can ask around and collect more people to test Intel Arc GPU support so we can add to a list? I'm pretty sure i915 GPUs will work. But I'll as some A and B dGPU users to test. Were just looking for SYCL in the `ramalama bench -ngl 99` command?

FWIW, I think i915 GPUs will work also, but often it's not worth it as they can be weak and have very limited VRAM. So sometimes on these machines it's more appropriate to use CPU-based inferencing.

hanthor:
Confirmed that a B580 works.

Had to force the intel-gpu image for it to use it. But it did 50 t/s on Vulkan and 54 t/s on SYCL so not a night and day difference

cgruver:
@hanthor I'm working on a PR that will use a list of GPU device IDs.

I'll go ahead and finish it and get you to test.

I had planned to only include GPUs with at least 128 EUs.  The older 13th gen and 12th gen chipsets just don't have enough juice...

PR #921, opened 2025-03-08
Allow user to specify the images to use per hardware

rhatdan:
## Summary by Sourcery

Enhancements:
- Allows users to specify different container images to be used based on the detected hardware.

### DIFF
--- a/docs/ramalama.conf
+++ b/docs/ramalama.conf
@@ -44,6 +44,15 @@
 #
 #image = "quay.io/ramalama/ramalama:latest"
 
+# Alternative images to use when RamaLama recognizes specific hardware
+#
+#[ramalama.images]
+#HIP_VISIBLE_DEVICES="quay.io/ramalama/rocm"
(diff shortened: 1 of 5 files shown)


rhatdan:
Should we use a better name for these then XXX_VISIBLE_DEVICES, for example XXX_VISIBLE_IMAGES or just use XXX_IMAGE=

ericcurtin:
> Should we use a better name for these then XXX_VISIBLE_DEVICES, for example XXX_VISIBLE_IMAGES or just use XXX_IMAGE=

The naming scheme came from CUDA and ROCm libraries, with these libraries you can set CUDA_VISIBLE_DEVICES or HIP_VISIBLE_DEVICES environment variables to select a certain GPU to use, etc.

So we just re-used them so people familiar with using Ollama, llama.cpp, etc. would be able to use these env vars again in the same manner and get the same results.

I am open minded how we go about this, although this seems reasonable.

ericcurtin (approved):

Issue #925, opened 2025-03-10
invalid argument "--chat-template openchat"


benoitf:
when I run the llama-server images I receive

invalid argument "--chat-template openchat"

```
    llama-server \
        --model "${MODEL_PATH}" \
        --host "${HOST:=0.0.0.0}" \
        --port "${PORT:=8001}" \
        --gpu_layers "${GPU_LAYERS:=0}" \
        "${CHAT_FORMAT}"
```

it looks like the variable CHAT_FORMAT, as surrounded by quotes is not expanded as 2 arguments but as a single argument '--chat-template openchat'

PR #926, opened 2025-03-10
fix: CHAT_FORMAT variable should be expanded

benoitf:
fixes https://github.com/containers/ramalama/issues/925

## Summary by Sourcery

Bug Fixes:
- Fixes an issue where the CHAT_FORMAT variable was not being expanded, leading to incorrect behavior of the llama-server script.

### DIFF
--- a/container-images/scripts/llama-server.sh
+++ b/container-images/scripts/llama-server.sh
@@ -9,7 +9,7 @@ if [ -n "${MODEL_CHAT_FORMAT}" ]; then
 fi
 
 if [ -n "${MODEL_PATH}" ]; then
-    llama-server \
+    eval llama-server \
         --model "${MODEL_PATH}" \
         --host "${HOST:=0.0.0.0}" \
         --port "${PORT:=8001}" \


rhatdan:
The linters do not like this change?

benoitf:
replaced the expand of the variable to eval usage so linter would be happier

rhatdan:
What about doing something like:

https://paste.centos.org/view/f4a1bbf7

benoitf:
@rhatdan replaced the eval by the function calling

benoitf:
so there is still an issue with this pattern

as if I don't use quotes

```
shellcheck container-images/scripts/llama-server.sh

In container-images/scripts/llama-server.sh line 19:
	$(chat_format)
        ^------------^ SC2046 (warning): Quote this to prevent word splitting.

For more information:
  https://www.shellcheck.net/wiki/SC2046 -- Quote this to prevent word splitt...

```


and if I use quotes then if the chat option is not provided it gives an empty argument and then llama-server reports: 

`error: invalid argument:`

so it looks like the `eval` usage is the only working solution for now

rhatdan:
LGTM

rhatdan:
Merging, since none of the tests are going to hit this.

ericcurtin:
Another way to satisfy the linter here would have been to use a bash array...

is an example script that does this.

PR #933, opened 2025-03-10
Make compatible with the macOS system python3

ericcurtin:
populate variables first. Reproducible on systems with brew macOS also by switching shebang to #!/usr/bin/python3

## Summary by Sourcery

Improve error message handling and model name construction.

Bug Fixes:
- Fixes an issue where error messages were not properly formatted when fetching manifest or snapshot files.
- Fixes an issue where model names were not properly constructed when listing models in the model store.


rhatdan:
LGTM

PR #939, opened 2025-03-11
Handle CNAI annotation deprecation

s3rj1k:
## Summary by Sourcery

Update annotations to use the opencontainers standard and deprecate the CNAI annotations.

Enhancements:
- Update annotations to use the opencontainers standard.
- Deprecate the cnai annotations.

### DIFF
--- a/ramalama/annotations.py
+++ b/ramalama/annotations.py
@@ -1,5 +1,5 @@
 # These annotations are based off the standards:
-# https://github.com/opencontainers/image-spec/blob/main/specs-go/v1/annotations.go
+# https://github.com/opencontainers/image-spec/blob/main/annotations.md
 # https://github.com/CloudNativeAI/model-spec
 
 # ArtifactTypeModelManifest specifies the media type for a model manifest.
--- a/ramalama/annotations.py
+++ b/ramalama/annotations.py
@@ -73,46 +73,7 @@ AnnotationBaseImageDigest = "org.opencontainers.image.base.digest"
 # image's base image.
 AnnotationBaseImageName = "org.opencontainers.image.base.name"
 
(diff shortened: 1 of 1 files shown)


rhatdan:
I guess the question would be, will RamaLama eventually care about some of the CNAI annotions to allow developers of models, to tell RamaLama how to configure the OCI model.

s3rj1k:
> I guess the question would be, will RamaLama eventually care about some of the CNAI annotions to allow developers of models, to tell RamaLama how to configure the OCI model.

There is literally one annotation CNAI in current version of spec, the one that is filename related and it is put on each layer, basically to help form filetree of whole artifact.

I would prefer having opencontainers annotations for something like ‘zot’ registry to consume and have native UI visiblity.

ericcurtin:
@nalind @saschagrunert PTAL

benoitf:
I'm 👍 on my end

ramalama/annotations.py
```diff
@@ -1,4 +1,5 @@
-# These annotations are based off the proposed standard:
+# These annotations are based off the standards:
+# https://github.com/opencontainers/image-spec/blob/main/specs-go/v1/annotations.go
```
nalind, line 2: Is this supposed to be referring to the Go sources, or the spec itself, which is at https://github.com/opencontainers/image-spec/blob/main/annotations.md?

s3rj1k: I was looking at code, found it more readable but you are correct, should link to actual doc and not code implementation. Fixed.

rhatdan:
LGTM

rhatdan:
Thanks @s3rj1k

PR #946, opened 2025-03-11
Lets run container in all tests, to make sure it does not explode.

rhatdan:
Also switch to using smollm


ericcurtin (approved):

maxamillion:
This seems to explode the CI because the model gets itself into an infinite loop and eventually craps out. We probably need a prompt that this model is happy with to avoid this situation.

rhatdan:
Sure, got a suggestion?

rhatdan:
This seems to work, hopefully not flakey.  Will merge.

Issue #948, opened 2025-03-11
Feature Request: Generic llama.cpp/runtime option


edmcman:
It would be nice to be able to specify an arbitrary runtime argument to `ramalama serve` that is sent to the runtime.  For example: `ramalama serve --runtime-arg "--top-p 1.0"`.  At the moment, I'd really like the ability to do this for `--jinja`, but I think that should be built in to `serve` (#947).

I suppose `run` might need the same functionality.

ericcurtin:
Sounds like a good idea, wanna open a PR @edmcman ?

edmcman:
Sure, will do.

PR #949, opened 2025-03-12
Add --runtime-arg option for run and serve

edmcman:
See #948 

Useful for passing `--jinja` to llama.cpp at the moment to enable tool calling.

## Summary by Sourcery

Adds the `--runtime-args` option to the `run` and `serve` commands, allowing users to pass arguments directly to the underlying runtime (llama.cpp or vllm).

New Features:
- Adds the `--runtime-args` option to the `run` and `serve` commands.

Tests:
- Adds tests to verify that `--runtime-args` are passed to the runtime.

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -2,6 +2,7 @@ import argparse
 import glob
 import json
 import os
+import shlex
 import subprocess
 from datetime import datetime, timezone
 from pathlib import Path
@@ -186,6 +187,8 @@ def post_parse_setup(args):
         if resolved_model:
             args.UNRESOLVED_MODEL = args.MODEL
             args.MODEL = resolved_model
+    if hasattr(args, "runtime_args"):
+        args.runtime_args = shlex.split(args.runtime_args)
 
 
 def login_parser(subparsers):
@@ -689,6 +692,13 @@ def run_serve_perplexity_args(parser):
         dest="context",
         default=CONFIG['ctx_size'],
         help="size of the prompt context (0 = loaded from model)",
+    )
+    parser.add_argument(
+        "--runtime-args",
+        dest="runtime_args",
+        default="",
+        type=str,
+        help="arguments to add to runtime invocation",
     )
 
 
--- a/docs/ramalama-perplexity.1.md
+++ b/docs/ramalama-perplexity.1.md
@@ -96,6 +96,9 @@ not have more privileges than the user that launched them.
 - **never**: Never pull the image but use the one from the local containers storage. Throw an error when no image is found.
 - **newer**: Pull if the image on the registry is newer than the one in the local containers storage. An image is considered to be newer when the digests are different. Comparing the time stamps is prone to errors. Pull errors are suppressed if a local image was found.
 
+#### **--runtime-args**="*args*"
+Add *args* to the runtime (llama.cpp or vllm) invocation.
+
 #### **--seed**=
 Specify seed rather than using random seed model interaction
 
--- a/docs/ramalama-run.1.md
+++ b/docs/ramalama-run.1.md
@@ -100,6 +100,9 @@ Pull image policy. The default is **missing**.
 - **never**: Never pull the image but use the one from the local containers storage. Throw an error when no image is found.
 - **newer**: Pull if the image on the registry is newer than the one in the local containers storage. An image is considered to be newer when the digests are different. Comparing the time stamps is prone to errors. Pull errors are suppressed if a local image was found.
 
+#### **--runtime-args**="*args*"
+Add *args* to the runtime (llama.cpp or vllm) invocation.
+
 #### **--seed**=
 Specify seed rather than using random seed model interaction
 
--- a/docs/ramalama-serve.1.md
+++ b/docs/ramalama-serve.1.md
@@ -127,11 +127,14 @@ not have more privileges than the user that launched them.
 - **never**: Never pull the image but use the one from the local containers storage. Throw an error when no image is found.
 - **newer**: Pull if the image on the registry is newer than the one in the local containers storage. An image is considered to be newer when the digests are different. Comparing the time stamps is prone to errors. Pull errors are suppressed if a local image was found.
 
+#### **--runtime-args**="*args*"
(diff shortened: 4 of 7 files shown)


ramalama/cli.py
```diff
@@ -690,6 +690,14 @@ def run_serve_perplexity_args(parser):
         default=CONFIG['ctx_size'],
         help="size of the prompt context (0 = loaded from model)",
     )
+    parser.add_argument(
+        "--runtime-arg",
+        dest="runtime_args",
+        action='append',
+        type=str,
+        default=[],
+        help="argument to add to runtime invocation"
```
engelmi, line 699: NIT:
```suggestion
        help="arguments to add to runtime invocation"
```

ramalama/cli.py
```diff
@@ -690,6 +690,14 @@ def run_serve_perplexity_args(parser):
         default=CONFIG['ctx_size'],
         help="size of the prompt context (0 = loaded from model)",
     )
+    parser.add_argument(
+        "--runtime-arg",
+        dest="runtime_args",
+        action='append',
+        type=str,
```
engelmi, line 697:
```suggestion
        type=str,
        nargs="+",
```

ramalama/cli.py
```diff
@@ -690,6 +690,14 @@ def run_serve_perplexity_args(parser):
         default=CONFIG['ctx_size'],
         help="size of the prompt context (0 = loaded from model)",
     )
+    parser.add_argument(
+        "--runtime-arg",
```
engelmi, line 694: NIT:
```suggestion
        "--runtime-args",
```

engelmi (commented):
I think some of the end-to-end tests need to be updated regarding the expected output.

rhatdan:
Needs man page documention?

Sourcery has some good points.

Test required, basically just look at the dryrun tests.

rhatdan:
Thanks @edmcman

rhatdan:
You can run `make validate` locally to descover quickly issues that will prevent your PR from being merged.

ericcurtin:
You need to sign your commits to pass the DCO build, recommend reading the build logs for the others.

edmcman:
Thanks, I will try to address the other issues after work tonight.

edmcman:
FYI [ramalama-serve.md](https://github.com/containers/ramalama/blob/main/docs/ramalama-serve.1.md) seems pretty specific to llama.cpp.  Is this right?  Does `ramalama-serve` not support `vllm`?

edmcman:
Ah, I don't think `nargs="+"` will work here because we might want to pass things that look like arguments to `argparse`.  The only way I know how to do this is too have `nargs=1` and use `--runtime-args=--arg1 --runtime-args=--arg2` and use append.  But if there is a better way I am all ears...

engelmi:
> Ah, I don't think `nargs="+"` will work here because we might want to pass things that look like arguments to `argparse`. The only way I know how to do this is too have `nargs=1` and use `--runtime-args=--arg1 --runtime-args=--arg2` and use append. But if there is a better way I am all ears...

Makes sense. With `nargs="+"` and `action="append"` we'll get a list of lists with one element.

ericcurtin:
There are other implementations of arg appending tools, dracut is one off the top of my head (it's kargs, but the same concept), might be worth a little bit of research on existing implementations. Arg parsing, appending, etc. can get hairy (especially when adding multiple args at once), would be nice to get it right the first time.

ericcurtin:
> FYI [ramalama-serve.md](https://github.com/containers/ramalama/blob/main/docs/ramalama-serve.1.md) seems pretty specific to llama.cpp. Is this right? Does `ramalama-serve` not support `vllm`?

It can also be used with vllm.

edmcman:
> Well we could not support a list, and just force users to put all options into one command, which should work.
> 
> --runtime-arg="--arg1 --arg2=foo --arg3"

We then have to parse the the string to split it.  Though maybe that is easy with [this](https://docs.python.org/3/library/shlex.html#shlex.split).  I kind of like that plan, though as @ericcurtin suggested I'll take a look at dracut.

edmcman:
@ericcurtin Were you thinking of `--kernel-cmdline`?  I guess those don't really have dashes in them, so it seems like they didn't have to worry about that problem.

https://github.com/dracutdevs/dracut/blob/5d2bda46f4e75e85445ee4d3bd3f68bf966287b9/test/test-functions#L29

edmcman:
Can someone allow the CI to run the BATS test again?  I am having trouble running it locally.

edmcman:
Hmm.  I think there might be an issue with ` run_ramalama` passing the quoted argument properly...

edmcman:
Woohoo. My BATS tests are finally passing :sweat:  I don't think the current bats failures are my fault.  Seems like a flaky test?

ericcurtin:
> @ericcurtin Were you thinking of `--kernel-cmdline`? I guess those don't really have dashes in them, so it seems like they didn't have to worry about that problem.
> 
> https://github.com/dracutdevs/dracut/blob/5d2bda46f4e75e85445ee4d3bd3f68bf966287b9/test/test-functions#L29

I think I was actually thinking of grubby 😅  I think that can add kargs.

It was just the first example of a arg appending tool I could think of. There may be some pitfalls to consider like how we interpret " , etc.

edmcman:
Looks like grubby calls `poptParseArgvString` https://github.com/rhboot/grubby/blob/c01b0d5bb182bde35b464d14996acf354a3ada2e/grubby.c#L3804

> Although popt is usually used for parsing arguments already divided into an argv-style array, some programs need to parse strings that are formatted identically to command lines. To facilitate this, popt provides a function that parses a string into an array of strings, using rules similiar to normal shell parsing.

This sounds like `shlex.split`, which I am using now.

My thoughts:
1. Add a test or two involving quoted arguments
2. Move the `shlex.split` calls closer to the argument parsing, rather than in `model.py` where they are currently.

rhatdan:
You need to squash your commits.

git rebase -i origin 
git push --force

rhatdan:
LGTM

PR #953, opened 2025-03-12
Add specified nvidia-oci runtime

rhatdan:
Nvidia recommends using their nvidia-container-runtime when running containers with GPU support, so ramalama should use this feature as well.

Allow users to override the oci-runtime when appropriate.


ramalama/common.py
```diff
@@ -318,6 +319,9 @@ def check_asahi():
 
 
 def check_nvidia():
+    global _nvidia
+    if _nvidia != -1:
+        return _nvidia
     try:
         command = ['nvidia-smi']
         run_cmd(command).stdout.decode("utf-8")
```
ericcurtin, line 327: I think we are missing a:

`_nvidia = cuda`

before we return here

rhatdan:
@ericcurtin @engelmi @bmahabirbu PTAL

ericcurtin (approved):

PR #954, opened 2025-03-12
There must be at least one CDI device present to use CUDA

ericcurtin:
Otherwise we get failures like:

Error: setting up CDI devices: unresolvable CDI devices nvidia.com/gpu=all

Co-authored-by: Brian <bmahabir@bu.edu>

### DIFF
--- a/ramalama/common.py
+++ b/ramalama/common.py
@@ -6,8 +6,6 @@ import json
 import logging
 import os
 import random
-
-# import re
 import shutil
 import string
 import subprocess
@@ -15,6 +13,7 @@ import sys
 import time
 import urllib.error
 import urllib.request
+from typing import List
 
 import ramalama.console as console
 from ramalama.http_client import HttpClient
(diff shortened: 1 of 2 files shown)


ramalama/common.py
```diff
@@ -322,7 +322,12 @@ def check_nvidia():
         command = ['nvidia-smi']
         run_cmd(command).stdout.decode("utf-8")
         os.environ["CUDA_VISIBLE_DEVICES"] = "0"
-        return "cuda"
+        command = [container_manager(), 'info', '--format', 'json']
+        result = run_cmd(command)
+        info = json.loads(result.stdout)
+        if 'host' in info and 'cdi' in info['host'] and 'devices' in info['host']['cdi']:
```
rhatdan, line 328: Did you check to see if this works with Docker?

ericcurtin: Kinda, there is no host.cdi.devices in either docker or podman for me.

But I don't know about the case where host.cdi.devices has a value, is this portable for docker/podman.

ericcurtin:
"ramalama run smollm" test is flakey, need to fix.

rhatdan:
Should we use a better model like granite and pay the price?

"ramalama run smollm" test is flakey, need to fix.

Why would this go into infinite loop?

ramalama/common.py
rhatdan: Do we need cdi?  If I run with `docker run --gpus all ...` it seems to run.
Also runs with podman.

ericcurtin:
> Should we use a better model like granite and pay the price?
> 
> "ramalama run smollm" test is flakey, need to fix.
> 
> Why would this go into infinite loop?

Sure, although if it was me, I'd go the opposite way and try an even smaller model, to keep pull times fast:

snowflake-arctic-embed:22m

is by a company called Snowflake Inc. , it's tiny.

ramalama/common.py
ericcurtin: Currently seeing this on an Ampere aarch64 system with Nvidia, although we don't want to use Nvidia here, hence why it's not fully configured. I saw this on an x86_64 machine before also

ericcurtin:
Wait I found a granite one that's tiny :)

granite-embedding:30m

didn't know this existed :)

rhatdan:
The problem I see with the smaller models, is they seem to go into infinite loops which ends up blowing up the CI.  Do not know if this is just the nature of small models not having good guardrails or something else.

ramalama/common.py
ericcurtin: I would like to see if there is a non-null value here in the correct case before we merge, because I can't test the case for the correctly configured system, maybe the json is wrong

ericcurtin:
@lsm5 any idea what's up with the testing-farm builds?

@rhatdan this is ready for review.

Issue #957, opened 2025-03-13
[Bug] Execute "ramalama info" failed, with error information: go2jinja.py line 398   f"{SYMBOL_REMOVE_WHITESPACE SyntaxError: EOL while scanning string literal


leo-pony:
My python version is 3.9.9.

[root@ascend910b-02 ramalama]# ramalama info
ramalama module not found in sys.path: ['./', '/usr/local/lib/python3.9/site-packages', '/usr/local/bin', '/usr/lib64/python39.zip', '/usr/lib64/python3.9', '/usr/lib64/python3.9/lib-dynload', '/root/.local/lib/python3.9/site-packages', '/usr/local/lib64/python3.9/site-packages', '/usr/local/lib/python3.9/site-packages', '/usr/lib64/python3.9/site-packages', '/usr/lib/python3.9/site-packages']
Traceback (most recent call last):
  File "/usr/local/bin/ramalama", line 99, in <module>
    main(sys.argv[1:])
  File "/usr/local/bin/ramalama", line 52, in main
    import ramalama
  File "/data/disk3/mnj/code/remallama/ramalama/./ramalama/__init__.py", line 5, in <module>
    from ramalama.cli import HelpException, init_cli, print_version
  File "/data/disk3/mnj/code/remallama/ramalama/./ramalama/cli.py", line 9, in <module>
    import ramalama.oci
  File "/data/disk3/mnj/code/remallama/ramalama/./ramalama/oci.py", line 9, in <module>
    from ramalama.model import Model
  File "/data/disk3/mnj/code/remallama/ramalama/./ramalama/model.py", line 24, in <module>
    from ramalama.model_store import ModelStore
  File "/data/disk3/mnj/code/remallama/ramalama/./ramalama/model_store.py", line 11, in <module>
    import ramalama.go2jinja as go2jinja
  File "/data/disk3/mnj/code/remallama/ramalama/./ramalama/go2jinja.py", line 398
    f"{SYMBOL_REMOVE_WHITESPACE
                               ^
SyntaxError: EOL while scanning string literal

I test with ramalama  serve, the same error information with above.

ericcurtin:
Fix https://github.com/containers/ramalama/pull/959

engelmi:
@leo-pony Could you check if #959 solves the issue? If yes we can close this one.

leo-pony:
> @leo-pony Could you check if #959 solves the issue? If yes we can close this one. 

Thank you very much, it solves this issue.
I closed this one.

PR #958, opened 2025-03-13
Fix go2jinja.py file f-string format error under some python version, such as: py3.9.9.

leo-pony:
Fix the issue #957 
Signed-off-by: leo-pony <nengjunma@outlook.com>

## Summary by Sourcery

Bug Fixes:
- Fixes f-string formatting errors in go2jinja.py that occur in some Python versions.

### DIFF
--- a/ramalama/go2jinja.py
+++ b/ramalama/go2jinja.py
@@ -395,12 +395,10 @@ def translate_continue_nodes(root_nodes: list[Node]) -> list[Node]:
         continue_node.end = -1
         continue_node.content = (
             f"{GO_SYMBOL_OPEN_BRACKETS}"
-            f"{SYMBOL_REMOVE_WHITESPACE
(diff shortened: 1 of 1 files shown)


ericcurtin:
LGTM just need to satisfy CI.

CI is complaining about line length.

If we only use variables in f-strings moving forward we can avoid this problem.

@engelmi just for visibility.

It would be nice if we could catch this in CI.

ericcurtin:
I think this PR should ensure this gets checked in CI moving forward, lets see:

https://github.com/containers/ramalama/pull/959

if the build goes red in the above PR, it's probably working.

ericcurtin:
Yup it worked 🥳

engelmi:
> Yup it worked 🥳

Nice! 
I guess then this PR can be closed and the fix(es) will be provided in #959?

leo-pony:
A more reasonable solution to fix this problem has been implemented in #959. This PR is closed.

leo-pony:
A more reasonable solution to fix this problem has been implemented in https://github.com/containers/ramalama/pull/959. This PR is closed.

PR #959, opened 2025-03-13
python3 validator

ericcurtin:
We are encountering issues where newer python3 features are breaking systems with older versions of python3, such as macOS, this should ensure we validate this in CI.

## Summary by Sourcery

Adds a compileall check to the linting process to ensure compatibility with older Python 3 versions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -114,6 +114,10 @@ docs:
 
 .PHONY: lint
 lint:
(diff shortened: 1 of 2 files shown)


ericcurtin:
We should merge this, once the python3 compatibility fixes get merged.

ericcurtin:
@engelmi PTAL

CI will catch these syntax compatibility problems now in future.

PR #961, opened 2025-03-13
Adds Rag

bmahabirbu:
A thing to note about containerizing the vector database

Line 25
self.client = qdrant_client.QdrantClient(path=os.getcwd()+"/db")

This is the path to save the database on host. From here we can containerize it. As long as the database folder is already there the code will use it and wont start a new db

for qdrant a vector data base must have a name. Line 29-34 creates a new database if non exists

## Summary by Sourcery

Introduces a Retrieval-Augmented Generation (RAG) framework, including a new Python script for RAG functionality, a build script, and a Dockerfile for containerization. The framework allows users to insert PDF files into a vector database and query them using natural language.

New Features:
- Adds a RAG framework with document insertion and querying capabilities.
- Implements a document converter using Docling for PDF processing.
- Integrates with Qdrant for vector storage and retrieval.
- Uses OpenAI for generating answers based on retrieved context.
- Provides a command-line interface for inserting documents and running the RAG.


container-images/rag/Containerfile
```diff
@@ -0,0 +1,20 @@
+ARG IMAGE=quay.io/ramalama/ramalama:latest
+FROM $IMAGE
+RUN mkdir /db
+COPY ../rag/rag_framework.py /
+COPY ../rag/build_rag.sh /tmp/
+RUN sh /tmp/build_rag.sh
+
+# buils container
```
rhatdan, line 8: Remove commented out sections.

container-images/rag/build_rag.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
ericcurtin, line 1: The easy way to get this into all the containers, it just to add this stuff to:

you could "git mv" that script to a more generic name also if you wanted, it originally just built llama.cpp and whisper.cpp

container-images/rag/rag_framework.py
```diff
@@ -0,0 +1,222 @@
...
+# Needed for mac to not give errors
+os.environ["TOKENIZERS_PARALLELISM"] = "true"
+
+
+# Helper Classes and Functions
+
+class Rag:
+    def __init__(self):
+        # Setup vector database
+        self.client = qdrant_client.QdrantClient(path=os.getcwd()+"/db")
```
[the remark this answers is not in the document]
rhatdan: Just make this a static path  /vector/db or /db

container-images/rag/build_rag.sh
bmahabirbu: Sounds good! but I dont think we want to include docling in the build so ill remove that for now.

PR #962, opened 2025-03-14
[NPU][Fix] only specify device num, but without ascend-docker-runtime installed, running ramalama/cann container image will failing

leo-pony:
1. Keep the environment variable of visible Ascend device in ramalama consistent with ascend-docker-runtime. 
2. Temporarily remove the default value of using device 0 when no ascend device is specified. The reason is that currently, if you only specify device 0 without using ascend-docker-runtime, it cannot be offloaded to NPU normally.

Signed-off-by: leo-pony <nengjunma@outlook.com>

## Summary by Sourcery

This PR ensures that the environment variable for visible Ascend devices is consistent with ascend-docker-runtime and updates the documentation to include instructions for installing Ascend docker runtime.

### DIFF
--- a/docs/ramalama-cann.7.md
+++ b/docs/ramalama-cann.7.md
@@ -42,6 +42,7 @@ This provides NPU acceleration using the AI cores of your Ascend NPU. And [CANN]
 For more information about Ascend NPU in [Ascend Community](https://www.hiascend.com/en/).
 
 Make sure to have the CANN toolkit installed. You can download it from here: [CANN Toolkit](https://www.hiascend.com/developer/download/community/result?module=cann)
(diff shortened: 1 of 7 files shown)


leo-pony:
> Hey @leo-pony - I've reviewed your changes - here's some feedback:
> 
> **Overall Comments**:
> * Consider adding a warning message when the user specifies device 0 without ascend-docker-runtime, instead of removing the default value.

The default device number supporting of Asend NPU will be added next. The current default value implementation mechanism has problems and is temporarily removed.

ericcurtin (approved):

ericcurtin (approved):

leo-pony:
> Is cann multi-arch or only x86?

Currently, for CANN, linux x86 and linux arm are supported.
For the Ramalama llama.cpp backend, linux x86 and linux arm are also supported.

Issue #964, opened 2025-03-14
ramalama whisper image is broken


jeffmaury:
When startung the quay.io/ramalama/ramalama-whisper-server image,  we have the error:

`
Error: preparing container 9a10f1549088c8f1ccfe83ff22d52c8fbe2433c3a9e7e6ae014345aa0844e99f for attach: crun: executable file `/usr/bin/whisper-server.sh` not found in $PATH: No such file or directory: OCI runtime attempted to invoke a command that was not found
`

The image defines `/usr/bin/whisper-server.sh`as entrypoint but in /usr/bin there is whisper-run.sh but no whisper-server.sh

nettyso:
I fixed this temporarily by reverting to a previous version, from quay.io/ramalama/asahi:latest to quay.io/ramalama/asahi:0.6.2

rhatdan:
https://github.com/containers/ramalama/pull/965 fixes this issue, we will have new images on monday.

PR #965, opened 2025-03-14
Fix ENTRYPOINTS of whisper-server and llama-server

rhatdan:
Fixes:https://github.com/containers/ramalama/issues/964

## Summary by Sourcery

Bug Fixes:
- Fixes an issue where the wrong executable was being called due to an incorrect entrypoint in the container build.

### DIFF
--- a/container_build.sh
+++ b/container_build.sh
@@ -41,10 +41,13 @@ add_entrypoint() {
     containerfile=$(mktemp)
     cat > "${containerfile}" <<EOF
 FROM $2
-ENTRYPOINT [ "/usr/bin/$3.sh" ]
+ENTRYPOINT [ "/usr/bin/$3" ]
 EOF
     echo "$1 build --no-cache -t $2-$3 -f ${containerfile} ."
     eval "$1 build --no-cache -t $2-$3 -f ${containerfile} ."
+    # verify entrypoints created correctly by running help on them.
+    echo "$1 run --rm $2-$3 -h"
+    eval "$1 run --rm $2-$3 -h"
     rm "${containerfile}"
 }
 


rhatdan:
@jeffmaury PTAL

rhatdan:
@benoitf @ericcurtin PTAL

container_build.sh
```diff
@@ -41,10 +41,13 @@ add_entrypoint() {
     containerfile=$(mktemp)
     cat > "${containerfile}" <<EOF
 FROM $2
-ENTRYPOINT [ "/usr/bin/$3.sh" ]
+ENTRYPOINT [ "/usr/bin/$3" ]
 EOF
     echo "$1 build --no-cache -t $2-$3 -f ${containerfile} ."
     eval "$1 build --no-cache -t $2-$3 -f ${containerfile} ."
+    # verify entrypoints created correctly by running help on them.
```
ericcurtin, line 48: I hope this manipulating of Containerfiles will work on konflux.

ericcurtin (approved):

container_build.sh
rhatdan: We will have to break it out into something Konflux will understand or just add additional Containerfiles.

The issue is we have the same pattern for each different GPU type.

benoitf:
this PR is breaking the llama-server images as the entrypoint is the bash script https://github.com/containers/ramalama/blob/main/container-images/scripts/llama-server.sh

not the server



```
podman inspect quay.io/ramalama/ramalama-llama-server:0.6.4 | jq -r '.[0].Config.Entrypoint'
[
  "/usr/bin/llama-server"
]
```

```
podman inspect quay.io/ramalama/ramalama-llama-server:0.6.3 | jq -r '.[0].Config.Entrypoint'
[
  "/usr/bin/llama-server.sh"
]
```

Issue #934, opened 2025-03-10
[Feature Request] CPU inferencing with Llama.cpp uses only 4 cores


antheas:
Running the following command on version v0.6.3 ends up using around 4 cores (4 processes using 100% CPU each).

```bash
ramalama run instructlab/merlinite-7b-lab --ngl 0 "Testing testing testing"
```

Adding `--env LLAMA_ARG_THREADS=32` or something similar does not have an effect.

This is on an AMD HX MAX 395+ (x86) CPU with 16c/32t. But has also been verified on an Ampere platform. So it is very easy to reproduce.

Suspected cause is core autodetection within `llama-run` being broken for some reason and not being able to supply the number of cores manually.

antheas:
cc @KyleGospo @castrojo

hanthor:
```
podman run --rm -i --label ai.ramalama --name ramalama_WUNG8Wm7HY --env=HOME=/tmp --init --security-opt=label=disable --cap-drop=all --security-opt=no-new-privileges --label ai.ramalama.model=ollama://llama3.2:3b --label ai.ramalama.engine=podman --label ai.ramalama.runtime=llama.cpp --label ai.ramalama.port=8080 --label ai.ramalama.command=serve --pull=newer -t --env LLAMA_ARG_THREADS=18 -p 8080:8080 --device /dev/dri -e INTEL_VISIBLE_DEVICES=1 --mount=type=bind,src=/var/home/james/.local/share/ramalama/models/ollama/llama3.2:3b,destination=/mnt/models/model.file,ro quay.io/ramalama/vulkan:latest llama-server --port 8080 -m /mnt/models/model.file -c 2048 --temp 0.8 -ngl 0 --host 0.0.0.0 
```

passing LLAMA_ARG_THREADS=18 via the new `--env` works for llama-server but llama-run doesn't appear to support that. I will investigate further

hanthor:
@ericcurtin is there a way to add env var support to llama-run similar to llama-server? thing about the Intel warning

`get_memory_info: [warning] ext_intel_free_memory is not supported (export/set ZES_ENABLE_SYSMAN=1 to support), use total memory as free memory` also

ericcurtin:
> [@ericcurtin](https://github.com/ericcurtin) is there a way to add env var support to llama-run similar to llama-server? thing about the Intel warning
> 
[...]

PR #966, opened 2025-03-14
feat(cpu): add --threads option to specify number of cpu threads

antheas:
Adds the `--threads` (and `-t`) option allowing to specify CPU threads for models.

Needs https://github.com/ggml-org/llama.cpp/pull/12370 (perhaps in that one it would be better to read `LLAMA_ARG_THREADS` for consistency)

Closes https://github.com/containers/ramalama/issues/934

TODO:
 - Wait for https://github.com/ggml-org/llama.cpp/pull/12370 to merge so that `ramalama run --threads 32` works.

## Summary by Sourcery

New Features:
- Introduces the `--threads` option (and its shorthand `-t`) to allow users to specify the number of CPU threads for model inferencing.

## Summary by Sourcery

New Features:
- Introduces the `--threads` (and `-t`) option to allow users to specify the number of CPU threads for model inferencing.


ramalama/model.py
```diff
@@ -294,6 +294,9 @@ def add_env_option(self, conman_args, args):
         for env in args.env:
             conman_args += ["--env", env]
 
+        if args.runtime == "llama.cpp" and args.subcommand != "run" and args.threads and args.threads > 0:
```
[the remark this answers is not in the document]
antheas: Do we want `--threads` to work on everything or just `llama-run`? The problem here is that for all `llama-*` commands the environment variable works, except `llama-run`.

So we need this dirty check to avoid adding a useless env var and to use the argument instead.

This function still needs an else that configures the vllm runtime. But it did not run on my system so I will try to do it later today/tomorrow.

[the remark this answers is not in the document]
antheas: My first attempt was `exec_args`, but `exec_args` happen too late to accept `--env`. Unless we do something dirty like prefix the `exec_args` with env

ericcurtin:
This will add it to llama-run:

https://github.com/ggml-org/llama.cpp/pull/12370

I'd rather just do everything via CLI args. Half configuring things via env vars and half configuring things via CLI options is not ideal Would rather consolidate on one approach.

antheas:
> This will add it to llama-run, should be merged by end of day:
> 
> [ggml-org/llama.cpp#12370](https://github.com/ggml-org/llama.cpp/pull/12370)
> 
> I'd rather just do everything via CLI args. Half configuring things via env vars and half configuring things via CLI options is not ideal Would rather consolidate on one approach.

Ok I just checked, both `llama-bench` and `llama-server` have a `--threads` argument, so I will move the code to `exec_args`

antheas:
Placed the code under `gpu_args`. Seems like `--runtime vllm` is happy with this as well (it crashes for a different reason afterwards). I also tried to fix the unit test I broke.

ericcurtin:
Related PR:

https://github.com/containers/ramalama/pull/967

Could we default to (number of cores)/(2) threads based on the performance data we gathered for CPU inferencing?

antheas:
> Related PR:
> 
> #967
> 
> Could we default to (number of cores)/(2) threads based on the performance data we gathered for CPU inferencing?

I think `llama-bench` and `llama-serve` seem to do this already. So do we want to do this through ramalama or defer to the scripts themselves?

docs/ramalama-bench.1.md
```diff
@@ -57,6 +57,10 @@ set the network mode for the container
 number of gpu layers, 0 means CPU inferencing, 999 means use max layers (default: -1)
 The default -1, means use whatever is automatically deemed appropriate (0 or 999)
 
+#### **--threads**
```
rhatdan, line 60: Should document ,**-t**

docs/ramalama-perplexity.1.md
```diff
@@ -62,6 +62,10 @@ set the network mode for the container
 number of gpu layers, 0 means CPU inferencing, 999 means use max layers (default: -1)
 The default -1, means use whatever is automatically deemed appropriate (0 or 999)
 
+#### **--threads**
```
rhatdan, line 65: Should document ,**-t**

docs/ramalama-run.1.md
```diff
@@ -65,7 +65,10 @@ set the network mode for the container
 number of gpu layers, 0 means CPU inferencing, 999 means use max layers (default: -1)
 The default -1, means use whatever is automatically deemed appropriate (0 or 999)
 
-Pull image policy. The default is **missing**.
+
+#### **--threads**
```
rhatdan, line 69: Should document ,**-t**

docs/ramalama-serve.1.md
```diff
@@ -89,6 +89,10 @@ set the network mode for the container
 number of gpu layers, 0 means CPU inferencing, 999 means use max layers (default: -1)
 The default -1, means use whatever is automatically deemed appropriate (0 or 999)
 
+#### **--threads**
```
rhatdan, line 92: Should document ,**-t**

docs/ramalama.conf.5.md
```diff
@@ -110,6 +110,11 @@ In some cases this is needed to access the gpu from a rootless container
 number of gpu layers, 0 means CPU inferencing, 999 means use max layers (default: -1)
 The default -1, means use whatever is automatically deemed appropriate (0 or 999)
 
+**threads**=-1
```
rhatdan, line 113: Should document ,**-t**

ramalama/cli.py
```diff
@@ -713,6 +713,13 @@ def bench_run_serve_perplexity_args(parser):
         default=CONFIG["ngl"],
         help="number of layers to offload to the gpu, if available",
     )
+    parser.add_argument(
+        "-t",
+        "--threads",
+        type=int,
+        default=CONFIG["ngl"],
```
rhatdan, line 720: ngl? Should be threads.

docs/ramalama-bench.1.md
antheas: That's a good question. So `ramalama serve` without a `--threads` argument uses 16 cores on a 32 core machine. With a `--threads 0` it uses 32. So it kind of has a differentiation, which makes me want to keep 0 as an option.

I cant test `--runtime vllm` or `llama run` (yet).

docs/ramalama.conf.5.md
rhatdan: Yes I got over zealous

rhatdan:
I think the default should be 0 not -1, since -t 0 means nothing as far as I see, it would be an invalid value?

antheas:
> I think the default should be 0 not -1, since -t 0 means nothing as far as I see, it would be an invalid value?

`--threads 0` and not having `--threads` has different behavior in `llama-serve`. Not having `--threads` on my machine uses 16 threads, and having `--threads 0` uses 32 threads.

So I would tend on having `--threads -1` use no `--threads` argument and `--threads 0` being passed through to the program.

rhatdan:
Ok you need to alphabetically sort your man page options, and then we can get this in.

PR #967, opened 2025-03-14
Update llama.cpp to contain threads features

ericcurtin:
So we can specify CPU threads for llama-run.

## Summary by Sourcery

Chores:
- Update llama.cpp SHA to 9f2250ba722738ec0e6ab684636268a79160c854


rhatdan:
LGTM

PR #971, opened 2025-03-16
Only set this environment variable if we can resolve CDI

ericcurtin:
We don't want to use Nvidia/CUDA just because nvidia-smi is present


rhatdan:
LGTM

PR #973, opened 2025-03-16
Update llama.cpp for some Gemma features

ericcurtin:
We want to get some new Gemma features added to llama.cpp .

## Summary by Sourcery

Chores:
- Update llama.cpp submodule to version 8ba95dc.


rhatdan:
LGTM

PR #975, opened 2025-03-17
Fix handling of whisper-server and llama-server entrypoints

rhatdan:
Entrypoint tests are blowing up so remove for now.

Fixes:  https://github.com/containers/ramalama/issues/977


benoitf (approved):
I was able to check for ramala-server image it works as expected

I've noticed that MODEL_PATH is hardcoded in the image
https://github.com/containers/ramalama/blob/e5249621fb96222be4a49ef01a8452e8b869be85/container-images/ramalama/Containerfile#L3

while the entrypoint expect that the user provides the path and report an error message
https://github.com/containers/ramalama/blob/e5249621fb96222be4a49ef01a8452e8b869be85/container-images/scripts/llama-server.sh#L11

if MODEL_PATH is not set, which is never the case
Maybe, follow up, should we remove that ENV var for this image ?


for whisper server, the fix is launching the entrypoint as expected

I tried the entrypoint and on my end it reported

```
[root@567655a5bf7c /]# whisper-server -tr       --model "${MODEL_PATH}"         --convert \
        --host "${HOST:=0.0.0.0}" \
        --port "${PORT:=8001}"
sh: line 1: ffmpeg: command not found
ffmpeg is not found. Please ensure that ffmpeg is installed and that its executable is included in your system's PATH. [root@567655a5bf7c /]#
```

so I'm wondering if all dependencies of whispercpp are inside the VM ?

I noticed https://github.com/containers/ai-lab-recipes/blob/418d2f3bf22b7520f0786a79e3ea57a7c78d71aa/model_servers/whispercpp/base/Containerfile#L19-L20 had these 2 lines

PR #979, opened 2025-03-17
Add docling support version 2

rhatdan:
Remove pragmatic, and move to using local implementation until llama-stack version is ready.

 python3 container-images/scripts/doc2rag.py --help
usage: docling [-h] target source [source ...]

process source files into RAG vector database

positional arguments:
  target
  source

options:
  -h, --help  show this help message and exit

## Summary by Sourcery

Enhancements:
- Replaces the pragmatic tool with a local doc2rag implementation.

### DIFF
--- a/container-images/scripts/build_rag.sh
+++ b/container-images/scripts/build_rag.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+set -exu -o pipefail
+
+export PYTHON_VERSION="python3 -m"
+pyversion=$(python3 --version)
+
+version_greater() {
+        string="$1
+$2"
+    [ "$string" != "$(sort --version-sort <<< "$string")" ]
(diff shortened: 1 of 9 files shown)


rhatdan:
@bmahabirbu PTAL

bmahabirbu:
Quick question I see that we're using the vectordb.name as part of the arguments. Shouldn't this be a folder path instead? From what I've tested when qdrant client output path is a folder that containers all the files.

bmahabirbu:
other than that it looks good to me! I can give it a quick test later today to confirm

rhatdan:
Have not tested yet, but yes I think it should be a directory not a file.  Also I am thinking that it should be a known file name.

ericcurtin (approved):

PR #982, opened 2025-03-18
Default the number of threads to (nproc)/(2)

ericcurtin:
The llama.cpp default for threads is hardcoded to 4. This changes that harcoding so instead we use the (number of cpu cores)/(2).

## Summary by Sourcery

Enhancements:
- Updates the default number of threads to be the number of CPU cores divided by 2.

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -710,7 +710,39 @@ def run_serve_perplexity_args(parser):
     )
 
 
+def default_threads():
+    if CONFIG["threads"] < 0:
+        nproc = os.cpu_count()
+        if nproc and nproc > 4:
+            return int(nproc / 2)
+
+        return 4
+
+    return CONFIG["threads"]
+
+
+def add_exec_arguments(parser):
+    parser.add_argument(
+        "--ngl",
+        dest="ngl",
+        type=int,
+        default=CONFIG["ngl"],
+        help="number of layers to offload to the gpu, if available",
+    )
+
+    def_threads = default_threads()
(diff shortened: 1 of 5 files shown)


ramalama/cli.py
```diff
@@ -190,6 +190,11 @@ def post_parse_setup(args):
     if hasattr(args, "runtime_args"):
         args.runtime_args = shlex.split(args.runtime_args)
 
+    if hasattr(args, "threads") and args.threads < 0:
+        nproc = os.cpu_count()
```
rhatdan, line 194: Can you default this in the CLI not at this point.  Basically if the user does not specify the --threads command it should show a default of nproc/2

rhatdan: Also need to update the man page.

ericcurtin: The thing about defaulting it directly in the CLI, is it makes the documentation and configurations kinda funny.

You are trying to kinda say: "if you want to use an automatically assigned number use -1, that's the true default although we don't display it, I know the CLI says the default is 6 but that's only kinda true and it depends on the machine."

ericcurtin: Changed

ericcurtin:
I think the macOS failure shows we are not ready for this change, need to let the:

"--threads"

option in llama-run percolate around a bit a little more.

ericcurtin:
@rhatdan this is ready, I'm not so confident our builds will go red in any PR at the moment, builds have gotten super flakey.

Issue #980, opened 2025-03-18
llama-server images mismatch between entrypoint and inherited env variable


benoitf:
I've noticed that MODEL_PATH is hardcoded in the image
https://github.com/containers/ramalama/blob/e5249621fb96222be4a49ef01a8452e8b869be85/container-images/ramalama/Containerfile#L3

while the entrypoint expect that the user provides the path and report an error message
https://github.com/containers/ramalama/blob/e5249621fb96222be4a49ef01a8452e8b869be85/container-images/scripts/llama-server.sh#L11

but MODEL_PATH is always set due to the parent env definition

the env variable should not be defined in this image so entrypoint could report an error asking to pass the environment variable of Model path

PR #984, opened 2025-03-18
Default whisper-server.sh, llama-server.sh to /mnt/models/model.file

rhatdan:
Fixes: https://github.com/containers/ramalama/issues/980

## Summary by Sourcery

Bug Fixes:
- Fixes the issue where the MODEL_PATH was not being correctly set, causing the scripts to fail.

### DIFF
--- a/container-images/scripts/llama-server.sh
+++ b/container-images/scripts/llama-server.sh
@@ -9,15 +9,13 @@ if [ -n "${MODEL_CHAT_FORMAT}" ]; then
 fi
 
 if [ -n "${MODEL_PATH}" ]; then
-    eval llama-server \
-        --model "${MODEL_PATH}" \
-        --host "${HOST:=0.0.0.0}" \
-        --port "${PORT:=8001}" \
-        --gpu_layers "${GPU_LAYERS:=0}" \
-	"${CHAT_FORMAT}"
-    exit 0
+    MODEL_PATH="/mnt/models/model.file"
 fi
-
-echo "Please set a MODEL_PATH"
-exit 1
+eval llama-server \
+     --model "${MODEL_PATH}" \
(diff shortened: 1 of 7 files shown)


rhatdan:
@benoitf PTAL

container-images/scripts/llama-server.sh
```diff
@@ -9,15 +9,13 @@ if [ -n "${MODEL_CHAT_FORMAT}" ]; then
...
 
 if [ -n "${MODEL_PATH}" ]; then
-    eval llama-server \
-        --model "${MODEL_PATH}" \
-        --host "${HOST:=0.0.0.0}" \
-        --port "${PORT:=8001}" \
-        --gpu_layers "${GPU_LAYERS:=0}" \
-	"${CHAT_FORMAT}"
-    exit 0
+    MODEL_PATH="/mnt/models/model.file"
```
benoitf, line 12: the issue is that the condition

```
if [ -n "${MODEL_PATH}" ]; then
    MODEL_PATH="/mnt/models/model.file"
fi
```

is never reached as the image has env variable defined in its Containerfile

https://github.com/containers/ramalama/blob/main/container-images/ramalama/Containerfile#L3

so either:

1. llama-cpp-server / whisper-image should unset this default env var binding 
or
2. no need to check the case where MODEL_PATH is undefined as it's always defined

rhatdan: Well  not all of the Containerfiles have that set, I am thinking of removing the MODEL_PATH setting from the Containerfiles that do have it.

Also this would allow users to specify an alternative MODEL_PATH environment with ramalama run --env MODEL_PATH=/PATH

benoitf: yes that would be the best way

benoitf (approved):

PR #985, opened 2025-03-18
whisper.cpp requires ffmpeg

ericcurtin:
Installing ffmpeg-free from Fedora/EPEL, ffmpeg-free includes only the FOSS/patent free bits of ffmpeg.

## Summary by Sourcery

Chores:
- Install ffmpeg-free in the container images.

### DIFF
--- a/container-images/scripts/build_llama_and_whisper.sh
+++ b/container-images/scripts/build_llama_and_whisper.sh
@@ -21,6 +21,89 @@ dnf_remove() {
   dnf -y clean all
 }
 
+dnf_install_asahi() {
+  dnf copr enable -y @asahi/fedora-remix-branding
+  dnf install -y asahi-repos
+  dnf install -y mesa-vulkan-drivers "${vulkan_rpms[@]}" "${rpm_list[@]}"
+}
+
+dnf_install_cuda() {
+  dnf install -y "${rpm_list[@]}" gcc-toolset-12
+  # shellcheck disable=SC1091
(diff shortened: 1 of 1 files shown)


ericcurtin:
@rhatdan @benoitf I think this should build everywhere, the only one I'm uncertain about is maybe openeuler, don't know what their repo layout looks like.

Maybe you wanna test/review @leo-pony ?

ericcurtin:
I also added a function:

`rm_non_ubi_repos`

That removes all non-UBI repos as soon as we've installed a package from the non-UBI place. The goal is to at least be deliberate about including non-UBI content.

rhatdan:
LGTM

leo-pony:
> @rhatdan @benoitf I think this should build everywhere, the only one I'm uncertain about is maybe openeuler, don't know what their repo layout looks like.
> 
> Maybe you wanna test/review @leo-pony ?

Thanks for the information. There is a problem with openeuler. The openeuler source contains the ffmpeg package, but not the ffmpeg-free package. However, the ffmpeg package in the openeuler source is LGPL/GPL. Since whisper has already used ffmpeg (GPL/LGPL), can the LGPL/GPL of the ffmpeg package meet ramalama's requirements?

ericcurtin:
The non-openeuler will continue to use ffmpeg-free.

leo-pony:
> The non-openeuler images will continue to use ffmpeg-free.
Thanks, I'll submit a PR soon to use ffmpeg for openeuler.

PR #999, opened 2025-03-20
Remove Ollama from macOS build

ericcurtin:
We still will test this stuff on Linux.

## Summary by Sourcery

CI:
- Removes the installation of Ollama on macOS.


ericcurtin:
@kush-gupt might have a better solution

PR #1003, opened 2025-03-20
Don't use relative paths for destination

rhatdan:
This can cause the file to not be installed in a subdir of the destination (/docs) within the container.

## Summary by Sourcery

Bug Fixes:
- Fixes an issue where files were not being installed in the correct subdirectory within the container due to the use of relative paths.


ericcurtin (approved):

bmahabirbu (approved):
LGTM

PR #1008, opened 2025-03-20
Use this container if we detect ROCm accelerator

ericcurtin:
Should help get ROCm + vLLM working

## Summary by Sourcery

Use the ROCm vLLM container image when a ROCm accelerator is detected.


ericcurtin:
Got all sorts of errors when it tried to load .gguf file

rhatdan:
LGTM

PR #1013, opened 2025-03-21
Improve UX for ramalama-client

ericcurtin:
I couldn't figure out why things weren't printing word by word yesterday as I was going for an evening walk it dawned on me, we were not flushing the buffers. Also adds color to the response like llama-run.

## Summary by Sourcery

Improves the user experience of ramalama-client by flushing the buffers to ensure words are printed word by word and adding color to the response.

Enhancements:
- Flushes the buffers to ensure words are printed word by word.
- Adds color to the response.

### DIFF
--- a/libexec/ramalama-client-core
+++ b/libexec/ramalama-client-core
@@ -23,6 +23,8 @@ def request_data(prompt, use_completions):
 
 
 def req(prompt):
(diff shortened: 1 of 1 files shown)


rhatdan (approved):
LGTM

PR #1014, opened 2025-03-21
Keep conversation history

ericcurtin:
Don't treat every prompt like a separate prompt, keep the conversation history.

## Summary by Sourcery

Enable the client to maintain conversation history across multiple prompts.


rhatdan:
LGTM

PR #1016, opened 2025-03-21
Rag condition should be and instead of or

ericcurtin:
We want both of these things to be true to execute rag functionality.

## Summary by Sourcery

Change the RAG condition to be an AND instead of an OR.


ericcurtin:
@bmahabirbu PTAL

rhatdan:
LGTM

PR #1017, opened 2025-03-21
Color each word individually

ericcurtin:
We don't want the color yellow to leak into the terminal if the process dies suddenly.

## Summary by Sourcery

Bug Fixes:
- Fixes a bug where the terminal color could persist if the process dies suddenly, leading to unintended color display in the terminal.


rhatdan:
LGTM

PR #1020, opened 2025-03-22
Add "ramalama open-webui"

ericcurtin:
ramalama open-webui granite3-moe for an example will just spin up a whole inferencing evironment, including inferencing server and webui server.

## Summary by Sourcery

New Features:
- Introduce a new CLI command 'open-webui' to spin up an inferencing environment, including an inferencing server and a web UI server.

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -163,6 +163,7 @@ def configure_subcommands(parser):
     list_parser(subparsers)
     login_parser(subparsers)
     logout_parser(subparsers)
+    open_webui_parser(subparsers)
     perplexity_parser(subparsers)
     pull_parser(subparsers)
     push_parser(subparsers)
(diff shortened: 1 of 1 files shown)


ramalama/cli.py
```diff
@@ -898,6 +899,13 @@ def version_parser(subparsers):
     parser.set_defaults(func=print_version)
 
 
+def open_webui_parser(subparsers):
+    parser = subparsers.add_parser(
+        "open-webui",
+        help="run open-webui",
```
rhatdan, line 905: Not quite sure where this is going, but I would just call this webui?  Is this launching a webservice or launching a web browser?

ericcurtin: It would run a ramalama server and a open webui podman container like in the below documentation and connect them together:

https://github.com/open-webui/docs/pull/455/files

I'm ok if this feature gets shot down FWIW

rhatdan:
I am going to close this PR, since I don't think we are going forward with it.  If you still believe we should then lets open an issue on it.

PR #1023, opened 2025-03-24
Fix up building of images

rhatdan:
## Summary by Sourcery

Fixes the image building process by correctly setting the install prefix and ensuring that ramalama and RAG are built with the correct prefix. Also installs wheel package to avoid errors when installing other packages. Removes the entrypoint script from the intel-gpu container.

Build:
- Fixes the image building process by correctly setting the install prefix.
- Installs wheel package to avoid errors when installing other packages.
- Removes the entrypoint script from the intel-gpu container.

### DIFF
--- a/container-images/scripts/build_llama_and_whisper.sh
+++ b/container-images/scripts/build_llama_and_whisper.sh
@@ -116,8 +116,8 @@ dnf_install() {
                   "procps-ng" "git" "dnf-plugins-core" "libcurl-devel" "gawk")
   local vulkan_rpms=("vulkan-headers" "vulkan-loader-devel" "vulkan-tools" \
                      "spirv-tools" "glslc" "glslang")
-  if [[ "${containerfile}" = "ramalama" ]] || [[ "${containerfile}" =~ rocm* ]] || \
-    [[ "${containerfile}" = "vulkan" ]]; then # All the UBI-based ones
+  if [ "${containerfile}" = "ramalama" ] || [[ "${containerfile}" =~ rocm* ]] || \
+    [ "${containerfile}" = "vulkan" ]; then # All the UBI-based ones
     if [ "${ID}" = "fedora" ]; then
       dnf install -y "${rpm_list[@]}"
     else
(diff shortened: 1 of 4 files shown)


container-images/scripts/build_llama_and_whisper.sh
```diff
@@ -270,17 +269,16 @@ main() {
   set -ex
 
   local containerfile="$1"
-  local install_prefix
+  install_prefix=$(set_install_prefix)
```
ericcurtin, line 272: I had been trying to keep a mantra, scope all variable in this script, even if it's local in main.

I'm actually curious why we needed to change the install_prefix stuff.

This change is fine, I just wonder why

container-images/scripts/doc2rag
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
```
ericcurtin, line 1: Good catch, some containers or OSes in the past didn't like "/usr/bin/python3"

container-images/scripts/build_llama_and_whisper.sh
```diff
@@ -253,14 +253,13 @@ clone_and_build_ramalama() {
...
   git submodule update --init --recursive
-  pip install . --prefix=/usr
+  python3 -m pip install . --prefix="$1"
   cd ..
   rm -rf ramalama
 }
 
 build_rag() {
-    python3 -m pip install qdrant_client fastembed openai fastapi uvicorn
-    rag_framework load
```
ericcurtin, line 263: What does "rag_framework load" actually do?

I fixed this another why, by copying rag_framework into the container image

container-images/scripts/build_llama_and_whisper.sh
rhatdan: Not sure it is necessary. I will add it back. I was trying to get the pip installs correct and they were not using the install_prefix, but I think it should work with local now.

container-images/scripts/build_llama_and_whisper.sh
rhatdan: It pulls images required for doc2rag, so it should only be called in the -rag containerfiles, not in all containerfiles.

container-images/scripts/doc2rag
ericcurtin: Adding a check for this to "make validate"

PR #1032, opened 2025-03-25
Minor bugfix remove self. from self.prompt

ericcurtin:
Not needed in non-constructor scope

## Summary by Sourcery

Removes unnecessary 'self.' prefix from a variable within a non-constructor scope.


rhatdan:
LGTM

PR #1039, opened 2025-03-26
Typo in the webui

ericcurtin:
Somebody noticed a typo in the built-in webui in llama.cpp, it was fixed in upstream llama.cpp. This just ensures we get the fix downstream!

## Summary by Sourcery

Chores:
- Update llama.cpp to ef19c71769681a0b3dde6bc90911728376e5d236


rhatdan:
LGTM

PR #1046, opened 2025-03-26
Never use entrypoint

rhatdan:
Turn off all use of entrypoints when running and serving containers.
    
    Entrypoints have the chance of screwing up the way containers run, and
    if a user provides their own image with an entrypoint this could become
    tough to diagnose errors.

### DIFF
--- a/ramalama/model.py
+++ b/ramalama/model.py
@@ -198,6 +198,7 @@ class Model(ModelBase):
             name,
             "--env=HOME=/tmp",
             "--init",
+            "--entrypoint=[]",
         ]
 
     def add_privileged_options(self, conman_args, args):


ericcurtin:
I'm not sure we ever want to use an entrypoint when execute from RamaLama CLI

rhatdan:
It is debateable, it does give you a good idea that you have the right container when running on a Cuda system.  But we do not do similar in other platforms.

ericcurtin:
I encountered an issue in out vllm container images where one has an entrypoint and one doesn't.

Happy to merge this regardless, might be worth checking out the sorcery list vs string concern though

rhatdan:
After further thought, I agree entrypoints bring a risk, and should be disabled by default.

ericcurtin:
String vs list issue fixed too, nice

ericcurtin:
We need to work on build stability before this gets out of hand

PR #1048, opened 2025-03-26
Install openvino whereever we can

ericcurtin:
It's only available in Rawhide x86_64


ericcurtin:
@rhatdan could we get a quick merge here? It's breaking other PRs

ericcurtin:
In reality this won't work in any of our container images right now as none of them are rawhide AFAIK

rhatdan:
We can try pip install openvino?

rhatdan:
openvino is available on Fedora 42.

ericcurtin:
> We can try pip install openvino?

Yeah I guess, I wonder should we do it only for x86_64 we only package it for x86_64 🤔 

Have to head offline.

rhatdan:
Going to close this, since we just want to remove openvino for now, and try other ways.

PR #1051, opened 2025-03-27
fixed rag argument error

bmahabirbu:


bmahabirbu:
Some build reds do to openvino but should be good to merge

RamaLama_rag_gyrc9nqe/RamaLama_rag_gshthiz3/.lock
```diff
@@ -0,0 +1 @@
+tmp lock file
```
ericcurtin, line 1: Need to get rid of the junk files

ericcurtin: If this branch is rebased after you remove the leftover files should get past the build issues

rhatdan:
Lets go with https://github.com/containers/ramalama/pull/1049 for the build fix.

But lets get the other fix from this PR merged.

rhatdan:
The build RAG issue is fixed but the other fix in this PR looks reasonable, rebase and resubmit.

Issue #988, opened 2025-03-18
jinja mode not enabled when using custom chat template in llama-server image


benoitf:
I have an example using a custom model chat

https://github.com/containers/ai-lab-recipes/tree/main/recipes/natural_language_processing/function_calling

It is giving MODEL_CHAT_FORMAT environment variable there
https://github.com/containers/ramalama/blob/787f3558b0c8fdd781173e39b06be89880c4be36/container-images/scripts/llama-server.sh#L8

the chat format is `chatml`

the server is starting well.

But then, when performing queries on the REST endpoint, I receive an error from llama-server that ninja is not specified:  'message': 'tools param requires --jinja flag', 'type': 'server_error'}}

![Image](https://github.com/user-attachments/assets/b4990100-df0c-4f31-9f9b-d70abbecefef)

so in case of a custom chat-format, maybe llama-server should be started with `--jinja` flag

ericcurtin:
I think we should work with upstream llama.cpp on llama-server to ensure we use jinja by default and fallback to the other template technique built into llama.cpp

Anything else we do would end up being a short-term fix.

engelmi:
I agree. This should be in llama.cpp - passing `--chat-template` should implicitly set `--jinja` if this is required. For example, in [llama-run](https://github.com/ggml-org/llama.cpp/blob/master/examples/run/run.cpp#L182-L186) we automatically turn the `use_jinja` option on when passing `--chat-template-file`.

rhatdan:
For now should we turn on the --jinja flag when user specifies --chat-template?

PR #1053, opened 2025-03-27
feat: add --jinja to the list of arguments if MODEL_JINJA env var is true

benoitf:
fixes https://github.com/containers/ramalama/issues/988

## Summary by Sourcery

Add support for the --jinja flag when MODEL_JINJA environment variable is set

New Features:
- Introduce conditional --jinja flag based on MODEL_JINJA environment variable in the llama-server.sh script

Bug Fixes:
- Resolve issue #988 by adding flexibility to enable Jinja templating for model chat formats


ericcurtin (approved):

ericcurtin:
This file isn't checked in our CI, we need to work on build stability again after the recent sprint to get rag out

PR #1059, opened 2025-03-27
fixed chunk error

bmahabirbu:
## Summary by Sourcery

Bug Fixes:
- Resolved a chunk-related error in the doc2rag script


bmahabirbu:
if running doc2rag local it errors out without this fix, fixing now to avoid errors after a new build

PR #1061, opened 2025-03-27
Don't display server port when using run --rag

rhatdan:
## Summary by Sourcery

Modify the serving port display behavior to suppress port information when running in RAG mode

Bug Fixes:
- Prevent unnecessary port information from being printed when using run --rag

Enhancements:
- Add a quiet parameter to control port information display during server startup

### DIFF
--- a/ramalama/model.py
+++ b/ramalama/model.py
@@ -654,9 +654,9 @@ class Model(ModelBase):
                 )
             raise NotImplementedError(file_not_found % {"cmd": exec_args[0], "error": str(e).strip("'")})
 
-    def serve(self, args):
+    def serve(self, args, quiet=False):
         self.validate_args(args)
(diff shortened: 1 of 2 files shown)


ramalama/model.py
```diff
@@ -774,5 +774,6 @@ def compute_serving_port(port: str, debug: bool) -> str:
 
     if target_port == 0:
         raise IOError("no available port could be detected. Please ensure you have enough free ports.")
-    print(f"serving on port {target_port}")
+    if not quiet:
+        print(f"serving on port {target_port}")
```
ericcurtin, line 778: You could simplify this and delete this completely if you wanted. llama-server and vllm tend to display the IP address and port anyway

ericcurtin (approved):

Issue #1040, opened 2025-03-26
[MacOS] GPU acceleration not enabled when `RAMALAMA_CONTAINER_ENGINE` is set


kpouget:
When I launch `ramalama` with `RAMALAMA_CONTAINER_ENGINE` (because my `podman` isn't in the `PATH`), `ramalama` doesn't set the GPU acceleration flag `podman_machine_accel` and the `--ngl` flag isn't passed to `llama.cpp`

---

Actual result
------------------

```
$ ramalama --debug run llama3.2
exec_cmd:  podman run --rm -i --label ai.ramalama --name ramalama_lPHimFZsfF --env=HOME=/tmp --init --security-opt=label=disable --cap-drop=all --security-opt=no-new-privileges --label ai.ramalama.model=llama3.2 --label ai.ramalama.engine=podman --label ai.ramalama.runtime=llama.cpp --label ai.ramalama.command=run --env LLAMA_PROMPT_PREFIX=🦭 >  --pull=newer -t --network none --mount=type=bind,src=.../.local/share/ramalama/models/ollama/llama3.2:latest,destination=/mnt/models/model.file,ro quay.io/ramalama/ramalama:latest llama-run -c 2048 --temp 0.8 -v /mnt/models/model.file
```

Expected result
------------------------

```
$ sed -i 's/podman_machine_accel = False/podman_machine_accel = True/g' ramalama/common.py
```

```
$ ramalama --debug run llama3.2
exec_cmd:  podman run --rm -i --label ai.ramalama --name ramalama_8Kd8tghKmk --env=HOME=/tmp --init --security-opt=label=disable --cap-drop=all --security-opt=no-new-privileges --label ai.ramalama.model=llama3.2 --label ai.ramalama.engine=podman --label ai.ramalama.runtime=llama.cpp --label ai.ramalama.command=run --env LLAMA_PROMPT_PREFIX=🦭 >  --pull=newer -t --device /dev/dri --network none --mount=type=bind,src=/Users/kevinpouget/.local/share/ramalama/models/ollama/llama3.2:latest,destination=/mnt/models/model.file,ro quay.io/ramalama/ramalama:latest llama-run -c 2048 --temp 0.8 -v --ngl 999 /mnt/models/model.file

load_tensors: offloading 28 repeating layers to GPU
load_tensors: offloading output layer to GPU
load_tensors: offloaded 29/29 layers to GPU
load_tensors:   CPU_Mapped model buffer size =   308.23 MiB
[...]

PR #1065, opened 2025-03-28
Fix handling of $RAMALAMA_CONTAINER_ENGINE

rhatdan:
apple_vm has a side effect of setting the podman_machine_accel global variable which is used when running and serving models. Currently if the user sets RAMALAMA_CONTAINER_ENGINE to podman in an alternative path the apple_vm code is not called, so the global variable is not set.

Fixes: https://github.com/containers/ramalama/issues/1040

## Summary by Sourcery

Improve container engine detection and handling for Podman on macOS, ensuring proper configuration when using custom Podman paths

Bug Fixes:
- Fix detection of Podman when using a custom path on macOS by modifying the apple_vm function to accept the engine path

Enhancements:
- Update get_engine() to handle Podman path detection more flexibly on macOS

### DIFF
--- a/ramalama/common.py
+++ b/ramalama/common.py
@@ -42,9 +42,12 @@ podman_machine_accel = False
 def get_engine():
     engine = os.getenv("RAMALAMA_CONTAINER_ENGINE")
     if engine is not None:
+        if os.path.basename(engine) == "podman" and sys.platform == "darwin":
+            # apple_vm triggers setting global variable podman_machine_accel side effect
+            apple_vm(engine)
         return engine
 
-    if available("podman") and (sys.platform != "darwin" or apple_vm()):
+    if available("podman") and (sys.platform != "darwin" or apple_vm("podman")):
         return "podman"
 
     if available("docker") and sys.platform != "darwin":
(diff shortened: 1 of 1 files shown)


rhatdan:
@ericcurtin PTAL
The docker blow ups are not related to this PR, I have a fix PR up to handle the Docker issues.

ericcurtin:
Our logic doesn't seem to line up with the documentation FWIW:

```
  --engine ENGINE       run RamaLama using the specified container engine.
                        The RAMALAMA_CONTAINER_ENGINE environment variable modifies default behaviour.
```
                   
Seems like we treat RAMALAMA_CONTAINER_ENGINE environment variable differently to --engine.

We should try and clean this up at some point and take a clear precedence:

CLI argument flag takes preference over
environment variable which takes preference over
config file

and set the actual option early in execution and respect it, it's sort of messy right now.

ericcurtin (approved):

kpouget:
you guys are fast :D 
LGTM, I tested it and solves [the issue](https://github.com/containers/ramalama/issues/1040) I observed in my environment 👍🏻

kpouget:
in my perf test harness ([TOPSAIL](https://github.com/openshift-psap/topsail/)), I learned to rely on end to end testing to move fast. It's a team-internal, mostly single-person, non-critical repo, so not the same context, but having just (at least) one sanity test being executed before merging anything helped me ensuring the consistency of the code base

Coderabbit.ai (same as sourcery.ai it seems) is also very good at detecting silly coding mistakes :D

ericcurtin:
coderabbit.ai vs sourcery.ai could be interesting to compare I don't think we explored coderabbit.ai

copilot also does the same thing, I haven't been able to get copilot to review automatically though, only on request when you click the button which is not great. Preferably you want the review bots to comment on PRs even when people sleep, no clicks.

PR #1066, opened 2025-03-28
Docker running of containers is blowing up

rhatdan:
Seems Docker does not support
--entrypoint=[] like podman.  --entrypoint "" seems to work on both platforms.

When attempting to mimic --pull=newer on Docker we were pulling the wrong image, we should be attempting to pull the accellerated image not the default.

For some reason llama-run --threads X is blowing up in a docker container with the option not being supported. This could be something being masked inside of Docker containers that is not masked inside of podman containers. Someone who understands what llama-run is doing with the --threads option would need to look further into this.

This should fix the issue in CI that is blowing up Docker tests.

## Summary by Sourcery

Fix Docker container running issues by adjusting container configuration and image pulling behavior

Bug Fixes:
- Correct Docker-specific container entrypoint configuration to use an empty string instead of an empty list
- Fix image pulling mechanism to use the accelerated image when running on Docker
- Disable --threads option for Docker containers to prevent runtime errors

Enhancements:
- Improve container engine compatibility between Docker and Podman
- Add conditional handling for Docker-specific container runtime limitations

Chores:
- Refactor container configuration logic to handle different container engines more robustly

### DIFF
--- a/ramalama/model.py
+++ b/ramalama/model.py
@@ -199,7 +199,8 @@ class Model(ModelBase):
             name,
             "--env=HOME=/tmp",
             "--init",
-            "--entrypoint=[]",
+            "--entrypoint",
+            "",
         ]
 
     def add_privileged_options(self, conman_args, args):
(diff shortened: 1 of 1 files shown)


ramalama/model.py
```diff
@@ -386,7 +388,11 @@ def gpu_args(self, args, runner=False):
 
             gpu_args += [f'{args.ngl}']
 
-        if args.threads != -1:
+        # for some reason the --threads option is blowing up on Docker,
```
ericcurtin, line 391: This is probably something worth visiting something. It sometimes can take while for podman/docker to pull new images. I notice on podman on my machine an updated images can be posted to quay.io and podman with --pull=newer seems to not recognise there's a new "latest" image. Don't know if there's caching somewhere or what.

Even when I delete all the images and re-pull from scratch sometimes it still pulls n-1 ":latest" image.

ericcurtin (approved):

PR #1068, opened 2025-03-28
Introduce wrapper for serve and run

ericcurtin:
We are coming to the limits of what we can do in a "podman run" line. Create wrapper functions so we can do things like forking processes and other similar things that you need to do inside a container in python3. There are some features coming up where rather than upstreaming separate solutions to all our engines like vLLM and llama.cpp we want to solve the problem in the python3 layer.

The "if True:"'s will remain for a while, we may need to wait for containers to be distributed around the place before we turn things on.

## Summary by Sourcery

Introduce a wrapper mechanism for serve and run commands to provide more flexibility in container execution and prepare for future enhancements

New Features:
- Add wrapper scripts for core operations (ramalama-run-core, ramalama-serve-core) to enable more advanced container execution

Enhancements:
- Implement a configurable wrapper mechanism with a USE_RAMALAMA_WRAPPER flag
- Extend exec_cmd function to support stderr redirection
- Add flexibility in locating and executing wrapper scripts

Build:
- Update setup.py to include libexec/ramalama directory and wrapper scripts
- Modify installation script to support new libexec directory structure

Tests:
- Update system tests to expect new wrapper script names in various generation scenarios

Chores:
- Modify RPM spec file to include libexec ramalama scripts


ramalama/model.py
```diff
@@ -530,7 +530,12 @@ def build_exec_args_run(self, args, model_path, prompt):
         if EMOJI and "LLAMA_PROMPT_PREFIX" not in os.environ:
             os.environ["LLAMA_PROMPT_PREFIX"] = "🦙 > "
 
-        exec_args = ["llama-run", "-c", f"{args.context}", "--temp", f"{args.temp}"]
+        if True:
```
kpouget, line 533: instead of `if True`, you may want to "prepare" the way for the next step:
```
# at the top level
USE_RAMALAMA_IN_CONTAINER = False
...
if USE_RAMALAMA_IN_CONTAINER:
    exec_args = ["ramalama-run-core", "--backend", "llama.cpp"]
else:
    exec_args = ["llama-run"]
```

ericcurtin: True I'll use some variable (see what I did there)

"if True:" is also driving static analysis and the bots crazy, even though it's deliberate :)

The if statement will be deleted in a couple of weeks

ericcurtin:
@bmahabirbu @rhatdan this is ready for review, to use one must flip a boolean flag to True. I think it needs to stay this way for a release or two, to leave new container images get around.

libexec/ramalama/ramalama-serve-core
```diff
@@ -1,9 +1,15 @@
 #!/usr/bin/env python3
 
+import argparse
 import sys
 
 
 def main(args):
```
ericcurtin, line 7: This will come into play more in future PRs, it will become the proxy server for llama.cpp, vllm, etc.

ericcurtin:
@rhatdan @bmahabirbu if this goes green I'd like to get it in

PR #1069, opened 2025-03-28
args.engine can be None in this code path

ericcurtin:
The code will then crash, ensure args.engine has a True value of some type.

## Summary by Sourcery

Bug Fixes:
- Prevent potential runtime error by adding a check for args.engine before accessing its basename


ericcurtin:
I hope we are close to green builds again

ericcurtin:
@bmahabirbu PTAL

maxamillion (approved):
LGTM 👍

ericcurtin:
Wow a green build, some sanity is restored :)

PR #1088, opened 2025-04-01
Remove files with colon in their name

engelmi:
The verify_checksum unit tests use files with a colon in their name. This causes issues for Windows machines since file names/paths can not contain this symbol. Therefore, these files have been removed and the tests create these on the fly and only when not run on Windows machines.

## Summary by Sourcery

Modify verify_checksum unit tests to create test files dynamically and avoid issues with file names containing colons on Windows

Bug Fixes:
- Resolve cross-platform compatibility issue with file naming in unit tests

Tests:
- Refactor verify_checksum test to dynamically create test files instead of using pre-existing files with colons in their names
- Add logic to skip tests with colon-containing file names on Windows platforms


axel7083 (approved):
LGTM

ericcurtin (approved):

PR #1102, opened 2025-04-02
Be verbose about pulling image

### LINKED ISSUE #1043: UI output when downloading Docker containers
When I run `ramalama run smollm` I get *zero* output.  Same for `ramalama --debug --dryrun run smollm`.  Looking at my system, I can see that ramalama is running `docker pull -q quay.io/ramalama/ramalama` but it would be nice to print a message to the user!

edmcman:
Re-close #1043

## Summary by Sourcery

Improve image pulling verbosity by adding an optional quiet mode for image version retrieval

Bug Fixes:
- Provide more informative output when attempting to pull a specific image version

Enhancements:
- Add a quiet parameter to control verbosity during image pulling process

### DIFF
--- a/ramalama/common.py
+++ b/ramalama/common.py
@@ -503,13 +503,13 @@ def accel_image(config, args):
     image = images.get(gpu_type, args.image)
     if hasattr(args, "rag") and args.rag:
         image += "-rag"
-    if attempt_to_use_versioned(conman, image, vers, args.debug):
+    if attempt_to_use_versioned(conman, image, vers, args.quiet, args.debug):
         return f"{image}:{vers}"
 
     return f"{image}:latest"
 
 
-def attempt_to_use_versioned(conman, image, vers, debug):
+def attempt_to_use_versioned(conman, image, vers, quiet, debug):
     try:
         # check if versioned image exists locally
         if run_cmd([conman, "inspect", f"{image}:{vers}"], ignore_all=True, debug=debug):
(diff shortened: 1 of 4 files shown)


ericcurtin:
Not sure about this one, the docker/podman info should be enough

edmcman:
> Not sure about this one, the docker/podman info should be enough

At least with docker, nothing is printed to the screen.  I ran `ramalama serve` and the only output was "serving on port 8080".  But it was actually silently pulling the new docker image.

rhatdan:
The Podman output on a MAC does not show up.  So this is probably good.

rhatdan:
A couple of tests need to be fixed.

ericcurtin:
> > Not sure about this one, the docker/podman info should be enough
> 
> At least with docker, nothing is printed to the screen. I ran `ramalama serve` and the only output was "serving on port 8080". But it was actually silently pulling the new docker image.

Man that sucks, these images are GBs, should really have progress bars for pulls that large

edmcman:
Hmm, I am having to change a lot of these tests.  Should we be pulling docker images when `--dryrun` flag is set?

ericcurtin:
In my opinion we should print the line on --dryrun, no pulling

rhatdan (approved):
LGTM

rhatdan:
Images should never be  pulled when running --dryrun

PR #1106, opened 2025-04-02
Enable --jinja for all llama-server instances

ericcurtin:
--jinja has been around a few months now, enable it everywhere.

## Summary by Sourcery

Enhancements:
- Unconditionally add the --jinja flag to the execution arguments for llama-server instances

### DIFF
--- a/ramalama/model.py
+++ b/ramalama/model.py
@@ -605,6 +605,7 @@ class Model(ModelBase):
                 f"{args.context}",
                 "--temp",
                 f"{args.temp}",
+                "--jinja",
             ] + args.runtime_args
             if chat_template_path != "":
                 exec_args.extend(["--chat-template-file", chat_template_path])


ericcurtin:
@edmcman @engelmi will this break any models?

maxamillion (approved):
+1

edmcman:
I'm not sure.  @ochafik is there a reason `--jinja` isn't enabled by default?

rhatdan:
LGTM

rhatdan:
Can you find an example of this?  Do you have an model that breaks?

benoitf:
before you were able to provide the `--chat-template llama2`  and then this template was working

now, as --jinja is provided, it'll try to consider this chat template as a ninja template and fails the output (garbage output)

it implies that you would need to provide now a ninja chat template

rhatdan:
SO --chat-template should imply --jinja=false?

benoitf:
it's a little bit more complex
it's how the template is interpreted

using `--jinja` flag, it means you need to provide a .jinja file as template like one from this template directory
https://github.com/ggml-org/llama.cpp/tree/f01bd02376f919b05ee635f438311be8dfc91d7c/models/templates

but people are also mostly using --chat-template with the non-jinja template and as RamaLama is setting --jinja by default, we can't no longer provide a non-jinja template

ericcurtin:
There's multiple template formats, jinja is kinda the best one and often embedded in .gguf files

PR #1107, opened 2025-04-02
No longer install podman remote and openvino in general images

rhatdan:
Remove other packages that are not necessary when running the
containers.

Remove leftover build_rag load command and move openvino to only the
intel container.

## Summary by Sourcery

Streamline container image build script by removing unnecessary packages and build steps

Bug Fixes:
- Clean up unnecessary package installations to reduce image size and complexity

Enhancements:
- Simplify the build script by removing unused components like podman-remote and RamaLama

Chores:
- Remove unnecessary build dependencies and packages from the container image build process

### DIFF
--- a/container-images/scripts/build_llama_and_whisper.sh
+++ b/container-images/scripts/build_llama_and_whisper.sh
@@ -17,7 +17,13 @@ dnf_install_intel_gpu() {
 dnf_remove() {
   dnf remove -y \
       python3-devel \
-      libcurl-devel
+      libcurl-devel \
+      git \
+      gcc \
+      gcc-c++ \
+      make \
+      cmake \
+      findutils
   dnf -y clean all
 }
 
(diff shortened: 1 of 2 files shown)


benoitf (approved):
if whisper could also be only in the whisper image (and not in the llama-server image) and the opposite (llama-server not included in the whispercpp image)

container-images/scripts/build_llama_and_whisper.sh
```diff
@@ -257,21 +263,6 @@ clone_and_build_llama_cpp() {
   rm -rf llama.cpp
 }
 
-clone_and_build_ramalama() {
-  # link podman-remote to podman for use by RamaLama
```
ericcurtin, line 261: I'm actually using this in another PR for the libexec scripts

rhatdan:
I am not sure if whisper-cpp adds a lot different then adding llama-cpp, @ericcurtin what do you think?

container-images/scripts/build_llama_and_whisper.sh
rhatdan: You using RamaLama or using podman-remote?

ericcurtin: ramalama, not podman remote

rhatdan: What feature of RamaLama are you needing? It is pretty small, so we can probably leave it.  But no Container support since Podman-remote is kind of chunky.

 podman-remote                      x86_64    5:5.4.1-1.fc42                      updates-testing        33.7 MiB

ericcurtin:
> I am not sure if whisper-cpp adds a lot different then adding llama-cpp, @ericcurtin what do you think?

If this question is about size, no whisper.cpp is small

ericcurtin (approved):

Issue #904, opened 2025-03-03
Converting image generates (non-removable): oci://<none>:<none> images


dwrobel:
Converting image from `ollama` to `ramalama` generates the following images:
```
$ ramalama ls | grep none
oci://<none>:<none>                       3 minutes ago 8.75 KB
oci://<none>:<none>                       6 minutes ago 8.75 KB
```

which can't be deleted:
```
$ ramalama rm 'oci://<none>:<none>'
Error: parsing reference "<none>:<none>": invalid reference format
Error: parsing reference "<none>:<none>": invalid reference format
Error: Command '['podman', 'rmi', '--force=False', '<none>:<none>']' returned non-zero exit status 125.
```

`ramalama` version:

```$ rpm -qv python3-ramalama
python3-ramalama-0.6.2-1.fc40.noarch
```

Command which generates `oci://<none>:<none>`  images:
```
$ ramalama convert file:///home/dw/.ollama/models/blobs/sha256-19314ef0159c739868860d0ee15851e7b19a0433a94c1e0afa727a8a013bd0fd pllum8b-instr-q4km
```

The file path I got from `ollama`:
```
$ ollama show --modelfile pllum8b-instr-q4km
# Modelfile generated by "ollama show"
# To build a new Modelfile based on this, replace FROM with:
# FROM pllum8b-instr-q4km:latest

FROM /home/dw/.ollama/models/blobs/sha256-19314ef0159c739868860d0ee15851e7b19a0433a94c1e0afa727a8a013bd0fd
TEMPLATE {{ .Prompt }}
```

```
$ rpm -qv ollama
ollama-0.4.4-4.fc40.x86_64
```

rhatdan:
Try this with the `podman image prune --force` command, this are not AI Models, but container images.  

We might need to add a ramalama rmi and or ramalama-prune command, right now we are relying on users using Docker and Podman to clean up container images.

PR #1110, opened 2025-04-02
Don't leak intermediate OCI image when converting model to OCI

rhatdan:
Fixes: https://github.com/containers/ramalama/issues/904

We are currently leaking a <none><none> image every time we convert an image with ramalama convert.

There will still be a <none><none> image but this is assocated with the Manifest list created for the OCI image.

## Summary by Sourcery

Improve OCI image conversion by optimizing container build process to reduce intermediate image leakage

Bug Fixes:
- Prevent leaking of unnecessary <none><none> images during model conversion process

Enhancements:
- Modify container build command to minimize intermediate image creation
- Add Podman-specific build flag to reduce layer generation

### DIFF
--- a/ramalama/oci.py
+++ b/ramalama/oci.py
@@ -194,18 +194,21 @@ LABEL {ociimage_car}
                 c.write(model_car)
             else:
                 c.write(model_raw)
+        build_cmd = [
+            self.conman,
+            "build",
+            "--no-cache",
+            "--network=none",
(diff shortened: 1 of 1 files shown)


rhatdan:
@flouthoc @nalind FYI

flouthoc (approved):
LGTM

ericcurtin (approved):

PR #1111, opened 2025-04-02
Fix all links to ramalama-cuda

rhatdan:


rhatdan:
@maxamillion PTAL

ericcurtin:
Note someone cloned ramalama on Windows recently and the "git clone" failed.

I remember doing a "git clone" on windows in the past and symlinks broke the clone because it's a Unix concept.

Windows does have some alternate concept of a symlink though, it may be fixed now

rhatdan:
The Windows clone issue was caused by files with a `:` in the name.

Issue #1103, opened 2025-04-02
Command-line parsing problem with `ramalama --use-model-store serve`


edmcman:
```
ramalama --use-model-store serve file:///home/ed/.cache/llama.cpp/bartowski_Qwen2.5-7B-Instruct-GGUF_Qwen2.5-7B-Instruct-Q4_K_M.gguf
serving on port 8080
Traceback (most recent call last):
  File "/home/ed/.local/bin/ramalama", line 101, in <module>
    main(sys.argv[1:])
  File "/home/ed/.local/bin/ramalama", line 74, in main
    args.func(args)
  File "/home/ed/Projects/ramalama/./ramalama/cli.py", line 851, in serve_cli
    model.serve(args)
  File "/home/ed/Projects/ramalama/./ramalama/model.py", line 667, in serve
    model_path = self.get_model_path(args)
  File "/home/ed/Projects/ramalama/./ramalama/model.py", line 497, in get_model_path
    model_path = self.pull(args)
  File "/home/ed/Projects/ramalama/./ramalama/url.py", line 52, in pull
    return self._pull_with_model_store()
  File "/home/ed/Projects/ramalama/./ramalama/url.py", line 112, in _pull_with_model_store
    self.store.new_snapshot(tag, snapshot_hash, files)
  File "/home/ed/Projects/ramalama/./ramalama/model_store.py", line 441, in new_snapshot
    self._ensure_chat_template(model_tag, snapshot_hash, snapshot_files)
  File "/home/ed/Projects/ramalama/./ramalama/model_store.py", line 425, in _ensure_chat_template
    info: GGUFModelInfo = GGUFInfoParser.parse("model", "registry", model_file_path)
TypeError: GGUFInfoParser.parse() missing 1 required positional argument: 'cli_args'
```

engelmi:
Hi @edmcman, 
PR https://github.com/containers/ramalama/pull/1113 should fix this. PTAL

PR #1113, opened 2025-04-03
Removed unused cli_args param from GGUF parse function

engelmi:
Fixes: https://github.com/containers/ramalama/issues/1103

Removed unused cli_args param from GGUFInfoParser.parse function which caused also the call in the model store to fail since it wasn't passed in.

/cc @edmcman

### DIFF
--- a/ramalama/gguf_parser.py
+++ b/ramalama/gguf_parser.py
@@ -160,7 +160,7 @@ class GGUFInfoParser:
             return value
         raise ParseError(f"Unknown type '{value_type}'")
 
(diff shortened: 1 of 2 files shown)


edmcman:
Seems fine to me.  You could add a test to be sure.

ericcurtin (approved):

ericcurtin:
Test in a follow on PR would be good

engelmi:
I agree. Unfortunately, dont have much capacity to work on ramalama lately.

PR #1123, opened 2025-04-06
Add ability to pull via hf://user/repo:tag syntax

### LINKED ISSUE #1057: Feature request: Support huggingface://user/repo:tag format for huggingface models
For example, to run a HF quant on llama.cpp, I run `llama-cli -hf mradermacher/ToolACE-2-Llama-3.1-8B-GGUF:Q4_K_M`.  It would be nice if the huggingface transport supported that syntax.  In many cases, this is probably equivalent to appending ".gguf", but not for models split into multiple files.

edmcman:
Closes #1057

## Summary by Sourcery

Enhance Hugging Face model pulling capabilities by adding support for user/repo:tag syntax and improving the model retrieval process

New Features:
- Add support for pulling Hugging Face models using the hf://user/repo:tag syntax
- Implement a more robust model retrieval mechanism with fallback options

Enhancements:
- Refactor the Hugging Face model pulling logic to handle different retrieval scenarios
- Improve error handling for model pulls

Chores:
- Rename ollama's init_pull function to ollama_repo_pull for better clarity

### DIFF
--- a/ramalama/huggingface.py
+++ b/ramalama/huggingface.py
@@ -8,6 +8,7 @@ import urllib.request
 from ramalama.common import available, download_file, exec_cmd, generate_sha256, perror, run_cmd, verify_checksum
 from ramalama.model import Model
 from ramalama.model_store import SnapshotFile, SnapshotFileType
+from ramalama.ollama import ollama_repo_pull
 
 missing_huggingface = """
 Optional: Huggingface models require the huggingface-cli module.
@@ -192,11 +193,8 @@ class Huggingface(Model):
                 repo_name = self.directory + "/" + model_name
                 registry_head = f"{Huggingface.REGISTRY_URL}{repo_name}"
 
-                # XXX: refactor
-                from ramalama.ollama import init_pull
-
                 show_progress = not args.quiet
-                return init_pull(
+                return ollama_repo_pull(
                     os.path.join(args.store, "repos", "huggingface"),
                     Huggingface.ACCEPT,
                     registry_head,
--- a/ramalama/huggingface.py
+++ b/ramalama/huggingface.py
@@ -144,13 +144,6 @@ def handle_repo_info(repo_name, repo_info, runtime):
             f"- https://huggingface.co/models?other=base_model:quantized:{repo_name} \n"
             "- https://huggingface.co/spaces/ggml-org/gguf-my-repo"
         )
-    if "gguf" in repo_info:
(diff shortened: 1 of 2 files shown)


edmcman:
This should work for GGUFs specified by filename or tag.

**What should happen if ramalama is pointed to a non-GGUF repository?**  Is this supported with the vllm runtime?

rhatdan:
We want to support this with alternate runtimes include vllm, so yes this is supported.

edmcman:
Can you provide an example command for this?

rhatdan:
Nope, currently we are not supporting our own version of vllm, perhaps the upstream version of vllm.
@ericcurtin WDYT?

rhatdan:
I think the sourcery-ai has good feedback, I think you should make those changes.

ericcurtin:
@rhatdan yes I think if someone prepares a Fedora-based vLLM image and it works decently well, we might build it.

ramalama/huggingface.py
```diff
@@ -187,18 +185,46 @@ def pull(self, args):
...
-                repo_name = self.directory + "/" + self.filename
-                repo_info = get_repo_info(repo_name)
-                handle_repo_info(repo_name, repo_info, args.runtime)
 
+                model_name, model_tag, _ = self.extract_model_identifiers()
+                repo_name = self.directory + "/" + model_name
+                registry_head = f"{Huggingface.REGISTRY_URL}{repo_name}"
+
+                show_progress = not args.quiet
+                return ollama_repo_pull(
```
engelmi, line 197: On first sight, it seems off to use an ollama function inside huggingface. If this is common functionality, could we extract it to a better place?

engelmi (commented):
It would be great to add this to the model store pull as well:
https://github.com/containers/ramalama/blob/main/ramalama/huggingface.py#L341-L371
This would enable downloading multiple files, e.g. a `config.json` (containing meta information such as chat template) specified in the manifest.

edmcman:
> It would be great to add this to the model store pull as well: https://github.com/containers/ramalama/blob/main/ramalama/huggingface.py#L341-L371 This would enable downloading multiple files, e.g. a `config.json` (containing meta information such as chat template) specified in the manifest.

I will look at this when I have a chance.  I did not immediately do this because I wasn't sure how to change it.

rhatdan:
Let's refactor in a separate PR.

PR #1128, opened 2025-04-07
Give --image option higher precedence over config variables

ueno:
While the manual page mentions that the --image option can be used to override the default image, the option only has effect when the value is not equal to the hard-coded default ("quay.io/ramalama/ramalama"), meaning that it is not possible to easily enforce CPU-only acceleration when a GPU is detected:
```console
  $ ramalama --image quay.io/ramalama/ramalama run deepseek
  Attempting to pull quay.io/ramalama/rocm:0.7...
```
This patch removes the check and slightly changes the logic to determine the image: first try the --image option, then the "images" configuration, and finally the "image" configuration and the RAMALAMA_IMAGE envvar.

## Summary by Sourcery

Modify the image selection logic to give the --image command-line option higher precedence when determining the container image to use

New Features:
- Implement a more flexible image selection hierarchy: command-line option, config 'images', then default image

Bug Fixes:
- Fix issue where --image option was not always effective in overriding the default image

Enhancements:
- Improve image selection logic to prioritize command-line image option over configuration variables

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -112,7 +112,6 @@ Needed to access gpu on some systems, but has security implications.""",
     )
     parser.add_argument(
         "--image",
-        default=CONFIG["image"],
         help="OCI container image to run with the specified AI model",
     )
     parser.add_argument(
--- a/ramalama/common.py
+++ b/ramalama/common.py
@@ -483,9 +483,6 @@ def rm_until_substring(input, substring):
 
 
 def accel_image(config, args):
-    if args.image != DEFAULT_IMAGE:
-        return args.image
-
     env_vars = get_accel_env_vars()
 
     if not env_vars:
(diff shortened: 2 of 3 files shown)


ueno:
Sorry, I realized that this need a bit more tweaks as the images supplied with `--image` or "image" may be suffixed with a version (e.g., ":latest").

rhatdan:
How about just doing:
https://paste.centos.org/view/87801339

ueno:
@rhatdan thank you; that would resolve the immediate issue, though after a second though, I suspect it might be more consistent if:
- the default behavior (i.e., choose the image based on the hardware) can be overridden in this precedence order: `--image` , RAMALAMA_IMAGE, and "image" in config
- if any of those is selected and if it includes a tag, use it as-is
- otherwise, append a tag with the later logic in `accel_image`

What do you think?

rhatdan:
1. If image is specified by user --image=FOO, then use it.
2. If image is specified by RAMALAMA_IMAGE then use it.
3. If image is specified by config then use it.
4. Otherwise generate use accerated image with RamaLama Minor Version tag

Makes the most sense to me.

rhatdan:
BTW I lover the unit tests to cement this.

ueno:
@rhatdan thank you, I opened a new PR for testing (#1142).

PR #1138, opened 2025-04-07
Build fix container image parsing

ericcurtin:
There's two cases we enter this function at the end of a rag command or when manually specifing a container as inferencing engine.

## Summary by Sourcery

Fix container image parsing logic to handle image tag specification more robustly

Bug Fixes:
- Improve handling of container image tag specification to correctly process images with and without explicit tags

Enhancements:
- Simplify image tag resolution logic in the accel_image function


rhatdan:
LGTM

Issue #1147, opened 2025-04-08
ramalama new rocm image is broken due to missing libhipblas.so.2


drizzt:
```
tredaell@aldebaran ~ ₿ podman images               
REPOSITORY             TAG         IMAGE ID      CREATED       SIZE
quay.io/ramalama/rocm  latest      319e4ad6981c  22 hours ago  2.92 GB
tredaell@aldebaran ~ ₿ ramalama --debug run granite
run_cmd:  podman inspect quay.io/ramalama/rocm:0.6
Working directory: None
Ignore stderr: False
Ignore all: True
exec_cmd:  podman run --rm -i --label ai.ramalama --name ramalama_HPmSyppEIA --env=HOME=/tmp --init --security-opt=label=disable --cap-drop=all --security-opt=no-new-privileges --label ai.ramalama.model=ollama://granite3.1-dense --label ai.ramalama.engine=podman --label ai.ramalama.runtime=llama.cpp --label ai.ramalama.command=run --env LLAMA_PROMPT_PREFIX=🦭 >  --pull=newer -t --device /dev/dri --device /dev/kfd -e HIP_VISIBLE_DEVICES=0 --network none --mount=type=bind,src=/home/tredaell/.local/share/ramalama/models/ollama/granite3.1-dense:latest,destination=/mnt/models/model.file,ro quay.io/ramalama/rocm:latest llama-run -c 2048 --temp 0.8 -v --ngl 999 /mnt/models/model.file
llama-run: error while loading shared libraries: libhipblas.so.2: cannot open shared object file: No such file or directory
tredaell@aldebaran ~ ₿ 
```

PR #1148, opened 2025-04-08
Removing git breaks rocm images

rhatdan:
Rocm requres a couple of -devel packages which require git.

Removing git removed these packages.

## Summary by Sourcery

Bug Fixes:
- Prevent breaking rocm images by ensuring git package is not removed, as it is required for -devel package dependencies


rhatdan:
Fixes: https://github.com/containers/ramalama/issues/1147

PR #1151, opened 2025-04-08
feat: Add ramalama client command with basic implementation

ericcurtin:
## Summary by Sourcery

Add a new 'client' subcommand to the Ramalama CLI for interacting with AI Model servers

New Features:
- Introduce a new 'client' CLI command that allows users to interact with AI Model servers by specifying a host and optional arguments

Enhancements:
- Refactor the get_cmd_with_wrapper method to be a standalone function in the common module
- Modify the CLI to support a new client subcommand with flexible argument handling

Documentation:
- Add man page documentation for the new ramalama-client command, explaining its usage and providing an example


.gitignore
```diff
@@ -13,4 +13,5 @@ venv/
 *.container
 *.image
 *.volume
-__pycache__/
\ No newline at end of file
+__pycache__/
+.aider*
```
ericcurtin, line 17: I actually used this aider tool to rewrite a lot of the boilerplate associated with introducing a new command.

rhatdan:
What kind of commands is the user going to send to the server?  Is this replacing `ramalama run`?

ericcurtin:
> What kind of commands is the user going to send to the server? Is this replacing `ramalama run`?

"ramalama run" runs client + server and terminates both on completion. This runs only the client, it's useful for testing servers.

rhatdan:
Might be useful for testing but difficult for us to support.  Perhaps make it hidden for now, or need to document that this is only for experimenting at this point.

ericcurtin:
Added "(experimental)", I could leave it as an undocumented option possible, would have to see if argparse allows that

PR #1158, opened 2025-04-09
Turn on client/server implementation of run

ericcurtin:
Now that we've had one release with the wrapper scripts included in the container images it should be safe to turn this on everywhere.

## Summary by Sourcery

Enable the client/server implementation of run by removing conditional wrapper script usage

Enhancements:
- Simplify the model execution logic by removing conditional checks for wrapper script usage

Chores:
- Remove the `USE_RAMALAMA_WRAPPER` flag and its associated conditional logic

### DIFF
--- a/libexec/ramalama/ramalama-client-core
+++ b/libexec/ramalama/ramalama-client-core
@@ -72,7 +72,9 @@ def req(conversation_history, url, color):
             response = urllib.request.urlopen(request)
             break
         except Exception:
-            print(f"\r{c}", end="", flush=True)
+            if sys.stdout.isatty():
+                print(f"\r{c}", end="", flush=True)
+
             time.sleep(i)
             i *= 2
 
--- a/ramalama/common.py
+++ b/ramalama/common.py
@@ -12,6 +12,7 @@ import shutil
 import string
 import subprocess
 import sys
+import sysconfig
 import time
 import urllib.error
 from typing import List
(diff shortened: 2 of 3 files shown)


rhatdan:
Once we turn this on, we need to bump to 0.8.0 and warn people to test it out. Then allow people to continue to work with 0.7.* for a while, when we work out the issues.

ericcurtin:
This might pass a build in future, the problem is our CI pulls the latest container image from quay.io so we typically have to wait a week or two for changes like:

https://github.com/containers/ramalama/pull/1242

to hit CI. Open issue on this:

https://github.com/containers/ramalama/issues/1248

ericcurtin:
This is needed for this PR:

https://github.com/containers/ramalama/pull/1250

rhatdan:
If we drop a ramalama.conf file into place, then we can change the default to
pull="missing" rather then pull="newer"

This would only pull images if they did not exists locally on the CI/CD system.

ericcurtin:
Rebasing, seeing where we are again

ericcurtin:
@rhatdan this is good to go now

ericcurtin:
Hit a flake, rebasing and running CI again

ericcurtin:
@bmahabirbu @swarajpande5 @cgruver @engelmi @maxamillion @jhjaggars PTAL

cgruver (approved):

PR #1160, opened 2025-04-09
Also hardcode version into version.py as fallback

ericcurtin:
This way we should always return the current version

## Summary by Sourcery

Hardcode a fallback version into version.py to ensure a consistent version is always returned

Bug Fixes:
- Ensure version() function always returns a valid version, even when package metadata is not found

Tests:
- Update CI version checks to verify the version is consistent across version.py

### DIFF
--- a/test/ci.sh
+++ b/test/ci.sh
@@ -31,6 +31,7 @@ version_checks() {
   version=$(get_version)
   grep "version.*$version" setup.py
   grep "version.*$version" rpm/python-ramalama.spec
+  grep "version.*$version" ramalama/version.py
 
   local arg_llama_cpp
   arg_llama_cpp=$(get_arg_llama_cpp)
--- a/.github/workflows/install_ramalama.yml
+++ b/.github/workflows/install_ramalama.yml
@@ -34,14 +34,7 @@ jobs:
             exit 1
           fi
 
-      - name: Run RamaLama Installer (Ubuntu)
-        if: matrix.os == 'ubuntu-latest'
-        run: |
-          chmod +x install.sh
-          sudo ./install.sh -l
-
-      - name: Run RamaLama Installer (macOS)
-        if: matrix.os == 'macos-latest'
+      - name: Run RamaLama Installer
         run: |
           chmod +x install.sh
           ./install.sh -l
--- a/install.sh
+++ b/install.sh
@@ -1,7 +1,1831 @@
 #!/bin/bash
 
-cleanup() {
-  rm -rf "$TMP" &
+# Licensed under the MIT license
+# <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+usage() {
+    # print help (this cat/EOF stuff is a "heredoc" string)
+    cat <<EOF
+uv-installer.sh
+
+The installer for uv 0.5.9
+
+This script detects what platform you're on and fetches an appropriate archive from
+https://github.com/astral-sh/uv/releases/download/0.5.9
+then unpacks the binaries and installs them to the first of the following locations
+
+    \$XDG_BIN_HOME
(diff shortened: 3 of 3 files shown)


test/ci.sh
```diff
@@ -31,6 +31,7 @@ version_checks() {
   version=$(get_version)
   grep "version.*$version" setup.py
   grep "version.*$version" rpm/python-ramalama.spec
+  grep "version.*$version" ramalama/version.py
```
ericcurtin, line 34: I hope this part of the CI is actually running, parts of our CI tend to disappear at times.

rhatdan: We need to make this easier, currently this constant is defined in 3 other places.

pyproject.toml, setup.py, rpm/python-ramalama.spec

Need a release script top populate everything.

rhatdan:
I am still worried about the install.sh script not handling python the correct way, basically why is the python project not being coffigured correctly?  Are we missing something else that could cause issues.  pip install and brew install are handling it properly.

ericcurtin:
The problem with pip is it doesn't work on a lot of platforms Ubuntu/macOS. brew solves macOS. I did install something called aider recently via:

https://github.com/Aider-AI/aider/blob/main/aider/website/install.sh

script. They seem to use this new kid on the block uv:

https://github.com/astral-sh/uv

which is getting quite popular quickly. Might be worth embedding uv in the install script, generally recommending it over pip. Then we just have to maintain the rpm and the pypi package.

WDYT @rhatdan ?

ericcurtin:
We could try something like this also:

`pip3 install --break-system-packages -U --force-reinstall ramalama`

the -U means user which is supposed to not make it overwrite system files. Install things in places like ~/.local/bin

rhatdan:
I am fine with uv, but it does mean you have to install it.

The best thing would be to get to native installers for Mac and Windows.

ericcurtin:
@rhatdan if you want to try it out I embedded uv in the install script here. The first thing it does is install a portable uv binary for your given platform

I like it because it seems like the most portable pypi solution around, not because it's new and shiny FWIW.

rhatdan:
Now that you fixed the version, we will have to fix the tests to use the minor version.

rhatdan:
This needs a rebase.

ericcurtin:
Removing the available ramalama check, install outputs like this:

```
_____                       _
 |  __ \                     | |
 | |__) |__ _ _ __ ___   __ _| |     __ _ _ __ ___   __ _
 |  _  // _` | '_ ` _ \ / _` | |    / _` | '_ ` _ \ / _` |
 | | \ \ (_| | | | | | | (_| | |___| (_| | | | | | | (_| |
 |_|  \_\__,_|_| |_| |_|\__,_|______\__,_|_| |_| |_|\__,_|

downloading uv 0.5.9 x8[6](https://github.com/containers/ramalama/actions/runs/14450413331/job/40521437910?pr=1160#step:5:7)_64-unknown-linux-gnu
no checksums to verify
installing to /root/.local/bin
  uv
  uvx
uv is installed!
Resolved 2 packages in 42ms
Prepared 2 packages in 3[7](https://github.com/containers/ramalama/actions/runs/14450413331/job/40521437910?pr=1160#step:5:8)8ms
Installed 2 packages in 5ms
 + argcomplete==3.6.2
 + ramalama==0.7.3
Installed 1 executable: ramalama
warning: `/root/.local/bin` is not on your PATH. To use installed tools, run `export PATH="/root/.local/bin:$PATH"` or `uv tool update-shell`.

To add $HOME/.local/bin to your PATH, either restart your shell or run:

    source $HOME/.local/bin/env (sh, bash, zsh)
    source $HOME/.local/bin/env.fish (fish)
Installation failed. Please check the previous output for errors.
```

there will be a manual step for some to add x directory to PATH, but that will be printed.

ericcurtin:
@rhatdan this is good to go I think, failed on flakey test

rhatdan:
LGTM

PR #1166, opened 2025-04-10
Update llama.cpp add llama 4

ericcurtin:
To pick up llama 4 support among other things.

## Summary by Sourcery

Enhancements:
- Update llama.cpp to a newer commit to potentially add support for Llama 4 and other improvements


rhatdan:
LGTM

PR #1169, opened 2025-04-10
Build images for llama-stack

rhatdan:
## Summary by Sourcery

Add build support for llama-stack container image and update release and build scripts to handle the new image type

New Features:
- Create a new Containerfile for llama-stack image based on Fedora, installing llama-stack and llama_stack-client via pip

Build:
- Modify container build script to add conditional handling for llama-stack image
- Update release script to include specific handling for llama-stack image

### DIFF
--- a/container-images/llama-stack/Containerfile
+++ b/container-images/llama-stack/Containerfile
@@ -1,0 +1,6 @@
+FROM registry.fedoraproject.org/fedora:42
+
+RUN dnf -y install pip && \
+    pip install -U llama-stack --prefix=/usr
+
+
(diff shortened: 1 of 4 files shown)


container-images/llama-stack/Containerfile
```diff
@@ -0,0 +1,6 @@
+FROM registry.fedoraproject.org/fedora:42
+
+RUN dnf -y install pip && \
+    pip install -U llama-stack llama_stack-client --prefix=/usr
```
nathan-weinberg, line 4:
```suggestion
    pip install -U llama-stack --prefix=/usr
```

`llama-stack-client` is a dependency of `llama-stack`: https://github.com/meta-llama/llama-stack/blob/main/requirements.txt#L24

PR #1174, opened 2025-04-11
Add check for toolbox

ericcurtin:
If we are in toolbox, don't attempt to run nested containers. We then have to rely on the user to install llama.cpp in the container themselves. It's tempting to do an even more generic attempt to see if we are already inside a container, so we never attempt to do nested containers, whether toolbox, podman, docker, etc.


ericcurtin:
@bmahabirbu any ideas on how we can stabilise this rag test.

ramalama/common.py
```diff
@@ -46,6 +46,14 @@ def get_engine():
             apple_vm(engine)
         return engine
 
+    if (
```
rhatdan, line 49: This will break the use case of running RamaLama with podman-remote, which Podman Desktop uses for running RamaLama on Windows.

rhatdan: Just check for /run/.toolboxenv

ericcurtin: Done

rhatdan:
LGTM

PR #1180, opened 2025-04-12
Improve model store

engelmi:
This PR does two small improvements:

- Copying the change from https://github.com/containers/ramalama/pull/1164 to the `list` command in the model store
- Splitting the model source (e.g. ollama) so it prints `ollama://<path>` instead of `ollama/<path>`

## Summary by Sourcery

Improve the model store listing functionality by enhancing model name formatting and adding an option to control container model listing

Enhancements:
- Modify model name formatting to use source://path format
- Add a new parameter to control listing of container models

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -431,7 +431,9 @@ def get_size(path):
 def _list_models(args):
     mycwd = os.getcwd()
     if args.use_model_store:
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
We need to start testing with this by default.  I think it is time to throw the switch and run this by default (if we are not already).  After Monday's release.

ericcurtin:
A good first step would be to start a PR, might take a few hours to get the tests green

PR #1183, opened 2025-04-14
Create openvino model server image and add it quay.io/ramalama

bmahabirbu:
fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/2622

@rhatdan This will replace having openvino as part of the intelgpu container so we can remove it! 

I haven't tested if this container works properly yet I'm basing it off of the github repo https://github.com/openvinotoolkit/model_server/tree/main?tab=readme-ov-file for now specifically the docker.redhat file.


container-images/openvino/Containerfile
```diff
@@ -0,0 +1,19 @@
+FROM registry.access.redhat.com/ubi9/ubi:9.5
+
+RUN dnf install -y \
+    python3-devel \
+    libcurl-devel \
+    gcc \
```
rhatdan, line 6: Why does this need gcc, gcc-c++, make, cmake, libcurl-devel if this is just untaring executables into a containerfile?

bmahabirbu: Good catch I just copied some of the installs from the base ramalama in case this image needed it but we shouldn't need anything but python3. I'll plan on seeing if this container works and ill update it!

bmahabirbu: Updated and fixed to match offical container

container-images/openvino/Containerfile
```diff
@@ -0,0 +1,19 @@
+FROM registry.access.redhat.com/ubi9/ubi:9.5
+
+# Install Python development dependencies
+RUN yum install -y python3-devel wget compat-openssl11.x86_64
+RUN pip3 install "Jinja2==3.1.6" "MarkupSafe==3.0.2"
```
ericcurtin, line 5: We have python3-jinja2, python3-markupsafe rpms that could be worth considering

container-images/openvino/Containerfile
```diff
@@ -0,0 +1,19 @@
+FROM registry.access.redhat.com/ubi9/ubi:9.5
+
+# Install Python development dependencies
+RUN yum install -y python3-devel wget compat-openssl11.x86_64
```
ericcurtin, line 4: Please use dnf, yum is deprecated quite a while now. Can we remove ".x86_64" from this package? openvino has some arm support.

container_build.sh
```diff
@@ -101,11 +101,15 @@ build() {
...
       case ${target} in
-	  ramalama-cli | llama-stack)
+    # Exclude openvino
+	  ramalama-cli | llama-stack | openvino)
 	  ;;
 	  *)
 	      add_entrypoints "${conman[@]}" "${REGISTRY_PATH}"/"${target}"
-	      add_rag "${conman[@]}" "${target}"
+        # Exclude openvino for rag for now
+	      if [ "$target" != "openvino" ]; then
```
ericcurtin, line 110: If statement could be formatted properly

ericcurtin (approved):

rhatdan:
@bmahabirbu Erics comments above have not been fixed.

container_build.sh
bmahabirbu: Actually not completely sure what is wrong here?

container-images/openvino/Containerfile
bmahabirbu: Fixed!

container-images/openvino/Containerfile
bmahabirbu: Switched to dnf instead of pip for those packages!

bmahabirbu:
Just updated!

container_build.sh
bmahabirbu: I see now! The if statement isnt even needed if i add it up here `ramalama-cli | llama-stack | openvino)`

bmahabirbu: Fixed!

container_build.sh
```diff
@@ -101,11 +101,12 @@ build() {
       echo "${conman_show_size[@]}"
       "${conman_show_size[@]}"
       case ${target} in
-	  ramalama-cli | llama-stack)
+    # Exclude openvino
```
rhatdan, line 104: Remove comment.

container_build.sh
```diff
@@ -101,11 +101,12 @@ build() {
...
+	  ramalama-cli | llama-stack | openvino)
```
rhatdan, line 105: This is not tabbed correctly.

container_build.sh
```diff
@@ -101,11 +101,12 @@ build() {
...
 	  ;;
 	  *)
 	      add_entrypoints "${conman[@]}" "${REGISTRY_PATH}"/"${target}"
-	      add_rag "${conman[@]}" "${target}"
+        add_rag "${conman[@]}" "${target}"
```
rhatdan, line 109: Ditto.  Should blow up on `make validate`

container_build.sh
bmahabirbu: Fixed!

PR #1188, opened 2025-04-14
ux: add vllm runtime suggestion when downloading safetensors

nathan-weinberg:
currently when downloading safetensors and using llama.cpp ramalama suggests using a or converting to a GGUF format

this commit updates the text to inform users they can also switch to use the vLLM engine if they want to run safetensors

## Summary by Sourcery

Update the user guidance for safetensors model downloads to include vLLM runtime as an alternative to GGUF conversion

Bug Fixes:
- Improve user guidance when downloading safetensors models with llama.cpp runtime

Enhancements:
- Expand runtime recommendation for safetensors models to provide users with more flexibility in model execution

### DIFF
--- a/ramalama/huggingface.py
+++ b/ramalama/huggingface.py
@@ -220,7 +220,7 @@ def handle_repo_info(repo_name, repo_info, runtime):
     if "safetensors" in repo_info and runtime == "llama.cpp":
         print(
             "\nllama.cpp does not support running safetensor models, "
-            "please use a/convert to the GGUF format using:\n"
+            "please switch your runtime to vLLM or use a/convert to the GGUF format using:\n"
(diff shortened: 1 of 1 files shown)


ericcurtin:
Nack, will approve if there is a test that actually performs vllm + safetensors inferencing

nathan-weinberg:
> Nack, will approve if there is a test that actually performs vllm + safetensors inferencing

is this not already something that is tested or assumed to be supported? Since vLLM supports safetensors and is already a runtime available in RamaLama, I assumed so

nathan-weinberg:
Never got an answer and I'm unlikely to return to this

bmahabirbu:
Makes sense, I think we're looking past adding vllm support to Ramalama.

Could be worth having llama.cpp do the conversation from safe sensors to gguf im not sure if this feature was added

PR #1208, opened 2025-04-16
Ship nvidia and cann man pages

rhatdan:
## Summary by Sourcery

Add support for generating and installing man7 pages for NVIDIA and CANN documentation

New Features:
- Introduce a new function `generate_man7_pages()` to handle installation of man7 documentation pages

Build:
- Update setup.py to include man7 pages in the package data files installation process

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,6 +25,16 @@ def generate_man5_pages(share_path, docs):
     return data_files
 
 
+def generate_man7_pages(share_path, docs):
+    data_files = []
+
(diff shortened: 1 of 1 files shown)


lsm5 (approved):
> Consider consolidating the man page generation functions into a single function with a parameter for the man page type.

I would echo this comment by sourcery, but I see we're only extending on what's implemented earlier, so no biggie.

RPM builds seem to be going fine with this change, so LGTM AFAICT.

PR #1219, opened 2025-04-17
Tag images on push with digests, so they are permanent

rhatdan:
We have accidently overwridden the images release version if we also tag by digest, then we will not destroy the image or manifest list. Since Podman Desktop AI Lab Recipes relies on the image digest this makes it safer for them.


rhatdan:
@benoitf PTAL

benoitf (dismissed):
thanks a lot

scripts/release-image.sh
```diff
@@ -31,6 +31,7 @@ release-arm() {
 }
 
 release-ramalama() {
+    digest=$(podman image inspect "${REPO}/$1" --format '{{ .Digest }}' | cut -f2 -d':')
```
benoitf, line 34: I don't know how this script is used but here digest is not used ?

rhatdan: Oops, cut and paste failure. Added missing line.

benoitf (approved):

PR #1226, opened 2025-04-19
fix: intel-gpu-rag build

afazekas:
* intle-gpu the only rag user container with f41, moving to f42
* dependencies referenced by git url, adding git package
* numpy compile requires gcc-c++
* f42 has python3-sentencepiece same version (no compile) Fixes issues with several other rag containers too


container-images/intel-gpu/Containerfile
```diff
@@ -1,10 +1,10 @@
-FROM quay.io/fedora/fedora:41 as builder
```
ericcurtin, line 1: Did you test this build? Fedora 42 upgrade broke some of our builds last time, may not have been intel-gpu though

afazekas: Containers were building, not tested other thing.

container-images/scripts/build_rag.sh
```diff
@@ -12,9 +14,14 @@ $2"
     [ "$string" != "$(sort --version-sort <<< "$string")" ]
 }
 
-packages=""
+packages="git gcc gcc-c++"
```
afazekas, line 17: Add python-devel

container-images/scripts/build_llama_and_whisper.sh
```diff
@@ -18,8 +18,6 @@ dnf_remove() {
   dnf remove -y \
       python3-devel \
       libcurl-devel \
-      gcc \
```
afazekas, line 21: Need to not remove more.

afazekas: Per discussion the relative gain is small for not keeping smaller stiff, so just removing the whole statement for now.

container-images/intel-gpu/Containerfile
afazekas: $ REGISTRY_PATH=quay.io/ramalama_rpc time ./container_build.sh build

Successfully tagged quay.io/ramalama_rpc/rocm-ubi-rag:latest
2320d96ab988d76861833be1cab49de9a886d2f02867885ff2d297669fb38f2e
51689.13user 2185.82system 2:19:03elapsed 645%CPU (0avgtext+0avgdata 2345616maxresident)k
25355192inputs+887504640outputs (6412major+468807028minor)pagefaults 0swaps


Current PR makes **all** container build.

rhatdan (approved):
LGTM

PR #1227, opened 2025-04-19
Enable llama.cpp rpc feature in containers

afazekas:
Enable both server and client support for rpc.
The feature currently PoC in llama.cpp, but can work in practice. Required for distributed inference.

## Summary by Sourcery

Enable RPC support for llama.cpp in container build scripts

New Features:
- Add RPC server binary installation for llama.cpp in container build process

Build:
- Modify build flags to enable GGML RPC feature during llama.cpp compilation
- Update build script to install RPC server binary


afazekas:
"Client" on that picture is the llama-cli or llama-server, they can talk to multiple rpc-server ("backend services") which can use the same tensor ops as the "Client" from the llama.cpp. Each rpc-server can use completely different type of accelerator .

ericcurtin (approved):

PR #1228, opened 2025-04-19
[PoC] RAG support for `ramalama client` command

Tojaj:
This is a PoC that adds RAG support to the experimental `ramalama client` command.

I would like to know your opinion, if this is something worth pursuing?

When no `--rag` is specified the original behavior of running  `ramalama-client-core` directly on the host is preserved.
When `--rag` option is used, a container that mounts the specified RAG container is run. There in the container a newly added script called `ramalama-client-rag-core` is used.

The `ramalama-client-rag-core` script implements a Retrieval-Augmented Generation (RAG) client that enhances LLM responses with relevant context from a vector database.

Notes:
* In this PoC, I have extended the `client_cli()` method in place in the `cli.py` but I think it would worth to extract that one piece to a separate module.
* The `dry_run()` which is part of `client_cli()` is a 1:1 copy of the same function from `modules.py` so it would likely worth to move this one into `common.py`.
* IMHO going forward, it would make sense to merge `ramalama-client-rag-core`, `rag_framework` and `ramalama-client-core` into a single script as they all provide the similar functionality and merge would reduce maintenance overhead.

### Examples of usage

Knowledge used for building the vector DB in the RAG image:

    $ cat my_rag/knowledge.md
    # RAG Support in RamaLama

    RamaLama is a tool that facilitates local management and serving of AI Models. It provides several commands for working with models, including:

    - `ramalama run` - Run a model as a chatbot
    - `ramalama serve` - Serve a model as a REST API
    - `ramalama client` - Connect to an OpenAI-compatible API endpoint

    ## RAG Implementation

    Retrieval-Augmented Generation (RAG) is a technique that enables generative AI models to retrieve and incorporate new information. RamaLama supports RAG through the following commands:

    - `ramalama rag` - Generate a RAG vector database from documents
    - `ramalama run --rag` - Run a model with RAG support
    - `ramalama serve --rag` - Serve a model with RAG support
    - `ramalama client --rag` - Connect to an API with RAG support

    ## Client RAG Support

    The `ramalama client --rag` command allows users to connect to any OpenAI-compatible API endpoint while using a local RAG vector database to enhance queries with relevant context. This provides the benefits of RAG even when using remote API services.

    ### Usage

    ```
    ramalama client --rag <rag-database> <api-endpoint> [options]
    ```

    ### Example

    ```
    ramalama client --rag localhost/myrag:latest https://api.openai.com/v1
    ```

    This command will:
    1. Mount the RAG vector database
    2. Connect to the specified API endpoint
    3. Enhance queries with context from the RAG database
    4. Return responses that incorporate the retrieved context

Query to a hosted LLM model via API:
(`localhost/ramalama-rag:latest` is my dev image that includes the `ramalama-client-rag-core` script)

    $ ramalama --debug client --rag localhost/myrag:latest --image localhost/ramalama-rag:latest --api-key "$(cat OR_API_KEY.txt)" --model "openai/gpt-4o-mini" https://openrouter.ai/api/v1 "What RAG commands does RamaLama support?"
    run_cmd:  podman image inspect localhost/myrag:latest
    Working directory: None
    Ignore stderr: False
    Ignore all: False
    Command finished with return code: 0
    exec_cmd:  podman run --rm -i --label ai.ramalama --name ramalama_KR3igZLrV6 --env=HOME=/tmp --init --env=LLM_API_KEY=/REDACTED/ --network bridge -t --mount=type=image,source=localhost/myrag:latest,destination=/rag,rw=true localhost/ramalama-rag:latest ramalama-client-rag-core --model openai/gpt-4o-mini /rag/vector.db https://openrouter.ai/api/v1 "What RAG commands does RamaLama support?"
    RamaLama supports the following RAG commands:  
    - `ramalama rag`  
    - `ramalama run --rag`  
    - `ramalama serve --rag`  
    - `ramalama client --rag`


rhatdan:
Please sign all of your commits.

rhatdan:
@bmahabirbu PTAL

rhatdan:
@ericcurtin PTAL
@Tojaj Please take into account the changes suggesed by AI Surgery.

@Tojaj I like where this is going.  Might be worth breaking apart this PR into multiple different PRs and getting the low easier parts in quickly.

Long range goal is to wire everything up with llama-stack, BTW.

Tojaj:
@rhatdan

> Please sign all of your commits.

Ack, is signing with SSH key enough or do I need to go with GPG key?

> @Tojaj Please take into account the changes suggesed by AI Surgery.

Yeah, will do.

Tojaj:
> I also like where this is going, but I don't want to get into a scenario where `ramalama client` is forking from `ramalama run`... `ramalama run` is basically client + server that die on completion, please consolidate the code as much as we can.
> 
> We have a similar scenario in llama.cpp llama-run and llama-server kept forking away from each other to the point where we have stopped using llama-run.

@ericcurtin ack, can take a look on that. How would you envision the use of `run` for the `client` use-case in the terms of consolidation? This PR expands the already existing client command: https://github.com/containers/ramalama/pull/1151/files that has not been using `ramalama run` in any way.

bmahabirbu:
hi @Tojaj sorry for the late response! Ive been wrestling with this idea myself. I like what your doing here but do we want to have the same endpoint v1/chat/completions enhanced with RAG or have a seperate endpoint all together. Maybe it'll become clearer once I finish adding a few more features to our current version of RAG and ill get back to you

ericcurtin:
As an example `class RamaLamaShell` and `class RagClient` should just be one class.

Tojaj:
@bmahabirbu ack, let me know once you're done :+1: 

About the endpoint, in the OpenAI compatible API I think mainly the chat completions API is where RAG makes sense.

The use-case I wanted to cover is that a user has an access to a hosted model with OpenIA compatible API but without Responses API (i.e. no native support for RAG) so `ramalama client --rag` would be able to overcome the gap and would enable users to leverage the hosted model with their own data.

rhatdan:
We have moved to ramalama chat, which has some rag support.  Closing this since it seems like it is not being worked on.  Reopen if you are making progress.

PR #1231, opened 2025-04-20
Switch all f41 container to f42

afazekas:
Actually only 1 container left on f41,
making all to use f42.

## Summary by Sourcery

Update base container image from Fedora 41 to Fedora 42

Enhancements:
- Add sentencepiece-devel package to the container build

Chores:
- Migrate the llama-stack container to use the latest Fedora base image


rhatdan:
We tried this a couple of weeks ago, and something went wrong, although I do not recall what.  I will merge this for consistency and see if everything works now.

Issue #643, opened 2025-01-28
Login to huggingface with a token asks for username and password


pbabinca:
I would like to use a model from huggingface. Since the model is hidden behind authentication, I follow [examples section of Documentation for ramalama login](https://github.com/containers/ramalama/blob/main/docs/ramalama-login.1.md#examples) and run:

```
$ ramalama login --token="${TOKEN}" huggingface
```

but instead of logging me in, I'm being asked for `Username: `.

---

I have found a [trivial typo](https://github.com/containers/ramalama/blob/3e90238d9cafb8e4d29eaed45a4e3b8ae390fb01/ramalama/cli.py#L282) on this line:

```
if registry in ["ollama", "hf" "huggingface"]:
```

which python interprets as:

```
>>> repr(["ollama", "hf" "huggingface"])
"['ollama', 'hfhuggingface']"
```

so I assume it should be:

```
if registry in ["ollama", "hf", "huggingface"]:
```

---

This is not enough though, because the registry from the command line (`huggingface`) is not [one of the expected prefixes in factory function New()](https://github.com/containers/ramalama/blob/3e90238d9cafb8e4d29eaed45a4e3b8ae390fb01/ramalama/cli.py#L859). I'm not sure how to fix this properly, but temporary workaround with:

```
if model.startswith("huggingface") or model.startswith("huggingface://") or model.startswith("hf://") or model.startswith("hf.co/"):
```

led me to an error:

```
Error: huggingface-cli not available, skipping login.
```

I haven't seen a requirement of this CLI to be installed on the system anywhere in the documentation. I _guess_ best place would be in the examples section of `ramalama-login`, either using the command (`pip install -U "huggingface_hub[cli]"`) or linking to https://huggingface.co/docs/huggingface_hub/en/guides/cli where I found it.

---

After all of this I was able to login.

ericcurtin:
I think the best fix is to implement ramalama login in python3 from scratch it shouldn't be too bad, I've seen at least 3 implementations of it.

ericcurtin:
But yes a documentation addition to install huggingface would be enough to close this issue for now. It's deliberately not a strong dependancy because it can be hard to install it on some platforms.

Would you like to open a PR @pbabinca with a documentation update?

If you are really interested in solving this, a PR with the python code is the longer term fix.

pbabinca:
> Would you like to open a PR [@pbabinca](https://github.com/pbabinca) with a documentation update?

Opened #645 which should solve the last issue.

---

[...]

PR #1232, opened 2025-04-21
Fix bug in login_cli and update huggingface or hf registry behavior

melodyliu1986:
I was investigating the https://github.com/containers/ramalama/issues/643, and tried to write some unit tests for the ramalama/cli.py, then I was so confused that the test always failed no matter when I used "huggingface" or "hf". Finally, I found there is a tiny bug here, let me first fix it before moving on:

Before the change:
```
if registry in ["ollama", "hf" "huggingface"]:
        return registry
```

After the change:
```
if registry in ["ollama", "hf", "huggingface"]:
        return registry
```

## Summary by Sourcery

Bug Fixes:
- Corrected the syntax in the `normalize_registry` function by adding a missing comma in the registry list, which was causing incorrect registry validation

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -254,7 +254,7 @@ def normalize_registry(registry):
     if not registry or registry == "" or registry.startswith("oci://"):
         return "oci://"
 
-    if registry in ["ollama", "hf" "huggingface"]:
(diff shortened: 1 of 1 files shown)


rhatdan:
Thanks @melodyliu1986 
LGTM

PR #1236, opened 2025-04-21
Move model and rag to use shared Engine implementation

rhatdan:
Shrink the size of cly.py and model.py by moving all engine related functions into new engine.py python module.

## Summary by Sourcery

Refactor container engine interactions by introducing a new Engine class to centralize and simplify container management logic across the project

Enhancements:
- Create a new Engine class in engine.py to encapsulate container engine operations
- Reduce code duplication by centralizing container-related methods
- Simplify method calls for container management across model.py and rag.py

Chores:
- Remove redundant methods from model.py and rag.py
- Move container engine utility functions to a centralized location

### DIFF
--- a/ramalama/engine.py
+++ b/ramalama/engine.py
@@ -1,0 +1,272 @@
+import glob
+import json
+import os
+import subprocess
+import sys
+
+from ramalama.common import (
+    check_nvidia,
+    exec_cmd,
+    get_accel_env_vars,
+    perror,
+    podman_machine_accel,
+    run_cmd,
+)
+from ramalama.console import EMOJI
+
+
+class Engine:
+
+    def __init__(self, args):
+        self.exec_args = [
+            args.engine,
+            "run",
+            "--rm",
+        ]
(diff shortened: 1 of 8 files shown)


bmahabirbu:
`Could not load model jinaai/jina-embeddings-v2-small-en from any source.
Error: building at STEP "RUN /usr/bin/build_rag.sh cpu": while running runtime: exit status 1`

Could this be to to an internet error when starting the rag container?

ramalama/engine.py
```diff
@@ -0,0 +1,272 @@
...
+
+class Engine:
+
+    def __init__(self, args):
+        self.exec_args = [
+            args.engine,
+            "run",
+            "--rm",
+        ]
+        self.use_docker = os.path.basename(args.engine) == "docker"
```
ericcurtin, line 26: Super minor, but could call os.path.basename once. Micro-optimization

rhatdan: Let's merge and do that as a cleanup.

ericcurtin (approved):

Issue #1212, opened 2025-04-17
ramalama does not work with granite instruct model


tyll:
### Issue Description

Trying to run ramalama with granite instruct fails.

### Steps to reproduce the issue

Steps to reproduce the issue
1. ramalama run hf://ibm-granite/granite-8b-code-instruct-4k-GGUF/granite-8b-code-instruct.Q4_K_M.gguf
2. write `hello` into the prompt

### Describe the results you received

terminate called after throwing an instance of 'std::runtime_error'
  what():  this custom template is not supported

### Describe the results you expected

I should be able to interact with the model or get an actionable error message. 

### ramalama info output

```yaml
{
    "Accelerator": "cuda",
    "Engine": {
        "Info": {
            "host": {
                "arch": "amd64",
                "buildahVersion": "1.39.4",
                "cgroupControllers": [
                    "cpu",
                    "io",
                    "memory",
                    "pids"
                ],
                "cgroupManager": "systemd",
                "cgroupVersion": "v2",
                "conmon": {
                    "package": "conmon-2.1.13-1.fc42.x86_64",
                    "path": "/usr/bin/conmon",
                    "version": "conmon version 2.1.13, commit: "
                },
                "cpuUtilization": {
                    "idlePercent": 99.5,
                    "systemPercent": 0.18,
                    "userPercent": 0.31
                },
                "cpus": 24,
                "databaseBackend": "sqlite",
                "distribution": {
                    "distribution": "fedora",
                    "variant": "workstation",
                    "version": "42"
                },
                "eventLogger": "journald",
                "freeLocks": 2041,
                "hostname": "genius",
                "idMappings": {
                    "gidmap": [
                        {
                            "container_id": 0,
                            "host_id": 1000,
                            "size": 1
                        },
                        {
                            "container_id": 1,
                            "host_id": 524288,
                            "size": 65536
                        }
                    ],
                    "uidmap": [
                        {
                            "container_id": 0,
                            "host_id": 1000,
                            "size": 1
                        },
                        {
                            "container_id": 1,
                            "host_id": 524288,
                            "size": 65536
                        }
                    ]
                },
                "kernel": "6.14.2-300.fc42.x86_64",
                "linkmode": "dynamic",
                "logDriver": "journald",
                "memFree": 87334432768,
                "memTotal": 134821318656,
                "networkBackend": "netavark",
                "networkBackendInfo": {
                    "backend": "netavark",
                    "dns": {
[...]

PR #1241, opened 2025-04-22
Always use --jinja with run too

afazekas:
`ramalama serve` already using --jinja by default, `ramalama run` should do it too.

fixes: #1212

## Summary by Sourcery

Bug Fixes:
- Ensure consistent Jinja template handling between ramalama serve and ramalama run commands


afazekas:
Hmm, the test does not uses --temp 0, How can I recheck the test ?

rhatdan:
LGTM

PR #1242, opened 2025-04-22
Fix for using client/server version of "ramalama run"

ericcurtin:
When using containers we need this check so the code doesn't start going into some pulling logic.

## Summary by Sourcery

Bug Fixes:
- Add a direct check for existing local model path before attempting to retrieve model through existing methods


afazekas:
LGTM

rhatdan:
LGTM

PR #1246, opened 2025-04-23
preload docling models

bmahabirbu:
Adds the ability to preload the docling models for PDF conversion! I tested this locally, but running the code doesn't actually output any indication that it loads, although I tested it and it works. Would love a second tester to confirm it works! Thank you!

## Summary by Sourcery

New Features:
- Add capability to preload docling models before PDF conversion process


rhatdan:
Does this speed up the process of conversion?

PR #1247, opened 2025-04-23
Working on reducing ram of docling convert

bmahabirbu:
This code aims to parse a large pdf in single page batches. This however doesn't reduce the ram at all. Docling convert seems to keep adding vram/ram no matter if its one page at a time or loading the whole pdf at once.

https://github.com/docling-project/docling/issues/1343

Based on this and another similar issue it seems docling in a container has a lot of memory ram issues. I think I've exhausted all the straightforward fixes. Theres still some investigation that needs to be done for the pdf pipeline to reduce ram. Maybe i don't have to use easyocr for example.

Otherwise, until a fix has happened on the docling side, maybe I can add an error for large PDFs.

## Summary by Sourcery

Enhancements:
- Attempt to process large PDFs in single-page batches to address high memory usage, though this does not resolve the underlying RAM consumption issue.

### DIFF
--- a/container-images/scripts/doc2rag
+++ b/container-images/scripts/doc2rag
@@ -5,6 +5,9 @@ import hashlib
 import os
 import sys
 import uuid
+import gc
+from pypdf import PdfReader, PdfWriter
+import tempfile
 
 import qdrant_client
 from qdrant_client import models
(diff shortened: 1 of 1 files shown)


tyll:
There are tools like pdftohtml to convert PDF to a different format. Does it make sense to do this for large PDFs? Does docling handle other formats than PDF better?

rhatdan:
Do we have any docling experts we could tag with this issue?

PR #1250, opened 2025-04-23
Allow jinja argument

ericcurtin:
It's on everywhere anyway, this just ensures these wrapper scripts don't crash if jinja gets passed.

## Summary by Sourcery

Enhancements:
- Add support for Jinja argument to prevent script crashes

### DIFF
--- a/libexec/ramalama/ramalama-client-core
+++ b/libexec/ramalama/ramalama-client-core
@@ -130,6 +130,7 @@ def main(args):
         default=2048,
         help="size of the prompt context (0 = loaded from model)",
     )
+    parser.add_argument("--jinja", action="store_true", help="enable jinja")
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
We need to add some tests for the client server operations.

ericcurtin:
> LGTM We need to add some tests for the client server operations.

It's well tested already, it's a core part of RamaLama, trying very hard to make the tests green. Ever "run" is a client server operation.

ericcurtin:
> > LGTM We need to add some tests for the client server operations.
> 
> It's well tested already, it's a core part of RamaLama, trying very hard to make the tests green. Ever "run" is a client server operation.

These libexec scripts, I would never expect a user to run manually FWIW. It just makes the transition from host OS to "podman run" easier. We have really come to the limits of what we can stick in a "podman run".

PR #1255, opened 2025-04-23
Initial draft model support

afazekas:
Allows to pass draft model to serve and fetching it when needed. 'run' does not supports passing draft_model.
You should also pass draft related args tuned to your combination and do not forget to set the sampling parameters like top_k on the UI.

## Summary by Sourcery

Add support for draft models in the model serving and loading process, enabling users to specify and use draft models with additional configuration options.

New Features:
- Introduce ability to pass and use draft models during model serving
- Add new CLI arguments for specifying draft models and GPU layer offloading for draft models

Enhancements:
- Extend model factory to support draft model initialization
- Modify GPU argument handling to accommodate draft model configurations

Chores:
- Update configuration defaults to include draft model-related settings

### DIFF
--- a/docs/ramalama-serve.1.md
+++ b/docs/ramalama-serve.1.md
@@ -79,6 +79,15 @@ show this help message and exit
 #### **--host**="0.0.0.0"
 IP address for llama.cpp to listen on.
 
+#### **--model-draft**
+The model to be used for speculative decoding, typically significantly smaller than
+the main model with the same vocabulary. It proposes next tokens the main model verify
+the proposal cheaply in case it is good the generation will be faster.
+Server logs contains statistics about it's usefulness.
+Use --runtime-arg to pass the other draft model related parameters.
+Make sure you set the sampling parameters like top_k on the web UI correctly
+as well.
+
 #### **--name**, **-n**
 Name of the container to run the Model in.
 
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -868,6 +868,8 @@ def serve_parser(subparsers):
     parser.add_argument(
         "--rag", help="RAG vector database or OCI Image to be served with the model", completer=local_models
     )
+    parser.add_argument("--model-draft", help="Draft model", completer=local_models)
+
     parser.add_argument("MODEL", completer=local_models)  # positional argument
     parser.set_defaults(func=serve_cli)
 
--- a/ramalama/common.py
+++ b/ramalama/common.py
@@ -23,6 +23,7 @@ logging.basicConfig(level=logging.WARNING, format="%(asctime)s - %(levelname)s -
 
 MNT_DIR = "/mnt/models"
 MNT_FILE = f"{MNT_DIR}/model.file"
+MNT_FILE_DRAFT = f"{MNT_DIR}/draft_model.file"
 MNT_CHAT_TEMPLATE_FILE = f"{MNT_DIR}/chat_template.file"
 
 RAG_DIR = "/rag"
--- a/ramalama/model.py
+++ b/ramalama/model.py
@@ -10,6 +10,7 @@ from ramalama.common import (
     MNT_CHAT_TEMPLATE_FILE,
     MNT_DIR,
     MNT_FILE,
+    MNT_FILE_DRAFT,
     accel_image,
     check_metal,
     check_nvidia,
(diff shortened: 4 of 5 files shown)


ramalama/cli.py
```diff
@@ -956,6 +956,16 @@ def serve_parser(subparsers):
     parser.add_argument(
         "--rag", help="RAG vector database or OCI Image to be served with the model", completer=local_models
     )
+    parser.add_argument("--model_draft", help="Draft model", completer=local_models)
+    parser.add_argument(
```
ericcurtin, line 960: --ngld option is too much I think...

How many users will want a different --ngl value to --ngld value? Maybe 0.00001% will actually use this option.

I think just a `--model_draft` option is good for this PR and just use the --ngl value. We have to be careful of maintenance of the codebase.

We do not want as many options as llama-server .

ericcurtin:
> 'run' does not supports passing draft_model.

This is gonna change very soon with:

https://github.com/containers/ramalama/pull/1158

lets make sure this option is ready for when "ramalama run" moves to the client/server technique.

We also have to be mindful we will also likely allow users to switch from using other non-llama.cpp backends, so we don't have to implement every single llama.cpp-specific option.

Have we measured how much of a difference this option makes?

ramalama/cli.py
afazekas: ack

rhatdan:
Needs man page with a full description of what a draft model is.

afazekas:
> > 'run' does not supports passing draft_model.
> 
> This is gonna change very soon with:
> 
> #1158
> 
> lets make sure this option is ready for when "ramalama run" moves to the client/server technique.
> 
> We also have to be mindful we will also likely allow users to switch from using other non-llama.cpp backends, so we don't have to implement every single llama.cpp-specific option.
> 
> Have we measured how much of a difference this option makes?

It likely needs llama.cpp upstream change.

ericcurtin:
> > > 'run' does not supports passing draft_model.
> > 
> > 
> > This is gonna change very soon with:
> > #1158
> > lets make sure this option is ready for when "ramalama run" moves to the client/server technique.
> > We also have to be mindful we will also likely allow users to switch from using other non-llama.cpp backends, so we don't have to implement every single llama.cpp-specific option.
> > Have we measured how much of a difference this option makes?
> 
> It likely needs llama.cpp upstream change.

Just to clarify, could you point out exactly which part of the current implementation requires an upstream change in llama.cpp?

afazekas:
neither llama-cli or llama-run has this argument, I only know about llama-server .

ramalama/cli.py
afazekas: It is possible you want to have higher ngld when you are dealing with big model which does not fully fits into VRAM.

ericcurtin:
> neither llama-cli or llama-run has this argument, I only know about llama-server .

We don't plan on using llama-cli or llama-run in RamaLama. Set this to true:

```
USE_RAMALAMA_WRAPPER = True
```

ensure this works with `ramalama run` then, you will see what I mean

afazekas:
Simple usage:
RAMALAMA_LLAMACPP_RPC_NODES=192.168.142.5:50052 ramalama serve --ctx 8192 --runtime-args="--draft 2  " --model-draft hf://mradermacher/QwQ-0.5B-GGUF/QwQ-0.5B.Q4_K_S.gguf hf://lmstudio-community/QwQ-32B-GGUF/QwQ-32B-Q6_K.gguf

top_p = 0
top_k = 1
temperature = 0
repeat_penalty = 1 (default)

prompt eval time =     104.74 ms /     6 tokens (   17.46 ms per token,    57.28 tokens per second)
       eval time =   92815.97 ms /  2399 tokens (   38.69 ms per token,    25.85 tokens per second)
      total time =   92920.71 ms /  2405 tokens
slot print_timing: id  0 | task 688 | 
draft acceptance rate = 0.77133 (  904 accepted /  1172 generated)

Question: "How many r's does strrrrrrrrrawberry have? "

without draft <20 t/s .

ramalama/cli.py
afazekas: Would it be better to either default it to 999 or remove it from here ?

ericcurtin (approved):

docs/ramalama-serve.1.md
```diff
@@ -79,6 +79,15 @@ show this help message and exit
 #### **--host**="0.0.0.0"
 IP address for llama.cpp to listen on.
 
+#### **--model-draft**
```
rhatdan, line 82: A draft model is a smaller, faster model that helps accelerate the decoding
process of larger, more complex models, like Large Language Models (LLMs). It
works by generating candidate sequences of tokens that the larger model then
verifies and refines. This approach, often referred to as speculative decoding,
can significantly improve the speed of inferencing by reducing the number of
times the larger model needs to be invoked.

Use --runtime-arg to pass the other draft model related parameters.
Make sure the sampling parameters like top_k on the web UI are set correctly.

PR #1257, opened 2025-04-23
Default use_model_store=false for release. Will revert for next

rhatdan:
Also add micro improvement on speed.

## Summary by Sourcery

Change the default configuration for model store usage to be disabled by default for release

Enhancements:
- Simplified engine initialization by extracting engine name extraction into a separate variable

Chores:
- Set default 'use_model_store' configuration to False


ericcurtin:
Ah yes, now the tests fail, I guess another case where a config file like this would prove useful:

https://github.com/containers/ramalama/pull/1254

Have model store on for the tests.

PR #1276, opened 2025-04-25
Add minor changes to enhance pylint mark

sarroutbi:
## Summary by Sourcery

Enhance the TOML parser by adding comprehensive docstrings and making minor code improvements

Enhancements:
- Refactored code to improve readability and maintainability
- Simplified error handling in value parsing

Documentation:
- Added detailed docstrings to the TOMLParser class and its methods
- Included a module-level docstring explaining the TOML parser's capabilities

PR #1285, opened 2025-04-25
Fix minor issues

sarroutbi:
- Include test rule in global Makefile
- Use GO variable in doc/Makefile

Signed-off-by: Sergio Arroutbi <sarroutb@redhat.com>

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -59,10 +59,9 @@ install: docs
 	install -d ${DESTDIR}/${DATADIR}/ramalama
 	install -m 0644 ramalama.conf ${DESTDIR}/${DATADIR}/ramalama
 
-
 .PHONY: install-tools
 install-tools:
-	go install github.com/cpuguy83/go-md2man@latest
+	$(GO) install github.com/cpuguy83/go-md2man@latest
 
 .PHONY: clean
 clean:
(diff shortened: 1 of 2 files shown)


docs/Makefile
```diff
@@ -46,23 +54,31 @@ docs:  $(patsubst %.md,%,$(wildcard *[157].md))
...
+.PHONY: install
+install: docs install_1 install_5 install_7
 	install -d ${DESTDIR}/${DATADIR}/ramalama
 	install -m 0644 ramalama.conf ${DESTDIR}/${DATADIR}/ramalama
 
-
 .PHONY: install-tools
 install-tools:
-	go install github.com/cpuguy83/go-md2man@latest
+	$(GO) install github.com/cpuguy83/go-md2man@latest
```
ericcurtin, line 81: This change makes sense. I question all the other changes checkmate has suggested. Same for this PR:

https://github.com/containers/ramalama/pull/1277

sarroutbi: > This change makes sense. I question all the other changes checkmate has suggested. Same for this PR:
> 
> #1276

May I ask to at least keep the 'test' rule in global Makefile? I think it is more "standard" than "tests" ...

ericcurtin: What you seems good alias test and tests, no need to break workflows people are using

ericcurtin (approved):

PR #1290, opened 2025-04-25
Fix to check correct directory for shortnames file

ericcurtin:
If installed with uv the correct directory wasn't being looked for

## Summary by Sourcery

Bug Fixes:
- Add the installation-specific data directory path to the list of locations searched for 'shortnames.conf'.


rhatdan:
LGTM

PR #1291, opened 2025-04-25
Exit ramalama if user types exit

ericcurtin:
To behave like bash or python3

## Summary by Sourcery

Enhancements:
- Allow the `ramalama` client to terminate when the user inputs "exit".


rhatdan:
LGTM

We should add some info to the ramalama run man page explaining this and ^C behaviour. In another PR.

PR #1292, opened 2025-04-26
Pass args to ramalama run core

ericcurtin:
Ensure arguments are passed to *core scripts

### DIFF
--- a/libexec/ramalama/ramalama-run-core
+++ b/libexec/ramalama/ramalama-run-core
@@ -70,9 +70,9 @@ def main(args):
 
         args = Namespace(
             container=False, dryrun=False, engine=None, podman_keep_groups=False,
-            image='quay.io/ramalama/ramalama', runtime='llama.cpp',
+            image=None, runtime='llama.cpp',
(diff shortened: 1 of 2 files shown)


libexec/ramalama/ramalama-run-core
```diff
@@ -58,7 +58,7 @@ def main(args):
         container=False, dryrun=False, engine=None, podman_keep_groups=False,
         image='quay.io/ramalama/ramalama', runtime='llama.cpp',
```
rhatdan, line 59: Shold this be hard coded? or runtime? or use_model_store?

rhatdan:
Needs a rebase.

libexec/ramalama/ramalama-run-core
ericcurtin: Thanks must fix this

ericcurtin: Changed to None, to be honest this field is just misleading more than anything, at this stage we are already in a container, so regardless of what's in this parameter, we are not spinning up a new container

PR #1293, opened 2025-04-26
Change CLI behaviour

ericcurtin:
Don't exit on Ctrl-C, cut response short or print an info message to the user telling them how they may exit.

## Summary by Sourcery

Enhancements:
- Handle Ctrl-C by truncating output and informing the user how to exit, instead of immediately terminating.


rhatdan:
LGTM

Although do we use this now?  Or are we still using llama-run?

rhatdan:
Also it looks like there is hard coded values in llama-core-run?

ericcurtin:
We don't use this yes, it doesn't seem worth it implementing this stuff in C++ also

PR #1294, opened 2025-04-26
Make --no-cache optional for make build

rhatdan:
## Summary by Sourcery

Allow enabling the build cache for container builds.

Build:
- Add a `-C` flag to `container_build.sh` to optionally enable the build cache.
- Update Makefile targets (`build`, `build-rm`, `build_multi_arch`) to accept a `CACHE` variable for controlling build caching.
- Enable the build cache automatically when the `-r` (remove after build) flag is used in `container_build.sh`.
- Default container builds to use `--no-cache` unless explicitly overridden.

Documentation:
- Document the new `CACHE=-C` option in the Makefile help output.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,8 @@ help:
 	@echo "  - make build IMAGE=ramalama"
 	@echo "  - make multi-arch"
 	@echo "  - make multi-arch IMAGE=ramalama"
+	@echo "  Build using build cache, for development only"
+	@echo "  - make build IMAGE=ramalama CACHE=-C"
 	@echo
 	@echo "Build docs"
 	@echo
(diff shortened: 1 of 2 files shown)


benoitf (commented):
the doc is saying

`make build IMAGE=ramalama CACHE=-C`

Using that on macOS it's still calling with --no-cache on my end

```
 $ make build IMAGE=ramalama CACHE=-C
./container_build.sh  build ramalama -v ""
podman build --no-cache --platform linux/arm64 --volume=/Users/benoitf/git/containers/ramalama:/run/ramalama --security-opt=label=disable -t quay.io/ramalama/ramalama -f ramalama/Containerfile .
STEP 1/3: FROM registry.access.redhat.com/ubi9/ubi:9.5-1744101466
...
`

rhatdan:
Try
 make build IMAGE=ramalama NOCACHE=-C

benoitf:
`make build IMAGE=ramalama NOCACHE=-C` works, thanks

should this line be updated ?

https://github.com/containers/ramalama/blob/715dffbb53f45ba20849537541af4d6a0cda4b55/Makefile#L22

that said, if I modify a line of RamaLama python script and run again the build, the cache layers is not invalidated as it should be expected as a file has been changed (ramalama is provided with a volume, not with like a COPY instruction which invalidates the cache on a change) , then the image is still the same as the previous run

PR #1298, opened 2025-04-28
Fix the error: ramalama login can NOT get the value of RAMALAMA_TRANSPORT

melodyliu1986:
**Problem before the modification:**

```
$ export RAMALAMA_TRANSPORT=huggingface
$ ramalama login --token=hf_mytoken 
Authenticating with existing credentials for registry.fedoraproject.org
Existing credentials are valid. Already logged in to registry.fedoraproject.org
```

The value of the env var RAMALAMA_TRANSPORT was not recognized and handled in the code.

**Test after the modification**
test 1:
```
$ export RAMALAMA_TRANSPORT=huggingface
$ ramalama login --token=hf_mytoken 
The token has not been saved to the git credentials helper. Pass `add_to_git_credential=True` in this function directly or `--add-to-git-credential` if using via `huggingface-cli` if you want to set the git credential as well.
Token is valid (permission: write).
Your token has been saved to ~/.cache/huggingface/token
Login successful
$ ramalama logout
Successfully logged out.
```

test 2:
```
$ export RAMALAMA_TRANSPORT=huggingface
$ ramalama login --token=hf_mytoken huggingface
The token has not been saved to the git credentials helper. Pass `add_to_git_credential=True` in this function directly or `--add-to-git-credential` if using via `huggingface-cli` if you want to set the git credential as well.
Token is valid (permission: write).
Your token has been saved to ~/.cache/huggingface/token
Login successful
$ ramalama logout huggingface
Successfully logged out.
```

## Summary by Sourcery

Bug Fixes:
- The `login` command now correctly uses the `RAMALAMA_TRANSPORT` environment variable to determine the registry if no explicit registry argument is provided.

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -250,6 +250,9 @@ def login_parser(subparsers):
 
 
 def normalize_registry(registry):
+    # Determine the registry to use. Check the value of `RAMALAMA_TRANSPORT` env var if no `registry` argument is set.
+    registry = registry or os.getenv("RAMALAMA_TRANSPORT") or CONFIG['transport']
+
     if not registry or registry == "" or registry.startswith("oci://"):
         return "oci://"
 
--- a/docs/ramalama-login.1.md
+++ b/docs/ramalama-login.1.md
@@ -9,6 +9,8 @@ ramalama\-login - login to remote registry
 ## DESCRIPTION
 login to remote model registry
 
+By default, RamaLama uses the Ollama registry transport. You can override this default by configuring the `ramalama.conf` file or setting the `RAMALAMA_TRANSPORTS` environment variable. Ensure a registry transport is set before attempting to log in.
+
 ## OPTIONS
 Options are specific to registry types.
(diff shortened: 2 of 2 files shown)


ramalama/cli.py
```diff
@@ -251,6 +251,9 @@ def login_parser(subparsers):
 
 
 def normalize_registry(registry):
+    # Determine the registry to use. Check the value of `RAMALAMA_TRANSPORT` env var if no `registry` argument is set.
+    registry = registry or os.getenv("RAMALAMA_TRANSPORT") or "ollama"
```
rhatdan, line 255: This needs to take into account config('transport') rather then hard code ollama.

rhatdan:
Thanks @melodyliu1986 
Needs a couple of changes.

melodyliu1986:
> Thanks @melodyliu1986 Needs a couple of changes.

Thanks @rhatdan made changes according to the comments.

PR #1307, opened 2025-04-29
Catch possible error when parsing Go to Jinja template

engelmi:
Parsing a chat template in Go-syntax to a Jinja template might raise an exception. Since this is only a nice-to-have feature and we fallback to the chat template specified in the backend, lets silently skip it.

## Summary by Sourcery

Add error handling for Go template to Jinja template conversion to prevent potential failures during template processing

Bug Fixes:
- Prevent exceptions from breaking template conversion process
- Gracefully handle errors when converting Go templates to Jinja templates

Enhancements:
- Add logging for template conversion errors
- Improve template file type handling for Go templates

### DIFF
--- a/ramalama/model_store.py
+++ b/ramalama/model_store.py
@@ -455,9 +455,14 @@ class ModelStore:
                 if not go2jinja.is_go_template(chat_template):
                     return
 
-                jinja_template = go2jinja.go_to_jinja(chat_template)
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

rhatdan:
Is this something important to put a release out for, or can we wait til next week?

nathan-weinberg:
Maybe worth a `0.8.1`? for me at least this is a blocker for https://github.com/containers/ramalama-stack/pull/40

but not a huge deal if not, can keep using `0.7.5` in the meantime

PR #1311, opened 2025-04-30
common: adjust nvidia-smi for check cuda version

dougsland:
Be compatible with both versions of nvidia-smi
with version flag or not.


rhatdan:
LGTM

rhatdan:
Thanks @dougsland

PR #1312, opened 2025-04-30
Simplify installer

ericcurtin:
Use uv installer from uv itself

## Summary by Sourcery

Simplify the uv installer by using the official uv installer script instead of the custom implementation

Chores:
- Remove complex custom installer script
- Replace with direct curl and sh invocation of official installer


rhatdan:
Does this work fine if uv is already installed?

rhatdan:
LGTM

PR #1327, opened 2025-05-01
Expose http line in man pages

rhatdan:
The llama.cpp documentation links is lost when we convert markdown to nroff format. This change will expose the link in man pages.

## Summary by Sourcery

Documentation:
- Update ramalama-convert man page to explicitly display the full URL for llama.cpp quantization documentation


rhatdan:
@kush-gupt PTAL
@ericcurtin

kush-gupt:
Looks fantastic to me!

PR #1329, opened 2025-05-01
Add shortnames for mistral-small3.1 model

ericcurtin:
Another Ollama model that's only compatible with Ollama's forking of llama.cpp

## Summary by Sourcery

Add shortnames for the Mistral Small 3.1 24B model to the shortnames configuration

New Features:
- Added shortname 'mistral-small3.1' for Mistral Small 3.1 24B Instruct model
- Added alternative shortname 'mistral-small3.1:24b' for the same model

### DIFF
--- a/shortnames/shortnames.conf
+++ b/shortnames/shortnames.conf
@@ -32,6 +32,8 @@
   "mistral:7b-v3" = "huggingface://MaziyarPanahi/Mistral-7B-Instruct-v0.3-GGUF/Mistral-7B-Instruct-v0.3.Q4_K_M.gguf"
   "mistral:7b-v2" = "huggingface://TheBloke/Mistral-7B-Instruct-v0.2-GGUF/mistral-7b-instruct-v0.2.Q4_K_M.gguf"
   "mistral:7b-v1" = "huggingface://TheBloke/Mistral-7B-Instruct-v0.1-GGUF/mistral-7b-instruct-v0.1.Q5_K_M.gguf"
+  "mistral-small3.1" = "hf://bartowski/mistralai_Mistral-Small-3.1-24B-Instruct-2503-GGUF/mistralai_Mistral-Small-3.1-24B-Instruct-2503-IQ2_M.gguf"
+  "mistral-small3.1:24b" = "hf://bartowski/mistralai_Mistral-Small-3.1-24B-Instruct-2503-GGUF/mistralai_Mistral-Small-3.1-24B-Instruct-2503-IQ2_M.gguf"
   "hermes" = "huggingface://NousResearch/Hermes-2-Pro-Mistral-7B-GGUF/Hermes-2-Pro-Mistral-7B.Q4_K_M.gguf"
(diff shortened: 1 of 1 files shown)


nathan-weinberg (approved):

shortnames/shortnames.conf
```diff
@@ -32,6 +32,8 @@
   "mistral:7b-v3" = "huggingface://MaziyarPanahi/Mistral-7B-Instruct-v0.3-GGUF/Mistral-7B-Instruct-v0.3.Q4_K_M.gguf"
   "mistral:7b-v2" = "huggingface://TheBloke/Mistral-7B-Instruct-v0.2-GGUF/mistral-7b-instruct-v0.2.Q4_K_M.gguf"
   "mistral:7b-v1" = "huggingface://TheBloke/Mistral-7B-Instruct-v0.1-GGUF/mistral-7b-instruct-v0.1.Q5_K_M.gguf"
+  "mistral-small3.1" = "hf://bartowski/mistralai_Mistral-Small-3.1-24B-Instruct-2503-GGUF/mistralai_Mistral-Small-3.1-24B-Instruct-2503-IQ2_M.gguf"
```
rhatdan, line 35: How about just mistral-small?

rhatdan:
$ ramalama pull mistral-small
Downloading ollama://mistral-small:latest ...
Trying to pull ollama://mistral-small:latest...

shortnames/shortnames.conf
ericcurtin: It's deliberately named this, so users don't download:

https://ollama.com/library/mistral-small3.1

we can do both of course

ericcurtin: mistral-small on Ollama is actually 3.0 rather than 3.1

https://ollama.com/library/mistral-small

ericcurtin: This bartowski namespace in general is a good goto on huggingface, it's very popular

bmahabirbu (approved):

rhatdan:
I am fine with removing them.  But it is constant with pulling OCI Images.

shortnames/shortnames.conf
rhatdan:
```
mistral-small on Ollama is actually 3.0 rather than 3.1

https://ollama.com/library/mistral-small
```

Isn't this likely to change, and at that point mistral-small would pull a broken Model?

ericcurtin:
Ah ok fail enough

shortnames/shortnames.conf
ericcurtin: Most often this doesn't change, I don't know exactly why, but companies just tend to create new names for models rather than changing the existing ones (I guess part of the reason is the trend to use the tag for x billion parameters, etc.)

ericcurtin: Some versions of models can be a complete flop also and companies actually recommend people to use the older version

PR #1331, opened 2025-05-01
Fixup use of /.venv

rhatdan:
## Summary by Sourcery

Update the Containerfile and entrypoint script to use /.venv instead of /venv for the Python virtual environment

Enhancements:
- Modify virtual environment creation and activation to use a consistent /.venv path

Chores:
- Simplify the ramalama-stack installation process in the container

### DIFF
--- a/container-images/llama-stack/Containerfile
+++ b/container-images/llama-stack/Containerfile
@@ -9,9 +9,9 @@ RUN dnf -y update && \
     dnf -y install uv cmake gcc gcc-c++ python3-devel pkg-config sentencepiece-devel && \
     dnf -y clean all
 
-RUN uv pip install --system ramalama-stack
-
-RUN uv run llama stack run ~/.llama/distributions/ramalama/ramalama-run.yaml --image-type venv --image-name /venv
+RUN uv venv  && \
+    uv pip install ramalama-stack && \
+    uv run llama stack run  --image-type venv --image-name /.venv ~/.llama/distributions/ramalama/ramalama-run.yaml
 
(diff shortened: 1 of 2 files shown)


container-images/llama-stack/Containerfile
```diff
@@ -9,9 +9,9 @@ RUN dnf -y update && \
     dnf -y install uv cmake gcc gcc-c++ python3-devel pkg-config sentencepiece-devel && \
     dnf -y clean all
 
-RUN uv pip install --system ramalama-stack
-
-RUN uv run llama stack run ~/.llama/distributions/ramalama/ramalama-run.yaml --image-type venv --image-name /venv
+RUN uv venv  && \
```
ericcurtin, line 12: Could be interesting to do a plain-old pip install here.

In terms of uv, I'm kinda liking this command recently `uv tool install`:

uv tool install --force --python python3.12 ramalama

if the python package in an executable in puts the tool in your PATH somewhere so you don't have to prefix "uv run".

ericcurtin (approved):

container-images/llama-stack/Containerfile
rhatdan: Seems like llama-stack is insisting on being run in a /.venv type environment.

ericcurtin: Yeah not sure how we do it for this case I think it's because ramalama-stack is just a library and llama-stack is a dependancy, if we install llama-stack as a "tool" it works.

PR #1345, opened 2025-05-05
Use CLI ramalama serve here

ericcurtin:
It's easier to debug, we can do ps -ef, etc. Easier to code also.

## Summary by Sourcery

Enhancements:
- Improve debugging capabilities by switching to CLI serve approach

### DIFF
--- a/libexec/ramalama/ramalama-client-core
+++ b/libexec/ramalama/ramalama-client-core
@@ -52,7 +52,7 @@ def res(response, color):
     return assistant_response
 
 
-def req(conversation_history, url, color):
+def req(conversation_history, url, parsed_args):
     data = construct_request_data(conversation_history)
     json_data = json.dumps(data).encode("utf-8")
     headers = {
@@ -64,10 +64,8 @@ def req(conversation_history, url, color):
 
     # Send request
     i = 0.01
+    response = None
     for c in itertools.cycle(['⠋', '⠙', '⠹', '⠸', '⠼', '⠴', '⠦', '⠧', '⠇', '⠏']):
-        if i > 10:
-            break
-
         try:
             response = urllib.request.urlopen(request)
             break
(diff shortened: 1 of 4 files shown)


libexec/ramalama/ramalama-run-core
```diff
@@ -68,19 +68,11 @@ def main(args):
...
-            quiet=False, debug=False, subcommand='serve', ngl=parsed_args.ngl, threads=parsed_args.threads,
-            temp=parsed_args.temp, authfile=None, env=[], device=None, name=None,
-            oci_runtime=None, privileged=False, pull='newer', seed=None,
-            tlsverify=True, context=parsed_args.context, runtime_args=[], network=None,
-            detach=False, host='127.0.0.1', webui='on', generate=None, port=str(port),
-            rag=None, MODEL=parsed_args.MODEL, func=lambda: None,
-            UNRESOLVED_MODEL=parsed_args.MODEL
-        )
-        serve_cli(args)
+        with open(os.devnull, 'w') as devnull:
```
rhatdan, line 71: Looks like  you duplicated this code?

ericcurtin: It's deliberate, one redirects stderr to null, the other redirects stdout to null.

Could create a function for this with a parameter for stderr/stdout seems I just didn't because it seemed like overkill for something that's 2 lines and called twice

ericcurtin:
This PR is to help me figure out the broken test, I don't have a way of fixing --nocontainer tests locally on Linux right now:

https://github.com/containers/ramalama/pull/1347

libexec/ramalama/ramalama-run-core
rhatdan: Ok I missed the subtle difference.

rhatdan:
Is the failure a flake?

ericcurtin:
> Is the failure a flake?

Sadly I think not, it's occurred multiple times 😢

ericcurtin:
Trying to re-produce outside CI

ericcurtin:
I think I figured it out, missed passing the port

libexec/ramalama/ramalama-run-core
ericcurtin: Changed it anyway to try and make it more clear

ericcurtin:
I'm not sure but I think we must just increase the timeouts now

libexec/ramalama/ramalama-client-core
```diff
@@ -64,22 +64,28 @@ def req(conversation_history, url, color):
...
+                break
+
             time.sleep(i)
             i *= 2
 
-    return res(response, color)
+    if response:
+        return res(response, parsed_args.color)
 
+    print(f"\rError: could not connect to: {url}")
```
rhatdan, line 85: perror

PR #1347, opened 2025-05-05
Only execute this if /run/ramalama exists

ericcurtin:
Using this script to install llama.cpp and whisper.cpp bare metal on a bootc system, the build stops executing here:

+ ln -sf /usr/bin/podman-remote /usr/bin/podman
+ python3 -m pip install /run/ramalama --prefix=/usr ERROR: Invalid requirement: '/run/ramalama': Expected package name at the start of dependency specifier
    /run/ramalama
    ^
Hint: It looks like a path. File '/run/ramalama' does not exist. Error: building at STEP "RUN chmod a+rx /usr/bin/build_llama_and_whisper.sh && build_llama_and_whisper.sh "rocm"": while running runtime: exit status 1


rhatdan:
LGTM

Issue #1325, opened 2025-05-01
model store (partial breaking functionality)


ericcurtin:
### Issue Description

 ✗ [] ramalama list - rm -a removes all models [249]
   (from function `bail-now' in file test/system/helpers.podman.bash, line 122,
    from function `die' in file test/system/helpers.podman.bash, line 848,
    from function `run_ramalama' in file test/system/helpers.bash, line 186,
    in test file test/system/010-list.bats, line 47)
     `run_ramalama rm -a' failed

   [13:19:31.013447000] $ /Users/ecurtin/git/ramalama/bin/ramalama rm -a
   [13:19:31.138270000] Error: removing gemma3/gemma3:latest (partial): [Errno 2] No such file or directory: '/Users/ecurtin/.local/share/ramalama/store/ollama/gemma3/gemma3/refs/latest (partial)
   [13:19:31.143426000] [ rc=1 (** EXPECTED 0 **) ]
   #/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
   #| FAIL: exit code is 1; expected 0
   #\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   # [teardown]

### Steps to reproduce the issue

 ✗ [] ramalama list - rm -a removes all models [249]
   (from function `bail-now' in file test/system/helpers.podman.bash, line 122,
    from function `die' in file test/system/helpers.podman.bash, line 848,
    from function `run_ramalama' in file test/system/helpers.bash, line 186,
    in test file test/system/010-list.bats, line 47)
     `run_ramalama rm -a' failed

   [13:19:31.013447000] $ /Users/ecurtin/git/ramalama/bin/ramalama rm -a
   [13:19:31.138270000] Error: removing gemma3/gemma3:latest (partial): [Errno 2] No such file or directory: '/Users/ecurtin/.local/share/ramalama/store/ollama/gemma3/gemma3/refs/latest (partial)
   [13:19:31.143426000] [ rc=1 (** EXPECTED 0 **) ]
   #/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
   #| FAIL: exit code is 1; expected 0
   #\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   # [teardown]

### Describe the results you received

[...]

PR #1350, opened 2025-05-06
Fix partial model listing

engelmi:
Relates to: https://github.com/containers/ramalama/issues/1325
    
In the list models function only the url:// prefix is present. Passing a listed model to the factory can not map this model input correctly to the URL model class. Therefore, this gets extended and the unit tests updated by appropriate cases.

Also, instead of appending the (partial) identifier directly, the returned ModelFile class is extended to indicate if the file is partially downloaded or not.

## Summary by Sourcery

Improve model listing and handling by adding support for partial model downloads and extending URL model detection

New Features:
- Add support for detecting 'url://' prefix in model factory
- Introduce 'is_partial' flag for ModelFile to indicate partially downloaded models

Bug Fixes:
- Fix model listing to correctly handle and identify partially downloaded models

Enhancements:
- Modify model listing to track partial model downloads more accurately
- Update model removal to work with the new model listing approach

Tests:
- Add test case for 'url://' model input validation

### DIFF
--- a/ramalama/model_factory.py
+++ b/ramalama/model_factory.py
@@ -49,7 +49,12 @@ class ModelFactory:
             return Ollama, self.create_ollama
         if self.model.startswith("oci://") or self.model.startswith("docker://"):
             return OCI, self.create_oci
-        if self.model.startswith("http://") or self.model.startswith("https://") or self.model.startswith("file://"):
+        if (
+            self.model.startswith("http://")
+            or self.model.startswith("https://")
+            or self.model.startswith("file://")
+            or self.model.startswith("url://")
+        ):
             return URL, self.create_url
 
         if self.transport == "huggingface":
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -1030,8 +1030,11 @@ def rm_cli(args):
     if len(args.MODEL) > 0:
         raise IndexError("can not specify --all as well MODEL")
 
-    models = [k['name'] for k in _list_models(args)]
-    return _rm_model(models, args)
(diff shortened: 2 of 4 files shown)


engelmi:
This should fix #1325. 
Partially downloaded models are still displayed - without the "(partial)" identifier. Should these be displayed or not?

ramalama/model_factory.py
```diff
@@ -49,7 +49,12 @@ def detect_model_model_type(
             return Ollama, self.create_ollama
         if self.model.startswith("oci://") or self.model.startswith("docker://"):
             return OCI, self.create_oci
-        if self.model.startswith("http://") or self.model.startswith("https://") or self.model.startswith("file://"):
+        if (
+            self.model.startswith("http://")
+            or self.model.startswith("https://")
+            or self.model.startswith("file://")
+            or self.model.startswith("url://")
```
ericcurtin, line 56: This change looks great! My one question... Why "url://"? anything with "://" in it is a url. What is before "://" defines the protocol. A url protocol doesn't exist.

engelmi: This is basically the `model_type` which is getting passed to the store here:
https://github.com/containers/ramalama/blob/main/ramalama/model_factory.py#L86

So the protocols http, https and file get mapped to url - so we partially lose the protocol information. This seems like fixing the wrong thing when thinking about it more. Instead of passing `type(model)` to the store, it should probably be `model.type` (which is the url scheme). This would lead to models being re-downloaded again, unfortunately.

ericcurtin (approved):

PR #1355, opened 2025-05-06
Set all available GPUs in CUDA_VISIBLE_DEVICES env var

mcornea:
The CUDA_VISIBLE_DEVICES environment variable was hardcoded to '0', limiting the application to a single GPU. This commit updates it to include all available GPUs detected by nvidia-smi, allowing the application to utilize multiple GPUs.

Alternatively, removing the `CUDA_VISIBLE_DEVICES` environment variable when starting the container would also make all devices available. However, given its use in multiple places, this change opts for a less intrusive approach.

Fixes https://github.com/containers/ramalama/issues/1354


ericcurtin:
What we actually want to do here is pick the GPU with the most VRAM, and we must have at least 1GB VRAM like what we do for ROCm

bmahabirbu:
@ericcurtin I believe in the past we had this feature, but relying on nvidia-smi became an issue if I remember correctly!

I believe a streamlined approach would be to allow the user to add their env var to pick what GPU they want (architecture, device). Maybe we can have a command like ramalama hardware that showcases the users GPU hardware so its easy for them to pick from?

mcornea:
I am actually looking to use multiple GPUs so I can serve a model which doesn't fit on a single GPU. This would address the error I am seeing in https://github.com/containers/ramalama/issues/1354

ericcurtin:
> I am actually looking to use multiple GPUs so I can serve a model which doesn't fit on a single GPU. This would address the error I am seeing in #1354

I have read of people attempting this in llama.cpp but failing miserably. If you test it and it actually works well with very little gotcha's SGTM (I don't have multiple Nvidia GPUs personally)

ericcurtin:
You need to sign your commits by the way it's why the DCO build is failing

rhatdan:
We could check if the user sets CUDA_VISIBLE_DEVICES and then not attempt to set it.
export CUDA_VISIBLE_DEVICES=1 ramalama ...

We should not modify the value.

mcornea:
I updated the change to allow overriding the `CUDA_VISIBLE_DEVICES` env var when running ramalama command and it seems to work fine with both GPUs, attaching a nvtop screenshot:

```
exec_cmd:  podman run --rm --label ai.ramalama.model=deepseek-r1:70b --label ai.ramalama.engine=podman --label ai.ramalama.runtime=llama.cpp --label ai.ramalama.port=8080 --label ai.ramalama.command=serve --device /dev/dri --device nvidia.com/gpu=all -e CUDA_VISIBLE_DEVICES=0,1 --runtime /usr/bin/nvidia-container-runtime -p 8080:8080 --security-opt=label=disable --cap-drop=all --security-opt=no-new-privileges --pull newer -t -i --label ai.ramalama --name ramalama_tn4xiCG48i --env=HOME=/tmp --init --label ai.ramalama.model=deepseek-r1:70b --label ai.ramalama.engine=podman --label ai.ramalama.runtime=llama.cpp --label ai.ramalama.port=8080 --label ai.ramalama.command=serve --mount=type=bind,src=/var/lib/ramalama/store/ollama/deepseek-r1/deepseek-r1/blobs/sha256-4cd576d9aa16961244012223abf01445567b061f1814b57dfef699e4cf8df339,destination=/mnt/models/model.file,ro --mount=type=bind,src=/var/lib/ramalama/store/ollama/deepseek-r1/deepseek-r1/snapshots/sha256-4cd576d9aa16961244012223abf01445567b061f1814b57dfef699e4cf8df339/chat_template_converted,destination=/mnt/models/chat_template.file,ro quay.io/ramalama/cuda:0.8 /usr/libexec/ramalama/ramalama-serve-core llama-server --port 8080 --model /mnt/models/model.file --alias deepseek-r1:70b --ctx-size 2048 --temp 0.8 --jinja --cache-reuse 256 -v --flash-attn -ngl 999 --threads 32 --host 0.0.0.0
```

nvtop

rhatdan:
LGTM

I actually like the original one as well.  IE Set to multiple GPUs, and if only one works users could specify the environment variable.

Would love to know what @ggerganov thinks.

PR #1365, opened 2025-05-08
Groom coverage rules, genreate xml/lcov reports

sarroutbi:
## Summary by Sourcery

Improve test coverage reporting by modifying Makefile targets for coverage generation

Enhancements:
- Restructure coverage test targets to separate running tests from generating reports
- Add support for generating additional coverage report formats

Tests:
- Modify coverage test targets to generate XML, LCOV, and JSON coverage reports

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,5 +16,5 @@ venv/
 __pycache__/
 .aider*
 .coverage
-coverage.json
+coverage.*
 htmlcov/
--- a/Makefile
+++ b/Makefile
@@ -182,15 +182,21 @@ ci:
 unit-tests:
 	PYTHONPATH=. pytest test/unit/
 
-.PHONY: cov-tests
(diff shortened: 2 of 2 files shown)


rhatdan:
LGTM

ericcurtin:
I would prefer if we didn't have strict rules on code coverage, I think the stats will be nice to have, but I don't want to get into a situation where we say, we can't merge this PR because overally code coverage dropped by 1%

sarroutbi:
> I would prefer if we didn't have strict rules on code coverage, I think the stats will be great to have, but I don't want to get into a situation where we say, we can't merge this PR because overall code coverage dropped by 1%

That is not my intention. I just want to provide coverage tools and ease execution of coverage reports. I don't have in mind adding a GH action or similar to restrict PR merges up on low coverage reports.

PR #1369, opened 2025-05-09
Use all GPUs in CUDA_VISIBLE_DEVICES as default

mcornea:
Currently the CUDA_VISIBLE_DEVICES environment variable defaults to '0' when it's not overidden by the user. This commit updates it to include all available GPUs detected by nvidia-smi, allowing the application to utilize multiple GPUs by default.

## Summary by Sourcery

Update the default GPU selection behavior to use all available GPUs detected by nvidia-smi instead of defaulting to only GPU 0

Enhancements:
- Modify GPU detection to automatically use all available GPUs when CUDA_VISIBLE_DEVICES is not set

Documentation:
- Update documentation to reflect the new default GPU selection behavior

### DIFF
--- a/docs/ramalama-cuda.7.md
+++ b/docs/ramalama-cuda.7.md
@@ -126,7 +126,7 @@ Follow the installation instructions provided in the [NVIDIA Container Toolkit i
 
 ### CUDA_VISIBLE_DEVICES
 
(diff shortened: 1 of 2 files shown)


ericcurtin:
Suspect it's a flake

rhatdan:
LGTM
Thanks @mcornea

PR #1371, opened 2025-05-09
Add output path to generate quadlet/kube

engelmi:
This PR adds the `--output-dir` option to the `serve` command when generating kube/quadlet files so the user can choose where these files are being places (instead of the fixed, current location).

TODO:
- [ ] Adding system and unit tests.


ramalama/quadlet.py
```diff
@@ -25,130 +26,118 @@ def __init__(self, model, chat_template, image, args, exec_args):
...
-{name_string}\
-{port_string}
-
-[Install]
-# Start by default on boot
-WantedBy=multi-user.target default.target
-"""
-            )
+        file = IniFile(file_name)
+        file.add("Unit", "Description", "RamaLama {self.model} Kubernetes YAML - AI Model Service")
```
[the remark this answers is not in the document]
engelmi: Good catch. Fixed.

ramalama/file.py
```diff
@@ -29,3 +31,44 @@ def __del__(self):
...
+        self.filename = filename
+        self.config = ConfigParser()
+        self.config.optionxform = lambda option: option
+
+    def add(self, section: str, key: str, value: str):
+        if section not in self.config:
+            self.config[section] = {}
+        self.config[section][key] = value
+
+    def write_to_file(self, dirpath: str):
```
rhatdan, line 66: Simplify these function names to write and read?  I don't think you are going to have write_to_XYZ other then file.

ramalama/file.py
```diff
@@ -29,3 +31,44 @@ def __del__(self):
...
+        with open(os.path.join(dirpath, self.filename), "w") as f:
+            self.config.write(f, space_around_delimiters=False)
+            f.flush()
+
+    def from_file(dirpath: str, filename: str) -> "IniFile":
```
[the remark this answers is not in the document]
rhatdan: Or should this be IniRead?

ramalama/file.py
```diff
@@ -29,3 +31,44 @@ def __del__(self):
...
             self.file.close()
+
+
+class PlainFile:
+
+    def __init__(self, filename: str, content: str = ""):
+        self.filename = filename
+        self.content = content
+
+    def write_to_file(self, dirpath: str):
```
rhatdan, line 42: write()

ramalama/file.py
```diff
@@ -29,3 +31,44 @@ def __del__(self):
...
+        with open(os.path.join(dirpath, self.filename), "w") as f:
+            f.write(self.content)
+            f.flush()
+
+    def from_file(dirpath: str, filename: str) -> "PlainFile":
```
rhatdan, line 47: read()

docs/ramalama-serve.1.md
```diff
@@ -111,6 +111,10 @@ use. Using this option RamaLama will override these defaults.
 On Nvidia based GPU systems, RamaLama defaults to using the
 `nvidia-container-runtime`. Use this option to override this selection.
 
+#### **--output-dir**=path
```
rhatdan, line 114: Not crazy about options with - in them. --output or --out-dir?

Should throw an error when using --output-dir without --generate.

Rather then just adding another option, what do you think about adding :PATH to --generate option?  --generate quadlet:dest=/tmp/myquadlets?

engelmi: Adding a :PATH to the generate option seems good to me. The output directory only makes sense with this specific option. By default, :PATH is not specified and defaults to the current directory from where ramalama is executed (as it currently is).

ramalama/file.py
engelmi: renamed :heavy_check_mark:

ramalama/file.py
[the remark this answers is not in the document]
engelmi: removed it (it isn't used anyway)

ramalama/file.py
engelmi: removed it (it isn't used anyway)

docs/ramalama-serve.1.md
engelmi: Pushed an update to extend the `--generate` with a path, e.g. `--generate quadlet:/etc/containers/systemd`.
PTAL @rhatdan @ericcurtin

docs/ramalama-serve.1.md
```diff
@@ -273,6 +276,17 @@ $ cat tiny.image
 Image=quay.io/rhatdan/tiny:latest
 ```
 
+### Generate quadlet service off of tiny OCI Model and output to directory
+```
+$ ramalama --runtime=vllm serve --name tiny --generate=quadlet:/etc/containers/systemd oci://quay.io/rhatdan/tiny:latest
```
rhatdan, line 281: This is misleading, should use sudo, otherwise it looks like a rootless process can write to /etc. Alternately you could write to some where in the users homedir.
~/.config/containers/systemd/

rhatdan:
Please add tests for this.

Should it work if the path does not exist?  Or should it create the path?

rhatdan:
LGTM

PR #1393, opened 2025-05-12
This script is not macOS only

ericcurtin:
This script works for many platforms and generally does the right thing.

## Summary by Sourcery

Clarify installation instructions to cover multiple platforms and use the new centralized install script URL

Documentation:
- Clarify that the install script supports both Linux and macOS and update its header
- Update the install.sh URL to point to ramalama.ai


rhatdan:
LGTM

PR #1397, opened 2025-05-12
Add dev dependency groups 

ieaves:
## Additions
* Expands the dev dependency group to include all dev requirements.
* Introduces dependency groups for cov and cov-detailed
* Uses uv to run tests in workflows
* Modifications to configuration files to support the newly added dependencies groups and installation process

## Notes
Some of the dev dependencies listed in the Makefile are incompatible with python3.8, the required versions were decremented to maintain compatibility but tell me if that's the wrong decision.

## Questions
These are really just ideas to help improve developer experience. If any of this isn't desirable that's totally okay!

It looked like `pytest-cov` and a few assorted dependencies were references in the Makefile as well -- would those belong in the dev group or even a `cov` group?

## Summary by Sourcery

Group dev dependencies under pyproject.toml’s dev extras, simplify installation via the Makefile, and introduce pre-commit hooks for automatic code formatting.

Enhancements:
- Consolidate all development dependencies into a single ‘dev’ extras group and update the Makefile to install them via pipx using “.[dev]”.
- Add a pre-commit configuration to run black and isort on commits for consistent code formatting.
- Pin dev dependency versions to maintain compatibility with Python 3.8.


rhatdan:
Why do we need to support python3.8?  We only support python 3.11 or greater.

rhatdan:
Well that was a mistake.  We are looking for python3.11 or greater inside of our container-images so we should just stick with that.

ieaves:
Got it! I can update accordingly then.

engelmi:
@ieaves We already apply `black` and `isort` in the make target `format`/`check-format` ([see here](https://github.com/containers/ramalama/blob/main/Makefile#L132-L135)) in the CI during the Lint Code job [here](https://github.com/containers/ramalama/blob/main/.github/workflows/ci.yml#L22-L24). 
I'd consider the results in the CI the source of truth, so the pre-commit hook should use the same rules (not sure if it can just call the make target?) in order to prevent a back and forth with the CI.

ericcurtin:
> @ieaves We already apply `black` and `isort` in the make target `format`/`check-format` ([see here](https://github.com/containers/ramalama/blob/main/Makefile#L132-L135)) in the CI during the Lint Code job [here](https://github.com/containers/ramalama/blob/main/.github/workflows/ci.yml#L22-L24). I'd consider the results in the CI the source of truth, so the pre-commit hook should use the same rules (not sure if it can just call the make target?) in order to prevent a back and forth with the CI.

Agree with @engelmi not sure we should invade on someones local dev environment here, same goes for when people start checking in their personalised IDE configuration files.

ericcurtin:
Like for example if I do a quick documentation change and I want to commit it quickly and push, I don't want this to occur everytime I do a micro commit, it's invasive

ieaves:
I feel that tension as well, what are your thoughts about a ci job that pushes lint updates back to the commit rather than failing the workflow?

ieaves:
@rhatdan I've updated the project build to only accept python 3.11+!

---

I'm trying to consolidate the dev dependencies into the `pyproject.toml` so they can be centrally managed. We are using pipx pretty extensively in the make file but pipx doesn't naturally symlink dependency entrypoints which is causing subsequent tests to fail and I don't want to break existing workflows. I have a few ideas for how to proceed

## Forget centralized dev dependencies

Leave them duplicated in both the pyproject.toml and in the Makefile. 

## Use venv in the tests

Self explanatory

## Switch from pipx to uv

It looks like uv, unlike pipx,  automatically adds tools into the past so rather than `pipx install ".[dev]"` we could switch to `uv pip install ".[dev]"`. This requires us to install uv in the make command though.

## Add pipx to path

Add something like the following in CI / the Makefile so that all installed tools are in PATH.

```
export PATH="$HOME/.local/pipx/venvs/ramalama/bin:$PATH"
```

--- 

What do y'all think?

ericcurtin:
> @rhatdan I've updated the project build to only accept python 3.11+!
> 
> I'm trying to consolidate the dev dependencies into the `pyproject.toml` so they can be centrally managed. We are using pipx pretty extensively in the make file but pipx doesn't naturally symlink dependency entrypoints which is causing subsequent tests to fail and I don't want to break existing workflows. I have a few ideas for how to proceed
> 
> ## Forget centralized dev dependencies
> Leave them duplicated in both the pyproject.toml and in the Makefile.
> 
> ## Use venv in the tests
> Self explanatory
> 
> ## Switch from pipx to uv
> It looks like uv, unlike pipx, automatically adds tools into the path so rather than `pipx install ".[dev]"` we could switch to `uv pip install ".[dev]"`. This requires us to install uv in the make command though.
> 
> ## Add pipx to path
> Add something like the following in CI / the Makefile so that all installed tools are in PATH.
> 
> ```
> export PATH="$HOME/.local/pipx/venvs/ramalama/bin:$PATH"
> ```
> 
> What do y'all think?

My vote is switch to uv, I find it a better tool than pipx and easier to use, don't have to worry about venv's it's convenient the way it hides those complexities.

We already install uv in our installer if there is no package available for your platform:

https://github.com/containers/ramalama/blob/main/install-uv.sh

rhatdan:
Looks like this PR is kind of a mess, need to rebase and squash down to a single commit.

PR #1399, opened 2025-05-12
Fix cuda builds installation of python3.11

rhatdan:
## Summary by Sourcery

Update CUDA container and build scripts to support Python 3.11 installation, correct GPU package patterns, and clean up package caches.

Bug Fixes:
- Fix Python 3.11 and pip installation in the CUDA Containerfile and ensure the python3 symlink points to python3.11
- Correct the glob pattern for cuda-cupti-12 packages in build_rag.sh

Enhancements:
- Clean dnf cache after installing Python in the Containerfile
- Include GPU-related packages in the update_python function install command without stray quotes

### DIFF
--- a/container-images/cuda/Containerfile
+++ b/container-images/cuda/Containerfile
@@ -15,6 +15,8 @@ COPY --chmod=755 ../scripts /usr/bin
 # Workaround for CUDA libraries not in the ld path in base container
 RUN echo "/usr/local/cuda-12.8/compat" > /etc/ld.so.conf.d/99_cuda_compat.conf && ldconfig
 
-RUN dnf install -y python3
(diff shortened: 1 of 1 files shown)


container-images/scripts/build_rag.sh
```diff
@@ -30,7 +30,7 @@ $2"
 
 packages="git-core gcc gcc-c++"
 if [ "${GPU}" = "cuda" ]; then
-    packages+=" libcudnn9-devel-cuda-12 libcusparselt0 cuda-cupti-12\*"
```
ericcurtin, line 33: You can drop all this if you want, I caught this in the other PR, when we want to do a list of args, arrays are safer, less error prone to globbing type errors. It shuts up all the linters, etc. mentioning this also.

PR #1402, opened 2025-05-13
Fix issues reported by pylint for cli.py

sarroutbi:


ramalama/cli.py
```diff
@@ -323,34 +324,31 @@ def logout_cli(args):
...
+    if d < 4838400:
         return "1 month"
-    elif d < 31536000:
+    if d < 31536000:
         return f"{d // 2419200} months"
-    elif d < 63072000:
-        return "1 year"
-    else:
-        return f"{d // 31536000} years"
+    return "1 year" if d < 63072000 else f"{d // 31536000} years"
```
ericcurtin, line 351: I think this is less clear to be honest.

Not sure about this linter to be honest, there's no clear advantage to A or B.

Maybe get rid of the else, so it's clear there's a guaranteed return

ramalama/cli.py
```diff
@@ -78,7 +79,7 @@ def local_env(**kwargs):
 
 
 def local_models(prefix, parsed_args, **kwargs):
-    return [model['name'] for model in engine.list_models(parsed_args)]
+    return [model['name'] for model in _list_models(parsed_args)]
```
ericcurtin, line 82: @sarroutbi are we sure this is correct, if so, why?

PR #1411, opened 2025-05-15
python package fixes

ericcurtin:
Going to switch libexec files to just bin ones.

There are breakages because the python installers are not updating the shebangs in the libexec files, moving to bin solves this.


rhatdan:
Is this still active, or should we close and open a new one?

rhatdan:
Since these files no longer exist, closing.

PR #1412, opened 2025-05-15
Multmodal demo

ericcurtin:
Not to be merged

## Summary by Sourcery

Update container build process, consolidate Vulkan flags, and add multi-modal demo script

New Features:
- Add a multi-modal demo script to run SmolVLM with Podman

Enhancements:
- Remove automatic entrypoint, RAG injection, and image cleanup from container builds for non-core targets
- Replace GGML_KOMPUTE flags with a unified GGML_VULKAN compile flag

### DIFF
--- a/bin/ramalama
+++ b/bin/ramalama
@@ -1,18 +1,10 @@
-#!/usr/bin/env python3
+#!/bin/bash
 
-import sys
+set -euo pipefail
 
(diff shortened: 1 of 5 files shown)


bin/ramalama
```diff
@@ -1,18 +1,10 @@
-#!/usr/bin/env python3
+#!/bin/bash
```
rhatdan, line 1: did you plan on overwriting bin/ramalama?  Wny not create new file bin/ramalama-demo?

rhatdan: Or drop this  file in docs/demo subdir.

ericcurtin: I will move, although I'm never gonna check in the PR to be honest, it's hacky as hell

PR #1415, opened 2025-05-16
Add more debug for non starting servers with "ramalama run"

ericcurtin:
Sometimes the server doesn't start

## Summary by Sourcery

Enhancements:
- Use the long-form "--debug" flag instead of "-v" for run commands to provide more detailed diagnostic information

### DIFF
--- a/libexec/ramalama/ramalama-run-core
+++ b/libexec/ramalama/ramalama-run-core
@@ -18,6 +18,10 @@ def initialize_parser():
         default=2048,
         help="size of the prompt context (0 = loaded from model)",
     )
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

ericcurtin:
If our package plus container version pulling works correctly, it shouldn't be an issue, upgrading the python package should result in the user automatically pulling a new container image.

Also about .1% of users will use this option, the ones that are already broken, we want to poke those users to upgrade.

ericcurtin:
Oh the CI though, dammit, I'll fix for the CI

PR #1420, opened 2025-05-16
Don't throw Exceptions, be more specific

rhatdan:
Fixes: https://github.com/containers/ramalama/issues/1419


ericcurtin:
It's a pity merging the bot suggestion results in unsigned commits. Would be really cool if sourcery ai signed/authored/committed the commit from the button presser that accepted the change, rather than the bot

engelmi (approved):

PR #1424, opened 2025-05-18
Add smolvlm vision models

ericcurtin:
For multimodal usage


ericcurtin:
@olliewalsh I added these aliases for run/serve commands, switch to gemma3 ones that can act as a LLM and a multimodal model. Will these "just work" in the next release? Don't want to screw around with my machine too much, because it's prepared for a few demos.

olliewalsh:
> @olliewalsh I added these aliases for run/serve commands, switch to gemma3 ones that can act as a LLM and a multimodal model. Will these "just work" in the next release? Don't want to screw around with my machine too much, because it's prepared for a few demos.

Yeah.. gemma3 4b just works for me, thinks I look like Brad Pitt 😉 Need to take a closer lookat smolvlm, was not working for me

PR #1425, opened 2025-05-19
Add support for Hugging Face token authentication

olliewalsh:
Follow-up from https://github.com/containers/ramalama/pull/1416 to add support for pulling private Hugging Face repos/models.

## Summary by Sourcery

Enable authenticated pulls of private Hugging Face models by loading a local token and adding Authorization headers to API requests, and improve robustness by cleaning up snapshots on failures.

New Features:
- Support reading cached Hugging Face token from ~/.cache/huggingface/token
- Include Bearer token authorization header in Hugging Face API requests (checksum, manifest, metadata, blob)

Enhancements:
- Wrap store.new_snapshot calls in Hugging Face and ModelScope pull routines with error handling to clean up failed snapshots
- Remove redundant snapshot cleanup code after file list errors

### DIFF
--- a/ramalama/huggingface.py
+++ b/ramalama/huggingface.py
@@ -470,22 +470,24 @@ class Huggingface(Model):
                 hf_repo = HuggingfaceRepository(name, organization, tag)
             snapshot_hash = hf_repo.model_hash
             files = hf_repo.get_file_list(cached_files)
-            self.store.new_snapshot(tag, snapshot_hash, files)
+            try:
+                self.store.new_snapshot(tag, snapshot_hash, files)
+            except Exception as e:
+                # Cleanup failed snapshot
+                try:
+                    self.store.remove_snapshot(tag)
+                except Exception as exc:
+                    if args.debug:
+                        perror(f"ignoring failure to remove snapshot: {exc}")
+                    # ignore any error when removing snapshot
+                    pass
+                raise e
         except Exception as e:
             if not self.hf_cli_available:
                 perror("URL pull failed and huggingface-cli not available")
                 raise KeyError(f"Failed to pull model: {str(e)}")
             if args.debug:
                 perror(f"ignoring failure to get file list: {e}")
-
-            # Cleanup previously created snapshot
-            try:
-                self.store.remove_snapshot(tag)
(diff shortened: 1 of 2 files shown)


ericcurtin (approved):

ericcurtin:
Thanks it's about time someone looked at this 😊 Wonder what modelscope does

olliewalsh:
> Thanks it's about time someone looked at this 😊 Wonder what modelscope does

Looking at modelscope next....

ramalama/huggingface.py
```diff
@@ -449,23 +470,25 @@ def _pull_with_model_store(self, args, debug: bool = False):
...
+                try:
+                    self.store.remove_snapshot(tag)
+                except Exception as exc:
+                    if args.debug:
+                        perror(f"ignoring failure to remove snapshot: {exc}")
+                    # ignore any error when removing snapshot
+                    pass
+                raise e
         except Exception as e:
             if not self.hf_cli_available:
```
olliewalsh, line 486: @ericcurtin I think the huggingface-cli fallback logic could now be dropped but not 100% sure if there are other use-cases that it covers so leaving it in place for now...

ericcurtin: Which do you consider the fallback logic? :)

I was never a big fan of the hf cli compatibilty, I think we should do the hf client ourselves and store our versions of all the files somewhere here:

~/.local/share/ramalama/

kinda like what llama.cpp does.

olliewalsh: > Which do you consider the fallback logic? :)

This block.. catches all exceptions and tries again using huggingface-cli (if available). I suspect it was added to support pulling by repo.vs model but now that is implemented so suspect this can be dropped. Just wondering if its still required for something else.... Pulling safetensor models maybe?

rhatdan:
LGTM

ramalama/huggingface.py
ericcurtin: I think what you did makes sense, we should do our own huggingface client :)

engelmi: ...and maybe clean-up the "old" model store code?!

PR #1429, opened 2025-05-20
Build a Vulkan image

kpouget:
This PR adds the `ggml-vulkan` backend image to the images built by Ramalama, until this backend becomes part of the default image.

Build tested locally with the Fedora base image (I don't access the UBI image on this machine)

## Summary by Sourcery

Build:
- Add Containerfile for the ggml-vulkan backend image based on Fedora 42


ericcurtin:
We are better off just waiting a few days until podman desktop takes on podman 5.5 and then we will switch the main ramalama container image to GGML_VULKAN ... We want to use @slp's mesa with some page size patches also... It's also one less container image to maintain...

PR #1433, opened 2025-05-21
Switch python3-ramalama to ramalama

ericcurtin:
Because we are a python application rather than a python library, we have this option, it's more intuitive in general and puts us in line with `brew install ramalama`. We want to alias python3-ramalama.

## Summary by Sourcery

Alias python3-ramalama to ramalama across installation instructions and scripts

Enhancements:
- Modify install script to install the `ramalama` package name with dnf

Documentation:
- Update README install commands to use `ramalama` instead of `python3-ramalama`

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ On systems with Moore Threads GPUs, see [ramalama-musa](docs/ramalama-musa.7.md)
 ### Install on Fedora
 RamaLama is available in [Fedora 40](https://fedoraproject.org/) and later. To install it, run:
 ```
-sudo dnf install python3-ramalama
+sudo dnf install ramalama
 ```
 
 ### Install via PyPi
--- a/install.sh
+++ b/install.sh
@@ -143,7 +143,7 @@ main() {
   local sudo=""
   check_platform
   if ! $local_install && [ -z "$BRANCH" ]; then
-    if available dnf && dnf_install "python3-ramalama"; then
+    if available dnf && dnf_install "ramalama"; then
       return 0
     fi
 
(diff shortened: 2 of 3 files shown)


ericcurtin:
Was chatting to @carlwgeorge and we think it makes sense to change the primary name of the package to be the application/executable name and make python3-ramalama be an alias. Looping @smooge in.

Opening this draft PR so I don't forget about it.

carlwgeorge:
While I was discussing this with @ericcurtin I think I mistakenly said that renaming the package didn't require a re-review, but it actually does.

https://docs.fedoraproject.org/en-US/package-maintainers/Package_Renaming_Process/

I'm happy to perform that official review in bugzilla once someone submits it.

Should this PR include renaming `rpm/python-ramalama.spec` to `rpm/ramalama.spec`, along with the necessary changes to the spec file?

smooge:
I was wondering if it would be ok to do
``` 
Provides: ramalama
```
which should(?) allow for `dnf install ramalama` to work. 

Changing the package in Fedora from `python-ramalama` to `ramalama` would require changes in the spec file, the Fedora package system, and packit to make it work. I would actually put this Patch Set as the last one to be done in that series. I can start working on this but want to make sure it is what is wanted

rhatdan:
Either this can be merged now or this has been supplanted by a different PR?

smooge:
This needs to wait until I get the other one done in Fedora. Currently these changes would break someone in Fedora or EPEL. So first #1498 , then I get it reviewed and do the releng doe-see-do dance and then #1433

rhatdan:
@ericcurtin can this move forward now?

rhatdan:
This has been done in the main branch.

Issue #1445, opened 2025-05-29
bug(ramalama-run): AttributeError: 'RamaLamaShell' object has no attribute 'do_hello'. Did you mean: 'do_help'?


taronaeo:
### Issue Description

While running the command `podman exec -it ramalama ramalama run hf://lmstudio-community/Qwen2.5-1.5B-Instruct-GGUF/Qwen2.5-1.5B-Instruct-Q4_K_M.gguf`, chatting with the model with a message of "hello world" causes it to return `None` and crash entirely.

### Steps to reproduce the issue

1. `podman run -td --name ramalama ramalama`
2. `podman exec -it ramalama ramalama run hf://lmstudio-community/Qwen2.5-1.5B-Instruct-GGUF/Qwen2.5-1.5B-Instruct-Q4_K_M.gguf`
3. Type in `hello world` and notice that it crashes with `None` and throws `AttributeError: 'RamaLamaShell' object has no attribute 'do_hello'. Did you mean: 'do_help'?`

### Describe the results you received

```
podman exec -it ramalama ramalama run hf://lmstudio-community/Qwen2.5-1.5B-Instruct-GGUF/Qwen2.5-1.5B-Instruct-Q4_K_M.gguf
Downloading hf://lmstudio-community/Qwen2.5-1.5B-Instruct-GGUF/Qwen2.5-1.5B-Instruct-Q4_K_M.gguf:latest ...
Trying to pull hf://lmstudio-community/Qwen2.5-1.5B-Instruct-GGUF/Qwen2.5-1.5B-Instruct-Q4_K_M.gguf:latest ...
 99% |████  |  933.18 MB/ 940.37 MB 110.93 MB/s        0s
🦙 > hello world
NoneTraceback (most recent call last):
  File "/usr/lib64/python3.11/cmd.py", line 214, in onecmd
    func = getattr(self, 'do_' + cmd)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'RamaLamaShell' object has no attribute 'do_hello'. Did you mean: 'do_help'?

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/libexec/ramalama/ramalama-client-core", line 193, in <module>
    main(sys.argv[1:])
  File "/usr/libexec/ramalama/ramalama-client-core", line 188, in main
    run_shell_loop(ramalama_shell)
  File "/usr/libexec/ramalama/ramalama-client-core", line 170, in run_shell_loop
[...]

PR #1447, opened 2025-05-29
Choice could be not set and should not be used

rhatdan:
Fixes: https://github.com/containers/ramalama/issues/1445

## Summary by Sourcery

Prevent using unset choice in the client core by adding validation and error handling for missing choices

Bug Fixes:
- Add checks to ensure a choice is set before it is used
- Emit an appropriate error when attempting to use an unset choice and exit with failure


taronaeo (approved):
Tested against #1445, LGTM

PR #1451, opened 2025-05-29
Eliminate selinux-policy packages from containers

rhatdan:
## Summary by Sourcery

Enhancements:
- Exclude selinux-policy and container-selinux RPMs from dnf installs in container build scripts

### DIFF
--- a/container-images/scripts/build_llama_and_whisper.sh
+++ b/container-images/scripts/build_llama_and_whisper.sh
@@ -106,8 +106,9 @@ dnf_install_mesa() {
 }
 
 dnf_install_epel() {
+  local rpm_exclude_list="selinux-policy,container-selinux"
   local url="https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm"
-  dnf reinstall -y "$url" || dnf install -y "$url"
+  dnf reinstall -y "$url" || dnf install -y "$url" --exclude "${rpm_exclude_list}"
   crb enable # this is in epel-release, can only install epel-release via url
 }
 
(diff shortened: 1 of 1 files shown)


container-images/scripts/build_llama_and_whisper.sh
```diff
@@ -106,8 +106,9 @@ dnf_install_mesa() {
 }
 
 dnf_install_epel() {
+  local rpm_exclude_list=("selinux-policy,container-selinux")
```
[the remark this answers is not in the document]
ericcurtin: Probably don't need an array if it's just one arg also.... Typed like this, it's just one common separated arg in terms of bash... A single element array...

container-images/scripts/build_llama_and_whisper.sh
```diff
@@ -106,8 +106,9 @@ dnf_install_mesa() {
 }
 
 dnf_install_epel() {
+  local rpm_exclude_list="selinux-policy,container-selinux"
   local url="https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm"
-  dnf reinstall -y "$url" || dnf install -y "$url"
+  dnf reinstall -y "$url" || dnf install -y "$url" --exclude "${rpm_exclude_list}"
```
ericcurtin, line 111: If this was me I'd create a dnf_install function (and rename the old one) and only call that to ensure --exclude is passed every single time to dnf install, to ensure we never install these by mistake...

Also, can the "dnf reinstall" install it by mistake?

rhatdan: I have no idea on the reinstall and why that was added.  I will merge this and then we can revisit by creating a dnf install bash command that does what you suggest.

PR #1459, opened 2025-06-03
feat: s390x build commands

taronaeo:
currently it builds correctly on s390x but we should enforce the -DGGML_VXE=ON flag. we also want to disable whisper.cpp for now until we can bring up support for it, otherwise it will be a product that none of us have experience in.

## Summary by Sourcery

Extend container build script to enforce Release builds, enable GGML VXE support on s390x, and disable Whisper.cpp builds on s390x.

New Features:
- Add s390x-specific build flags (`-DGGML_VXE=ON`, `-DGGML_BLAS=ON`, `-DGGML_BLAS_VENDOR=OpenBLAS`) under the `ramalama` container flow.

Enhancements:
- Enforce `-DGGML_CMAKE_BUILD_TYPE=Release` in common CMake flags.

Build:
- Conditionally skip cloning and building Whisper.cpp on s390x architectures.

### DIFF
--- a/container-images/scripts/build_llama_and_whisper.sh
+++ b/container-images/scripts/build_llama_and_whisper.sh
@@ -217,7 +217,7 @@ set_install_prefix() {
 }
 
 configure_common_flags() {
-  common_flags=("-DGGML_NATIVE=OFF")
+  common_flags=("-DGGML_NATIVE=OFF" "-DGGML_CMAKE_BUILD_TYPE=Release")
   case "$containerfile" in
   rocm*)
     if [ "${ID}" = "fedora" ]; then
@@ -314,12 +314,16 @@ main() {
   fi
 
   setup_build_env
-  clone_and_build_whisper_cpp
+  if [ "$uname_m" != "s390x" ]; then
+    clone_and_build_whisper_cpp
+  fi
   common_flags+=("-DLLAMA_CURL=ON" "-DGGML_RPC=ON")
   case "$containerfile" in
   ramalama)
     if [ "$uname_m" = "x86_64" ] || [ "$uname_m" = "aarch64" ]; then
       common_flags+=("-DGGML_VULKAN=ON")
+    elif [ "$uname_m" = "s390x" ]; then
+      common_flags+=("-DGGML_VXE=ON" "-DGGML_BLAS=ON" "-DGGML_BLAS_VENDOR=OpenBLAS")
     else
       common_flags+=("-DGGML_BLAS=ON" "-DGGML_BLAS_VENDOR=OpenBLAS")
     fi


container-images/scripts/build_llama_and_whisper.sh
```diff
@@ -314,12 +314,16 @@ main() {
...
+  if [ "$uname_m" != "s390x" ]; then
+    clone_and_build_whisper_cpp
+  fi
   common_flags+=("-DLLAMA_CURL=ON" "-DGGML_RPC=ON")
   case "$containerfile" in
   ramalama)
     if [ "$uname_m" = "x86_64" ] || [ "$uname_m" = "aarch64" ]; then
       common_flags+=("-DGGML_VULKAN=ON")
+    elif [ "$uname_m" = "s390x" ]; then
+      common_flags+=("-DGGML_VXE=ON" "-DGGML_BLAS=ON" "-DGGML_BLAS_VENDOR=OpenBLAS")
```
taronaeo, line 326: I was thinking if this should be merged with the `else` condition below so we only need to add `common_flags+=("-DGGML_VXE=ON")`

rhatdan: @ericcurtin thoughts?

taronaeo:
Update: rebased with `rhatdan/VERSION` to hopefully fix `ci / bats (pull_request)` erroring out

taronaeo:
Btw, we're missing GitHub release version `v0.9.0` to pass all the `rpm-build*` and `testing-form*` tests. Once we release the new version, all of them would pass I presume.

container-images/scripts/build_llama_and_whisper.sh
ericcurtin: I guess we can, I guess...

Should these apply to PPC and RISC-V?

Looking at the code, this is on by default anyway:

https://github.com/ggml-org/llama.cpp/blob/ea1431b0fa3a8108aac1e0a94a13ccc4a749963e/ggml/CMakeLists.txt#L133

The only reference I see to it being off is older z platforms:

https://github.com/ggml-org/llama.cpp/blob/ea1431b0fa3a8108aac1e0a94a13ccc4a749963e/ggml/src/ggml-cpu/CMakeLists.txt#L377

ericcurtin: Merging for now, we can catch things in a follow-on PR, I wonder should things like:

`"-DGGML_VXE=ON" "-DGGML_BLAS=ON" "-DGGML_BLAS_VENDOR=OpenBLAS"`

be the default in the else statement. It feels like they should be the default in llama.cpp if universally suitable.

ericcurtin (approved):

ericcurtin:
Merging on green build, maybe we should rebase/repush this sometime to make it green

container-images/scripts/build_llama_and_whisper.sh
taronaeo: > Should these apply to PPC and RISC-V?

Unfortunately, I don't maintain the PPC/PPCLE codebase or have any knowledge or access to POWER systems, so perhaps we should bring in Ayappan for that. Likewise for RISC-V - maybe someone more knowledgeable.

> Looking at the code, this is on by default anyway:
>
> https://github.com/ggml-org/llama.cpp/blob/ea1431b0fa3a8108aac1e0a94a13ccc4a749963e/ggml/CMakeLists.txt#L133

Yes, correct - only if the machine type is `s390x`. But I wrote that specifically for Llama.cpp, where we can maintain backwards interoperability at the engine level, albeit not supported by our internal team and extremely slow performance without SIMD (approx 1 t/s) - whoever runs it on z14 and older will be on their own.

My idea for Ramalama is to keep the availability of the current support that our internal team can provide (i.e., z15 and newer machines) via forcing the `-DGGML_VXE=ON` flag, so if their s390x Ramalama builds fail, we'll immediately know that they're running on older, "unsupported" hardware.

Let me know your thoughts, or if we should approach this differently!

taronaeo:
>  Merging on green build, maybe we should rebase/repush this sometime to make it green

Rebase with main to incorporate the latest changes. Lets see if CI goes green before merging

container-images/scripts/build_llama_and_whisper.sh
ericcurtin: We can merge with the else, I would remove the else keeping the elif and merge them that way...

So that if we introduce another platform risc-v, ppc, etc. That we are not setting random flags from Z, etc.

taronaeo: Oh, I meant merging in this way:

```diff
diff --git a/container-images/scripts/build_llama_and_whisper.sh b/container-images/scripts/build_llama_and_whisper.sh
index 432b836..952ff0f 100755
--- a/container-images/scripts/build_llama_and_whisper.sh
+++ b/container-images/scripts/build_llama_and_whisper.sh
@@ -322,10 +322,12 @@ main() {
   ramalama)
     if [ "$uname_m" = "x86_64" ] || [ "$uname_m" = "aarch64" ]; then
       common_flags+=("-DGGML_VULKAN=ON")
-    elif [ "$uname_m" = "s390x" ]; then
-      common_flags+=("-DGGML_VXE=ON" "-DGGML_BLAS=ON" "-DGGML_BLAS_VENDOR=OpenBLAS")
     else
       common_flags+=("-DGGML_BLAS=ON" "-DGGML_BLAS_VENDOR=OpenBLAS")
+
+      if [ "$uname_m" = "s390x" ]; then
+        common_flags+=("-DGGML_VXE=ON")
+      fi
     fi
     ;;
   esac
```

So we're only declaring the flags `("-DGGML_BLAS=ON" "-DGGML_BLAS_VENDOR=OpenBLAS")` once and s390x can add its own flags thereafter.

Btw, iirc Apple uses the `Accelerate` framework over `OpenBLAS`. We might also want to fix that because I don't recall any MacOS having either OpenBLAS and/or OpenMP bundled. Let me re-check this and create a new PR if necessary.

ericcurtin: Just remove the else, it's not very clear why we have that else anyway

PR #1461, opened 2025-06-03
docs: update container_build.sh help information

### LINKED ISSUE #1460: Improve `container_build.sh` Help Information

While running the `container_build.sh` or `container_build.sh --help` command, we see that only the `-d` and `-r` commands are listed in the help information. But looking through `parse_arguments()` in `container_build.sh`, there are more flags available than listed.

We should update the help information to include the additional flags, especially the `-C` flag that would have saved me a lot of time from rebuilding the container images.

FYI:
```
parse_arguments() {
  while [[ $# -gt 0 ]]; do
    case "$1" in
      -h|--help)
        print_usage
        exit 0
        ;;
      -c)
        ci="true"
        shift
        ;;
      -C)
        nocache=""
        shift
        ;;
      -d)
        dryrun="$1"
        shift
        ;;
      -r)
        rm_after_build="true"
        nocache=""
        shift
        ;;
      -v)
        version="$2"
        shift
        shift
        ;;
      build|push|multi-arch)
        command="$1"
        shift
        ;;
      *)
        target="$1"
        shift
        ;;
    esac
  done
}
```

Include the missing flags into the help information

_No response_

### Additional context

_No response_

taronaeo:
Fixes #1460

## Summary by Sourcery

Update the help text in container_build.sh to include new flags and comprehensive examples.

Enhancements:
- Add -h, -c, -C, and -v flags to the script's usage output
- Introduce an Examples section demonstrating CPU and CUDA builds with and without cache

### DIFF
--- a/container_build.sh
+++ b/container_build.sh
@@ -253,15 +253,33 @@ print_usage() {
   echo "  multi-arch   Build and Push multi-arch images with podman farm"
   echo
   echo "Options:"
+  echo "  -h           Print help information"
+  echo "  -c           Do not use cached images (default behaviour)"
+  echo "  -C           Use cached images"
   echo "  -d           Dryrun, print podman commands but don't execute"
   echo "  -r           Remove container image after build"
   echo
   echo "Targets:"
   echo "  Specify the target container image to build or push"
   echo "  If no target is specified, all container images will be processed"
+  echo
   echo "Destination Registry:"
   echo "  Override the target registry path by setting the env var REGISTRY_PATH"
   echo "  default - quay.io/ramalama"
+  echo
+  echo "Examples:"
+  echo "  Build CPU only image without cache:"
+  echo "    $(basename "$0") build ramalama"
+  echo
+  echo "  Build CPU only image WITH cache:"
+  echo "    $(basename "$0") build -C ramalama"
+  echo
+  echo "  Build CUDA image without cache:"
+  echo "    $(basename "$0") build cuda"
+  echo
+  echo "  Build CUDA image WITH cache:"
+  echo "    $(basename "$0") build -C cuda"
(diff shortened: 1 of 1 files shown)


container_build.sh
```diff
@@ -253,15 +253,34 @@ print_usage() {
   echo "  multi-arch   Build and Push multi-arch images with podman farm"
   echo
   echo "Options:"
+  echo "  -h           Print help information"
+  echo "  -c           Do not use cached images (default behaviour)"
+  echo "  -C           Use cached images"
   echo "  -d           Dryrun, print podman commands but don't execute"
   echo "  -r           Remove container image after build"
+  echo "  -v           Print version information"
```
taronaeo: Wait is this correct? I can't find proper references of `-v` and even running the `-v` command resulted in error

```sh
$ ./container_build.sh -v
./container_build.sh: line 203: $2: unbound variable
```

taronaeo:
@rhatdan PTAL

container_build.sh
ericcurtin: These is a -v flag, but this is not what it does:

`Print version information`

taronaeo: Yeah I thought the same after I submitted this PR also. The only thing that made me believe it would be something to do with versioning was this:

```
      -v)
        version="$2"
        shift
        shift
        ;;
```

But I guess I'm wrong here. What does `-v` do though, I can't get it to run neither did I catch an understanding from the codebase. Otherwise I'll just remove it from the help information.

taronaeo: Update: Removed `-v` from the print information in the latest push

ericcurtin: It is something along the lines of `I want to specify a version manually`

PR #1464, opened 2025-06-03
chore: remove unclear else from llama and whisper build

taronaeo:
Ref: https://github.com/containers/ramalama/pull/1459#discussion_r2124350835

## Summary by Sourcery

Chores:
- Remove the unclear else block that added default GGML_BLAS flags for non-s390x architectures


taronaeo:
@ericcurtin PTAL

PR #1465, opened 2025-06-03
fix: lock down ramalama-stack version in llama-stack Containerfile

nathan-weinberg:
We should probably be more explicit in how we are pulling in the ramalama-stack Python package and YAML files, to prevent regressions for users

Pin the versions for now, can explore a less manual strategy in the future

## Summary by Sourcery

Pin ramalama-stack remote YAML and Python package versions in the llama-stack Containerfile to prevent unintended updates.

Enhancements:
- Lock ramalama-stack YAML sources to tag v0.1.5
- Pin ramalama-stack Python package installation to version 1.5.0

PR #1467, opened 2025-06-04
llama-stack container build fails with == 1.5.0

rhatdan:


nathan-weinberg:
@rhatdan this directly contradicts the motivations I laid out in https://github.com/containers/ramalama/pull/1465

there also wasn't a newer version of `ramalama-stack` available when this was made, so this PR doesn't actually fix anything

PR #1468, opened 2025-06-04
Documentation improvements

waltdisgrace:


ericcurtin:
Not sure I would universally suggest use pip, sometimes you don't want to do this, particularly on some platforms. Commits need to be signed and repushed to satisfy the DCO build.

PR #1469, opened 2025-06-04
Change timeouts

ericcurtin:
Most we want to sleep between request attempts in 100ms, a request every 100ms isn't that expensive.

## Summary by Sourcery

Enhancements:
- Introduce a 100ms sleep interval between successive request attempts


rhatdan:
LGTM

PR #1473, opened 2025-06-05
Fix stopping of llama-stack based containers by name

rhatdan:
## Summary by Sourcery

Enable stopping of llama-stack based containers by name and update tests to verify the stop operation

Bug Fixes:
- Fix stopping of llama-stack based containers by adding a fallback inspect for pods with '-pod-model-server' suffix

Enhancements:
- Simplify inspect function by removing the debug flag from run_cmd invocation

Tests:
- Update serve system test to invoke 'ramalama stop' on the served container after startup


rhatdan:
Handling this in https://github.com/containers/ramalama/pull/1472.

PR #1477, opened 2025-06-05
Don't warmup by default

ericcurtin:
llama-server by default warms up the model with an empty run for performance reasons. We can warm up ourselves with a real query. Warming up was causing issues and delays start time.

## Summary by Sourcery

Disable model warmup by default in llama-server invocation

Bug Fixes:
- Prevent default empty-run warmup that caused startup delays and issues

Enhancements:
- Add `--no-warmup` flag to disable automatic warmup and allow real-query warmup

### DIFF
--- a/ramalama/model.py
+++ b/ramalama/model.py
@@ -524,7 +524,7 @@ class Model(ModelBase):
                 draft_model = self.draft_model.get_model_path(args)
                 draft_model_path = MNT_FILE_DRAFT if args.container or args.generate else draft_model
 
-            exec_args += ["llama-server", "--port", args.port, "--model", exec_model_path]
+            exec_args += ["llama-server", "--port", args.port, "--model", exec_model_path, "--no-warmup"]
             if mmproj_path:
                 exec_args += ["--mmproj", mmproj_path]
             else:


ericcurtin:
From touvan in matrix:

touvan
I'll leave a couple of questions, and answer them if I figure it out before someone else:

I'm running one instance on Bazzite, on all AMD hardware (5700X, 32GB RAM, RX 6800 16GB) - it doesn't have rocm, but rocm seems to work in the ramalama container. It does seem to start up, fill in to the GPU, but then it stalls forever on this message: common_init_from_params: warming up the model with an empty run - please wait ... (--no-warmup to disable)
I'm also trying to run it on macos on an M1 Max with 64GB RAM. This was pretty easy to get running, but it doesn't use the GPU - so it's quite slow (less than half as fast as the same model running on my 5700X). I had to tell it to use docker on macos, but that's the only thing I changed.
Also, is there a way to access other ports, or some way to find that information? I'd like to test running this for codegen with tools like cline. I know it'll be slow - just curious.

rhatdan:
Is this a reversal, I though we were wanting to warmup the server so when the user first types a query, it responds quicker?

ericcurtin:
> Is this a reversal, I though we were wanting to warmup the server so when the user first types a query, it responds quicker?

It was breaking a user's machine. Not sure if we want warmup on by default. Is the performance difference even noticeable to the human eye? It's also slowing down our startup performance. I think llama-server startup performance is even more important than warming up for RamaLama. But it's ok if I'm told otherwise 😄 After a few words printed we will likely be upto full speed anyway.

For me if users want want warmed up, speedy inferencing, they should leave llama-server processes pre-forked lying around ready for execution, so that they are always warm.

I see two usecases moving forward:

1. Model swapping server: for this case you want to favour startup time rather than warmup as loading a model can take seconds, don't want to delay any further.
2. Pre-forked server: processes should be warm anyway as they are around for a long time.

ericcurtin:
I should have posted the full matrix message above. --no-warmup didn't just accelerate things, it fixed a bug for that guy, his GPU started working. A micro-optimization like this that breaks some GPUs doesn't seem worth it to me. And it's not just an optimization, it's a tradeoff, we lose performance in terms of startup time.

PR #1486, opened 2025-06-09
Only print this in the llama-stack case

ericcurtin:
In the llama.cpp case it doesn't make as much sense, llama-server prints this string when it's ready to be served like so:

main: server is listening on http://0.0.0.0:8080 - starting the main loop

This can be printed seconds or minutes too early potentially in the llama.cpp case.

## Summary by Sourcery

Bug Fixes:
- Only print the OpenAI RESTAPI endpoint when using the llama-stack API

### DIFF
--- a/ramalama/model.py
+++ b/ramalama/model.py
@@ -744,5 +744,5 @@ def compute_serving_port(args, quiet=False) -> str:
         if args.api == "llama-stack":
             print(f"Llama Stack RESTAPI: {openai}")
             openai = openai + "/v1/openai"
-        print(f"OpenAI RESTAPI: {openai}")
+            print(f"OpenAI RESTAPI: {openai}")
     return str(target_port)


rhatdan:
What should we do in the `-d` case?
```
 ./bin/ramalama serve -d --pull=never granite
OpenAI RESTAPI: http://localhost:8080
e70bdc5981b30c91c6a327edf55c254342a32bfbd471e0a248f1dbd480cf88ae
```

ericcurtin:
I would say do nothing in the -d case keep it simple (and keep it like a normal podman run -d). I would hope in real life people don't use -d, only for development. I would hope they start and stop via an init system instead and log to somewhere.

rhatdan:
The issue is that the server could pick a different port then port 8080, and you need to examine the `ramalama ps` to see what ports were actually used.

If you feel strongly about this I am fine.  I was trying to be helpful for the user, in production, you should always specify the port and not leave this to chance.

PR #1490, opened 2025-06-10
Make sure llama-stack URL is shown to user

rhatdan:


rhatdan:
@olliewalsh PTAL

bin/ramalama
```diff
@@ -4,6 +4,9 @@ import sys
 
 
 def main():
+    if not sys.stdout.isatty():
+        sys.stdout.reconfigure(line_buffering=True)
```
olliewalsh, line 8: LGTM
I expect stdout buffering was also an issue when running under systemd so not just required for the bats test assertions

olliewalsh (approved):

PR #1497, opened 2025-06-10
This installs ramalama via uv if python3 version is too old

ericcurtin:
Lets say in the case of RHEL9.

## Summary by Sourcery

Improve the install.sh script to handle older Python versions by guarding system-package installs behind a version check and falling back to the uv-based installer when needed.

Enhancements:
- Guard DNF package install of ramalama with a Python 3.10 version check and fall back to uv-based installation for older versions
- Relocate and streamline Podman and APT install steps into the main flow, removing redundant platform-check logic
- Abstract the install-uv.sh URL host into a variable for a cleaner curl command

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -84,12 +84,6 @@ check_platform() {
 
       sudo="sudo"
     fi
-
-    if available dnf; then
-      dnf_install_podman
-    elif available apt; then
-      apt_update_install
-    fi
   else
     echo "This script is intended to run on Linux and macOS only"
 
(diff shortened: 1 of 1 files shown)


rhatdan:
The cann Containerfile and perhaps the MUSA one, they only have python3.10, (as I recall).  Not sure if this will break them or not.

Our builders seem to running out of space on the bats test.

ericcurtin:
Changed to greater than or equal to

rhatdan:
LGTM

PR #1499, opened 2025-06-10
This is not a multi-model model

ericcurtin:
Although the other gemma once are. Point the user towards a single gguf.


rhatdan:
LGTM

PR #1502, opened 2025-06-10
model: always pass in GPU offloading parameters

### LINKED ISSUE #1482: Fallback to Vulkan on unsupported AMD GPUs

When using GPUs that are not supported by ROCm (i.e. RX 580/gfx803), Ramalama should use the Vulkan image automatically instead.

`amdkfd` driver exports `gfx_target_version` in `/sys/class/kfd/kfd/topology/nodes/*/properties` that can be used to detect the GPU architecture. Use it to disable the ROCm image for pre-`gfx900` GPUs.

User can manually override the image in their configuration file, but I would prefer this working out of the box.

Alternatively, a custom ROCm build could be done to enable support for gfx8 GPUs, however the performance is ~20x slower than the Vulkan image in my own testing.

### Additional context

_No response_

alaviss:
This does nothing on systems with no GPUs, but on Vulkan-capable systems, this would automatically offload the model to capable accelerators.

Take this moment to claim Vulkan support in README also.

Closes #1482


README.md
```diff
@@ -61,8 +61,10 @@ RamaLama then pulls AI Models from model registries, starting a chatbot or REST
 | Apple Silicon GPU (podman-machine) | &check;                     |
 | Nvidia GPU (cuda)                  | &check; See note below      |
 | AMD GPU (rocm)                     | &check;                     |
+| AMD GPU (vulkan)                   | &check;                     |
```
ericcurtin: I would have done:

AMD GPU (rocm, vulkan)

but anyway, no big deal.

ramalama/model.py
```diff
@@ -252,33 +250,19 @@ def setup_container(self, args):
...
-                args.ngl = 999
+        if args.ngl < 0:
+            args.ngl = 999
 
-            if runner:
-                gpu_args += ["--ngl"]  # double dash
-            else:
-                gpu_args += ["-ngl"]  # single dash
+        if runner:
+            gpu_args += ["--ngl"]  # double dash
```
ericcurtin, line 257: We must remember to clean this up in a follow on PR, this runner variable isn't required anymore

README.md
```diff
@@ -61,8 +61,10 @@ RamaLama then pulls AI Models from model registries, starting a chatbot or REST
...
 | Ascend NPU (Linux)                 | &check;                     |
 | Intel ARC GPUs (Linux)             | &check; See note below      |
+| Intel GPUs (vulkan / Linux)        | &check                      |
```
ericcurtin: Looks like you are missing a semicolon here

ericcurtin:
I want to test this with an integrated GPU with 512MB I have at home before we merge this. I'm travelling so that could take a week.

kpouget:
@alaviss, I'm not sure to understand when this fix would change the behavior 🤔 

```
if args.ngl < 0:
    args.ngl = 999
```
for sure this simplification is good to have, but in my Vulkan system `ngl` was already set to 999:
```
ramalama --debug run ollama://llama3.2:latest
2025-06-11 08:53:17 - DEBUG - exec_cmd: podman run --rm --label ai.ramalama.model=ollama://llama3.2:latest ... --ngl 999 --threads 7 /mnt/models/model.file
ggml_vulkan: Found 1 Vulkan devices:
llama_model_load_from_file_impl: using device Vulkan0 (Virtio-GPU Venus (Apple M4 Pro)) - 49152 MiB free
load_tensors: loading model tensors, this can take a while... (mmap = true)
load_tensors: offloading 28 repeating layers to GPU
load_tensors: offloading output layer to GPU
load_tensors: offloaded 29/29 layers to GPU
load_tensors:      Vulkan0 model buffer size =  1918.35 MiB
load_tensors:   CPU_Mapped model buffer size =   308.23 MiB
...........................................................................
```

but maybe that's because of this part (both of them are `true` in my system), which doesn't catch your system
```
                # linux and macOS report aarch64 (linux), arm64 (macOS)
                ramalama.common.podman_machine_accel
                or (machine == "aarch64" and os.path.exists("/dev/dri"))
```

0cc4m:
> I want to test this with an integrated GPU with 512MB I have at home before we merge this. I'm travelling so that could take a week.

Integrated GPUs usually have access to between half and all of the CPU memory. The Vulkan backend will use this, it is not limited to the dedicated portion of RAM.

A larger issue might be small iGPUs that aren't really meant to do more than display a desktop, for example the ones AMD packs into their Ryzen 7000+ desktop CPUs. These you wouldn't really want to use over the CPU, unlike large APUs.

alaviss:
@kpouget 

> which doesn't catch your system

That's exactly the case. ARM64 systems were given that hack because it covers Apple Silicon, but not the rest of us.

@0cc4m

> A larger issue might be small iGPUs that aren't really meant to do more than display a desktop, for example the ones AMD packs into their Ryzen 7000+ desktop CPUs. These you wouldn't really want to use over the CPU, unlike large APUs.

While that is a potential issue, I think it's a lot better to just get this in and add heuristics to filter out those later. An user can always disable offloading by specifying `--ngl 0` on the command line, but they cannot turn offloading on if ramalama thinks they don't have a GPU.

0cc4m:
`--ngl 0` does not disable the GPU completely, it still gets used for large matrix multiplications and some other operations based on a heuristic.

alaviss:
> `--ngl 0` does not disable the GPU completely, it still gets used for large matrix multiplications and some other operations based on a heuristic.

That wouldn't change with or without this PR. The default image already have Vulkan support in it and RamaLama unconditionally forwards GPUs into the container. It's just picky at choosing when to pass `--ngl` in.

ramalama/model.py
rhatdan: Why not remove it now?

ericcurtin:
> > I want to test this with an integrated GPU with 512MB I have at home before we merge this. I'm travelling so that could take a week.
> 
> Integrated GPUs usually have access to between half and all of the CPU memory. The Vulkan backend will use this, it is not limited to the dedicated portion of RAM.
> 
> A larger issue might be small iGPUs that aren't really meant to do more than display a desktop, for example the ones AMD packs into their Ryzen 7000+ desktop CPUs. These you wouldn't really want to use over the CPU, unlike large APUs.

Guess what I have a crappy AMD iGPU to test with, that's exactly the test machine I want to run this by.

ericcurtin (approved):

ericcurtin:
Tested seems good

PR #1503, opened 2025-06-11
fix: remove unneeded dependency from Llama Stack container

nathan-weinberg:
partially reverts https://github.com/containers/ramalama/pull/1484

`blobfile` dependency is already included in ramalama-stack version 0.2.0

adding it explicitly is unnecessarily


ericcurtin (approved):

rhatdan:
It was not in the container when I built it.  I will try again.

rhatdan:
Ok I just tried it again with this change, and now I am seeing the blobfile, do not know what happened last week.

PR #1504, opened 2025-06-11
Remove libexec files

ericcurtin:
This is breaking nocontainer invocations, python package managers don't recognize libexec files and replace the shebang.


rhatdan:
Two out of three libexec files are now gone.  Lets figure out how to get rid of the third one.

It does almost nothing other then to rexec the server with the current path appended onto the python library paths.  Do we still need this?

Issue #1487, opened 2025-06-09
unusual podman error with CUDA CDI, but nvidia-smi works


jgforbes:
```
~$ ramalama version
ramalama version 0.9.1
```
The model and ramalama-cuda container get pulled correctly; however, we get a conmon error:
```
~$ ramalama run cogito
Error: container create failed (no logs from conmon): conmon bytes "": readObjectStart: expect { or n, but found , error found in #0 byte of ...||..., bigger context ...||...
```
in a fedora container, nvidia-smi works correctly.
```
~$ podman run --rm --device=nvidia.com/gpu=all docker.io/fedora nvidia-smi
Mon Jun  9 20:00:17 2025
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 570.133.20             Driver Version: 570.133.20     CUDA Version: 12.8     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA GeForce GTX 1080        Off |   00000000:9E:00.0 Off |                  N/A |
| 27%   35C    P8             10W /  180W |       5MiB /   8192MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+

+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI              PID   Type   Process name                        GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
|  No running processes found                                                             |
+-----------------------------------------------------------------------------------------+
Precision-7820-Tower:~$
```

alaviss:
You can pass `--debug` to `ramalama` like this:

```
ramalama --debug run cogito
```

To see exactly how ramalama executed `podman`. The error you're seeing appears to come from podman directly.

rhatdan:
You could also check previous cuda image to see if image update broke:

ramalama --image quay.io/ramalama/cuda:0.9.0 run cogito

rhatdan:
This usually means that you have a mismatch between the libraries on the host system and those within the container?

If you get the podman command and remove the "-t" option, it will show you the error.  Sadly the -t option is covering up the fact.

[...]

PR #1506, opened 2025-06-12
Do not run with --tty when not in interactive mode

rhatdan:
I have found that when running with nvidia the -t (--tty) option in podman is covering up certain errors.  When we are not running ramalama interactively, we do not need this flag set, and this would make it easier to diagnose what is going on with users systems.


rhatdan:
This should help diagnose issues like:

https://github.com/containers/ramalama/issues/1487

ericcurtin:
CI failing

PR #1507, opened 2025-06-12
Increase retry attempts to attempt to connect to server

ericcurtin:
increase i to 512

## Summary by Sourcery

Enhancements:
- Increase retry attempts for server connections to 512


rhatdan:
LGTM

Issue #1508, opened 2025-06-12
`ramalama serve smolvlm` bailing out


ericcurtin:
### Issue Description

This works on my machine, but at least two different users reported this bailing out @rhatdan and @cedricclyburn:

```
Failed to create new snapshot: Only one model supported, got 4: [<ramalama.huggingface.HuggingfaceCLIFile object at 0x10206e900>, <ramalama.huggingface.HuggingfaceCLIFile object at 0x102056850>, <ramalama.huggingface.HuggingfaceCLIFile object at 0x102056990>, <ramalama.huggingface.HuggingfaceCLIFile object at 0x10202a650>]
Removing snapshot...
Error: [Errno 2] No such file or directory: '/Users/cclyburn/.local/share/ramalama/store/huggingface/ggml-org/SmolVLM-500M-Instruct-GGUF/snapshots/
```

### Steps to reproduce the issue

NA

### Describe the results you received

NA

### Describe the results you expected

NA

### ramalama info output

```yaml
NA
```

### Upstream Latest Release

Yes

### Additional environment details

NA

ericcurtin:
@olliewalsh this feature has been working great on my machine, since it was introduced... For some other users, not so much

ericcurtin:
I reproduced this on @cedricclyburn 's machine by blowing out the whole model store.

He was on:

```
ramalama 0.9.1
```

I dunno if he had huggingface cli installed too, we have some code around that which I wouldn't mind deleting. Our huggingface client is fine.

olliewalsh:
Yes, looks like a dup of https://github.com/containers/ramalama/issues/1493.

I assume it is download using huggingface-cli in this case as it has all 4 gguf files from https://huggingface.co/ggml-org/SmolVLM-500M-Instruct-GGUF/tree/main

olliewalsh:
Ultimately the issue here is that it is failing to remove the snapshot. @engelmi Possibly a regression in https://github.com/containers/ramalama/commit/4f53c653864ed2b46a08bf76bcad17b7289f21ff#diff-452df1d3a56e954c748b85f61013db53d15918e86ce6326d9ebc000351a4cd46

olliewalsh:
Can reproduce this by forcing it to fallback to huggingface-cli.

@engelmi Only get the "No such file or directory" in the 2nd attempt, probably just need to handle the case where the snapshot doesn't exist when calling remove_snapshot()

```
[...]

PR #1511, opened 2025-06-12
Ignore errors when removing snapshot directory

engelmi:
Relates to: https://github.com/containers/ramalama/issues/1508

remove_snapshot should never fail and, therefore, the shutil.rmtree call needs ignore_errors=True.

/cc @olliewalsh


ericcurtin (approved):

ericcurtin:
LGTM... Will this fix or partially fix the issue @cedricclyburn saw? Was this tested?

engelmi:
> LGTM... Will this fix or partially fix the issue @cedricclyburn saw? Was this tested?

The failure to remove the snapshot directory is a subsequent error - and this PR makes the model store a bit more resilient. The failure in #1508 is due to the fallback to the hf cli, which we don't support multi-modal models. Ramalama uses the fallback probably due to an API error (rate limiting?) in the first, multi-modal supporting method.

PR #1512, opened 2025-06-12
Make minimum version of Python consistent

Hasnep:
Change documentation and code to consistently use 3.10 as the minimum python version

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -56,7 +56,7 @@ This section describes how to make a contribution to RamaLama.
 
 ### Prepare your environment
 
-The minimum version of Python required to use RamaLama is Python 3.11
+The minimum version of Python required to use RamaLama is Python 3.10
 
 ### Fork and clone RamaLama
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -61,7 +61,7 @@ include-package-data = true
 line-length = 120
 skip-string-normalization = true
 preview = true
-target-version = ["py311", "py312", "py313"]
+target-version = ["py310", "py311", "py312", "py313"]
 include = '\.pyi?$'
 exclude = '''
 /(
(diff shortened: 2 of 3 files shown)


ericcurtin:
I'm not sure we should obsess over this, why try and be consistent and default to the higher one. Are there 3.11 features that make this worth it vs 3.10 ?

ericcurtin:
Notice also for things that run outside of the container, the python3 requirement is lower, this can be a good thing

Hasnep:
My experience with Python is that if you're not actively testing your code with a version of Python, you can't be 100% sure whether you actually support that version. It's so easy to silently introduce a feature from a later version of Python, e.g. someone uses a match statement or imports `tomllib` and it works in CI but not in production.

ericcurtin:
Whilst I agree about the testing, 3.10 I think is a better minimum for us now see:

https://github.com/containers/ramalama/pull/1497

To be honest I would have stayed on 3.9 for RHEL9 support, these small increments are not worth the loss in compatibility, unless we absolutely need it for a new dependancy (and I think we should be careful about dragging in dependancies too). And even then, the code outside the container can and should be a lower python version ideally, that's important for compatibility. Note for toml we specifically wrote this to be cross-compatible across many versions of python3:

https://github.com/containers/ramalama/blob/main/ramalama/toml_parser.py

rhatdan:
A couple of out Containerfiles Cann and perhaps Musa only support 3.10.

Hasnep:
Okay, in that case I will change all the places that say that the minimum supported Python version is 3.11 to 3.10.

rhatdan:
I still see >=3.11 and should be >=3.10

Hasnep:
First force push was just rebasing on the latest main, just force pushed again with the new changes :)

rhatdan:
LGTM
Will merge once test pass, and get it into todays release.

PR #1513, opened 2025-06-12
Update black target version

Hasnep:
The `black` formatter can make better formatting suggestions if it knows the correct versions of Python being targeted.

## Summary by Sourcery

Enhancements:
- Extend Black's target-version in pyproject.toml to include Python 3.10 through 3.13 instead of Python 3.6


ericcurtin:
I think we have to be careful about updating python version required for the sake of linters. I'm gonna let the build run anyway

PR #1518, opened 2025-06-12
Rm huggingfacecli

ericcurtin:
## Summary by Sourcery

Remove the huggingface-cli dependency and related methods from the Huggingface class, simplifying model download and upload to rely solely on HTTP endpoints.

Enhancements:
- Remove is_huggingface_cli_available utility and associated availability checks
- Eliminate get_cli_command, get_cli_download_args, and CLI-based push/download methods
- Drop in_existing_cache symlinking logic tied to the CLI cache

### DIFF
--- a/ramalama/huggingface.py
+++ b/ramalama/huggingface.py
@@ -21,11 +21,6 @@ pip install huggingface_hub
 """
 
 
-def is_huggingface_cli_available():
-    """Check if huggingface-cli is available on the system."""
-    return available("huggingface-cli")
-
-
 def huggingface_token():
     """Return cached Hugging Face token if it exists otherwise None"""
     token_path = os.path.expanduser(os.path.join("~", ".cache", "huggingface", "token"))
@@ -145,10 +140,6 @@ class Huggingface(HFStyleRepoModel):
     def __init__(self, model):
         super().__init__(model)
         self.type = "huggingface"
-        self.hf_cli_available = is_huggingface_cli_available()
-
-    def get_cli_command(self):
-        return "huggingface-cli"
 
     def get_missing_message(self):
         return missing_huggingface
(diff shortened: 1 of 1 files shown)


ramalama/huggingface.py
```diff
@@ -145,10 +140,6 @@ class Huggingface(HFStyleRepoModel):
     def __init__(self, model):
         super().__init__(model)
         self.type = "huggingface"
-        self.hf_cli_available = is_huggingface_cli_available()
-
-    def get_cli_command(self):
```
olliewalsh, line 150: required for login IIRC

ericcurtin: Do you (or anybody else) use the hf login? And couldn't that be something they do external to ramalama

We could reimplement this ourselves too

olliewalsh: I expect hf login is the only way to avoid rate limiting - and maybe a pro/enterprise account

olliewalsh: > 
> We could reimplement this ourselves too

Or could use the hf python APIs if the module is installed? Any particular reason it's shelling out to the CLI?

rhatdan: Yes I think we got to worry about login.  I would figure the HF has a mechanism for storing private models, that require you to login, to gain access.

taronaeo: Yeah login is very much required to avoid rate limiting and most importantly, download gated models like Meta-Llama.

taronaeo:
Side question, are we using HF XET to download models by any chance? iirc pulling models via HTTP calls alone is quite slow, and HF did document this somewhere, and from my testing, HF XET did improve download speeds quite dramatically

kush-gupt:
Or maybe switching from hfcli to [xet-core](https://github.com/huggingface/xet-core?tab=readme-ov-file) as an option...

rhatdan:
I think these are all options, and should be taken advantage if possible. BUT we should not require them.  We should always be able to fall back to simple URL handling, if there are not packages available.

rhatdan:
Closing since this is required for logging.

The huggingface_cli is an optional requirement, so no platform should be forced to install it.

Issue #1525, opened 2025-06-13
Invalid argument to --image silently ignored


nikic:
### Issue Description

Passing an invalid argument to `--image` silently ignores the argument and falls back to the default, in my case `quay.io/ramalama/cuda:0.9`.

The case I actually ran into is of course not `--image foobar`, but rather `--image quay.io/ramalama/ramalama` trying to force execution on the CPU. This worked on ramalama 0.8, but on ramalama 0.9 it only works if the version is specified as `--image quay.io/ramalama/ramalama:0.9`, otherwise it just fails silently.

### Steps to reproduce the issue

```
ramalama --image foobar run granite:8b
```

### Describe the results you received

No error, fallback to `quay.io/ramalama/cuda:0.9`.

### Describe the results you expected

Error that the image is invalid.

### ramalama info output

```yaml
{
    "Accelerator": "cuda",
    "Engine": {
        "Info": {
            "host": {
                "arch": "amd64",
                "buildahVersion": "1.40.0",
                "cgroupControllers": [
                    "cpu",
                    "io",
                    "memory",
                    "pids"
                ],
                "cgroupManager": "systemd",
                "cgroupVersion": "v2",
                "conmon": {
                    "package": "conmon-2.1.13-1.fc42.x86_64",
                    "path": "/usr/bin/conmon",
                    "version": "conmon version 2.1.13, commit: "
                },
                "cpuUtilization": {
                    "idlePercent": 96.01,
                    "systemPercent": 0.47,
                    "userPercent": 3.51
                },
                "cpus": 32,
                "databaseBackend": "boltdb",
                "distribution": {
                    "distribution": "fedora",
                    "variant": "workstation",
                    "version": "42"
                },
                "eventLogger": "journald",
                "freeLocks": 1901,
                "hostname": "fedora.fritz.box",
                "idMappings": {
                    "gidmap": [
                        {
                            "container_id": 0,
                            "host_id": 1000,
                            "size": 1
                        },
                        {
                            "container_id": 1,
                            "host_id": 100000,
                            "size": 65536
                        }
                    ],
                    "uidmap": [
                        {
                            "container_id": 0,
                            "host_id": 1000,
                            "size": 1
                        },
                        {
                            "container_id": 1,
                            "host_id": 100000,
                            "size": 65536
                        }
                    ]
                },
                "kernel": "6.14.9-300.fc42.x86_64",
                "linkmode": "dynamic",
                "logDriver": "journald",
                "memFree": 30134116352,
[...]

Issue #1521, opened 2025-06-13
quay.io/ramalama/ramalama-rag:0.8.5 is broken


thom311:
### Issue Description

On Fedora 42, I have python3-ramalama.noarch 0.8.5-1.fc42.

Try:
```
$ echo "Something something about something." > out.md
$ ramalama rag out.md localhost/myrag:0.1

Building localhost/myrag:0.1 ...
adding vectordb ...
77cb9e6bb889865a049f6e595a5e6105a70febba556fad87ad45ca54f13a8a22
$ ramalama run --rag localhost/myrag:0.1 ollama://gemma3/gemma3:4b
> hello
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/httpx/_transports/default.py", line 101, in map_httpcore_exceptions
    yield
  File "/usr/lib/python3.9/site-packages/httpx/_transports/default.py", line 250, in handle_request
    resp = self._pool.handle_request(req)
  File "/usr/lib/python3.9/site-packages/httpcore/_sync/connection_pool.py", line 256, in handle_request
    raise exc from None
  File "/usr/lib/python3.9/site-packages/httpcore/_sync/connection_pool.py", line 236, in handle_request
    response = connection.handle_request(
  File "/usr/lib/python3.9/site-packages/httpcore/_sync/connection.py", line 101, in handle_request
    raise exc
  File "/usr/lib/python3.9/site-packages/httpcore/_sync/connection.py", line 78, in handle_request
    stream = self._connect(request)
  File "/usr/lib/python3.9/site-packages/httpcore/_sync/connection.py", line 124, in _connect
    stream = self._network_backend.connect_tcp(**kwargs)
  File "/usr/lib/python3.9/site-packages/httpcore/_backends/sync.py", line 215, in connect_tcp
    sock.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
  File "/usr/lib64/python3.9/contextlib.py", line 137, in __exit__
    self.gen.throw(typ, value, traceback)
  File "/usr/lib/python3.9/site-packages/httpcore/_exceptions.py", line 14, in map_exceptions
    raise to_exc(exc) from exc
httpcore.ConnectError: [Errno 111] Connection refused

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/openai/_base_client.py", line 969, in request
    response = self._client.send(
[...]

PR #1526, opened 2025-06-13
[th/honor-image] cli: honor --image parameter in rag mode

thom311:
Previously, to account for "--rag" option and the "rag" mode, the
args.image was mangled at a late point. At that point, args.image does
not indicate whether the value was chosen by default (and needs
adjustment for "-rag") or whether it was explicitly specified by the
user. In the latter case, we would not want to overwrite the user's
choice.

In those cases, the parameter was not ignored entirely, but rather the
tag was stripped, "-rag" appended, and the default tag added.
Nonetheless, we should honor the original image. Granted, the user must
choose an image that is suitable (e.g. for "--rag" mode).

Similarly, the "run" mode entirely ignored the "--image" argument.

Now, image selection happens early while parsing command line arguments.
It is based on CONFIG, the cli args and the hardware detection in
common.py.

---

Fixes: https://github.com/containers/ramalama/issues/1525
See-also: https://github.com/containers/ramalama/issues/1521#issue-3142998375

## Summary by Sourcery

Honor user-specified --image in both rag and run modes by moving image resolution upstream and eliminating late-stage overrides

Bug Fixes:
- Ensure --image is no longer ignored or overwritten in run mode
- Prevent automatic replacement of a user-specified image when using --rag

Enhancements:
- Centralize image selection in post_parse_setup based on args.image, args.rag, and CONFIG
- Remove legacy handle_rag_mode logic and inline accel_image calls across model and rag modules
- Change CLI --image default to None and update help text to display the default no-rag image

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -190,8 +190,8 @@ The RAMALAMA_CONTAINER_ENGINE environment variable modifies default behaviour.""
     )
     parser.add_argument(
         "--image",
-        default=accel_image(CONFIG, None),
-        help="OCI container image to run with the specified AI model",
+        default=None,
+        help=f"OCI container image to run with the specified AI model. The default (no-rag) image is {repr(accel_image(CONFIG, None))}.",
         action=OverrideDefaultAction,
         completer=local_images,
     )
@@ -264,6 +264,13 @@ def parse_arguments(parser):
 
 def post_parse_setup(args):
     """Perform additional setup after parsing arguments."""
+
+    if args.func is rag_cli:
+        args.rag = "rag"
+
+    if args.image is None:
+        args.image = accel_image(CONFIG, args)
+
     if hasattr(args, "MODEL") and args.subcommand != "rm":
         resolved_model = shortnames.resolve(args.MODEL)
         if resolved_model:
--- a/ramalama/model.py
+++ b/ramalama/model.py
@@ -13,7 +13,6 @@ from ramalama.common import (
     MNT_FILE,
     MNT_FILE_DRAFT,
     MNT_MMPROJ_FILE,
-    accel_image,
     check_metal,
     check_nvidia,
     exec_cmd,
@@ -21,7 +20,7 @@ from ramalama.common import (
     get_cmd_with_wrapper,
     set_accel_env_vars,
 )
-from ramalama.config import CONFIG, DEFAULT_PORT, DEFAULT_PORT_RANGE
+from ramalama.config import DEFAULT_PORT, DEFAULT_PORT_RANGE
 from ramalama.console import EMOJI
 from ramalama.engine import Engine, dry_run
 from ramalama.gguf_parser import GGUFInfoParser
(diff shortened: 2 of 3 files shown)


thom311:
@sourcery-ai summary

thom311:
https://github.com/containers/ramalama/blob/6556b04df90545730afa5e647142eeb839b5b03e/ramalama/quadlet.py#L26 doesn't look right. I don't think that `args.rag` is ever set to anything but `"rag"` (or where does that happen?). Any suggestion for how to fix this?

ericcurtin:
@bmahabirbu PTAL

rhatdan:
What is supposed to happen is -rag is appended to the end of rag images.  So if user does not specify an  image, then they are supposed to ACCELL_IMAGE-RAG:$VERSION.
For example on a Nvidia system, they should run the image in 

quay.io/ramalama/cuda-rag:0.9 Image (With ramalama 0.9.*)

I have opened a different PR to attempt to fix this.

thom311:
>  So if user does not specify an image, then they are supposed to ACCELL_IMAGE-RAG:$VERSION.

Sure. I don't think the current variant of the patch changes that (does it?).

This PR is mostly about what happens when the user specifies `--image`.

> What is supposed to happen is -rag is appended to the end of rag images.

Do you mean, also if the user specifies the `--image` on the command line?

Maybe add:
```diff
diff --git i/ramalama/cli.py w/ramalama/cli.py
index 32abef1d5739..244d828bd3ce 100644
--- i/ramalama/cli.py
+++ w/ramalama/cli.py
@@ -191,7 +191,7 @@ The RAMALAMA_CONTAINER_ENGINE environment variable modifies default behaviour.""
     parser.add_argument(
         "--image",
         default=None,
-        help=f"OCI container image to run with the specified AI model. The default (no-rag) image is {repr(accel_image(CONFIG, None))}.",
+        help=f"OCI container image to run with the specified AI model. The default (no-rag) image is {repr(accel_image(CONFIG, None))}. For rag mode, the suffix \"-rag\" is appended to the image name.",
         action=OverrideDefaultAction,
         completer=local_images,
     )
@@ -270,6 +270,15 @@ def post_parse_setup(args):
 
     if args.image is None:
         args.image = accel_image(CONFIG, args)
+    else:
+        if hasattr(args, "rag") and args.rag:
+            parts = args.image.rsplit(":", 1)
+            image = parts[0]
+            if not image.endswith("-rag"):
+                image += "-rag"
+            if len(parts) > 1:
+                image += ":" + parts[1]
+            args.image = image
 
     if hasattr(args, "MODEL") and args.subcommand != "rm":
         resolved_model = shortnames.resolve(args.MODEL)
```

This at least preserves the tag. Because, all the current code will replace `--image repo/image:tag` with `repo/image-rag:0.9`.

---

However, I find it wrong trying to be smarter than the user. If the user makes an invocation that requires a rag image, then they should specify the right name with the `--image` parameter. They then should not be bothered by the convention that their image needs to have a "-rag" suffix (or getting that suffix added automatically).

This patch no longer automatically adds the "-rag" suffix to the `--image` option, so there is a change in behavior. Problem?

If we assume that the user might not easily anticipate whether ramalama ends up needing a rag image or not, then there should be a separate option `--image-rag`. The user can then specify both images, and the right one will be used. For example,  if you write an `alias myramalama='ramalama --image ... --image-rag ...'`.

We could also combine both: if the user specifies only `--image`, they automatically get "-rag` added. If they specify both options, they can be explicit about the full names. That preserves current behavior.

rhatdan:
I think we need to add some tests to make sure this is preserved. I added a test to my version of this PR.

With --dryrun we want to make sure `--image` is honored. We also want to make sure that if the user specified an `--rag OCI` that the default_image for their GPU is pulled with the `-rag` version and the correct tag maintained.

rhatdan:
Closing in favor of: https://github.com/containers/ramalama/pull/1527

Please review that one, to make sure it matches  your expectation.

PR #1527, opened 2025-06-14
honor the user specifying the image

rhatdan:
Currently we are ignoreing the user specified image if it does not contain a ':'

Fixes: https://github.com/containers/ramalama/issues/1525

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -926,6 +926,7 @@ def run_cli(args):
         model.serve(args, quiet=True) if args.rag else model.run(args)
 
     except KeyError as e:
+        logger.debug(e)
         try:
             args.quiet = True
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -92,6 +103,8 @@ jobs:
 
       - name: run bats
         run: |
+	   mkdir /mnt/tmp
+	   export TMPDIR=/mnt/tmp
```
lsm5, line 107: @rhatdan ya, this would work for github actions. But not if we're running it on any other CI env. Long term, might be best to make it a general change in `test/system/helpers.bash` if possible.

ericcurtin:
Cleanup PR, noticed the red in the UI

https://github.com/containers/ramalama/pull/1538

ericcurtin:
This broke the build, never passed

PR #1529, opened 2025-06-14
Add colors to "ramalama serve" if we can

ericcurtin:
I don't notice any difference but a lot of things are LOG_INFO in llama.cpp

## Summary by Sourcery

Add automatic detection and enabling of colored logs in the `ramalama serve` command when run in a color-capable terminal.

New Features:
- Enable colored logging in `ramalama serve` by appending the `--log-colors` flag when the terminal supports ANSI colors.

Enhancements:
- Introduce a `should_colorize` helper that checks the `TERM` environment variable and `isatty()` status to determine if color output is supported.

### DIFF
--- a/ramalama/model.py
+++ b/ramalama/model.py
@@ -50,6 +50,11 @@ with the default RamaLama
 $(error)s"""
 
 
+def should_colorize():
+    t = os.getenv("TERM")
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

PR #1531, opened 2025-06-15
Add ramalama chat command

rhatdan:
For now we will just add the chat command, next PR will remove the external chat command and just use this internal one.

## Summary by Sourcery

Introduce a new internal "ramalama chat" CLI command to enable OpenAI chat over REST API and update documentation accordingly.

New Features:
- Add "chat" subcommand and associated CLI logic (chat_parser and chat_cli) for interacting with OpenAI via REST API.

Documentation:
- Add ramalama-chat(1) man page and integrate it into the docs index.
- Update README and main man page to list the new "ramalama chat" command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1041,6 +1041,7 @@ $ cat /usr/share/ramalama/shortnames.conf
 | ------------------------------------------------------ | ---------------------------------------------------------- |
 | [ramalama(1)](https://github.com/containers/ramalama/blob/main/docs/ramalama.1.md)                      | primary RamaLama man page                                  |
 | [ramalama-bench(1)](https://github.com/containers/ramalama/blob/main/docs/ramalama-bench.1.md)| benchmark specified AI Model                                         |
+| [ramalama-chat(1)](https://github.com/containers/ramalama/blob/main/docs/ramalama-chat.1.md)| chat with specified OpenAI RESTAPI                         |
 | [ramalama-containers(1)](https://github.com/containers/ramalama/blob/main/docs/ramalama-containers.1.md)| list all RamaLama containers                               |
 | [ramalama-convert(1)](https://github.com/containers/ramalama/blob/main/docs/ramalama-convert.1.md)      | convert AI Model from local storage to OCI Image           |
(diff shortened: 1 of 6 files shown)


ericcurtin (approved):

ericcurtin:
It's a good idea we changed to --url

We probably want the positional arg to be model (maybe optionally) in future when we start supporting model swapping

ericcurtin:
Be interesting if we can make --url work with:

--url unix://somesocketfile

rhatdan:
Yes I agree.  Next step is to remove libexec/ramalama-client-core.

I did turn on gemini for the repo as well, will see how well it does.

ericcurtin:
copilot suggested do something like this, that could be complete junk though, never tested, at least this solution uses urllib.request so one doesn't have to re-implement HTTP (even though the basics of http we need, wouldn't be that hard to re-implement if needs be):

```
import urllib.request
import socket
import json

DOCKER_SOCKET = "/var/run/docker.sock"
DOCKER_PATH = "/images/json"

class UnixHTTPConnection(urllib.request.HTTPConnection):
    def __init__(self, unix_socket_path):
        super().__init__('localhost')
        self.unix_socket_path = unix_socket_path

    def connect(self):
        self.sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
        self.sock.connect(self.unix_socket_path)

class UnixHTTPHandler(urllib.request.AbstractHTTPHandler):
    def __init__(self, unix_socket_path):
        self.unix_socket_path = unix_socket_path
        super().__init__()

    def http_open(self, req):
        return self.do_open(lambda host, **_: UnixHTTPConnection(self.unix_socket_path), req)

def get_docker_images():
    opener = urllib.request.build_opener(UnixHTTPHandler(DOCKER_SOCKET))
    url = f"http://localhost{DOCKER_PATH}"
    with opener.open(url) as response:
        data = response.read()
        return json.loads(data)

if __name__ == "__main__":
    images = get_docker_images()
    print(json.dumps(images, indent=2))
```

ericcurtin:
We will want to delete "ramalama client" also with the libexec, this is a copy of that with a different name

PR #1532, opened 2025-06-16
Suggest using uv pip install to get missing module

rhatdan:
## Summary by Sourcery

Update missing dependency error messages for Huggingface and ModelScope integrations and refactor exception handling to use a dynamic message method.

Enhancements:
- Add “uv” to the list of supported pip installation tools in dependency instructions
- Revise example commands to use “uv pip install <package>”
- Replace hardcoded missing-module messages with a call to self.get_missing_message() for consistency


rhatdan:
@ericcurtin PTAL

PR #1535, opened 2025-06-16
Not sure this is supposed to be here

ericcurtin:
Think it's only meant for the:

container-images/scripts/build-cli.sh

version, it's breaking podman on my bootc system and replacing /usr/bin/podman with a broken /usr/bin/podman-remote symlink.


rhatdan:
This is needed in the ramalama-cli install at least.

container-images/scripts/build_llama_and_whisper.sh
```diff
@@ -278,8 +278,7 @@ clone_and_build_llama_cpp() {
 }
 
 install_ramalama() {
-  # link podman-remote to podman for use by RamaLama
-  ln -sf /usr/bin/podman-remote /usr/bin/podman
```
rhatdan, line 282: You could just move this below the if /run/ramalama line below, or we should move it to the containerfile that installs podman-remote.

rhatdan:
Actually after further search this is  correct.
LGTM

PR #1538, opened 2025-06-16
Convert tabs to spaces

ericcurtin:
Saw this in github ui

## Summary by Sourcery

CI:
- Replace tabs with spaces in .github/workflows/ci.yml for Bats job steps

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,8 +103,7 @@ jobs:
 
       - name: run bats
         run: |
-	   mkdir /mnt/tmp
-	   export TMPDIR=/mnt/tmp
+           export TMPDIR=$(mktemp -d)
            make validate
            make bats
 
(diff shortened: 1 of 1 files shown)


sarroutbi (approved):
LGTM

ericcurtin:
/mnt doesn't exist on macOS:

```
export TMPDIR=$(mktemp -d)
```

is more portable

ericcurtin:
Merging this because the build is broken, it's more important to get the build green again

ericcurtin:
@rhatdan just checking you didn't pick /mnt for a specific reason? Hoping it didn't have any extra disk space or something, merged this quickly as the mkdir /mnt/tmp broke our macOS build

rhatdan:
The issue is on Linux systems, the /mnt is a large amount of space as I understand it.

If you look at the docker one, it says:

      # /mnt has ~ 65 GB free disk space. / is too small.


So for at lease Linux tests we should be setting TMPDIR to point at /mnt.

ericcurtin:
@rhatdan maybe it works... We might need to if around it for macOS though, if definitely broke macOS... But there's so much instability now, it's hard to keep track

PR #1548, opened 2025-06-17
Run bats test with TMPDIR pointing at /mnt/tmp

rhatdan:
## Summary by Sourcery

CI:
- Create a /mnt/tmp directory and set TEMPDIR to /mnt/tmp before running bats tests in CI to avoid disk space issues

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,6 +103,9 @@ jobs:
 
       - name: run bats
         run: |
+	   # /mnt has ~ 65 GB free disk space. / is too small.
+           mkdir -p /mnt/tmp
+           TEMPDIR=/mnt/tmp
            make validate
            make bats
 


rhatdan:
@ericcurtin going to try this again, I am not sure why you reverted this one before.  This should only run on Ubuntu systems.

.github/workflows/ci.yml
```diff
@@ -103,6 +103,9 @@ jobs:
 
       - name: run bats
         run: |
+	   # /mnt has ~ 65 GB free disk space. / is too small.
```
ericcurtin, line 106: I dunno if this one makes a difference

ericcurtin:
There was several issues with the last version of this PR, this PR is different, failed on macOS, runroot path too long, I was battling with it for hours, this might be worth changing:

```
test/system/helpers.podman.bash
662:    for opt in root runroot tmpdir;do
```

to

```
test/system/helpers.podman.bash
662:    for opt in root tmpdir; do
```

to see if it works, if we hit this issue, this PR is different only does TMPDIR in one place. The runroot character length restriction was removed in the latest podman, I tried to install Ubuntu 25.04 podman on 24.04, it failed badly.

Hopefully we see a green build

.github/workflows/ci.yml
ericcurtin: I'm gonna assume no, we got a green build

PR #1549, opened 2025-06-17
Tabs to spaces

ericcurtin:
github UI showed red, changing just in case, incorrect tabs or spaces can cause github ui to skip builds.

## Summary by Sourcery

CI:
- Correct YAML indentation in .github/workflows/ci.yml to prevent CI parsing issues due to tabs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,7 +103,7 @@ jobs:
 
       - name: run bats
         run: |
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

rhatdan:
We should see if there is a way to check for this in lint.  Emacs is added these. Sorry

PR #1550, opened 2025-06-17
Replace ramalama-client-code with ramalama chat

rhatdan:
## Summary by Sourcery

Replace the legacy ramalama-client-core binary with the unified `ramalama chat` command, remove the old binary from packaging, and add a hidden PID kill flag.

New Features:
- Add a suppressed `--pid2kill` option to the chat parser

Enhancements:
- Invoke `ramalama chat` instead of `ramalama-client-core` in `client_cli`
- Remove `ramalama-client-core` from libexec and update pyproject.toml to drop it


rhatdan:
@ericcurtin PTAL, I have not tested this with a container with the command built in yet.  Will try it once I get back to a WIFI signal.

ericcurtin:
Build failure:

```
mkdir: cannot create directory ‘/mnt/tmp’
```

LGTM, this is a well tested code path in CI, so if the build goes green, seems fine to me

rhatdan:
@ericcurtin this is ready for review.

A couple of big changes here
1: We can no longer run ramalama run containers with --network=none, since we are always running ramalama serve under the covers.
2: The ramalama serve is running in the client but the ramalama chat is running on the host when you run with --nocontainer.

ericcurtin (approved):

ericcurtin:
All makes sense to me

PR #1574, opened 2025-06-21
Make model argument mandatory

ericcurtin:
To be consistent with "ramalama run" experience. Inferencing
servers that have implemented model-swapping require this. In the
case of servers like llama-server that only load one server, any
value is sufficient.


rhatdan:
LGTM

PR #1578, opened 2025-06-22
API key support

ericcurtin:
If we pass --api-key, we can talk to OpenAI providers.

### DIFF
--- a/ramalama/chat.py
+++ b/ramalama/chat.py
@@ -70,7 +70,7 @@ class RamaLamaShell(cmd.Cmd):
         self.request_in_process = False
         self.prompt = args.prefix
 
-        self.url = f"{args.url}/v1/chat/completions"
+        self.url = f"{args.url}/chat/completions"
 
     def handle_args(self):
         if self.args.ARGS:
@@ -106,6 +106,13 @@ class RamaLamaShell(cmd.Cmd):
         headers = {
             "Content-Type": "application/json",
         }
+
+        if self.args.api_key:
+            if len(self.args.api_key) < 20:
+                print("Warning: Provided API key is invalid.")
+
+            headers["Authorization"] = f"Bearer {self.args.api_key}"
+
         request = urllib.request.Request(self.url, data=json_data, headers=headers, method="POST")
 
         return request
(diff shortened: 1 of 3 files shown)


ramalama/chat.py
```diff
@@ -70,7 +70,11 @@ def __init__(self, args):
         self.request_in_process = False
         self.prompt = args.prefix
 
-        self.url = f"{args.url}/v1/chat/completions"
+        if args.gemini:
```
rhatdan: I would prefer this to not be gemini option, but perhaps 
--cloud-api gemini? So that it becomes easy to add other cloud api servers that we might want to make guessable.

ramalama/chat.py
```diff
@@ -70,7 +70,11 @@ def __init__(self, args):
...
+            host = "https://generativelanguage.googleapis.com"
```
rhatdan: Why use a local variable rather then just putting it in the URL below.

The --gemini option needs to conflict with --url.

ericcurtin: Mainly because I'm trying to keep long length small but can combine into one variable again

ramalama/chat.py
ericcurtin: Looking at the docs of various providers:

https://docs.anthropic.com/en/api/openai-sdk
https://openrouter.ai/docs/quickstart
https://ai.google.dev/gemini-api/docs/openai#python

I think we can make it super generic just by adding an --api-key option

ramalama/cli.py
```diff
@@ -920,7 +926,7 @@ def chat_parser(subparsers):
         help='possible values are "never", "always" and "auto".',
     )
     parser.add_argument("--prefix", type=str, help="prefix for the user prompt", default=default_prefix())
-    parser.add_argument("--url", type=str, default="http://127.0.0.1:8080", help="the host to send requests to")
+    parser.add_argument("--url", type=str, default="http://127.0.0.1:8080/v1", help="the host to send requests to")
```
rhatdan: Should be `the url to sent requests to`

rhatdan: This is preexisting so will merge to get release out, but we should fix.

rhatdan: Actually this also needs to have man page changes, so need a new PR.

rhatdan:
LGTM

Issue #1587, opened 2025-06-23
`ramalama --help` pulls a large image before printing anything


ktdreyer:
### Feature request description

The `ramalama --help` command pulls a large image (1.3GB) before displaying anything. It would be great to print the help text immediately without pulling such a large image. (I thought it was broken for a while and ctrl-c'd and restarted a few times, until I noticed in the stack traces that `ramalama` was actually downloading an image.)

### Suggest potential solution

Since `ramalama` or `-h` or `--help` are common ways that users can explore the CLI sub-commands, it would be great to support printing output quickly to users, so that we don't have to pull large images in new environments.

### Have you considered any alternatives?

In playing around with this, I tried a few arguments to see if I could find an easy way to work around this. I tried specifying a non-default image, and the `--nocontainer` flag, but I guess those do not do what I want.

```
ramalama --image quay.io/podman/hello --nocontainer --help
```

Even with those args, `--help` still downloads the `quay.io/ramalama/ramalama:0.9` image.

rhatdan:
Ok what is happening here is during help we are trying to determine the default image to be used for the container based on the version of RamaLama.

For example on my system we are using quay.io/ramalama/cuda:0.9.

RamaLama checks if this image exists locally, if it does not exists locally then it attempts to pull the image from the registry, if it succeeds, then it reports the image name in the --help output.

  --image IMAGE         OCI container image to run with the specified AI model (default: **quay.io/ramalama/cuda:0.9**)

The issue is there is no way to know if this image exists with podman/docker without attempting to pull it.

We could attempt to use skopeo to check if it exists, but this would only fix the problem on systems with skopeo IE Non Macs and Windows systems.

Not sure what is the best way to fix this.

ktdreyer:
Thanks Dan!

I read `accel_image()` and I see that tries to pull the entire image. I would expect `--help` to not make network calls. 

What if `--help` did not print the default value?

I can get the default image value with `ramalama info | jq .Image`.

rhatdan:
Yup, the question is what is the bigger problem.  Once the image is pulled once, then --help will not cause the issue.  But I do agree it stinks to pull it at all.  The problem is there is no easy way to check at the registry, whether or not the image exists, without attempting to pull it from a Podman or Docker point of view.

rhatdan:
It might be best to add a don't pull option to accel_image, and then just guess that the image exists.

PR #1589, opened 2025-06-23
Don't pull image when doing ramalama --help call

rhatdan:
Fixes: https://github.com/containers/ramalama/issues/1587

## Summary by Sourcery

Add a `nopull` option to the image selection logic and apply it in the CLI so that running `ramalama --help` (or other non-runtime invocations) no longer triggers a container image pull.

Bug Fixes:
- Prevent unnecessary image pull when calling `ramalama --help`

Enhancements:
- Introduce a `nopull` parameter to `accel_image` to bypass pulling the container image
- Invoke `accel_image` with `nopull=True` when setting the CLI’s default `--image` value

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -192,7 +192,7 @@ The RAMALAMA_CONTAINER_ENGINE environment variable modifies default behaviour.""
     )
     parser.add_argument(
         "--image",
-        default=accel_image(CONFIG),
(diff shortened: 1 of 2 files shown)


ktdreyer:
Thanks Eric & Dan!

I proposed a small improvement here: https://github.com/containers/ramalama/pull/1611

PR #1591, opened 2025-06-24
Add CPU based vllm inferencing

ericcurtin:
So we can do vllm enablement and testing

## Summary by Sourcery

Enable CPU-based vllm inferencing in the ramalama container by updating the base image to a vllm CPU OpenAI UBI9 image, running as root, and clearing the default entrypoint.

Enhancements:
- Switch container base image to quay.io/rh-aiservices-bu/vllm-cpu-openai-ubi9:0.3 for CPU-based vllm support

Build:
- Add USER 0 to run container processes as root
- Remove the default ENTRYPOINT to allow custom command execution

PR #1596, opened 2025-06-24
seperate rag to its own python userspace using venv

bmahabirbu:
unblocks https://github.com/containers/ramalama/pull/1591

If moving to fedora 42 we can install uv with dnf and use that instead of manually creating a venv


rhatdan:
@bmahabirbu is this still something we want?  Needs a rebase and then answer the AI reviewers.

bmahabirbu:
Closing, the change to uv will happen when separating rag dependency from inference containers

PR #1603, opened 2025-06-25
container-images: pin mesa version to COPR

slp:
When building on RHEL-based systems make sure we install the mesa version from the COPR, which enables the virtio driver and has the patches to force alignment to 16K (needed for GPU acceleration on macOS, but harmless to other systems).

We also need to add "--nobest" to "dnf update" to ensure it doesn't get frustrated by being unable to install the mesa package from appstream.

## Summary by Sourcery

Update container image build scripts to install and maintain a specific COPR-provided mesa package for reliable GPU acceleration, preserve COPR repos during cleanup, and add the --nobest flag to dnf updates to prevent package resolution issues.

Build:
- Enable COPR and pin mesa-vulkan-drivers to a fixed version on Fedora builds
- Stop removing COPR repository files during cleanup to retain pinned repo entries
- Add --nobest to dnf update in build_rag.sh to avoid package resolution failures

### DIFF
--- a/container-images/scripts/build_llama_and_whisper.sh
+++ b/container-images/scripts/build_llama_and_whisper.sh
@@ -89,7 +89,7 @@ add_stream_repo() {
 
 rm_non_ubi_repos() {
   local dir="/etc/yum.repos.d"
-  rm -rf $dir/mirror.stream.centos.org_9-stream_* $dir/epel* $dir/_copr:*
+  rm -rf $dir/mirror.stream.centos.org_9-stream_* $dir/epel*
 }
 
 is_rhel_based() { # doesn't include openEuler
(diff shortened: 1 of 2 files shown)


slp:
I've just noticed the previous commit has switched to building on F42 by default. Let's put this one on hold for the moment.

ericcurtin:
@slp @rhatdan we do need a solution though:

https://github.com/containers/ramalama/pull/1595

If we released container images today, macOS support would be broken, we have at least two options:

1. Go back to UBI9.
2. Stay on Fedora:42 but @slp rebuilds the forked mesa for Fedora 42, (and 43, 44, when they get released if the changes haven't made it from upstream yet). If we are rebuilding @slp please pick up the Asahi changes from Fedora rawhide, means we can stop building a separate container image for Asahi.

rhatdan:
Can we get the mesa stuff updated for Fedora 42 or is the a reason to stick with ubi9?

slp:
> Stay on Fedora:42 but @slp rebuilds the forked mesa for Fedora 42, (and 43, 44, when they get released if the changes haven't made it from upstream yet). If we are rebuilding @slp please pick up the Asahi changes from Fedora rawhide, means we can stop building a separate container image for Asahi.

I've created a new COPR with a patched F42 mesa. I couldn't pick up the Asahi changes because would have meant switching to 25.1.x, and I haven't had the time to test it properly yet. I think it's safer to wait a bit first.

We should also see if it'd be possible to get the patch into mainline Fedora.

ericcurtin (approved):

rhatdan:
LGTM

PR #1604, opened 2025-06-25
Check if api key exists first

ericcurtin:
When "ramalama chat" is called api key always exists. When "ramalama run" is called api key doesn't exist causing "ramalama run" to be broken, prior to this change.

## Summary by Sourcery

Bug Fixes:
- Add a hasattr check around args.api_key to avoid attribute errors in commands without an API key.


ericcurtin:
Closing, duplicate of:

https://github.com/containers/ramalama/pull/1601

PR #1605, opened 2025-06-25
Switchout hasattr for getattr wherever possible

rhatdan:
This greatly simplifies the code logic.

## Summary by Sourcery

Enhancements:
- Replace hasattr checks with getattr calls providing default values in ramalama/engine.py, model.py, quadlet.py, kube.py, model_factory.py, cli.py, oci.py, and stack.py to streamline argument handling logic


ericcurtin (approved):

ericcurtin:
Shouldn't have merged, meant to approve, I'll keep an eye on the build and fix accordingly

ericcurtin:
Somewhat related, I like how a bot suggested, set the default as this for me:

```
    parser.add_argument(
        "--api-key",
        type=str,
        default=os.getenv("API_KEY"),
        help="OpenAI-compatible API key. Can also be set via the API_KEY environment variable.",
    )
 ```
 
ideally in default we should call a function that handles the, config file, env var precedence. With CLI option taking precedence over everything.

PR #1606, opened 2025-06-25
Allow std input

ericcurtin:
We used to have this feature, got dropped recently accidentally, can do things like:

`cat text_file_with_prompt.txt | ramalama run smollm:135m`

or

`cat some_doc | ramalama run smollm:135m Explain this document:`

## Summary by Sourcery

Allow reading from standard input for prompts, merging it with any provided CLI arguments and passing the combined text to the chat handler.

New Features:
- Restore support for reading prompt content from stdin when piped to the CLI

Enhancements:
- Combine command-line arguments and stdin input into a single prompt

### DIFF
--- a/ramalama/chat.py
+++ b/ramalama/chat.py
@@ -74,10 +74,19 @@ class RamaLamaShell(cmd.Cmd):
         self.url = f"{args.url}/chat/completions"
 
     def handle_args(self):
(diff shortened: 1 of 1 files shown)


ericcurtin:
This might help @ieaves

rhatdan:
LGTM

PR #1609, opened 2025-06-25
Separate build image into its own VM

rhatdan:
## Summary by Sourcery

Separate the container image build into its own CI job and enhance the build script for selective image construction.

Enhancements:
- Introduce `-s` flag and `build_all` variable in `container_build.sh` to support selective image building.

CI:
- Remove container build steps from existing `ci` and `docker` jobs.
- Add a new standalone `build` job on Ubuntu 24.04 that sets up Podman 5 and runs image construction.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,13 +34,52 @@ jobs:
         run: |
           make lint
 
+  build-image:
+    runs-on: ubuntu-24.04
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install the latest version of uv and activate the environment
+        uses: astral-sh/setup-uv@v6
+        with:
+          activate-environment: true
+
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -34,13 +34,44 @@ jobs:
...
+           # Update package lists
+           sudo apt-get update
+           sudo apt-get purge firefox
+           # Install specific podman version
+           sudo apt-get upgrade
+
+      - name: Build a container for CPU inferencing
+           make install-requirements
+	   run: |
+	   ./container_build.sh build -s ramalama
```
tstellar: It looks like something got mixed up here with the `run` step.

rhatdan: How do I fix it?

rhatdan:
OK I think I finally have the test running.

PR #1619, opened 2025-06-27
Use standard zsh completion directory

carlwgeorge:
We're currently using `/usr/share/zsh/vendor-completions` for zsh completions.  However, the RPM macro `%{zsh_completions_dir}` (which is required by the Fedora packaging guidelines) is defined as `/usr/share/zsh/site-functions`, so let's switch to that.

https://docs.fedoraproject.org/en-US/packaging-guidelines/ShellCompletions/


ericcurtin:
Thanks @carlwgeorge

PR #1623, opened 2025-06-27
Want to pick up support for gemma3n

ericcurtin:
And the other latest and greatest llama.cpp features


ericcurtin:
As requested by @giuseppe

rhatdan:
LGTM

Issue #1572, opened 2025-06-20
llama server is not available at localhost when running "ramalama serve --host 127.0.0.1"


lephuongbg:
### Issue Description

To try to serve a model inside localhost only, I attempted to run `RAMALAMA_CONTAINER_ENGINE=docker ramalama serve --host 127.0.0.1 qwen2.5-coder:3b`. Trying to curl to `http://127.0.0.1:8080` failed, because ramalama instead of telling docker to do port forward from 127.0.0.1, it tells llama.cpp to serve at 127.0.0.1, which in turn makes docker port forwarding no longer work because it needs llama.cpp to always serve at 0.0.0.0 inside the container.

### Steps to reproduce the issue

Step 1: Run `RAMALAMA_CONTAINER_ENGINE=docker ramalama serve --host 127.0.0.1 qwen2.5-coder:3b`
Step 2: 
```
curl -X POST http://localhost:8080/infill \
-H "Content-Type: application/json" \
-d '{
  "input_prefix": "import math\n\ndef calculate_factorial(n):\n    \"\"\"\n    Calculates the factorial of a non-negative integer.\n    \"\"\"\n    if not isinstance(n, int) or n < 0:\n        raise ValueError(\"Input must be a non-negative integer\")",
  "input_suffix": "\n    return result\n\n# Example Usage:\nprint(f\"The factorial of 5 is: {calculate_factorial(5)}\")\nprint(f\"The factorial of 0 is: {calculate_factorial(0)}\")"
}'
```

### Describe the results you received

Connection refused

### Describe the results you expected

can connect to http://localhost:8080 succesfully

### ramalama info output

```yaml
{
    "Accelerator": "intel",
    "Engine": {
        "Info": {
            "Architecture": "x86_64",
            "CDISpecDirs": [
                "/etc/cdi",
                "/var/run/cdi"
            ],
            "CPUSet": true,
            "CPUShares": true,
            "CgroupDriver": "systemd",
            "CgroupVersion": "2",
            "ClientInfo": {
                "Arch": "amd64",
                "BuildTime": "Fri May 30 00:00:00 2025",
                "Context": "default",
                "Debug": false,
                "DefaultAPIVersion": "1.50",
                "GitCommit": "1.fc42",
                "GoVersion": "go1.24.3",
                "Os": "linux",
                "Plugins": [
                    {
                        "Name": "buildx",
                        "Path": "/usr/libexec/docker/cli-plugins/docker-buildx",
[...]

PR #1628, opened 2025-06-28
Fix handling of --host option when running in a container

rhatdan:
When you run a Model server within a container and only wanted it bound to a certain port, the port binding should happen to the container not inside of the container.

Fixes: https://github.com/containers/ramalama/issues/1572

Also fix handling of -t option, should not be used with anything other then run command, and now I am not sure of that.

The LLAMA_PROMPT_PREFIX= environment variable should not be set within containers as an environment variable, since we are doing chat on the outside.

## Summary by Sourcery

Fix host binding behavior for serve commands in container mode by applying host options as Docker port mappings, restrict TTY option usage to run, and remove internal LLAMA_PROMPT_PREFIX injection

Bug Fixes:
- Convert --host flags in container mode into Docker port mappings instead of passing them to the inner process
- Prevent --host from being added inside the container at runtime
- Restrict the -t (TTY) option to only the run subcommand

Enhancements:
- Remove internal LLAMA_PROMPT_PREFIX injection to avoid setting it within containers
- Adjust use_tty logic to correctly determine when to allocate a TTY

Tests:
- Update system tests to assert the absence of --host in container mode and verify host:port mapping with -p

### DIFF
--- a/ramalama/engine.py
+++ b/ramalama/engine.py
@@ -7,7 +7,6 @@ import sys
 # Live reference for checking global vars
 import ramalama.common
 from ramalama.common import check_nvidia, exec_cmd, get_accel_env_vars, perror, run_cmd
-from ramalama.console import EMOJI
 from ramalama.logger import logger
 
 
@@ -31,7 +30,6 @@ class Engine:
         self.add_privileged_options()
         self.add_pull_newer()
         self.add_rag()
-        self.add_subcommand_env()
         self.add_tty_option()
         self.handle_podman_specifics()
         self.add_detach_option()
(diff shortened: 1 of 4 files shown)


rhatdan:
@ericcurtin Verify my assumptions are correct?  Does llama-server by default bind to ports 0.0.0.0?

rhatdan:
@lephuongbg could you verify that this fixes your issue?

ramalama/engine.py
```diff
@@ -91,14 +89,7 @@ def use_tty(self):
             return False
         if getattr(self.args, "ARGS", None):
             return False
-        return getattr(self.args, "subcommand", "") != "run"
-
-    def add_subcommand_env(self):
```
ericcurtin, line 96: This makes sense to me now that chat is outside the container

ericcurtin (approved):

Issue #1618, opened 2025-06-26
Tests can't be run on OSX for users using the applehv engine


ieaves:
### Issue Description

Tests are blocked by the no gpu confirmation prompt when on Mac.

### Steps to reproduce the issue

1. Be on OSX
2. Create a podman machine with the applehv runtime
3. run tests

### Describe the results you received

Tests are blocked by a user input prompt

### Describe the results you expected

Tests which don't require gpu acceleration should run.

### ramalama info output

```yaml
{
    "Accelerator": "none",
    "Engine": {
        "Info": {
            "Client": {
                "APIVersion": "5.5.1",
                "BuildOrigin": "brew",
                "Built": 1749147935,
                "BuiltTime": "Thu Jun  5 13:25:35 2025",
                "GitCommit": "",
                "GoVersion": "go1.24.4",
                "Os": "darwin",
                "OsArch": "darwin/arm64",
                "Version": "5.5.1"
            },
            "host": {
                "arch": "arm64",
                "buildahVersion": "1.40.1",
                "cgroupControllers": [
                    "cpuset",
                    "cpu",
                    "io",
                    "memory",
                    "pids",
                    "rdma",
                    "misc"
                ],
                "cgroupManager": "systemd",
                "cgroupVersion": "v2",
                "conmon": {
                    "package": "conmon-2.1.12-3.fc41.aarch64",
                    "path": "/usr/bin/conmon",
                    "version": "conmon version 2.1.12, commit: "
                },
                "cpuUtilization": {
                    "idlePercent": 98.56,
                    "systemPercent": 0.66,
                    "userPercent": 0.79
                },
                "cpus": 7,
                "databaseBackend": "sqlite",
                "distribution": {
                    "distribution": "fedora",
                    "variant": "coreos",
                    "version": "41"
                },
                "eventLogger": "journald",
                "freeLocks": 2048,
                "hostname": "localhost.localdomain",
                "idMappings": {
                    "gidmap": null,
                    "uidmap": null
                },
                "kernel": "6.12.13-200.fc41.aarch64",
                "linkmode": "dynamic",
                "logDriver": "journald",
                "memFree": 5219094528,
                "memTotal": 7222493184,
                "networkBackend": "netavark",
                "networkBackendInfo": {
                    "backend": "netavark",
                    "dns": {
                        "package": "aardvark-dns-1.14.0-1.fc41.aarch64",
                        "path": "/usr/libexec/podman/aardvark-dns",
[...]

PR #1632, opened 2025-06-30
Adds a user configuration setting to disable gpu prompting

ieaves:
This PR provides two pieces of functionality

1. The ability to nest configuration settings as structured objects
2. A user configuration field to disable no_gpu prompting. This was inspired by issue #1618 

If we like this path, I will modify the tests to disable the gpu prompting with a fixture.


ieaves:
This morning I thought this was going to be an easy fix but once I started running unit tests I belatedly realized the issue was more complicated than it first appeared because of a circular import issue and the dependency tree between common and config. In order to break the cyclical dependency I needed to pull apple_vm out of the get_default_engine logic being used in the config which is in any case a good idea as otherwise we end up calling it multiple times (which itself involves a subprocess call to the podman cli). 

The equivalent logic for OSX has been moved to the `initialize_environment` code under the root __init__.py.

Next up, I needed a place to store the new prompt disabling configuration setting. I went back and forth on either setting up a new local cache for user preferences using something like [shelve](https://docs.python.org/3/library/shelve.html) but anticipated some hesitation around splitting the config and introducing a second file which has to be managed. I landed, instead, on leaving it in the default config but under a nested `user` attribute. This required some additional machinery so that we could instantiate and type nested config objects.

It was at this point that I realized modifying the config _file_ in a testing environment was slightly complicated by the multiple config file locations we traverse and that, for the purposes of the initial issue, a better resolution would be to allow these settings to be defined through environment variables. In order to do that I implemented a pattern for the arbitrary resolution of config values in environment variables where the config value

```toml
[ramalama.subkey]
key = value
```

corresponds to the environment variable `RAMALAMA_SUBKEY__KEY=value` or equivalently `RAMALAMA__SUBKEY__KEY=value`.

Finally, I added runtime type validation logic for these environmental variables. 

With that machinery in place I added changes in conftest to preload specific environment variables for testing purposes and then restore the users environment once testing completed.

ramalama/config.py
```diff
@@ -32,6 +33,33 @@ def get_default_store() -> str:
     return os.path.expanduser("~/.local/share/ramalama")
 
 
+def coerce_to_bool(value: Any) -> bool:
+    if isinstance(value, bool):
+        return value
+    elif isinstance(value, str):
```
rhatdan, line 39: Nit: No need for elif, if works fine.

ramalama/config.py
```diff
@@ -32,6 +33,33 @@ def get_default_store() -> str:
...
+        val = value.strip().lower()
+        if val in {"true", "1", "yes", "y"}:
+            return True
+        elif val in {"false", "0", "no", "n"}:
```
rhatdan, line 43: Nit: No need for elif, if works fine.

rhatdan:
LGTM

PR #1635, opened 2025-07-01
Add command to list available models

ericcurtin:
With commands such as:

  ramalama chat --url https://generativelanguage.googleapis.com/v1beta/openai --ls

we can now list the various models available.

## Summary by Sourcery

Introduce a model listing command, streamline API key management, and refactor model selection flags in the chat interface

New Features:
- Add `--list`/`--ls` option to display available models from the API endpoint
- Support specifying the inference model via `--model`/`-m` flag instead of a positional argument

Enhancements:
- Consolidate API key header injection and validation into a new `add_api_key` helper
- Refactor chat request logic to use the lowercase `model` flag and centralized API key handling

Chores:
- Remove obsolete `AttributeError` exception handler in the CLI

### DIFF
--- a/docs/ramalama-chat.1.md
+++ b/docs/ramalama-chat.1.md
@@ -26,6 +26,12 @@ Possible values are "never", "always" and "auto". (default: auto)
 #### **--help**, **-h**
 Show this help message and exit
 
+#### **--list**
+List the available models at an endpoint
+
+#### **--model**=MODEL
+Model for inferencing (may not be required for endpoints that only serve one model)
+
 #### **--prefix**
 Prefix for the user prompt (default: 🦭 > )
 
--- a/ramalama/chat.py
+++ b/ramalama/chat.py
@@ -64,6 +64,20 @@ def default_prefix():
     return "🦙 > "
 
 
+def add_api_key(args, headers=None):
+    # static analyzers suggest for dict, this is a safer way of setting
+    # a default value, rather than using the parameter directly
+    headers = headers or {}
+    if getattr(args, "api_key", None):
+        api_key_min = 20
+        if len(args.api_key) < api_key_min:
+            print("Warning: Provided API key is invalid.")
+
+        headers["Authorization"] = f"Bearer {args.api_key}"
+
+    return headers
+
+
 class RamaLamaShell(cmd.Cmd):
     def __init__(self, args):
         super().__init__()
@@ -121,19 +135,17 @@ class RamaLamaShell(cmd.Cmd):
         data = {
             "stream": True,
             "messages": self.conversation_history,
-            "model": self.args.MODEL,
         }
+
+        if getattr(self.args, "model", False):
+            data["model"] = self.args.model
+
         json_data = json.dumps(data).encode("utf-8")
         headers = {
             "Content-Type": "application/json",
         }
 
-        if getattr(self.args, "api_key", None):
-            if len(self.args.api_key) < 20:
-                print("Warning: Provided API key is invalid.")
-
-            headers["Authorization"] = f"Bearer {self.args.api_key}"
-
+        headers = add_api_key(self.args, headers)
         logger.debug("Request: URL=%s, Data=%s, Headers=%s", self.url, json_data, headers)
         request = urllib.request.Request(self.url, data=json_data, headers=headers, method="POST")
 
@@ -212,11 +224,24 @@ def chat(args):
         signal.signal(signal.SIGALRM, alarm_handler)
         signal.alarm(convert_to_seconds(args.keepalive))
 
+    list_models = getattr(args, "list", False)
+    if list_models:
+        url = f"{args.url}/models"
+        headers = add_api_key(args)
+        req = urllib.request.Request(url, headers=headers)
+        with urllib.request.urlopen(req) as response:
+            data = json.loads(response.read())
+            ids = [model["id"] for model in data.get("data", [])]
+            for id in ids:
+                print(id)
+
     try:
         shell = RamaLamaShell(args)
         if shell.handle_args():
             return
-        shell.loop()
+
+        if not list_models:
+            shell.loop()
     except TimeoutException as e:
         logger.debug(f"Timeout Exception: {e}")
         # Handle the timeout, e.g., print a message and exit gracefully
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -904,10 +904,16 @@ def chat_parser(subparsers):
         choices=['never', 'always', 'auto'],
         help='possible values are "never", "always" and "auto".',
     )
+    parser.add_argument(
+        "--list",
+        "--ls",
+        action="store_true",
+        help="list the available models at an endpoint",
+    )
     parser.add_argument("--prefix", type=str, help="prefix for the user prompt", default=default_prefix())
     parser.add_argument("--rag", type=str, help="a file or directory to use as context for the chat")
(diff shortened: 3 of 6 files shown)


ramalama/chat.py
```diff
@@ -212,11 +217,23 @@ def chat(args):
...
 
+    if args.list:
+        url = f"{args.url}/models"
+        headers = add_api_key(args)
+        req = urllib.request.Request(url, headers=headers)
+        with urllib.request.urlopen(req) as response:
+            data = json.loads(response.read())
+            ids = [model["id"] for model in data.get("data", [])]
+            for id in ids:
+                print(id)
```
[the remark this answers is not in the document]
ericcurtin: It's intended

ramalama/cli.py
```diff
@@ -1179,11 +1185,6 @@ def eprint(e, exit_code):
         eprint(f"pulling {e.geturl()} failed: {e}", errno.EINVAL)
     except HelpException:
         parser.print_help()
-    except AttributeError as e:
```
ericcurtin, line 1182: I think we should remove this, I keep on doing it manually anyway, it keeps masking real issues with the misleading error:

"ramalama: requires a subcommand"

rhatdan: Fine, I just use --debug to see the error, but removing it permanently, is a good idea, so if it happens in release we can get an easy bug report.

ericcurtin: I'd be open to just changing the string to something like below also, I think they call it a traceback in python3 speak, I think the main issue is the string, it's misleading, I don't think I've even encountered this and the error has actually been "requires a subcommand":

"ramalama: AttributeError please re-run with --debug for full traceback"

ericcurtin: I found a way to cover both cases "ramalama: requires a subcommand" and the rest. Just check in advance if we have a subcommand.

rhatdan:
LGTM
Can you add a test for this?

ericcurtin:
Test added, green builds

docs/ramalama-chat.1.md
```diff
@@ -26,6 +26,12 @@ Possible values are "never", "always" and "auto". (default: auto)
 #### **--help**, **-h**
 Show this help message and exit
 
+#### **--list**
+List the available models at an endpoint
+
+#### **--model**=MODEL
```
rhatdan, line 32: Shouldn't this be available for `ramalama run` as well?

test/system/040-serve.bats
```diff
@@ -121,6 +121,9 @@ verify_begin=".*run --rm"
     run_ramalama ps
     is "$output" ".*${container1}" "list correct for container1"
 
+    run_ramalama chat --ls
+    is "$output" "ollama://smollm:135m" "list of models available correct"
+
```
rhatdan, line 126: While you are in, could you add test for

run_ramalama run --model bogus

And verify an error message.

docs/ramalama-chat.1.md
ericcurtin: It's tricky, for `ramalama run` we always used the very familiar podman/docker/ollama syntax:

```
ramalama run MODEL [ARGS]
```

For "chat" I guess there's a lot of cases where mandating a model to be specified doesn't make sense, like when the inferencing server is vllm/llama-server directly and model isn't necessary because these inference servers only support single model and ignore the field anyway...

And in the case of:

```
ramalama chat --url https://generativelanguage.googleapis.com/v1beta/openai --ls 
```

It doesn't make much sense either.

ericcurtin: `ramalama chat --ls` works too, but isn't as useful, as it's always one value assuming local port 8080 is running llama-server

test/system/040-serve.bats
ericcurtin: ramalama run bogus?

or

ramalama chat --model bogus?

`ramalama run --model bogus` would be:

unrecognized arguments: --model

ericcurtin: Funnily enough, I would regard this a bug:

```
ramalama run bogus
🦙 > Error: Manifest for bogus:latest was not found in the Ollama registry

Use Ctrl + d or /bye or exit to quit.
🦙 > ^D
```

if we fail to pull, we should just exit, not continue to the shell

rhatdan: Your right, I did not think this through.  Why do I need --model if chat already does
ramalama chat MODEL?

docs/ramalama-chat.1.md
rhatdan: I don't think we need this option, since we already take MODEL as an arg?

rhatdan: Or perhaps we should remove users specifying MODEL being required and have chat figure out what the model being served is and just handle it internally (I like this idea) especially when the service is only serving one model. When it is serving more then one model, then take the first if the API does not allow querying of the primary or current model.

ramalama/chat.py
```diff
@@ -64,6 +64,19 @@ def default_prefix():
     return "🦙 > "
 
 
+def add_api_key(args, headers=None):
+    # static analyzers suggest for dict, this is a safer way of setting
+    # a default value, rather than using the parameter directly
+    headers = headers or {}
+    if getattr(args, "api_key", None):
+        if len(args.api_key) < 20:
```
rhatdan: Should 20 be a constant defined somewhere?

ramalama/chat.py
```diff
@@ -121,19 +134,18 @@ def _make_request_data(self):
         data = {
             "stream": True,
             "messages": self.conversation_history,
-            "model": self.args.MODEL,
         }
+
+        model = getattr(self.args, "model", False)
```
rhatdan: Can we just do
```
if getattr(self.args, "model", False):
            data["model"] = self.args.model
```

docs/ramalama-chat.1.md
rhatdan: I see below you are doing this, but doesn't this mean their are additional changes to the man page?

ramalama/chat.py
rhatdan: Is this function going to be used by other uses?  If not then make it part of the class.

test/system/040-serve.bats
ericcurtin: This was the format prior to this PR (talking about chat only for a moment):

ramalama chat MODEL [ARGS]

It doesn't make sense that MODEL is a mandatory arg when trying to list various models via (and because it's positional before [ARGS] it's kinda forced to be mandatory):

ramalama chat --ls

Or when connecting to llama-server which already has a model loaded.

So I'm proposing making this breaking change to "ramalama chat" now, nobody really knows "ramalama chat" exists yet.

"ramalama run" is a little different because it's serving up a model.

ramalama/chat.py
ericcurtin: It's used outside the class in this PR, because "--ls" has no need to launch a RamaLamaShell

docs/ramalama-chat.1.md
ericcurtin: Yes you are correct will make the relevant man page changes

ericcurtin:
I opened an issue for `ramalama run bogus` which is unrelated to this PR, this is about `chat`:

https://github.com/containers/ramalama/issues/1644

`ramalama chat --model bogus` test is kinda irrelevant, because llama-server ignores the MODEL field, it's just going to run against whatever model is loaded no matter what field is passed as the "model". So positive and negative tests have the same outcome.

rhatdan:
LGTM

PR #1641, opened 2025-07-01
build layered images from Containerfiles

mikebonnet:
Move the `Containerfiles` for the entrypoint and rag images out of `container_build.sh` and into their own files. This is necessary so they can be built with Konflux.


rhatdan:
Nice work @mikebonnet

PR #1666, opened 2025-07-07
Inspect add safetensor support

engelmi:
This PR adds support to `inspect` for safetensors models based [the huggingface doc](https://huggingface.co/docs/safetensors/en/index#format). 

Unfortunately, adding system tests for safetensors isn't really possible as I haven't found a small model yet (smallest I've seen is ~1GB). This wouldn't be very practical in the CI.

# Summary by Sourcery

Add safetensor support to the inspect command and refactor model inspection into a common base module, while improving error handling for model pulls and parsing.

New Features:

    Add support for inspecting safetensors models via a new parser and info class

Bug Fixes:

    Clarify GGUF parsing warning message to reference GGUF models explicitly

Enhancements:

    Refactor model inspection logic into a shared base_info module for ModelInfoBase, Tensor, and formatting utilities
    Integrate safetensor parsing into the inspect command workflow
    Wrap snapshot creation calls in try/catch to report and handle failures during model pulls

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -29,6 +29,7 @@ from ramalama.config import CONFIG
 from ramalama.logger import configure_logger, logger
 from ramalama.model import MODEL_TYPES
 from ramalama.model_factory import ModelFactory, New
+from ramalama.model_inspect.error import ParseError
 from ramalama.model_store.global_store import GlobalModelStore
 from ramalama.shortnames import Shortnames
 from ramalama.stack import Stack
@@ -602,7 +603,7 @@ def pull_cli(args):
     if matching_files:
         return matching_files[0]
 
-    return model.pull(args)
+    model.pull(args)
 
 
 def convert_parser(subparsers):
@@ -1217,3 +1218,5 @@ def main():
         eprint(e, errno.EINVAL)
     except IOError as e:
         eprint(e, errno.EIO)
+    except ParseError as e:
+        eprint(f"Failed to parse model: {e}", errno.EINVAL)
--- a/ramalama/model_inspect/gguf_parser.py
+++ b/ramalama/model_inspect/gguf_parser.py
@@ -3,9 +3,9 @@ import struct
 from enum import IntEnum
 from typing import Any, Dict
 
-import ramalama.console as console
 from ramalama.endian import GGUFEndian
-from ramalama.model_inspect import GGUFModelInfo, Tensor
+from ramalama.model_inspect.error import ParseError
+from ramalama.model_inspect.gguf_info import GGUFModelInfo, Tensor
 
 
 # Based on ggml_type in
@@ -99,10 +99,6 @@ GGUF_NUMBER_FORMATS: list[GGUFValueType] = [
 ]
 
 
-class ParseError(Exception):
-    pass
-
-
 class GGUFInfoParser:
     @staticmethod
     def is_model_gguf(model_path: str) -> bool:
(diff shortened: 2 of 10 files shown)


rhatdan:
Please add tests.

engelmi:
> Please add tests.

I had trouble finding a small safetensors model for a new system test, but I could find this one:
https://huggingface.co/LiheYoung/depth-anything-small-hf
~90MB should be fine, I think, for pulling in the CI. Added two new system test cases based on that.

ramalama/url.py
```diff
@@ -105,6 +105,9 @@ def pull(self, args):
                 )
             )
 
-        self.model_store.new_snapshot(tag, snapshot_hash, files)
+        try:
+            self.model_store.new_snapshot(tag, snapshot_hash, files)
+        except Exception as ex:
```
rhatdan: Should we be catching all exceptions or a specific exception?

Is it safe to return a snapshot_file_path if creation of a new_snapshot fails?

engelmi: Should be ok, I think. For HF-like repos we also catch all errors. The cleanup on failure is taken care of inside the `new_snapshot`. 

The return value of `pull` is actually never used. I am removing these (replacing by simple early returns) in https://github.com/containers/ramalama/pull/1643

olliewalsh: Would be better to allow it to bubble up to the top IMO. Should only really catch specific exceptions and only if you can do something to address the exception in the handler e.g retry

engelmi: @olliewalsh Fair point. I moved handling the `ParseError` to the top-level `pull_cli` function. The handling task here is essentially to "beautify" the error. PTAL

ramalama/cli.py
```diff
@@ -602,7 +603,10 @@ def pull_cli(args):
     if matching_files:
         return matching_files[0]
 
-    return model.pull(args)
+    try:
+        model.pull(args)
+    except ParseError as ex:
+        perror(f"Failed to pull '{model.model_name}:{model.model_tag}' : {ex}")
```
rhatdan: We still want error raised, so the command exits with non 0 status.

ramalama/model_inspect/gguf_parser.py
```diff
@@ -99,19 +99,14 @@ class GGUFValueType(IntEnum):
...
     @staticmethod
     def is_model_gguf(model_path: str) -> bool:
         try:
             with open(model_path, "rb") as model_file:
                 magic_number = GGUFInfoParser.read_string(model_file, GGUFEndian.LITTLE, 4)
                 return magic_number == GGUFModelInfo.MAGIC_NUMBER
-        except Exception as ex:
-            console.warning(f"Failed to read model '{model_path}': {ex}")
+        except Exception:
             return False
```
rhatdan, line 112: Can you at least add a logger.debug(f"Caught exception: {ex}"

engelmi: Fair point. Exchanged the `console.warning` with a `logger.debug`. Changed :heavy_check_mark: 

Background: One reason for removing (or lowering the log level) is that applying `is_model_gguf` on a safetensor will result in an exception and since it is part of the usual flow (I can't know what model type we are checking) this shouldn't log a warning.

ramalama/cli.py
engelmi: Ups, you are right. Moved the except block to the cli's main function and use an `EINVAL` as exit code... doesn't really fit, but I couldn't find a better match.

ericcurtin (approved):

rhatdan:
LGTM

Issue #1662, opened 2025-07-04
Notify the user when pulling ramalama container


nyoxi:
### Feature request description

It would be nice if ramalama always printed some message when pulling the ramalama image. This action is far from instantaneous -- image has ~1.5GB -- and the application can appear stuck. Especially in situations where it is not expected/obvious this action can occur, e.g. when doing `ramalama rm ...`. 

### Suggest potential solution

Always write a message like `Pulling ramalama image` when doing fresh podman pull.

rhatdan:
I do not see ramalama pulling images without notifying users in v0.10.0?  
Are you on a MAC? Can  you get a repeater with v0.10.0?

kpouget:
for information, here is what I see with `v0.10.0` on MacOS, when using a custom `IMAGE`
```
% ramalama --image IMAGE run llama3.2
Trying to pull IMAGE...
Getting image source signatures
Copying blob sha256:SHA
Copying blob sha256:SHA
Copying blob sha256:SHA
Copying config sha256:SHA
Writing manifest to image destination
🦭 >
```

nyoxi:
```sh
$ podman rmi quay.io/ramalama/ramalama:0.10   # Remove ramalama image
Untagged: quay.io/ramalama/ramalama:0.10
Deleted: a07c27bac880ff1e42c783a983f5e85a9ad572d9b9770a994ec893098b99963f
$ time ramalama rm hf://microsoft/phi-1_5     # Remove model
ramalama rm hf://microsoft/phi-1_5  27,45s user 3,09s system 64% cpu 47,377 total
$ #    No messages from ramalam and notice the run time   -----------^^^^^^
$ podman image list | grep ramalama           # Check that ramalama image is back
quay.io/ramalama/ramalama                         0.10        7dfb440cffd4  2 hours ago    1.68 GB
$ podman version
Client:       Podman Engine
Version:      5.5.2
API Version:  5.5.2
Go Version:   go1.24.4
Git Commit:   e7d8226745ba07a64b7176a7f128e4ef53225a0e
Built:        Fri Jun 27 09:15:44 2025
OS/Arch:      linux/amd64
$ ramalama version
ramalama version 0.10.0
```

nyoxi:
Well I just checked and I don't have any messages when using `ramalama run` or `ramalama pull` either.

kpouget:
indeed, there's something strange:
```
ramalama --image CUSTOM_IMAGE run llama3.2
[...]

PR #1669, opened 2025-07-07
move --image & --keep-groups to run, serve, perplexity, bench commands

rhatdan:
This eliminates the need for pulling images by accident when not using containers.  Since these commands are only used for container commands, no need for them in other places.

Fixes: https://github.com/containers/ramalama/issues/1662

## Summary by Sourcery

Restrict image and podman keep-groups options to only containerized commands and update related behavior, documentation, and tests.

Bug Fixes:
- Prevent accidental image pulls when not using containers by removing global --image and --keep-groups options.

Enhancements:
- Move --image and --keep-groups arguments into run, serve, perplexity, and bench subcommands instead of the global CLI.
- Change the info command to derive the container image from the default configuration rather than a global argument.

Documentation:
- Update manual pages to document --image and --keep-groups only under relevant container commands.

Tests:
- Adjust the system test to set the image via RAMALAMA_IMAGE and relax the regex to match the new default image source.

### DIFF
--- a/docs/ramalama-bench.1.md
+++ b/docs/ramalama-bench.1.md
@@ -48,6 +48,33 @@ for a value and set the variable only if it is set on the host.
 #### **--help**, **-h**
 show this help message and exit
 
+#### **--image**=IMAGE
+OCI container image to run with specified AI model. RamaLama defaults to using
+images based on the accelerator it discovers. For example:
+`quay.io/ramalama/ramalama`. See the table below for all default images.
+The default image tag is based on the minor version of the RamaLama package.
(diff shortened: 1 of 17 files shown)


rhatdan:
This is a possible breaking change, which is why I kept the hidden --image option on ramalama,  I don't think it is worth having a --keep-groups hidden option, since I don't believe many people use it or know about it.

Since these options are only used for container operations, makes little sense to keep them global.  So moving them specifically to container operations eliminates the side effect of us pulling an image for commands like ramalama rm, pull rm ...

We could also move the --container and --nocontainer options, and perhaps should.

bmahabirbu:
LGTM! Once bat tests are done can merge!

bmahabirbu:
Tested works!

PR #1673, opened 2025-07-08
mlx fixes

ericcurtin:
mlx_lm.server is the only one in my path at least on my system.

Also, printing output like this which doesn't make sense:

Downloading huggingface://RedHatAI/Llama-3.2-1B-Instruct-FP8-dynamic/model.safetensors:latest ... Trying to pull huggingface://RedHatAI/Llama-3.2-1B-Instruct-FP8-dynamic/model.safetensors:latest ...

## Summary by Sourcery

Normalize model URIs and streamline subcommand invocation.

Bug Fixes:
- Prevent redundant huggingface:// prefix and :latest suffix from appearing in output logs

Enhancements:
- Introduce trim_model_name to abbreviate model URIs by replacing full prefixes and stripping default tags
- Use trim_model_name in CLI listings and download messages for consistent model naming
- Switch subcommand execution from python -m mlx_lm to direct mlx_lm.server invocation

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,11 +116,6 @@ RamaLama is available via PyPi at [https://pypi.org/project/ramalama](https://py
 pip install ramalama
 ```
 
-### Install via Homebrew
-```
-brew install ramalama
-```
-
 ### Install script (Linux and macOS)
 Install RamaLama by running:
 ```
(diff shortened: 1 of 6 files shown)


ericcurtin:
@kush-gupt PTAL

kush-gupt:
LGTM, the reason why I wanted to leave python in cmd was the explicit declaration that we're using that python mlx-lm for it, but it should be fine on path

rhatdan:
LGTM

PR #1677, opened 2025-07-08
Add vllm to cpu inferencing Containerfile

ericcurtin:
To be built upon "ramalama" image

### DIFF
--- a/container-images/ramalama-vllm/Containerfile
+++ b/container-images/ramalama-vllm/Containerfile
@@ -1,0 +1,17 @@
+FROM quay.io/ramalama/ramalama
+
+ENV PATH="/root/.local/bin:$PATH"
+ENV VIRTUAL_ENV="/opt/venv"
+ENV UV_PYTHON_INSTALL_DIR="/opt/uv/python"
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+
+ENV UV_HTTP_TIMEOUT=500
+
+ENV UV_INDEX_STRATEGY="unsafe-best-match"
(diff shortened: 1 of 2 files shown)


rhatdan:
@mikebonnet PTAL
LGTM

rhatdan:
@ericcurtin built for all arches correct?
Named vllm-cpu? Or just vllm?
Then we could add vllm-cuda, vllm-rocm, vllm-intel?

ericcurtin:
> @ericcurtin built for all arches correct? Named vllm-cpu? Or just vllm? Then we could add vllm-cuda, vllm-rocm, vllm-intel?

I only tested x86_64 and aarch64.

But there's a lot of active enablement work going on for ppc also, I got pinged to take a look at a ppc PR this morning in upstream vllm (the joys of open-source, you open one PR to enable aarch64 CPU inferencing and people then poke you to review things tangentially related).

I also know there had been some s390 enablement work also.

Might be worth attempting to build for all 4 arches. If ppc or s390 fail for any reason, maybe abandon them until they are more mature.

I suspect vllm-cuda and vllm-rocm will try and duplicate all the cuda and rocm libraries with other specific versions to vllm.

Apparently Intel works, they seem to call it xpu?

No vulkan support which is a pity.

ericcurtin:
If vulkan was added to vLLM... vLLM would almost have as much accelerated hardware support as llama.cpp, would only be missing native macOS, which could be worked around with @slp 's krunkit solution.

Ignoring things like cann and musa, as that hardware isn't crazy common.

PR #1679, opened 2025-07-09
Bugfix for chat

ericcurtin:
This was recently removed:

+            if getattr(self.args, "model", False):
+                data["model"] = self.args.model

it is required

## Summary by Sourcery

Restore conditional addition of the `model` field in chat request payloads based on the presence of the `--model` argument.

Bug Fixes:
- Reintroduce the model field in request data only when `self.args.model` is set instead of always using `self.args.MODEL`.

Enhancements:
- Switch to using `getattr(self.args, "model", False)` for safe attribute access.

### DIFF
--- a/ramalama/chat.py
+++ b/ramalama/chat.py
@@ -132,8 +132,9 @@ class RamaLamaShell(cmd.Cmd):
             "stream": True,
             "messages": self.conversation_history,
         }
-        if not (hasattr(self.args, 'runtime') and self.args.runtime == "mlx"):
-            data["model"] = self.args.MODEL
+        if getattr(self.args, "model", False):
+            data["model"] = self.args.model
+
         json_data = json.dumps(data).encode("utf-8")
         headers = {
             "Content-Type": "application/json",


ericcurtin:
@kush-gupt PTAL

kush-gupt:
This fixes the model error indeed!

For chatting with mlx, a wrong model name won't show any errors on the chat side. It just repeats the api request until infinity, and the only indication is the mlx server process logs

`127.0.0.1 - - [09/Jul/2025 12:06:27] "POST /v1/chat/completions HTTP/1.1" 404 -`

kush-gupt:
An agnostic model name that works is "default_model", as nothing else worked with my continue integration

ericcurtin:
On second thought I think we should just remove this bit and go back to the original code prior to mlx:

`not (hasattr(self.args, 'runtime') and self.args.runtime == "mlx")`

if the user passes some bogus model, that's user error, we shouldn't mask that

rhatdan:
LGTM

PR #1689, opened 2025-07-13
Only install if pyproject.toml exists

ericcurtin:
Otherwise skip

## Summary by Sourcery

Add a guard to skip the pip installation if the project definition file is missing

Enhancements:
- Wrap the pip install command in install_ramalama with a check for pyproject.toml
- Skip installation when pyproject.toml does not exist


rhatdan:
LGTM

PR #1704, opened 2025-07-16
Minor fixes to rpm builds by packit and spec file.

smooge:
* This removes epel9 from packit rules as epel9 does not currently build without many additional packages added to the distro.
* This fixes a breakage in epel10 by adding mailcap as a buildrequires.


rhatdan:
LGTM

bmahabirbu:
LGTM too!

PR #1706, opened 2025-07-16
Consolodate run and chat commands together also allow specification of prefix in ramalama.conf

rhatdan:
## Summary by Sourcery

Allow users to define a custom prompt prefix for chat and run commands, centralize shared CLI options into a helper function, and enhance BaseConfig with additional settings while updating documentation to cover the new prefix configuration.

New Features:
- Add --prefix flag to chat and run commands for custom prompt prefixes
- Introduce persistent prefix configuration in ramalama.conf via new prefix setting

Enhancements:
- Extract common flags (--color, --prefix, --rag) into shared chat_run_options for chat and run commands
- Extend and reorganize BaseConfig with new fields (api, default_image, dryrun, ocr, selinux)

Documentation:
- Document the prefix configuration in ramalama.conf and the man page with engine-based defaults

### DIFF
--- a/ramalama/chat.py
+++ b/ramalama/chat.py
@@ -52,6 +52,9 @@ def default_prefix():
     if not EMOJI:
         return "> "
 
+    if CONFIG.prefix:
+        return CONFIG.prefix
+
     engine = CONFIG.engine
 
     if engine:
--- a/docs/ramalama.conf
+++ b/docs/ramalama.conf
@@ -77,6 +77,16 @@
 #
 #port = "8080"
 
+# Specify default prefix for chat and run command. By default the prefix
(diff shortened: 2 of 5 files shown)


ramalama/chat.py
```diff
@@ -52,6 +52,9 @@ def default_prefix():
     if not EMOJI:
         return "> "
 
+    if CONFIG.prefix:
```
olliewalsh, line 55: should this override the prompt when emojis are not supported? E.g if running native windows python via cmd.exe/powershell (if that even works?)

rhatdan: Lets look at that as a followon.

PR #1712, opened 2025-07-17
Add support for Intel Iris Xe Graphics (46AA, 46A6, 46A8)

tonyjames:
This commit adds the Intel Iris Xe Graphics (46AA, 46A6, 46A8) chips to the `check_intel` function so that the `ramalama/intel-gpu` image is used. Without this the standard `ramalama/ramalama` image is used with very poor performance.


rhatdan:
LGTM

PR #1728, opened 2025-07-22
No virglrenderer in RHEL

ericcurtin:
In this build script we basically assume if it's a dnf based OS that is not Fedora. It's some RHEL-based OS. virglrenderer is required for a microVM feature in RamaLama. krun and virglrenderer are not in RHEL.

## Summary by Sourcery

Bug Fixes:
- Remove virglrenderer from the dnf install command on non-Fedora dnf-based OS


rhatdan:
LGTM
The AI Suggestion makes some sense.  Are these packages available via EPEL?

ericcurtin:
openEuler would only work if the host was Fedora, non-EPEL Fedora is the only OS that has crun-krun. virgilrender is not in EPEL. I'll add for openEuler nonetheless :)

rhatdan:
LGTM

PR #1736, opened 2025-07-23
Adds docs site

ieaves:
## Summary by Sourcery

Add a Docusaurus-powered documentation site, including manpage conversion, site configuration, and content generation.

New Features:
- Introduce a documentation site with Docusaurus and MDX pages auto-generated from manpages
- Add convert_manpages.py script and Makefile targets to convert and build documentation

Enhancements:
- Refine and merge note sections in the main ramalama manpage

Build:
- Add npm-based site build configuration (package.json, docusaurus.config.ts, tsconfig.json)
- Introduce Makefile commands for converting manpages, building, and serving the documentation site

Documentation:
- Bulk add auto-generated MDX documentation pages for all commands, configuration, and platform guides
- Add homepage layout, components, and custom CSS for the documentation site


docsite/convert_manpages.py
```diff
@@ -0,0 +1,327 @@
...
+            description = "RamaLama documentation"
+    
+    return title.strip(), description.strip()
+
+def get_sidebar_position(filename):
+    """Get sidebar position for files"""
+    base_name = os.path.basename(filename)
+    
+    # Define command priority order
+    command_order = {
```
rhatdan: Should we generate these automatically?
ramalama*1.md

docsite/convert_manpages.py
```diff
@@ -0,0 +1,327 @@
...
+        'ramalama-login.1.md': 15,
+        'ramalama-logout.1.md': 16,
+        'ramalama-containers.1.md': 17,
+        'ramalama-stop.1.md': 18,
+        'ramalama-version.1.md': 19,
+    }
+    
+    # Platform guides order
+    platform_order = {
+        'ramalama-cuda.7.md': 1,
```
rhatdan: Ditto

docsite/convert_manpages.py
```diff
@@ -0,0 +1,327 @@
...
+        'ramalama-macos.7.md': 2,
+        'ramalama-cann.7.md': 3,
+        'ramalama-musa.7.md': 4,
+    }
+    
+    # Configuration order
+    config_order = {
+        'ramalama.conf.5.md': 1,
```
rhatdan: Ditto

docsite/convert_manpages.py
```diff
@@ -0,0 +1,327 @@
...
+        'ramalama-oci.5.md': 2,
+    }
+    
+    if base_name in command_order:
+        return command_order[base_name]
+    elif base_name in platform_order:
```
rhatdan: Nit, elif should be just if.

docsite/convert_manpages.py
```diff
@@ -0,0 +1,327 @@
...
+    return frontmatter + content
+
+def get_output_filename(input_filename):
+    """Generate output filename from input filename"""
+    base = os.path.basename(input_filename)
+    
+    if base == 'ramalama.1.md':
+        # Base ramalama command goes in commands directory
+        return 'ramalama.mdx'
+    elif base.startswith('ramalama-') and base.endswith('.1.md'):
```
rhatdan: Nit: No need for elif, since the if does a return.  I think it is cleaner to just dow
```
if AAA:
   return
if BBB:
   return
if CCC:
   return
return 
```

docsite/docs/commands/ramalama/bench.mdx
```diff
@@ -0,0 +1,163 @@
+---
```
rhatdan, line 1: Are these generated files, should they be in the PR and stored in Github?

rhatdan:
Please squash your commits.

docsite/docs/commands/ramalama/bench.mdx
ieaves: I thought reusing the manpages would be the best place to start since a fair amount of effort had already gone into making them and there's testing to ensure correctness. Docusaurus requires mdx files so it's not a simple task of copying the existing markdown over to the docsite (breaking links, different formatting, some mdx incompatible strings, etc...).

Rather than managing two sources of truth with the likely future of divergence between them I figured we would keep the manpages as the single source of truth and build a script to automatically convert them to site pages. I'd prefer not to store both but I don't see another way. Have you seen a better approach on other projects? Definitely worth taking a look if so.

rhatdan: No this is fine, as long as make docs rebuilds these files.

rhatdan:
LGTM

PR #1764, opened 2025-07-28
Typing and bug squashes

ieaves:
## Summary by Sourcery

Introduce comprehensive static typing to core modules, convert ModelBase to an abstract ABC, tighten ModelFactory initialization, and resolve GGUF parsing bugs with accompanying tests

Bug Fixes:
- Raise ParseError for unknown GGUF value types and fix tensor type handling in GGUF parsing and serialization

Enhancements:
- Add type annotations across modules using TypeAlias, TypedDict, and explicit casts
- Convert ModelBase to an abstract ABC with @abstractmethod declarations
- Enforce non-null engine argument in ModelFactory and update function signatures

Tests:
- Add unit test for GGUFModelInfo serialization when tensor.type is a string

### DIFF
--- a/ramalama/common.py
+++ b/ramalama/common.py
@@ -13,8 +13,9 @@ import shutil
 import string
 import subprocess
 import sys
+from collections.abc import Callable, Iterable
 from functools import lru_cache
-from typing import TYPE_CHECKING, Callable, List, Literal, Protocol, cast, get_args
+from typing import TYPE_CHECKING, Literal, Protocol, TypeAlias, TypedDict, cast, get_args
 
 import ramalama.amdkfd as amdkfd
 from ramalama.logger import logger
@@ -230,7 +231,15 @@ def engine_version(engine: SUPPORTED_ENGINES) -> str:
     return run_cmd(cmd_args).stdout.decode("utf-8").strip()
 
 
-def load_cdi_yaml(stream) -> dict:
+class CDI_DEVICE(TypedDict):
+    name: str
+
+
+class CDI_RETURN_TYPE(TypedDict):
+    devices: list[CDI_DEVICE]
+
+
+def load_cdi_yaml(stream: Iterable[str]) -> CDI_RETURN_TYPE:
     # Returns a dict containing just the "devices" key, whose value is
     # a list of dicts, each mapping the key "name" to a device name.
     # For example: {'devices': [{'name': 'all'}]}
@@ -238,7 +247,7 @@ def load_cdi_yaml(stream) -> dict:
     # under "devices" and the value of the "name" key being on the
     # same line following a colon.
 
-    data = {"devices": []}
+    data: CDI_RETURN_TYPE = {"devices": []}
     for line in stream:
         if ':' in line:
             key, value = line.split(':', 1)
@@ -247,7 +256,7 @@ def load_cdi_yaml(stream) -> dict:
     return data
 
 
-def load_cdi_config(spec_dirs: List[str]) -> dict | None:
+def load_cdi_config(spec_dirs: list[str]) -> CDI_RETURN_TYPE | None:
     # Loads the first YAML or JSON CDI configuration file found in the
     # given directories."""
 
@@ -275,7 +284,7 @@ def load_cdi_config(spec_dirs: List[str]) -> dict | None:
     return None
 
 
-def find_in_cdi(devices: List[str]) -> tuple[List[str], List[str]]:
+def find_in_cdi(devices: list[str]) -> tuple[list[str], list[str]]:
     # Attempts to find a CDI configuration for each device in devices
     # and returns a list of configured devices and a list of
     # unconfigured devices.
@@ -327,11 +336,12 @@ def check_nvidia() -> Literal["cuda"] | None:
         return None
 
     smi_lines = result.stdout.splitlines()
-    parsed_lines = [[item.strip() for item in line.split(',')] for line in smi_lines if line]
+    parsed_lines: list[list[str]] = [[item.strip() for item in line.split(',')] for line in smi_lines if line]
+
     if not parsed_lines:
         return None
 
-    indices, uuids = zip(*parsed_lines) if parsed_lines else (tuple(), tuple())
+    indices, uuids = map(list, zip(*parsed_lines))
     # Get the list of devices specified by CUDA_VISIBLE_DEVICES, if any
     cuda_visible_devices = os.environ.get("CUDA_VISIBLE_DEVICES", "")
     visible_devices = cuda_visible_devices.split(',') if cuda_visible_devices else []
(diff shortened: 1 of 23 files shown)


ramalama/model_store/go2jinja.py
```diff
@@ -553,13 +550,13 @@
...
                 return (
                     node.content[: m.start(1)].replace(GO_SYMBOL_OPEN_BRACKETS, JINJA_SYMBOL_OPEN_BRACKETS)
                     + "endif"
                     + node.content[m.end(1) :].replace(GO_SYMBOL_CLOSE_BRACKETS, JINJA_SYMBOL_CLOSE_BRACKETS)
                 )
-            elif node.open_scope_node.type == NodeType.RANGE:
+            elif node.open_scope_node.type == NodeType.RANGE:  # type: ignore
                 loop_vars.pop()
                 if loop_index_vars:
                     loop_index_vars.pop()
```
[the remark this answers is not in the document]
ieaves: While it's true this is a legitimate error based on the type construction of the node class, I couldn't construct an example of this error occurring in practice (i.e. from a template string).

ramalama/model_store/go2jinja.py
```diff
@@ -376,6 +372,7 @@ def add_if_continue_check_block(parent_node: Node, start_index: int) -> None:
                     children=[],
                     artificial=True,
                 )
+                for_node.next = cast(Node, for_node.next)
```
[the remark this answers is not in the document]
ieaves: No, this block of code is triggered for a template like {{continue}}{{range .}}. The for_node is the parent making it's .next the range by construction from the parsing logic above.

While this error can be induced by artificially constructing a node, it is not triggered from a legitimate template string.

ramalama/common.py
```diff
@@ -327,11 +335,12 @@
...
 
     smi_lines = result.stdout.splitlines()
-    parsed_lines = [[item.strip() for item in line.split(',')] for line in smi_lines if line]
+    parsed_lines: list[list[str]] = [[item.strip() for item in line.split(',')] for line in smi_lines if line]
+
     if not parsed_lines:
         return None
 
-    indices, uuids = zip(*parsed_lines) if parsed_lines else (tuple(), tuple())
+    indices, uuids = cast(list[list[str]], zip(*parsed_lines))
```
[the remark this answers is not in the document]
ieaves: The issue isn't with the type checker failing to detect the type from zip. Without the cast, the type checker infer indices and uuids are tuples. Further down, `configured` will be identified as a list (the return type of `find_in_cdi`). Then on line 364 when configured is assigned to indices we generate an assignment issue where a list is assigned as a tuple. I'll switch the cast to `tuple[Sequence[str], Sequence[str]]` so that it's technically correct.

ramalama/model_inspect/gguf_info.py
```diff
@@ -53,7 +53,7 @@ def serialize(self, json: bool = False, all: bool = False) -> str:
             i = 0
             for tensor in self.Tensors:
                 ret = ret + adjust_new_line(
-                    f"      {i}: {tensor.name, tensor.type.name, tensor.n_dimensions, tensor.offset}"
+                    f"      {i}: {tensor.name, tensor.type, tensor.n_dimensions, tensor.offset}"
```
ieaves, line 56: I wanted to highlight this line @rhatdan. It looks like `tensor.type` was assumed to be an enum member in this line of code and a string elsewhere. I can see on line 220 of gguf_parser that the tensor was being passed an enum member as well but because dataclasses don't perform validation by default the discrepancy wasn't caught. 

In any case, since I'm not as familiar with the gguf parser I wanted to make sure it was the right change to make.

rhatdan: No idea.
@ericcurtin @engelmi ?

engelmi: Yes, the typing for the `type` member on the Tensor dataclass should remain a `str`. So your changes make sense to me - passing the `tensor_type.name` to the Tensor constructor. Good catch!

test/unit/test_gguf_info.py
```diff
@@ -0,0 +1,28 @@
+from ramalama.endian import GGUFEndian
+from ramalama.model_inspect.base_info import Tensor
+from ramalama.model_inspect.gguf_info import GGUFModelInfo
+
+
+def test_serialize_with_all_true_tensor_type_error():
```
engelmi: Is this test missing an assertion?

ramalama/model_store/store.py
```diff
@@ -97,9 +97,12 @@ def get_ref_file(self, model_tag: str) -> Optional[RefJSONFile]:
         return RefJSONFile.from_path(ref_file_path)
 
     def update_ref_file(
-        self, model_tag: str, snapshot_hash: str = "", snapshot_files: list[SnapshotFile] = []
+        self, model_tag: str, snapshot_hash: str = "", snapshot_files: list[SnapshotFile] | None = None
```
engelmi: I'd prefer using `Optional` in cases where `None` as a default is provided:
```suggestion
        self, model_tag: str, snapshot_hash: str = "", snapshot_files: Optional[list[SnapshotFile]] = None
```

ramalama/model_store/store.py
```diff
@@ -182,8 +185,8 @@ def _prepare_new_snapshot(self, model_tag: str, snapshot_hash: str, snapshot_fil
         snapshot_directory = self.get_snapshot_directory(snapshot_hash)
         os.makedirs(snapshot_directory, exist_ok=True)
 
-    def _download_snapshot_files(self, model_tag: str, snapshot_hash: str, snapshot_files: list[SnapshotFile]):
-        ref_file = self.get_ref_file(model_tag)
+    def _download_snapshot_files(self, model_tag: str, snapshot_hash: str, snapshot_files: Sequence[SnapshotFile]):
+        ref_file: None | RefJSONFile = self.get_ref_file(model_tag)
```
engelmi, line 189: At this point, `ref_file` is not allowed to be `None`. If it is, then its probably best to raise an Exception.

ramalama/model_factory.py
```diff
@@ -119,7 +127,7 @@ def create_url(self) -> URL:
         return model
 
 
-def New(name, args, transport: str = None) -> Union[Huggingface | ModelScope | Ollama | OCI | URL]:
+def New(name, args, transport: str | None = None) -> Union[Huggingface | ModelScope | Ollama | OCI | URL]:
```
engelmi: The return type could be replaced here, right?
```suggestion
def New(name, args, transport: str | None = None) -> CLASS_MODEL_TYPES:
```

test/unit/test_gguf_info.py
ieaves: We can actually remove that test. I was just iterating through the type errors mypy found and constructing test cases to validate whether they were real errors or not. I'll pull this one out!

ramalama/model_store/store.py
ieaves: Should the error be in `get_ref_file` in that case? It has a valid None return path where, for example, the ref file path doesn't exist.

engelmi: Yes, the `None` is valid for `get_ref_file` since it could be that the file is missing and we don't want to throw an error for that case. I think it would be best to raise an Exception here in `_download_snapshot_files` if the ref_file is None since the download function is part of a routine where we rely on the ref_file being created - if its missing then, we don't have a way to proceed.

rhatdan:
Please squash your commits.

rhatdan (approved):
LGTM

Issue #1766, opened 2025-07-29
Improve error message when huggingface-cli is not available


pbabinca:
### Feature request description

This is a follow-up on #643.

When user tries to log in using [current example in the documentation](https://github.com/containers/ramalama/blob/main/docs/ramalama-login.1.md#examples):

```bash
export RAMALAMA_TRANSPORT=huggingface
ramalama login --token=XYZ
```

while they don't have the `huggingface-cli` tool they get message:

```
Error:
Optional: Huggingface models require the huggingface-cli module.
This module can be installed via PyPI tools like uv, pip, pip3, pipx, or via
distribution package managers like dnf or apt. Example:
uv pip install huggingface_hub
```

This might not be clear to the user.

### Suggest potential solution

* "Error:" on an empty line might indicate that it introduces an error for the rest of the text. That's not the case - error is only the next line.
* "Optional: " prefix might indicate that this is not an error message. Which contradicts from the previous line. The tool is not optional if user wants to use huggingface ramalama transport.
* Currently README uses `pip` as one way to install the package, distribution package managers (and an installation script). Currently missing tool guides user to use `uv` which they might not have available (yet). Use `pip` and `dnf` instead as examples.

PR #1767, opened 2025-07-29
Improve error message when huggingface-cli is needed and missing

pbabinca:
Before:

    Error:
    Optional: Huggingface models require the huggingface-cli module.
    This module can be installed via PyPI tools like uv, pip, pip3, pipx, or via
    distribution package managers like dnf or apt. Example:
    uv pip install huggingface_hub

After:

    Error: Huggingface models require the huggingface-cli tool.

    This tool can be installed via PyPI tools like uv, pip, pip3, pipx. Example:

    pip install -U "huggingface_hub[cli]"

    Or via distribution package managers like dnf or apt. Example:

    sudo install python3-huggingface-hub

In particular:

    - separate error from the explanation and guidance by new line,

    - provide `pip` as first example how to solve the issue (that's also the tool mention in the README)

    - use same package extra as official documentation to install the CLI tool: https://huggingface.co/docs/huggingface_hub/en/guides/cli

    - provide example with dnf command to install the package (dnf is also mentioned in the README file)

    - `huggingface-cli` is a command line tool, not a module as it was referred before

Fixes: #1766

## Summary by Sourcery

Enhance the missing huggingface-cli error message to provide clearer guidance and up-to-date install instructions

Enhancements:
- Separate the error header from the explanatory guidance with a blank line
- Prioritize pip install example using `pip install -U "huggingface_hub[cli]"` to match official documentation
- Include distribution package install command (`sudo install python3-huggingface-hub`) alongside pip instructions
- Correct terminology from module to tool in the error message

### DIFF
--- a/ramalama/huggingface.py
+++ b/ramalama/huggingface.py
@@ -13,11 +13,15 @@ from ramalama.hf_style_repo_base import (
 from ramalama.logger import logger
 from ramalama.model_store.snapshot_file import SnapshotFileType
 
-missing_huggingface = """
-Optional: Huggingface models require the huggingface-cli module.
-This module can be installed via PyPI tools like uv, pip, pip3, pipx, or via
-distribution package managers like dnf or apt. Example:
-uv pip install huggingface_hub
+missing_huggingface = """This operation requires huggingface-cli which is not available.
+
+This tool can be installed via PyPI tools like uv, pip, pip3 or pipx. Example:
+
+pip install -U "huggingface_hub[cli]"
+
+Or via distribution package managers like dnf or apt. Example:
+
+sudo dnf install python3-huggingface-hub
 """
 
 
--- a/ramalama/modelscope.py
+++ b/ramalama/modelscope.py
@@ -10,11 +10,11 @@ from ramalama.hf_style_repo_base import (
 )
 from ramalama.model_store.snapshot_file import SnapshotFileType
 
-missing_modelscope = """
-Optional: ModelScope models require the modelscope module.
-This module can be installed via PyPI tools like uv, pip, pip3, pipx, or via
-distribution package managers like dnf or apt. Example:
-uv pip install modelscope
+missing_modelscope = """This operation requires modelscope which is not available.
+
+This tool can be installed via PyPI tools like uv, pip, pip3 or pipx. Example:
+
+pip install modelscope
 """
 
 


ramalama/huggingface.py
```diff
@@ -13,11 +13,15 @@
 from ramalama.logger import logger
 from ramalama.model_store.snapshot_file import SnapshotFileType
 
-missing_huggingface = """
-Optional: Huggingface models require the huggingface-cli module.
-This module can be installed via PyPI tools like uv, pip, pip3, pipx, or via
-distribution package managers like dnf or apt. Example:
-uv pip install huggingface_hub
+missing_huggingface = """Huggingface models require the huggingface-cli tool.
```
engelmi: The hf cli is optional, so I'd keep the "Optional: " in the beginning. 
There are even plans on removing the support of the hf cli completely and only use http calls / the REST API (afaik).

ramalama/huggingface.py
```diff
@@ -13,11 +13,15 @@
...
+
+This tool can be installed via PyPI tools like uv, pip, pip3, pipx. Example:
+
+pip install -U "huggingface_hub[cli]"
+
+Or via distribution package managers like dnf or apt. Example:
+
+sudo install python3-huggingface-hub
```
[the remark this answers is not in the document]
rhatdan: Please use suggestion.

ramalama/huggingface.py
pbabinca: > The hf cli is optional, so I'd keep the "Optional: " in the beginning.

If user selects ramalama transport with "export RAMALAMA_TRANSPORT=huggingface" and the tool isn't able to continue and exits because of missing hf CLI that does not seem optional to me.

>  There are even plans on removing the support of the hf cli completely and only use http calls / the REST API (afaik).

And when that's done this message could be updated or even removed. Until then I believe this is not optional.

ramalama/huggingface.py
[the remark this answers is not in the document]
pbabinca: Oh, I have noticed your comment only now. I fixed that already with a amended & force pushed change. I hope the project workflow is compatible with this approach.

ramalama/huggingface.py
olliewalsh: what is the exact command line you are using? Right now huggingface cli is only really required to get a login token. However if pulling the image fails then it will retry pulling the image using huggingface cli if available (which is likely to fail too).

pbabinca: > what is the exact command line you are using?

```
export RAMALAMA_TRANSPORT=huggingface
ramalama login --token=XYZ
```

This implements #1766 which has that command and also where I got that command from - [the examples section of docs](https://github.com/containers/ramalama/blob/main/docs/ramalama-login.1.md#examples).

And, I understand why you are asking now. I forgot to put #1766 in the original commit when opening PR and even though I updated it since then I didn't update the description of PR. I did so now. I'm sorry for the confusion.

ramalama/huggingface.py
```diff
@@ -13,11 +13,15 @@
...
-uv pip install huggingface_hub
+missing_huggingface = """Huggingface models require the huggingface-cli tool.
+
+This tool can be installed via PyPI tools like uv, pip, pip3, pipx. Example:
+
+pip install -U "huggingface_hub[cli]"
+
+Or via distribution package managers like dnf or apt. Example:
+
+sudo dnf install python3-huggingface-hub
```
olliewalsh, line 24: There is no deb/rpm AFAIK https://huggingface.co/docs/huggingface_hub/en/installation

ramalama/huggingface.py
olliewalsh: ok, yes as I mentioned it's currently required for login - hoping to remove that in the near future too

olliewalsh: message in modelscope.py should be updated to be consistent

rhatdan:
@pbabinca any update on this, would like to get it merged.

ramalama/huggingface.py
pbabinca: You are right that neither deb nor rpm are coming from official installation docs. On the other hand:

- Fedora ships a rpm [python3-huggingface-hub](https://packages.fedoraproject.org/pkgs/python-huggingface-hub/python3-huggingface-hub/index.html) in all its supported release (currently 41, 42).

- there is a deb [huggingface-hub](https://tracker.debian.org/pkg/huggingface-hub] shipped in Debian Unstable (Sid).

so I believe hint where to install the package from is still valid.

ramalama/huggingface.py
pbabinca: Do I read it correctly that you are fine with current wording?

pbabinca: > message in modelscope.py should be updated to be consistent

I'm not sure what do you mean by this. I haven't even tried to use ModelScope. This focuses only on Huggingface.

olliewalsh: > > message in modelscope.py should be updated to be consistent
> 
> I'm not sure what do you mean by this. I haven't even tried to use ModelScope. This focuses only on Huggingface.

ModelScope and HuggingFace support are essentially the same. So should update the missing modelscope cli message too i.e https://github.com/containers/ramalama/blob/47babb31c376234499a228b9fcc86bb8c3b9a90f/ramalama/modelscope.py#L13

olliewalsh: No, I agree with @engelmi - should keep "Optional: "

pbabinca: In https://github.com/containers/ramalama/pull/1767#discussion_r2240000289 I have explained why I believe this cannot be considered optional. Since that does not seem to be enough I'll reference the code:

Firstly module `hf_style_repo_base`:

- [`login()` method](https://github.com/containers/ramalama/blob/main/ramalama/hf_style_repo_base.py#L203-L204) first checks for existence of the CLI command and if it does not exist it basically propagates the error message through exception `NotImplementedError`.
- [`logout()` method](https://github.com/containers/ramalama/blob/main/ramalama/hf_style_repo_base.py#L211-L212) starts basically the same way.
- [`exec()` method](https://github.com/containers/ramalama/blob/main/ramalama/hf_style_repo_base.py#L249-L253) looks like safety check on any exec of the CLI tool and prints the same error. This time without propagating the exception so one could consider this optional. From what I can tell `exec()` is called only in `login()` and `logout()` though.

Then, there is a module `huggingface` which has class `Huggingface` based on `HFStyleRepoModel` from module `hf_style_repo_base`:

- [`push()` method](https://github.com/containers/ramalama/blob/main/ramalama/huggingface.py#L225-L226) begins with check for CLI tool and again uses `NotImplementedError` exception to propagate the message if the tool does not exist.

Finally the module `modelscope` has [very similar logic for `push()` method](https://github.com/containers/ramalama/blob/main/ramalama/modelscope.py#L117-L118).

---

I haven't found any other way to display the error message that I'm proposing to change.

---

When I look at the check for existence of CLI tool from the other way around - I can see that there is a method `in_existing_cache()` which checks for existence of the CLI tool and then returns False without any error or warning. Both [modelscope](https://github.com/containers/ramalama/blob/main/ramalama/modelscope.py#L97-L98) and [huggingface](https://github.com/containers/ramalama/blob/main/ramalama/huggingface.py#L197-L198) modules have this check.

I can tell the CLI is optional here. But because the message is not displayed I don't think this check is relevant to this PR.

---

@olliewalsh or @engelmi tell me in which workflow you would get a message (likely when some code calls `get_missing_message()`) where lack of "Optional: " prefix would be misleading and the CLI tool is indeed optional.

pbabinca: I see. Let's agree on the wording first (https://github.com/containers/ramalama/pull/1767#discussion_r2273729377 ) and then I'll update the modelscope.

olliewalsh: How about changing "Optional: Huggingface models require the huggingface-cli tool." to "Huggingface login requires the huggingface-cli tool."

AFAIK push is only supported for OCI models right now so can ignore that method.

pbabinca: > How about changing "Optional: Huggingface models require the huggingface-cli tool." to "Huggingface login requires the huggingface-cli tool."

Sounds better, but this message will be displayed also when user tries to logout which is misleading.

> AFAIK push is only supported for OCI models right now so can ignore that method.

Well, from what I read in the code one could call `push()` method from the `huggingface` module. And the message: "Huggingface login requires the huggingface-cli tool." would be misleading here as well.

pbabinca: What about message:

> Error: This operation requires huggingface-cli tool which is not available.

Followed by new line and the rest of the proposed message in the PR.

rhatdan:
LGTM

PR #1769, opened 2025-07-29
building stable diffusion.

jtligon:
I made a copy of build whisper/llama script and made it build leejet/sd
Not exactly sure about ALL of the boilerplate, but it built the binary.
also made the pip stuff in make work on my machine, but I should probably lose that. looking forward to feedback instead of just merging here.

assisted by cursor/claude.

## Summary by Sourcery

Add support for building and packaging stable diffusion in the container infrastructure and switch Makefile installs to use pipx.

New Features:
- Introduce a build_stable_diffusion.sh script to clone and compile leejet/stable-diffusion.cpp into an sd binary
- Add a Containerfile and entrypoint for a stable-diffusion image that runs sd by default

Enhancements:
- Include 'stable-diffusion' as a target in container_build.sh
- Refactor Makefile install commands to use MYPIP (pipx) instead of pip

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,7 @@ SHAREDIR ?= ${PREFIX}/share
 PYTHON ?= $(shell command -v python3 python|head -n1)
 DESTDIR ?= /
 PATH := $(PATH):$(HOME)/.local/bin
+MYPIP ?= pip
 IMAGE ?= ramalama
 PROJECT_DIR:=$(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
 PYTHON_SCRIPTS := $(shell grep -lEr "^\#\!\s*/usr/bin/(env +)?python(3)?(\s|$$)" --exclude-dir={.venv,venv} $(PROJECT_DIR) || true)
@@ -43,16 +44,16 @@ help:
 	@echo
 
 install-detailed-cov-requirements:
-	pip install ".[cov-detailed]"
+	${MYPIP} install ".[cov-detailed]"
 
 .PHONY: install-cov-requirements
 install-cov-requirements:
-	pip install ".[cov]"
+	${MYPIP} install ".[cov]"
 
 .PHONY: install-requirements
 install-requirements:
 	./install-uv.sh
-	pip install ".[dev]"
+	${MYPIP} install ".[dev]"
 
 .PHONY: install-completions
 install-completions: completions
@@ -86,7 +87,7 @@ completions:
 .PHONY: install
 install: docs completions
 	RAMALAMA_VERSION=$(RAMALAMA_VERSION) \
-	pip install . --no-deps --root $(DESTDIR) --prefix ${PREFIX}
+	${MYPIP} install . --no-deps --root $(DESTDIR) --prefix ${PREFIX}
 
 .PHONY: build
 build:
--- a/container-images/scripts/build_stable_diffusion.sh
+++ b/container-images/scripts/build_stable_diffusion.sh
@@ -1,0 +1,180 @@
+#!/bin/bash
+
+python_version() {
+  if python3 -c 'import sys; exit(not (sys.version_info.major == 3 and sys.version_info.minor == 11))'; then
+    echo "python3.11"
+  else
+    echo "python3"
+  fi
+}
+
+available() {
+  command -v "$1" >/dev/null
+}
+
+dnf_remove() {
+  dnf -y clean all
+}
+
+dnf_install_epel() {
+  local rpm_exclude_list="selinux-policy,container-selinux"
+  local url="https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm"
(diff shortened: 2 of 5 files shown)


container-images/scripts/build_stable_diffusion.sh
```diff
@@ -0,0 +1,186 @@
...
+  if [ -n "$pyversion" ]; then
+    string="$pyversion
+Python 3.10"
+    if [ "$string" == "$(sort --version-sort <<<"$string")" ]; then
+      echo "python3.11"
+      return
+    fi
+  fi
+  echo "python3"
+}
```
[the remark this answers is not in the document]
ericcurtin: I like this suggestion, although I'd do an equals twice instead:

```
if python3 -c 'import sys; exit(not (sys.version_info.major == 3 and sys.version_info.minor == 11))'; then
  echo "python3.11"
else
  echo "python3"
fi
```

container-images/scripts/build_stable_diffusion.sh
```diff
@@ -0,0 +1,186 @@
...
+      exit 1
+    fi
+  fi
+}
+
+cmake_steps() {
+  local cmake_flags=("$@")
+  cmake -B build "${cmake_flags[@]}" 2>&1 | cmake_check_warnings
+  cmake --build build --config Release -j"$(nproc)" 2>&1 | cmake_check_warnings
+  cmake --install build 2>&1 | cmake_check_warnings
```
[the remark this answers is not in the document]
ericcurtin: This is helping us catch warnings that were causing failures

Makefile
```diff
@@ -43,16 +44,16 @@ help:
 	@echo
 
 install-detailed-cov-requirements:
-	pip install ".[cov-detailed]"
+	${MYPIP} install ".[cov-detailed]"
```
ericcurtin, line 47: We are actually trying to standardize on uv these days... But I wouldn't say pipx is worse than pip, at least it works on macOS like uv

ericcurtin:
Need to sign commits to pass the DCO build

ericcurtin (approved):
If we can get this to pass the builds, it's good enough to merge

Makefile
```diff
@@ -7,6 +7,7 @@ SHAREDIR ?= ${PREFIX}/share
 PYTHON ?= $(shell command -v python3 python|head -n1)
 DESTDIR ?= /
 PATH := $(PATH):$(HOME)/.local/bin
+MYPIP := pipx
```
sarroutbi: Why not using `MYPIP ?= pipx` here?

It will be equivalent, but, in case anybody wants to override it externally, it will be possible to do so:

```bash
MYPIP=pip make unit-test
```

jtligon: Thanks! nice catch!

container-images/scripts/build_stable_diffusion.sh
```diff
@@ -0,0 +1,186 @@
...
+  git clone https://github.com/leejet/stable-diffusion.cpp
+  cd stable-diffusion.cpp
+  git submodule update --init --recursive
+  # Use master branch (latest stable)
+  cmake_steps "${common_flags[@]}"
+  
+  # Install the stable diffusion binary
+  install -m 755 build/bin/sd "$install_prefix"/bin/sd
+  cd ..
+  rm -rf stable-diffusion.cpp
```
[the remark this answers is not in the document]
ericcurtin: Disagree, it will fail anyway if build/bin/sd is not available:

```
install -m 755 build/bin/sd "$install_prefix"/bin/sd
install: cannot stat 'build/bin/sd': No such file or directory
```

container-images/scripts/build_stable_diffusion.sh
```diff
@@ -0,0 +1,186 @@
...
+  if [ -n "$containerfile" ]; then
+    install_ramalama "${install_prefix}"
+  fi
+
+  setup_build_env
+  # Build stable diffusion only (no whisper)
+  clone_and_build_stable_diffusion_cpp
+  
+  available dnf && dnf_remove
+  rm -rf /var/cache/*dnf*
```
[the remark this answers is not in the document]
ericcurtin: Disagree

Makefile
```diff
@@ -7,7 +7,7 @@ SHAREDIR ?= ${PREFIX}/share
 PYTHON ?= $(shell command -v python3 python|head -n1)
 DESTDIR ?= /
 PATH := $(PATH):$(HOME)/.local/bin
-MYPIP := pipx
+MYPIP ?= uv
```
rhatdan: Should this be `uv pip`?

rhatdan:
Could you squash your commits?

Not sure why MYPIP ends up being?

git rebase -i origin 
git push --force

Makefile
jtligon: I tried that, and it didn't pass, so I went back to just pip.

rhatdan:
LGTM

PR #1771, opened 2025-07-29
Include the host in the quadlet's PublishPort directive

Stebalien:
Without this:

  ramalama serve --generate=quadlet --host 127.0.0.1 --port 1234

Generates a container that binds to 0.0.0.0, not 127.0.0.1.

## Summary by Sourcery

Implement host specification in quadlet's PublishPort directive to bind containers to the provided host address

Enhancements:
- Include host parameter in PublishPort format as host:containerPort:containerPort, defaulting to 0.0.0.0

Tests:
- Add unit test for custom host mapping in quadlet port directive
- Add test fixtures for localhost-specific port mapping

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -997,6 +997,7 @@ def run_cli(args):
         except Exception as exc:
             raise e from exc
 
+
 def serve_parser(subparsers):
     parser = subparsers.add_parser("serve", help="serve REST API on specified AI Model")
     runtime_options(parser, "serve")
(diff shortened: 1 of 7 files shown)


rhatdan:
LGTM

rhatdan:
Thanks @Stebalien

rhatdan:
Some tests need to be updated for this change?  Should we handle the default case of 0.0.0.0 by not showing it and just showing ports?

Stebalien:
> Some tests need to be updated for this change?

Ah, I only ran the python tests.

> Some tests need to be updated for this change? Should we handle the default case of 0.0.0.0 by not showing it and just showing ports?

Up to you. I prefer an explicit 0.0.0.0 because the default is IPv4 only and explicitly stating 0.0.0.0 it clear, but I don't have a strong preference.

rhatdan:
I am fine with the explicit host in quadlets

rhatdan:
LGTM

PR #1773, opened 2025-07-30
De-duplicate bash build scripts

ericcurtin:
Create a common lib.sh

## Summary by Sourcery

Deduplicate container image build scripts by extracting shared bash functions into a common lib.sh and updating the scripts to source and use those utilities

Enhancements:
- Extract common utility functions into container-images/scripts/lib.sh
- Remove duplicate function definitions from build-vllm.sh and build_llama_and_whisper.sh
- Update build scripts to source lib.sh and use git_clone_specific_commit with parameters

### DIFF
--- a/container-images/scripts/build-vllm.sh
+++ b/container-images/scripts/build-vllm.sh
@@ -1,38 +1,5 @@
 #!/bin/bash
 
-available() {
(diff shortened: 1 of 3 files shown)


rhatdan (approved):
LGTM

PR #1777, opened 2025-07-30
Enable multiline chat

engelmi:
This PR enables multiline chat by using the backslack `\` symbol as control character. 
Users can either add `\` to the end of the line or paste a text block (with `\` at the end of each line) to the chat console:
```bash
🦭 > hi \
🦭 > tell me a funny story \
🦭 > please
sure, here's a funny story:
...
```

## Summary by Sourcery

Enable multiline user input in the chat console by treating a trailing backslash as a continuation marker and sending the accumulated lines as a single message once the backslash is omitted.

New Features:
- Allow users to enter multiline messages by ending lines with a backslash and pasting blocks with backslashes at line ends

Enhancements:
- Initialize and maintain a content buffer in the Chat class to accumulate and join input lines into one message

### DIFF
--- a/docs/ramalama-chat.1.md
+++ b/docs/ramalama-chat.1.md
@@ -53,6 +53,12 @@ $ ramalama chat
 Communicate with an alternative OpenAI REST API URL. With Docker containers.
 $ ramalama chat --url http://localhost:1234
 🐋 >
+
+Send multiple lines at once
+$ ramalama chat
+🦭 > Hi \
+🦭 > tell me a funny story \
+🦭 > please
 ```
 
 ## SEE ALSO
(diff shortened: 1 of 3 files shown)


ericcurtin (approved):
Neat! I guess it's a workaround for true multi-line, but it's better than nothing

ericcurtin:
Would be worth deleting this again if we do true multi-line

rhatdan:
Can you document this in the man page for ramalama chat and ramalama run, potentially just add your example above.

engelmi:
> Can you document this in the man page for ramalama chat and ramalama run, potentially just add your example above.

Added :heavy_check_mark:

engelmi:
> Neat! I guess it's a workaround for true multi-line, but it's better than nothing
> Would be worth deleting this again if we do true multi-line

Yes, I agee. It is really just a quick hack to enable multi-line.

PR #1784, opened 2025-08-01
Fix handling of configured_has_all

rhatdan:
## Summary by Sourcery

Bug Fixes:
- Define configured_has_all outside the if statement to prevent it from being undefined when checking for the “all” device.

### DIFF
--- a/ramalama/common.py
+++ b/ramalama/common.py
@@ -352,7 +352,8 @@ def check_nvidia() -> Literal["cuda"] | None:
 
     configured, unconfigured = find_in_cdi(visible_devices + ["all"])
 
-    if unconfigured and not (configured_has_all := "all" in configured):
+    configured_has_all = "all" in configured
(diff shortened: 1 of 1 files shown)


rhatdan:
@ieaves PTAL, your PR was crashing with configured_has_all was not set, the sourcery is complaining about my fix.

ieaves:
Ahh, it failed to set when the original if condition short-circuited. The elephant operator is a blessing and a curse at times.

There's nothing wrong with your fix AFAICT. It's all the same original logic, the `configured_has_all` flag was just meant to remove what was previously a duplicated `"all" in configured` check.

I don't know enough about the implementation to know why we call find_in_cdi like this `find_in_cdi(visible_devices + ["all"])`. It may be a redundant check but it's a subtle enough issue that it might be better to leave the check in place as is if that better reflects the actual intention.

rhatdan:
Well I am merging.

Issue #1778, opened 2025-07-31
Models downloaded to /tmp instead of $HOME/.local/share/ramalama


xek:
### Issue Description

Models are downloaded to /tmp instead of $HOME/.local/share/ramalama, resulting in OSError: [Errno 28] No space left on device

I have enough space in my home directory and I'd like ramalama to respect the [--store=STORE](https://www.mankier.com/1/ramalama#Global_Options---store=STORE) option .

### Steps to reproduce the issue

Download a large model, which doesn't fit into your /tmp, using `ramalama pull`.

### Describe the results you received

```
Downloading 'model-00035-of-00047.safetensors' to '/tmp/tmpud_9iexn/.cache/huggingface/download/qL_cQqWQBgT_FdmAuv2uJix1LI0=.1987ca16a5051008d42528a6090e0858cca95e96635c17630fc4b42be64fba2b.incomplete'
model-00030-of-00047.safetensors:  54%|█████████████████████████████████████████▎                                   | 1.26G/2.35G [00:13<00:11, 91.3MB/s]
model-00035-of-00047.safetensors:   0%|▎                                                                            | 10.5M/2.35G [00:00<00:30, 75.4MB/s]
model-00034-of-00047.safetensors:   0%|▎                                                                            | 10.5M/2.35G [00:00<00:36, 63.6MB/s]
model-00033-of-00047.safetensors:   0%|▎                                                                            | 10.5M/2.35G [00:00<00:40, 57.9MB/s]
Traceback (most recent call last):
  File "/usr/local/bin/huggingface-cli", line 8, in <module>                                                        | 10.5M/2.35G [00:00<00:40, 58.0MB/s]
    sys.exit(main())
             ~~~~^^
  File "/usr/local/lib/python3.13/site-packages/huggingface_hub/commands/huggingface_cli.py", line 61, in main
    service.run()
    ~~~~~~~~~~~^^
  File "/usr/local/lib/python3.13/site-packages/huggingface_hub/commands/download.py", line 157, in run
    print(self._download())  # Print path to downloaded files
          ~~~~~~~~~~~~~~^^
  File "/usr/local/lib/python3.13/site-packages/huggingface_hub/commands/download.py", line 191, in _download
    return snapshot_download(
        repo_id=self.repo_id,
    ...<10 lines>...
        max_workers=self.max_workers,
    )
  File "/usr/local/lib/python3.13/site-packages/huggingface_hub/utils/_validators.py", line 114, in _inner_fn
[...]

PR #1786, opened 2025-08-02
Set TMPDIR to /var/tmp if not set

rhatdan:
Currently we are defaulting to /tmp and in some cases this means pulling huge models to the /tmp directory which on a lot of linux platforms is a tmpfs. /var/tmp tends to have much more space, so setting this as a default is better.

Fixes: https://github.com/containers/ramalama/issues/1778

## Summary by Sourcery

Use /var/tmp as the default temporary directory if TMPDIR is unset and update documentation accordingly

Enhancements:
- Default TMPDIR to /var/tmp when not set to avoid tmpfs size constraints

Documentation:
- Document the new TMPDIR environment variable in the man page

### DIFF
--- a/bin/ramalama
+++ b/bin/ramalama
@@ -1,11 +1,16 @@
 #!/usr/bin/env python3
 
+import os
 import sys
 
 
 def main():
     if not sys.stdout.isatty():
         sys.stdout.reconfigure(line_buffering=True)
+
+    if sys.platform != 'win32':
+        # default TMPDIR environment to /var/tmp on non Windows platforms
+        os.environ["TMPDIR"] = os.getenv("TMPDIR", "/var/tmp")
 
     sys.path.insert(0, './')
     try:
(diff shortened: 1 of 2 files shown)


engelmi:
@rhatdan I saw this PR too late and created #1787 to address #1778. It uses the model store directory for the temporary files and since we use `with tempfile.TemporaryDirectory(...` its guaranteed to be cleaned up.

rhatdan:
I think this one should still go in, because huggingface-cli will default to pulling in /tmp as well.

I just merged your fix, but I think it is also safer to use /var/tmp rather then /tmp by default.

engelmi:
> I think this one should still go in, because huggingface-cli will default to pulling in /tmp as well.
> 
> I just merged your fix, but I think it is also safer to use /var/tmp rather then /tmp by default.

Good point. Lets merge this as well. 
(public testing farm has outage, which explains the failed provision)

engelmi (approved):
LGTM

Issue #1785, opened 2025-08-01
Erasing all containers of a contributor is not a friendly behavior and should not happen !


Annakan:
#### Rant ?

Sorry for the catchy scary title but it encapsulates my experience, with a much lower blood pressure than when it happened two days ago.

I was contributing a (very) minor functionality to Ramalama and trying to test (one should never tests is obviously the lesson here) the invocation parameters, thus using the "system tests" written around BATS.

At some point i saw this "
```ramalama on  Extra_Quadlet_Unit_Params is 📦 v0.11.2 via 🐍 v3.13.5 (ramalama) on ☁️  xxx
❯ RAMALAMA=./bin/ramalama bats -T test/system/040-serve.bats
   [teardown] $ podman pod rm -t 0 --all --force --ignore
   time="2025-07-31T11:01:14+02:00" level=warning msg="Found incomplete layer \"3cfb621c46982f9e5ff31b45c19f4b19dd946d7cf3ae35194ecefcf1cfa798be\", deleting it"
[...]
```

Besides the fact that I had problem with my podman installation after a hard reboot, the important point is this : 

```podman pod rm -t 0 --all --force --ignore```

This is NOT COOL. This is the container equivalent of `rm -rf ~/` !!!!

My podman installation problem (easily solvable if I still had coherent containers) triggered a recovery option in the test suite in ramalama/test/system/helpers.podman.bash :

```bash 
function clean_setup() {
    if [[ "${_RAMALAMA_TEST_OPTS}" == "--nocontainer" ]]; then
	return
    fi
    local actions=(
        "pod rm -t 0 --all --force --ignore"
            "rm -t 0 --all --force --ignore"
        "network prune --force"
        "volume rm -a -f"
    )
[...]
```
All containers, all pods/networks, all volumes ...

This is NOT COOL 😠.

On top of that this function activation is pretty loose : if the previous run of the tests did not end well, there can be a lot of reasons for that.

In `basic_teardown`:

```bash 
      if [ $exit_code -gt 0 ] || [ -z "$BATS_TEST_COMPLETED" ]; then
            clean_setup
            exit_code=$((exit_code + $?))
        fi
```

And in `setup_suite`

```bash 
[...]
	# The above does not handle errors. Do a final confirmation.
	assert "$PODMAN_LOGIN_REGISTRY_PORT" != "" \
               "Unable to set PODMAN_LOGIN_REGISTRY_PORT"
    fi
    clean_setup
[...]
```
[...]

PR #1792, opened 2025-08-04
Do not cleanup all podman contact if user runs make test.

rhatdan:
This was a hold over from copying the podman test suite when RamaLama was first created.  Will leave this function, since there might be some cleanup needed at some point.

Fixes: https://github.com/containers/ramalama/issues/1785

## Summary by Sourcery

Bug Fixes:
- Remove the blanket removal of containers, networks, volumes, and external resources in clean_setup so that make test no longer wipes all Podman content


rhatdan:
@mikebonnet @engelmi @olliewalsh @bmahabirbu PTAL

PR #1798, opened 2025-08-04
Fix the 050-pull test.

jwieleRH:
Remove the smollm:135m model with the URL used to pull it, otherwise the test will fail with:
 in test file test/system/050-pull.bats, line 25)
 `run_ramalama rm ollama://smollm:135m ollama://smollm:360m' failed
 ...
 Error: Model 'smollm:135m' not found

## Summary by Sourcery

Tests:
- Update 050-pull.bats to remove the 135m model using its full ollama.com/library URL instead of the shorthand schema


rhatdan:
LGTM

PR #1804, opened 2025-08-06
Explicitly pull of base container before converted container build

engelmi:
Fhe oci container uses the quiet option in order to get the image id properly. This, however, causes the convert command to seemingly sleep for a very long time since it might pull the latest base container image (which is quite large). In order to provide feedback to the user, lets pull explicityl before the build.

## Summary by Sourcery

Make the container build quiet mode optional by removing the hardcoded '-q' flag and conditionally adding '--quiet' when args.quiet is enabled, so that progress is shown when pulling new base images.

Enhancements:
- Remove default '-q' flag from both OCI and RAG build commands
- Add conditional '--quiet' flag based on args.quiet for podman/ramalama builds


rhatdan:
LGTM

PR #1806, opened 2025-08-06
Fixed error in readme to ensure consistency.

rhatdan:
Replaces: https://github.com/containers/ramalama/pull/1724

## Summary by Sourcery

Documentation:
- Update README.md and man page to use “does not exist locally” instead of “does not exists locally”

### DIFF
--- a/README.md
+++ b/README.md
@@ -874,7 +874,7 @@ $ cat /usr/share/ramalama/shortnames.conf
 	</summary>
 	<br>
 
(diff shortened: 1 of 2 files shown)


engelmi:
From the test log:
```
requests.exceptions.HTTPError: 429 Client Error: Too Many Requests for url: https://huggingface.co/api/models/ggml-org/SmolVLM-256M-Instruct-GGUF/revision/main
```
Seems like we hit the hf rate limit, so not related to this change.

Issue #598, opened 2025-01-16
Provide model info in chat ui & allow multiple models


sallyom:
Rather than serving all models with generic `model.file` filename, ramalama should provide more information about the currently served or loaded models.

Also, could allow passing a model-config file to allow to switch easily between models with a single instance of a server. 
https://llama-cpp-python.readthedocs.io/en/latest/server/#configuration-and-multi-model-support 

Here, I've renamed the file ramalama downloaded to `granite-code` so it shows at llamacpp:port/models.
![Image](https://github.com/user-attachments/assets/25ac6fed-f7e7-4fb7-9890-bca1210b25b1)

Currently, any model served by ramalama is listed as `model.file` 
![Image](https://github.com/user-attachments/assets/71c7316b-34d7-4233-9510-56ddd0175a9a)

ericcurtin:
We have just llama.cpp server and vllm server integrated but are open to llama-cpp-python server also.

ericcurtin:
This is how it should look when we've multiple models:

```
{"object":"list","data":[{"id":"smollm:360m","object":"model","created":1737073177,"owned_by":"library"},{"id":"smollm:135m","object":"model","created":1736344249,"owned_by":"library"}]}
```

ericcurtin:
@sallyom I think we should consider this:

https://github.com/BerriAI/litellm

it's compatible with vllm and llama.cpp, it sits in front of either.

vllm closed the door on this feature suggesting it:

https://github.com/vllm-project/vllm/issues/299#issuecomment-2327886605

considering we want switchable runtimes, it's ideal.

ericcurtin:
https://docs.litellm.ai/docs/providers/vllm

ericcurtin:
After reading a bit, I'm gonna propose we write our own proxy, litellm won't provide what we need. You can proxy/route/bridge (whichever terminology you prefer) requests to pre-spun up llama-servers or vllm servers running a single model with litellm. So using something like litellm we'd need to spin up a model server for every single model on a machine which would use too much memory.

So I'm proposing we write something called ramalama-server (which might be more correctly named ramalama-proxy but it may grow to have other features in future, I suggest python3, the http libraries we need should be built into it's stdlib, it will only be a couple of hundred lines of python3).

So when we call:

`ramalama serve some-model`

there's no functional change, except the python3 script we now call in the container would be "ramalama-server --runtime vllm/llama.cpp ...". We should make this run without containers too of course for macOS.

[...]

PR #1807, opened 2025-08-06
Initial model swap work

engelmi:
This PR adds initial work on the model swap feature.

## Basic idea

The core idea is roughly depicted in the following diagram:

image

The new `daemon` component is essentially a server providing two different REST APIs:
- Daemon API to control the models run in the container
- Proxy API to forward requests to the running models inside the container

The command `ramalama daemon setup` is starting the container with a running daemon inside, ready to handle requests. Currently, there are these endpoints for the Daemon API available:
- GET /api/tags - lists all models available in the mounted model store (excluding oci ones)
- GET /api/model - lists all running llama-server processes (so the running models) with additional information
- POST /api/serve - start a new llama-server process with the passed parameters (and defaults from ramalama)
- POST /api/stop - stop a llama-server process

## Trying it out

At the moment, the `daemon` package is missing in the latest ramalama container images. Therefore, one needs to "rebuild" a custom container image, for example:
```
FROM quay.io/ramalama/rocm:latest

COPY dist/ /
RUN pip install --force-reinstall /ramalama-0.11.3-py3-none-any.whl
```
And then build the wheel package and container image:
```bash
# in ramalama root directory
$ make pypi-build
$ podman build -f custom-rocm -t quay.io/ramalama/rocm:latest .
```
When done, one can start the container via 
```bash
./bin/ramalama daemon setup
```
and issue requests, e.g. via curl:
```bash
# list all available models
$ curl -X GET http://localhost:8080/api/tags
[
    {
        "name": "hf://mlx-community/Llama-3.2-1B-Instruct-4bit",
        "modified_at": "2025-07-24T14:32:16.156628+00:00",
        "size": 712575975,
        "digest": "",
        "details": {
            "format": "",
            "family": "",
            "families": [],
            "parameter_size": "",
            "quantization_level": ""
        }
    },
...
# serve smollm2:135m
curl  -X POST http://localhost:8080/api/serve -d '{"model_name":"smollm2:135m", "runtime":"llama.cpp", "exec_args": {"thinking": "true", "context": 2048, "temp": 0.8}}' -H "Content-Type: application/json"
{
    "model_id": "sha256-8cc4a7f5d0b22d87f9ca41680d72e87d62236588951180cdc479dbfa1c22a184",
    "serve_path": "/model/sha256-8cc4a7f5d0b22d87f9ca41680d72e87d62236588951180cdc479dbfa1c22a184"
}
# stop smollm2:135m
curl  -X POST http://localhost:8080/api/stop -d '{"model_name":"smollm2:135m"}' -H "Content-Type: application/json"
```

Navigating to `/model` in the browser (or again via curl) will list all running models:
image

Using the returned serve path from the serve call on the API enables to use the webui of llama-server:
Navigate to the webui:
image

## Summary by Sourcery

Introduce a containerized Ramalama daemon to dynamically manage and proxy multiple model serving processes via REST APIs.

New Features:
- Add `ramalama daemon` CLI with `setup` and `run` commands to launch a persistent model-serving daemon
- Implement an HTTP server providing /api endpoints for listing available models, starting new model processes, and stopping them
- Add a proxy handler under /model to list running models and forward requests to individual llama-server instances
- Develop a ModelRunner framework for spawning, tracking, and terminating managed model processes with dynamic port allocation
- Introduce CommandFactory and DTO classes to build serve commands and serialize/deserialize serve/stop request and response payloads
- Provide graceful shutdown handling and file-based logging for the daemon


ramalama/cli.py
```diff
@@ -1086,6 +1087,52 @@ def stop_container(args):
...
+            "--pull",
+            "never",
+            "-i",
+            "-t",
+            "-d",
+            "-p",
+            "8080:8080",
+            "-v",
+            f"{CONFIG.store}:/ramalama/models",
+            "quay.io/ramalama/rocm:latest",
```
bmahabirbu: I dont think we should have the image hardcoded here? Is it possible to add --image arg to this for easier testing?

bmahabirbu:
Nice work @engelmi!!! 

I'm trying to test this on my nvidia machine but running into this

```
brian@DESKTOP-SB69448:~/ramalama$ ./bin/ramalama --debug daemon setup
2025-08-16 00:24:54 - DEBUG - exec_cmd: podman run --pull never -i -t -d -p 8080:8080 -v /home/brian/.local/share/ramalama:/ramalama/models quay.io/ramalama/rocm:latest ramalama daemon run --store /ramalama/models
Error: quay.io/ramalama/rocm:latest: image not known
brian@DESKTOP-SB69448:~/ramalama$ 
```

Would it be possible to add --image to this argument?

ramalama/daemon/model_runner/command_factory.py
```diff
@@ -0,0 +1,114 @@
...
+
+        if check_nvidia() or check_metal(SimpleNamespace({"container": False})):
+            cmd.extend(["--flash-attn"])
+
+        # gpu arguments
+        ngl = self.request_args.get("ngl")
+        if ngl < 0:
+            ngl = 999
+        cmd.extend(["-ngl", f"{ngl}"])
+        cmd.extend(["--threads", f"{self.request_args.get("threads")}"])
```
bmahabirbu:
```
 File "/usr/local/lib/python3.11/site-packages/ramalama/daemon/model_runner/command_factory.py", line 112
    cmd.extend(["--threads", f"{self.request_args.get("threads")}"])
```

we can fix this by doing something like

```
        threads = self.request_args.get("threads")
        cmd += ["--threads", str(threads)]
```

bmahabirbu:
After fixing that error and resolving the image I got it to work!!

```
brian@DESKTOP-SB69448:~/ramalama$ curl -X GET http://localhost:8080/api/tags
[
    {
        "name": "ollama://llama3.2/llama3.2:latest",
        "modified_at": "2025-07-25T00:19:06.530056+00:00",
        "size": 2019379366,
        "digest": "",
        "details": {
            "format": "",
            "family": "",
            "families": [],
            "parameter_size": "",
            "quantization_level": ""
        }
    },
    {
        "name": "ollama://tinyllama/tinyllama:latest",
        "modified_at": "2025-05-14T02:46:44.949334+00:00",
        "size": 637700077,
        "digest": "",
        "details": {
            "format": "",
            "family": "",
            "families": [],
            "parameter_size": "",
            "quantization_level": ""
        }
    },
   ..etc

File: ramalama/cli.py (hunk 1 above)
Comment (bmahabirbu) [reply ramalama/cli.py]: maybe add the --image parser to daemon_parser then in the exec_args do args.image instead of the hardcoded image

Comment (engelmi) [reply ramalama/cli.py]: Yes, you are right. I pushed a new commit with improved CLI options (incl. the `--image`). PTAL

File: ramalama/daemon/model_runner/command_factory.py (hunk 2 above)
Comment (engelmi) [reply ramalama/daemon/model_runner/command_factory.py]: Ups, good catch! Changed

**Comment by rhatdan** [comment]
Fixes: https://github.com/containers/ramalama/issues/598

File: ramalama/daemon/handler/ramalama.py [hunk 3]
```diff
@@ -0,0 +1,106 @@
...
+        self.setup()
+        try:
+            self.handle()
+        except Exception as e:
+            self.send_error(500, f"Internal Server Error: {e}")
+            logger.error(f"Error handling request: {e}")
+        finally:
+            self.finish()
+
+    def do_GET(self):
```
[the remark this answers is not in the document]
engelmi: Can be done later. Not sure we will stick with this simplistic webserver implementation or move to something like Flask.

engelmi:
@rhatdan PTAL. I think it is ready to be reviewed/merged (sorry for the size of the PR).

Quick, updated summary of this PR:
* The REST API of the daemon has two path prefixes: `/api` and `/model`
* `/api` provides endpoints for REST operations, e.g. starting to serve a model and listing running ones (see curl examples in the description)
* `/api` endpoints are compliant with ollama cli (only listing available (`ollama ls`) and running models (`ollama ps`) at the moment) - provided the ramalama port is 11434 (used by ollama cli)
* `/model/<org>/<name>` provides the llama-server API (OpenAPI compliant)
* it also adds an expiration timeout, stopping served models when no traffic was redirected there (currently ~5min)

I'd try to get this PR merged since its already quite big. In follow-up PRs we can tackle things like:
* ramalama CLI integration, e.g. :
   * `serve`, `run` and `chat` starting the daemon and triggering the serve endpoint (if necessary) and then using REST API requests for communication
   * adding a nice web dashboard, e.g. in the REST API root path `/` (as discussed in [#598](https://github.com/containers/ramalama/issues/598))
   * defining a swap config - with sensible defaults - defining the behavior for swapping models
   * etc.
 
I can create GH issues for these to keep track of and plan them.

rhatdan:
@olliewalsh @mikebonnet @bmahabirbu @ericcurtin PTAL

ramalama/daemon/daemon.py
```diff
@@ -0,0 +1,108 @@
...
+        RamalamaHandler(self.model_store_path, self.model_runner, request, client_address, self)
+
+    def check_model_expiration(self):
+        curr_time = datetime.now()
+        for name, m in self.model_runner.managed_models.items():
+            if m.expiration_date > curr_time:
+                continue
+
+            try:
+                logger.error(f"Stopping expired model '{name}'...")
```
rhatdan: Should be Warn or Debug?

ramalama/daemon/daemon.py
```diff
@@ -0,0 +1,108 @@
...
+    parser.add_argument("--host", type=str, default="0.0.0.0")
+    parser.add_argument("--port", type=int, default=8080)
+    parser.add_argument("--model-store-path", type=str, default="/models")
+
+    return parser.parse_args()
+
+
+def run(host: str = "0.0.0.0", port: int = 8080, model_store_path: str = "/models"):
+    configure_logger(LogLevel.DEBUG)
+    logger.info(f"Starting Ramalama daemon on {host}:{port}...")
```
rhatdan: Either drop this to Debug or add a quiet option.

rhatdan:
I would like to discuss with you how you see this working from a user point of view?

With the case of containers, would the ramalama daemon execute within the container?

ericcurtin:
My only request would be to try and test for interoperability with Docker Model Runner... Lets say a user takes a client that implements Docker Model Runner support for the sake of argument, lets say it's something like AnythingLLM (a client could be a lot of things)... It would be nice if users could just switch the port the client points to between RamaLama/Docker Model Runner, etc. and everything "just works"... Same for OCI artifact standard, etc.

rhatdan:
I agree, I think we should have good integration between RamaLama and Docker Model Runner.

ramalama/daemon/daemon.py
engelmi: Changed to debug.

ramalama/daemon/daemon.py
engelmi: Changed to info since stopping an expired model is something expected and "normal" and debug seems not enough.

engelmi:
> My only request would be to try and test for interoperability with Docker Model Runner... Lets say a user takes a client that implements Docker Model Runner support for the sake of argument, lets say it's something like AnythingLLM (a client could be a lot of things)... It would be nice if users could just switch the port the client points to between RamaLama/Docker Model Runner, etc. and everything "just works"... Same for OCI artifact standard, etc.

Thanks for the feedback! @ericcurtin Definitely! So far I haven't looked at the [REST API of DMR](https://docs.docker.com/ai/model-runner/api-reference/), though. The [DMR endpoints](https://docs.docker.com/ai/model-runner/api-reference/#available-dmr-endpoints) should be easy to mimic after some reverse engineering (since I can't find details such as payload, unfortunately). The [provided OpenAI endpoints](https://docs.docker.com/ai/model-runner/api-reference/#available-openai-endpoints) are not yet clear to me. For example, we currently spin up multiple llama.cpp processes (containers without the daemon), so each llama.cpp instance has its own models... probably we need to improve the llama.cpp integration to achieve this, although I don't know yet if having multiple processes is better/worse than having one big process.
Regarding the OCI artifacts: Yeah, that is definitely missing at the moment - I only mount the local model store directory into the daemon container, but don't know how it could be done with OCI artifacts - that needs to be investigated. Or we just keep the listing of available models on the host.

So far I only looked at Ollama CLI and made the two endpoints `/api/ps` (listing running models) and `/api/ tags` (listing available models) compatible with it - so running the daemon on port 11434 and using `ollama ps` would work. But as I wrote, adding more path aliases (and using the user-agent field to distinguish) should be a viable option.

engelmi:
> I would like to discuss with you how you see this working from a user point of view?
> 
> With the case of containers, would the ramalama daemon execute within the container?

Yes, the ramalama daemon starts a simple HTTP server - either directly on the host or in the ramalama container. Because of the latter, it would not be usable directly since this code is missing in the current ramalama images. This is also the reason why I didn't do the integration with the ramalma serve/run/chat command yet. I think this delayed release of change is not good, but couldn't come up with a better idea other than pushing it to "build a new image on each merge".

From the user perspective, nothing would change for these three commands - it would all be under the hood. However, we could add then additional CLI options over time leveraging this daemon, e.g. `ramalama serve smollm2 --ttl=5min` to define a max. idle time. Or `ramalama serve smollm2 --group=xyz` allowing group definitions for model swapping.

rhatdan:
Ok lets merge and then we can continue to discuss and play with it.

LGTM

PR #1808, opened 2025-08-06
Untabify test/system/helpers.bash in accordance with PEP8.

jwieleRH:
gemini-code-assist suggests using spaces for all indentation in accordance with PEP8.

See https://peps.python.org/pep-0008/#tabs-or-spaces


rhatdan:
We should add a peb8 linter so these can not get in.
LGTM

PR #1809, opened 2025-08-06
Wait for container to be healthy with ramalama run

rhatdan:
## Summary by Sourcery

Improve `ramalama run` to wait for Podman containers to pass a built-in HTTP health check before completing, enhancing reliability and debuggability

New Features:
- Wait for Podman containers to become healthy in `ramalama run` before proceeding

Enhancements:
- Inject an HTTP health check into containers via `--health-cmd`
- Implement `wait_for_healthy` helper to poll Podman container health status
- Log errors and suggest using `ramalama serve` for detailed failure diagnostics on health-check failures

### DIFF
--- a/ramalama/engine.py
+++ b/ramalama/engine.py
@@ -3,6 +3,7 @@ import json
 import os
 import subprocess
 import sys
+import time
 
 # Live reference for checking global vars
 import ramalama.common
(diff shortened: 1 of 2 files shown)


ramalama/engine.py
```diff
@@ -333,3 +333,10 @@ def add_labels(args, add_label):
     for arg, label_prefix in label_map.items():
         if value := getattr(args, arg, None):
             add_label(f"{label_prefix}={value}")
+
+
+def wait_for_healthy(args):
+    conman_args = ["podman", "wait", "--condition", "healthy", args.name]
```
kpouget: I'm wondering what happens when the container doesn't turn healthy, like
- the container crashes because it cannot correctly access the GPU
- the container blocks forever (for whatever reason around the GPU initialization)

both cases happen to me from time to time 🙃

rhatdan: The goal of this is to catch he case when the container crashes, so it will throw an exception and report to the user.  I have taken the AI advise and am chaning to have a time out of 30 seconds to exit if the container hangs.

Sadly one of my test cases is that the container gets to healthy and then crashes, which the user will only see once they send a query to it.

kpouget: nice, that's a great enhancement, thanks 🙏🏻

ramalama/engine.py
```diff
@@ -333,3 +334,26 @@ def add_labels(args, add_label):
     for arg, label_prefix in label_map.items():
         if value := getattr(args, arg, None):
             add_label(f"{label_prefix}={value}")
+
+
+def wait_for_healthy(args, timeout=20):
```
kpouget, line 339: one thing I'm wondering regarding the timeout is if we're sure that podman will never pull the image while this `wait_for_healthy` is counting down

like
```
podman pull image
podman run image
```
and 
```
podman run image
```
these two blocks are semantically similar, but the second `podman run` may take longer before the container comes healthy

rhatdan: I might have to account for this.  That is probably what is failing in the test infrastructure.

rhatdan: Made this change:
```
+        self.ensure_model_exists(args)
         pid = os.fork()
```

PR #1822, opened 2025-08-11
fix "skopeo copy" command"

mikebonnet:
`--multi-arch all` conflicts with `--all`.

## Summary by Sourcery

Bug Fixes:
- Remove --multi-arch all flag from skopeo copy command to avoid conflict with --all.


mikebonnet:
Merging this before the Monday release. Extensively tested over the last week.

PR #1827, opened 2025-08-11
konflux: build on the fastest instance types available, and increase timeouts

mikebonnet:
Build on the fastest instance types available in Konflux to accelerate developer feedback.

Increase pipeline timeouts to avoid unnecessary failures.


rhatdan (approved):
LGTM

Issue #1126, opened 2025-04-06
ramalama chat inferencing on mac returns endless random text


edhaynes:
Installed ramalama on mac m3 using bash script from git page. A previous install from a month ago worked fine.  I tried a few chat llms and no matter which one i try any question returns an unending stream of nonsense text.  Happens using either the CLI or the webserver interface.  Also note ramalama "version" is returning zero which doesn't seem correct.

ehaynes@ehaynes-mac ~ % curl -fsSL https://raw.githubusercontent.com/containers/ramalama/s/install.sh | bash
  _____                       _
 |  __ \                     | |
 | |__) |__ _ _ __ ___   __ _| |     __ _ _ __ ___   __ _
 |  _  // _` | '_ ` _ \ / _` | |    / _` | '_ ` _ \ / _` |
 | | \ \ (_| | | | | | | (_| | |___| (_| | | | | | | (_| |
 |_|  \_\__,_|_| |_| |_|\__,_|______\__,_|_| |_| |_|\__,_|

==> Downloading https://formulae.brew.sh/api/formula.jws.json
==> Downloading https://formulae.brew.sh/api/cask.jws.json
Warning: llama.cpp 5060 is already installed and up-to-date.
To reinstall 5060, run:
  brew reinstall llama.cpp

🦙🦙🦙🦙🦙🦙🦙🦙🦙🦙🦙🦙🦙🦙🦙🦙🦙🦙🦙🦙🦙🦙🦙🦙🦙🦙🦙🦙
ehaynes@ehaynes-mac ~ % ramalama version
ramalama version 0
ehaynes@ehaynes-mac ~ % ramalama-version
zsh: command not found: ramalama-version
ehaynes@ehaynes-mac ~ % ramalama run granite3-moe
Attempting to pull quay.io/ramalama/ramalama:0...
Attempting to pull quay.io/ramalama/ramalama:0...
🦭 > Where is Chicago?
[...]

PR #1831, opened 2025-08-13
Revert "Switch default ramalama image build to use VULKAN"

booxter:
This change makes ramalama container generate garbage on MacOS, as was
originally reported at:

https://github.com/containers/ramalama/issues/1126

I believe the original promise was that VULKAN will work fine with
podman 5.5+, but I'm on 5.5.1 and still experience this problem.

DKOMPUTE_OPT_DISABLE_VULKAN_VERSION_CHECK is no longer supported by
upstream, so leaving it out of the revert.

This reverts commit b6d5e95e2c175311fc88cf18d337a5fd1727745b.

## Summary by Sourcery

Bug Fixes:
- Restore GGML_KOMPUTE as the default flag instead of GGML_VULKAN for x86_64 and aarch64 platforms in the build script for ramalama

### DIFF
--- a/container-images/scripts/build_llama_and_whisper.sh
+++ b/container-images/scripts/build_llama_and_whisper.sh
@@ -277,7 +277,7 @@ add_common_flags() {
   case "$containerfile" in
   ramalama)
     if [ "$uname_m" = "x86_64" ] || [ "$uname_m" = "aarch64" ]; then
-      common_flags+=("-DGGML_VULKAN=ON")
+      common_flags+=("-DGGML_KOMPUTE=ON")
     elif [ "$uname_m" = "s390x" ]; then
       common_flags+=("-DGGML_VXE=ON" "-DGGML_BLAS=ON")
       common_flags+=("-DGGML_BLAS_VENDOR=OpenBLAS")


rhatdan:
@slp Do you know if this is necessary?
@engelmi @olliewalsh Thoughts?

rhatdan:
Are you sure you have the latest krunkit?

kpouget:
@booxter , can you explain, maybe in a new ticket, what's going wrong for you exactly?

engelmi:
> @booxter , can you explain, maybe in a new ticket, what's going wrong for you exactly?

Yes, that would be great. Then we also know which model(s) you are trying to run etc. Please also attach the logs when run with `--debug` @booxter 
[This issue](https://github.com/ggml-org/llama.cpp/issues/14974) on llama.cpp could be related, maybe?

booxter:
Thanks for initial thoughts. I will report the issue with some details later this week and link back here. I will also try to pull the mesa commits in my env to see if they fix anything.

rhatdan:
Also check other models to see if those work.

booxter:
@kpouget @rhatdan @engelmi I tried to capture what I see in https://github.com/containers/ramalama/issues/1843 please let me know if I can collect any more info that would be of help.

PR #1840, opened 2025-08-14
konflux: migrate to per-component service accounts

mikebonnet:
Konflux is moving from a shared `appstudio-pipeline` service account to a separate service account for each component, to increase security and improve efficiency.


rhatdan:
LGTM

Issue #1844, opened 2025-08-15
NVIDIA checking on my machine is broken.


rhatdan:
### Issue Description

def load_cdi_yaml(stream: Iterable[str]) -> CDI_RETURN_TYPE:

Is finding no devices on my nvidia system.

### Steps to reproduce the issue

Run ramalama and get error aboud cdi devices.

```
 ./bin/ramalama version
Device all does not have a CDI configuration
No CDI configuration found for all
You can use the "nvidia-ctk cdi generate" command from the 
nvidia-container-toolkit to generate a CDI configuration.
See ramalama-cuda(7).
ramalama version 0.12.0
```

### Describe the results you received

Bogus complaint about nvidia-ctk

### Describe the results you expected

nvidia support.

### ramalama info output

```yaml
$ ./bin/ramalama info
Device all does not have a CDI configuration
No CDI configuration found for all
You can use the "nvidia-ctk cdi generate" command from the 
nvidia-container-toolkit to generate a CDI configuration.
See ramalama-cuda(7).
{
    "Accelerator": "none",
    "Config": {
        "selinux": true,
        "settings": {
            "config_files": [
                "/usr/share/ramalama/ramalama.conf",
                "/home/dwalsh/.config/ramalama/ramalama.conf",
                "/home/dwalsh/.config/ramalama/ramalama.conf.d/foo.conf"
            ]
        }
    },
    "Engine": {
        "Info": {
            "host": {
                "arch": "amd64",
                "buildahVersion": "1.40.1",
                "cgroupControllers": [
                    "cpu",
                    "io",
                    "memory",
                    "pids"
                ],
                "cgroupManager": "systemd",
                "cgroupVersion": "v2",
                "conmon": {
                    "package": "conmon-2.1.13-1.fc42.x86_64",
                    "path": "/usr/bin/conmon",
                    "version": "conmon version 2.1.13, commit: "
                },
                "cpuUtilization": {
                    "idlePercent": 95.72,
                    "systemPercent": 1.01,
                    "userPercent": 3.27
                },
                "cpus": 32,
                "databaseBackend": "sqlite",
                "distribution": {
                    "distribution": "fedora",
                    "variant": "workstation",
                    "version": "42"
                },
                "eventLogger": "journald",
                "freeLocks": 2043,
                "hostname": "danslaptop",
                "idMappings": {
                    "gidmap": [
                        {
                            "container_id": 0,
                            "host_id": 3267,
                            "size": 1
                        },
                        {
[...]

PR #1845, opened 2025-08-15
Fix handling of nvidia settings in /etc/cdi/nvidia.yaml

rhatdan:
Fixes: https://github.com/containers/ramalama/issues/1844

### DIFF
--- a/ramalama/common.py
+++ b/ramalama/common.py
@@ -251,7 +251,7 @@ def load_cdi_yaml(stream: Iterable[str]) -> CDI_RETURN_TYPE:
     for line in stream:
         if ':' in line:
             key, value = line.split(':', 1)
-            if key.strip() == "- name":
+            if key.strip() == "- name" or key.strip() == "name":
                 data['devices'].append({'name': value.strip().strip('"')})
     return data
 


ieaves:
I'm not 100% what sort of structure these CDI yaml files are supposed to abide by but... re-implementing yaml parsing ourselves is destined to be broken in one way or another. I'm assuming [this](https://github.com/cncf-tags/container-device-interface/blob/main/SPEC.md#cdi-json-specification) is the reference spec for which the key "name" only appears once which reduces the problem surface area dramatically. If it's the spec I don't see what circumstance would lead the key to be a list though.

EDIT: Ahh yeah, okay, if name is the first record in the list. I think in that case, we also need to check for `"-name"` so possibly

```python
key.strip() in {"name", "- name", "-name"}
```

or something like

```python
key.lstrip("- ").rstrip() == "name"
```

To avoid constructing the set? Would there ever exist a valid device name which starts with the character "-"?

jwieleRH:
A. I have a similar PR in the works that also improves the unit tests using more realistic test data extracted from actual cdi configuration files. I'm testing it on machines with 0, 1, and 4 NVIDIA devices.
B. I have run multiple versions of nvidia-ctk on several OSes with a variety of nvidia GPUs and I have never seen it generate a YAML CDI configuration file with "- name". So I am curious where that has been observed.
C. I have said more than once and I still believe that the way to get the parsing right is to import the pyYAML package and be done with it. It is not a large dependency. It wouldn't be the first external dependency. gemini-code-assist agrees with me FWTW.

mikebonnet:
I saw `- name` in the yaml generated by `nvidia-ctk cdi generate` on AWS gpu instances, using the latest version of the NVIDIA Container Toolkit (1.17.8 I believe), which prompted the change which broke previous behavior, sorry about that. FWIW I agree with @jwieleRH that we should just use `pyyaml`.

rhatdan:
I am fine with using pyyaml if people can guarantee this will not cause issues on MAC platforms.  We have been very conservative on RamaLama client to not pull in external modules, for fear of support on MAC and Windows python.

rhatdan:
Will go with pyyaml solution.

PR #1846, opened 2025-08-15
Fix progress bar not reaching 100%

olliewalsh:
Progress bar never reached 100% as the method was returning as soon as the download completed - skipping the final progress update.

## Summary by Sourcery

Fix progress bar not reaching 100% by replacing the early return with a loop break, restoring the final update_progress call, and adjusting the line-clearing behavior

Bug Fixes:
- Prevent early exit from download loop to allow final progress update
- Ensure final progress update occurs when show_progress is enabled

### DIFF
--- a/ramalama/http_client.py
+++ b/ramalama/http_client.py
@@ -59,24 +59,27 @@ class HttpClient:
         self.start_time = time.time()
         accumulated_size = 0
         last_update_time = time.time()
-        while True:
-            data = self.response.read(1024)
-            if not data:
-                return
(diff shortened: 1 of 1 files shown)


rhatdan (approved):
LGTM

rhatdan:
I think the MAC failure might be real.  I have tried to retest multiple times.  @olliewalsh WDYT?

olliewalsh:
Should be fixed in https://github.com/containers/ramalama/pull/1847

Issue #1857, opened 2025-08-19
chat template conversion generates invalid jinja for ollama://deepseek-r1


olliewalsh:
### Issue Description

Converting the ollama chat template to jinja produces invalid jinja. llama.cpp fails to parse it and falls back to chatml chat tamplate.

### Steps to reproduce the issue

ramalama --debug serve deepseek

### Describe the results you received

```
...
common_chat_templates_init: failed to parse chat template (defaulting to chatml): Failed to parse number: '.' ([json.exception.parse_error.101] parse error at line 1, column 1: syntax error while parsing value - invalid literal; last read: '.') at row 1, column 22:
{{- if .System }}{{ .System }}{{ end }}
                     ^
{{- range $i, $_ := .Messages }}
 
```

Generated jinja:
```
{%- if system %}{{ system }}{% endif %}
{%- for m in messages %}
{%- set last = ((messages)[loop.index0:])|length==1%}
{%- if m["role"]=="user" %}<｜User｜>{{ m["content"] }}
{%- elif m["role"]=="assistant" %}<｜Assistant｜>
  {%- if isthinkset and last and m["thinking"] and -%}
<think>
{{ m["thinking"] }}
</think>
{%- endif %}{{ m["content"] }}{%- if not last %}<｜end▁of▁sentence｜>{%- endif %}
{%- endif %}
{%- if last and ne .Role "assistant" %}<｜Assistant｜>
{%- if isthinkset and not think and -%}
<think>

</think>

{% endif %}
{%- endif -%}
{%- endfor %}
```

### Describe the results you expected

valid jinja

### ramalama info output

```yaml
latest main
```

### Upstream Latest Release

Yes

PR #1858, opened 2025-08-19
Fix deepseek-r1 chat template conversion

olliewalsh:
From https://github.com/engelmi/go2jinja/pull/5

Fixes https://github.com/containers/ramalama/issues/1857

## Summary by Sourcery

Improve Go-to-Jinja template conversion by refining operator mapping, expression grouping, variable parsing, and control-flow handling.

Bug Fixes:
- Change FunctionType.NEQUALS operator from 'neq' to 'ne'
- Update regex to allow dots in variable and local variable names
- Raise a generic exception for unmatched braces instead of IndexError

Enhancements:
- Add a brackets flag to FunctionNode to preserve parentheses in complex expressions
- Propagate bracket information from parse_functions to to_jinja output
- Simplify go_to_jinja variable transforms to strip prefixes and map nested properties via dot notation
- Remove legacy continue-node translation and related helper logic

### DIFF
--- a/ramalama/model_store/go2jinja.py
+++ b/ramalama/model_store/go2jinja.py
@@ -7,7 +7,7 @@ import re
 import sys
 from dataclasses import dataclass, field
 from enum import Enum
-from typing import Dict, Optional, cast
+from typing import Dict, Optional
 
 
 class NodeType(Enum):
@@ -50,7 +50,7 @@ class FunctionType(Enum):
     NOT = "not"
 
     EQUALS = "eq"
-    NEQUALS = "neq"
+    NEQUALS = "ne"
     LESSER = "lt"
     LESSEREQUALS = "le"
     GREATER = "gt"
@@ -87,7 +87,14 @@ class FunctionNode:
 
     operands: list["FunctionNode"] = field(default_factory=lambda: [])
 
+    brackets: bool = False
+
     def to_jinja(self) -> str:
+        prefix = ""
+        postfix = ""
+        if self.brackets:
+            prefix = "("
+            postfix = ")"
         if self.type in [
             FunctionType.EQUALS,
             FunctionType.NEQUALS,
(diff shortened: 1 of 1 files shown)


ramalama/model_store/go2jinja.py
```diff
@@ -229,18 +246,16 @@ def parse_go_template(content: str) -> list[Node]:
 
         end_pos = content.find(GO_SYMBOL_CLOSE_BRACKETS, start_pos) + len(GO_SYMBOL_CLOSE_BRACKETS)
         if end_pos == -1:
-            raise IndexError("Found opening without closing brackets")
+            raise Exception("Found opening without closing brackets")
```
rhatdan, line 249: Why did you change this to a generic Exception?

olliewalsh: Wasn't me - https://github.com/engelmi/go2jinja/pull/5. The copy in ramalama has diverged from the upstream version.

I'm not sure why ramalama is overloading these standard exception types such as IndexError and KeyError. I've never seen that done before and can't think of a valid reason to. Fixing this is a job for another day though...

For now I'll raise specific exception types in go2jinja, which ramalama can catch and reraise as IndexError etc..

engelmi: Yes, its a bit tricky to keep this in sync - missed the previous PR(s) doing changes here, unfortunately.
Both exception types are fine for me. It will bubble up to the call in ramalama where we handle all types equally... by just ignoring it :grin: (since its optional)

engelmi (approved):
LGTM

rhatdan:
/ok-to-test
Is this still draft, or can we merge?

engelmi:
> /ok-to-test Is this still draft, or can we merge?

I think so. Lets merge and if something is missing, we can tackle that in a follow-up PR - I hope this is fine @olliewalsh

PR #1861, opened 2025-08-20
Fix python environment in ci jobs

olliewalsh:
Was installing uv 2 different ways (github action and make install-requirement).
Use `uv run` explicitly instead of activating the virtualenv in the uv github action (it's docs advise not to use this option).

## Summary by Sourcery

Streamline CI workflows by switching to explicit uv run commands for tasks, consolidate dependency installation steps, and refactor Makefile with install-uv and man-check targets.

Enhancements:
- Add install-uv phony target and refactor install-requirements in the Makefile
- Introduce man-check phony target and include it in the validate make workflow

CI:
- Remove uv environment activation from GitHub Actions and use uv run to invoke make targets in all CI jobs
- Standardize apt-get installs across CI workflows by removing python3-argcomplete and adding podman and pipx where needed
- Add a new ‘Run man check’ step in CI to invoke the man-check make target

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,26 +13,28 @@ jobs:
     timeout-minutes: 20
     steps:
       - uses: actions/checkout@v5
-      
-      - name: Install the latest version of uv and activate the environment
+
+      - name: Install the latest version of uv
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

PR #1876, opened 2025-08-28
Changed from google-chrome to firefox

AlexonOliveiraRH:
The camera-demo.html app is failing when used with Google Chrome due to some restrictions with permissions, so Firefox will work better. Also, the path given was not working either, so it was changed to point to the current directory instead.

## Summary by Sourcery

Update docs/demo script to launch camera-demo in Firefox using the correct local path

Bug Fixes:
- Replace google-chrome invocation with firefox for camera-demo
- Fix camera-demo.html path to use the current directory

### DIFF
--- a/docs/demo/ramalama.sh
+++ b/docs/demo/ramalama.sh
@@ -176,7 +176,7 @@ multi-modal() {
     echo ""
 
     echo_color "Use web browser to show interaction"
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

rhatdan:
Thanks @AlexonOliveiraRH

PR #1881, opened 2025-08-29
Don't print llama stack api endpoint info unless --debug *is* passed

booxter:
I think the logic here should be in reverse: print debug info about API
endpoint used only when --debug is passed; not vice versa.

## Summary by Sourcery

Bug Fixes:
- Correct logic to suppress API endpoint information unless the debug flag is enabled


rhatdan:
I Agree.  LGTM

PR #1888, opened 2025-09-02
Fix linting of shell commands

rhatdan:


engelmi:
@rhatdan The fix for the shellcheck is included in #1879 (which LGTM). Can we close this one in favor of #1879?

rhatdan:
Ok merged the other one.

PR #1903, opened 2025-09-05
Improve NVIDIA CDI check.

jwieleRH:
Eliminate the superfluous load_cdi_yaml() function and call yaml.safe_load() directly.

Fix exception handling: log a debug message on failure to read or parse the CDI configuration.

Eliminate the test for the deleted load_cdi_yaml() function.

Test load_cdi_config() with realistic YAML and JSON data extracted from actual CDI configuration files generated by nvidia-ctk.

Test load_cdi_config() with a broken CDI configuration file.

## Summary by Sourcery

Simplify and harden CDI configuration loading by removing the custom YAML parser, inlining yaml.safe_load/json.load, adding debug logging on failures, and making find_in_cdi resilient to malformed configs. Enhance tests with realistic and broken CDI examples and remove obsolete test code.

Bug Fixes:
- Fix exception handling in load_cdi_config to catch and log YAML/JSON read or parse errors

Enhancements:
- Remove deprecated load_cdi_yaml and inline yaml.safe_load/json.load in load_cdi_config
- Log debug messages on failures to read or parse CDI config files and continue searching
- Harden find_in_cdi to handle invalid config structures by logging errors and treating all devices as unconfigured
- Prioritize /var/run/cdi over /etc/cdi when searching for CDI specifications

Tests:
- Add parametrized tests for load_cdi_config using realistic YAML and JSON fixtures
- Add tests for broken CDI configuration in find_in_cdi and remove obsolete load_cdi_yaml tests

### DIFF
--- a/ramalama/common.py
+++ b/ramalama/common.py
@@ -13,7 +13,7 @@ import shutil
 import string
 import subprocess
 import sys
-from collections.abc import Callable, Iterable
+from collections.abc import Callable
 from functools import lru_cache
 from typing import TYPE_CHECKING, Literal, Protocol, TypeAlias, TypedDict, cast, get_args
 
@@ -243,26 +243,6 @@ class CDI_RETURN_TYPE(TypedDict):
     devices: list[CDI_DEVICE]
 
 
-def load_cdi_yaml(stream: Iterable[str]) -> CDI_RETURN_TYPE:
-    # Returns a dict containing just the "devices" key, whose value is
-    # a list of dicts, each mapping the key "name" to a device name.
-    # For example: {'devices': [{'name': 'all'}]}
-    # This depends on the key "name" being unique to the list of dicts
-    # under "devices" and the value of the "name" key being on the
-    # same line following a colon.
-
-    data: CDI_RETURN_TYPE = {"devices": []}
-    parsed = yaml.safe_load(stream) or {}
-    devices = parsed.get("devices") or []
-    for device in devices:
-        if not isinstance(device, dict):
-            continue
-        for key, value in device.items():
-            if key == "name":
-                data['devices'].append({'name': value})
-    return data
-
-
 def load_cdi_config(spec_dirs: list[str]) -> CDI_RETURN_TYPE | None:
     # Loads the first YAML or JSON CDI configuration file found in the
     # given directories."""
@@ -275,18 +255,16 @@ def load_cdi_config(spec_dirs: list[str]) -> CDI_RETURN_TYPE | None:
                 if ext in [".yaml", ".yml"]:
                     try:
                         with open(file_path, "r") as stream:
-                            return load_cdi_yaml(stream)
-                    except OSError:
+                            return yaml.safe_load(stream)
+                    except (OSError, yaml.YAMLError) as e:
+                        logger.warning(f"Failed to load YAML file {file_path}: {e}")
                         continue
                 elif ext == ".json":
                     try:
                         with open(file_path, "r") as stream:
                             return json.load(stream)
-                    except json.JSONDecodeError:
-                        continue
-                    except UnicodeDecodeError:
-                        continue
-                    except OSError:
+                    except (OSError, json.JSONDecodeError, UnicodeDecodeError) as e:
+                        logger.warning(f"Failed to load JSON file {file_path}: {e}")
                         continue
     return None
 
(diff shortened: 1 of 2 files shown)


ramalama/common.py
```diff
@@ -273,18 +253,16 @@ def load_cdi_config(spec_dirs: list[str]) -> CDI_RETURN_TYPE | None:
                 if ext in [".yaml", ".yml"]:
                     try:
                         with open(file_path, "r") as stream:
-                            return load_cdi_yaml(stream)
-                    except OSError:
+                            return yaml.safe_load(stream)
+                    except (OSError, yaml.YAMLError) as e:
+                        logger.debug(f"Failed to read YAML file {file_path}: {e}")
```
rhatdan: If there is a YAMLERROR, shouldn't we at least warn?  In the current code doesn't this throw cause the application to exit with an error?

ramalama/common.py
```diff
@@ -293,9 +271,14 @@ def find_in_cdi(devices: list[str]) -> tuple[list[str], list[str]]:
...
     # unconfigured devices.
-    cdi = load_cdi_config(['/etc/cdi', '/var/run/cdi'])
-    cdi_devices = cdi.get("devices", []) if cdi else []
-    cdi_device_names = [name for cdi_device in cdi_devices if (name := cdi_device.get("name"))]
+    cdi = load_cdi_config(['/var/run/cdi', '/etc/cdi'])
+    try:
+        cdi_devices = cdi.get("devices", []) if cdi else []
+        cdi_device_names = [name for cdi_device in cdi_devices if (name := cdi_device.get("name"))]
+    except (AttributeError, KeyError, TypeError) as e:
+        logger.debug(f"Unable to process CDI configuration: {e}")
```
rhatdan: We are hiding the exception here, shouldn't we at least warn about this situation?

jwieleRH: These are good questions. How nuanced do we want to be? If there is no CDI file, or if a requested device is not listed in it, we print a message and continue with CPU. If there appears to be a CDI file, but it is unreadable, or is not valid YAML or JSON, should we log a warning message and continue with CPU or bail?
There are other places where we call external programs and are, shall we say, casual about non-success results. Any error from nvidia-smi is treated as no nvidia devices found. Similarly with npu-smi and mthreads-gmi. If we get an error trying to run "podman machine" we silently swallow the error and continue.

rhatdan: Right, which is why I would warn here.  I just figure someone with a nvidia GPU might be getting bad performance and not realize that his config file is misconfigured.  If we warn about it, they have the opportunity to check it.

ramalama/common.py
```diff
@@ -275,18 +255,16 @@ def load_cdi_config(spec_dirs: list[str]) -> CDI_RETURN_TYPE | None:
                 if ext in [".yaml", ".yml"]:
                     try:
                         with open(file_path, "r") as stream:
```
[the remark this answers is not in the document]
jwieleRH: Probably a good suggestion for  a future change, but out of scope for this change.

ramalama/common.py
```diff
@@ -275,18 +255,16 @@ def load_cdi_config(spec_dirs: list[str]) -> CDI_RETURN_TYPE | None:
...
-                            return load_cdi_yaml(stream)
-                    except OSError:
+                            return yaml.safe_load(stream)
+                    except (OSError, yaml.YAMLError) as e:
+                        logger.warning(f"Failed to load YAML file {file_path}: {e}")
                         continue
                 elif ext == ".json":
                     try:
                         with open(file_path, "r") as stream:
```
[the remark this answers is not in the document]
jwieleRH: Probably a good suggestion for  a future change, but out of scope for this change.

ramalama/common.py
jwieleRH: Logging change to warning.

test/unit/test_common.py
```diff
@@ -239,12 +349,26 @@ def test_load_cdi_config(mock_walk):
...
+    [
+        (["all"], [], ["all"]),
+        (["0", "all"], [], ["0", "all"]),
+        ([CDI_GPU_UUID, "all"], [], [CDI_GPU_UUID, "all"]),
+    ],
+)
+@patch("builtins.open", mock_open(read_data="asdf\n- ghjk\n"))
+@patch("os.walk", return_value=(("/etc/cdi", None, ("nvidia.yaml",)),))
+def test_find_in_cdi_broken(mock_walk, visible, conf, unconf):
+    assert find_in_cdi(visible) == (conf, unconf)
```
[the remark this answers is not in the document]
jwieleRH: I believe this does adequately test the try...except block  because it tests the results of executing the except clause for four examples of invalid yaml.

rhatdan:
LGTM

PR #1907, opened 2025-09-07
Reorganize transports and add new rlcr transport option

ieaves:
## Summary by Sourcery

Reorganize the codebase by moving all transport logic into a dedicated ramalama/transports package and extracting OCI helper functions into an oci_tools module. Introduce a new RLCR transport for rlcr:// model registries, update the ModelFactory and CLI to support it, add new --add-to-unit and --thinking options, and refresh documentation, configuration, and tests to reflect these changes.

New Features:
- Introduce an RLCR transport (RamalamaContainerRegistry) for models hosted at rlcr:// registries
- Add new CLI options: --add-to-unit for unit file customization and --thinking flag for reasoning models

Enhancements:
- Reorganize transport implementations into ramalama/transports package and factor out OCI helpers into oci_tools module
- Extend ModelFactory to recognize and instantiate the rlcr transport and centralize imports to use ramalama.transports
- Simplify config default loading in tests by mocking file and environment config loaders

Documentation:
- Propagate rlcr transport across all command docs and README and update documentation for new CLI flags, config search paths, TMPDIR, VLLM runtime, and bump default version to 0.12.1

Tests:
- Add a suite of unit tests for RLCR transport initialization, model file discovery, mount setup, and integration
- Update existing unit tests to use new transport import paths and improved config mocking

### DIFF
--- a/container-images/scripts/build_rag.sh
+++ b/container-images/scripts/build_rag.sh
@@ -62,9 +62,9 @@ rag() {
 }
 
 to_gguf() {
-    # required to build under GCC 15 until a new release is available, see https://github.com/google/sentencepiece/issues/1108 for details
-    export CXXFLAGS="-include cstdint"
-    ${python} -m pip install "numpy~=1.26.4" "sentencepiece~=0.2.0" "transformers>=4.45.1,<5.0.0" git+https://github.com/ggml-org/llama.cpp#subdirectory=gguf-py "protobuf>=5.27.2,<6.0.0"
+    ${python} -m pip install "numpy~=1.26.4" "sentencepiece~=0.2.1" "transformers>=4.45.1,<5.0.0" \
+              "git+https://github.com/ggml-org/llama.cpp#subdirectory=gguf-py" \
+              "protobuf>=5.27.2,<6.0.0" "mistral-common>=1.8.3"
 }
 
 main() {


ramalama/oci_tools.py
```diff
@@ -0,0 +1,121 @@
...
+        (
+            '{"name":"oci://{{ .Repository }}:{{ .Tag }}","modified":"{{ .CreatedAt }}",        "size":{{ .VirtualSize'
+            ' }}, "ID":"{{ .ID }}"},'
+        ),
+    ]
+    output = run_cmd(conman_args).stdout.decode("utf-8").strip()
+    if output == "":
+        return []
+
+    manifests = json.loads("[" + output[:-1] + "]")
```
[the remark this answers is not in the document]
rhatdan: Please make this change.

ramalama/transports/rlcr.py
```diff
@@ -0,0 +1,104 @@
...
+        self._model_type = 'oci'
+        self._src_image_path = find_model_file_in_image(self.conman, self.model)
+        self._model_filename = os.path.basename(self._src_image_path)
+        self._model_entrypoint = os.path.join(MNT_DIR, self._model_filename)
+
+    def _get_entry_model_path(self, *args, **kwargs) -> str:
+        return self._model_entrypoint
+
+    def _build_docker_mount_command(self) -> str:
+        """Builds"""
```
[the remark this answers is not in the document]
rhatdan: Seems like a good suggestion.

ramalama/transports/rlcr.py
```diff
@@ -0,0 +1,104 @@
...
+                "-x",
+                "-p",
+                "-f",
+                "-",
+                f"models/{self._model_filename}",
+            ]
+
+            with (
+                subprocess.Popen(export_cmd, stdout=subprocess.PIPE) as p_out,
+                subprocess.Popen(untar_cmd, stdin=p_out.stdout) as p_in,
```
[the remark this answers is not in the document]
rhatdan: Can you use the current run_cmd for this?

rhatdan:
What is rlcr?  Also please document this in the man pages.

rhatdan:
@olliewalsh @engelmi PTAL

ramalama/transports/__init__.py
```diff
@@ -0,0 +1,18 @@
+from .huggingface import Huggingface, HuggingfaceRepository
+from .modelscope import ModelScope, ModelScopeRepository
+from .oci import OCI
+from .ollama import Ollama, OllamaRepository
+from .rlcr import RamalamaContainerRegistry
+from .url import URL
+
+__all__ = [
```
olliewalsh, line 8: Is this used anywhere?

ieaves:
Sorry @rhatdan let me mark this as a draft. I only opened the PR so I could see the CI test outputs. I can't run some of the tests (like manpage validation) locally.

ramalama/transports/rlcr.py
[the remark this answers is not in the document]
ieaves: Curious about your general thoughts here Dan. I was trying to address the current docker [limitations](https://github.com/containers/ramalama/blob/2f9e143e401ba33f36828325da820901aa02ed93/ramalama/model.py#L362) which prevent us from mounting OCI models from image mounts in docker.

With respect to the actual implementation I had initially written to a temporary file on host as sourcery suggested so I don't have strong opinions either way. If the model objects were smaller I would bias towards the temporary file option but given the potential size thought minimizing intermediary copies would be the way to go.

ramalama/transports/__init__.py
ieaves: The only place I can see URL being used is a `create_url` method on ModelFactory.

olliewalsh: I meant is the module `__init__.py` used anywhere? I don't see anything importing it

ramalama/transports/rlcr.py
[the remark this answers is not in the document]
rhatdan: I am fine with the change, but I would rather see this handled by doing it in Common either extending the run_cmd or adding a new command.

test/unit/test_rlcr.py
```diff
@@ -0,0 +1,231 @@
...
+            patch(
+                'ramalama.transports.rlcr.run_cmd', side_effect=subprocess.CalledProcessError(1, "inspect")
+            ) as mock_run,
+            patch('warnings.warn') as mock_warn,
+        ):
+            result = find_model_file_in_image("docker", "rlcr.io/ramalama/test-model")
+
+            assert result == "/models/model.file"
+            assert mock_run.call_count == 1
+            mock_warn.assert_called_once_with(
```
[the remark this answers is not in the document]
ieaves: String isn't even a valid return type from `run_cmd`.

ramalama/transports/__init__.py
ieaves: Oh, sorry, misunderstood the question.

I'm in the habit of including an init on submodules because it gives you control over the default exports from `__all__` but it's otherwise not strictly required.

ieaves:
> What is rlcr? Also please document this in the man pages.

It's the Ramalama Labs Container Registry ([rlcr](https://registry.ramalama.com/projects/ramalama)). We are building out a collection of OCI artifacts including raw model artifacts (e.g. gguf), scratch compiled runtime distroless source images (for example, the llama server [cpu](https://registry.ramalama.com/projects/ramalama/repositories/ramalama%2Fllamacpp-cpu-distroless) image is only 20 MB), and runnable models.

rhatdan:
Ok, OCI Artifacts sound good to me.  Hopefully we can standardize on this along with docker models.  I have just started adding "artifact" support to ramalama convert as well.

ieaves:
@rhatdan the comment was marked as resolved but I could move the docker variant image mount implementation over onto the OCI Model class so that everyone can benefit. You mentioned wanting to pull the `Popen` code into a reusable component under common. I'm not sure what level of granularity you were considering for that but I think the closest to a reusable element I can imagine is something along the lines of.

```python
def populate_volume_from_image(model: Model) -> str:
    ...
    return volume_name
```

EDIT: Went ahead and switched to a common implementation.

ieaves:
Are the bats tests correct? I can't seem to reproduce the specific test failures locally. Likewise for the konflux builds.

olliewalsh:
> Are the bats tests correct? I can't seem to reproduce the specific test failures locally.

I can reproduce the bats-nocontainer failure on main so doesn't look related to this PR

>  Likewise for the konflux builds.

Probably due to https://status.testing-farm.io/issues/2025-09-08-baseos-ci-is-down/

test/unit/test_common.py
```diff
@@ -248,3 +249,108 @@ def test_find_in_cdi(mock_walk, visible, conf, unconf):
...
+        assert result.startswith("ramalama-models-")
+
+        assert mock_run_cmd.call_count >= 3
+        assert mock_popen.call_count == 2
+
+    @patch('subprocess.Popen')
+    @patch('ramalama.common.run_cmd')
+    def test_populate_volume_export_failure(self, mock_run_cmd, mock_popen, mock_model):
+        """Test handling of export process failure"""
+        mock_run_cmd  # satisfies the linter
```
olliewalsh: I expect using `_` for the kwarg name would satisfy the linter

ieaves: I actually need the `mock_run_cmd` fixture to be imported so that run_cmd is replaced with the mock. That way, it doesn't actually attempt to invoke docker / podman. The linter gets angry that `mock_run_cmd` was defined but unused.

olliewalsh: Yes, just need to name it `_`  to indicate that it is unused i.e:
```
    @patch('subprocess.Popen') 
    @patch('ramalama.common.run_cmd')
    def test_populate_volume_export_failure(self, _, mock_popen, mock_model):
        """Test handling of export process failure"""
        output_filename = "model.gguf"
        ...
```
This passed unit-tests and lint

rhatdan:
@olliewalsh any idea how to fix the smollm failure.  I built a ramalama image with the latest llama.cpp and it seems to be working fine.  Is this just a bad release of llama.cpp?  Are we building the latest llama.cpp release in our --nocontainer tests or are we just pip installing the latest on in PIP?

olliewalsh:
> @olliewalsh any idea how to fix the smollm failure. I built a ramalama image with the latest llama.cpp and it seems to be working fine. Is this just a bad release of llama.cpp? Are we building the latest llama.cpp release in our --nocontainer tests or are we just pip installing the latest on in PIP?

We are building the latest llama.cpp AFAICT - https://github.com/containers/ramalama/blob/main/.github/workflows/ci.yml#L166. I'll try to reproduce on a ubuntu 24.04 VM...

Lets move this to https://github.com/containers/ramalama/issues/1908

rhatdan:
I believe the smollvm test is finally fixed, rebase and repush and the test should clear.

ieaves:
@rhatdan this should be ready to go if we are good to merge. I added a small fix to the packit builds so tests would run. It looks like the version detection logic needed a fix after a transition to dynamic version builds.

ramalama/model.py
```diff
@@ -356,14 +355,19 @@ def exec_model_in_container(self, cmd_args, args):
...
-
         if self.model_type == 'oci':
             if not self.engine.use_podman:
-                raise NotImplementedError("Serving OCI models via image mount is only supported with Podman.")
-            self.engine.add([f"--mount=type=image,src={self.model},destination={MNT_DIR},subpath=/models,rw=false"])
+                output_filename = self._get_entry_model_path(args.container, True, args.dryrun)
+                volume = populate_volume_from_image(self, os.path.basename(output_filename))
+                mount_cmd = f"--mount=type=volume,src={volume},dst={MNT_DIR},readonly"
+            else:
+                mount_cmd = f"--mount=type=image,src={self.model},destination={MNT_DIR},subpath=/models,rw=false"
```
engelmi: NIT: The statement in the else branch could be moved  before the `if not self.engine.use_podman`. I try to avoid `else` if possible.

engelmi (commented):
In general, the PR LGTM.

I added a small nit (optional, up to you if you want to change). 
Otherwise, I am wondering about the relation between the Model class and HF, OCI, etc. The inheritance implies a `is-a` relation, which kind of contradicts the directory name `transports` of these child classes. Thinking of these classes as "transports" (or maybe "sources"?) seems to fit better than as a "model". 
What do you think of moving the `model.py` and `model_factory.py` to transports as well and renaming them accordingly? e.g. `model.py -> base.py` and `class Model -> class Transport`. @ieaves 
Could also be done in a follow-up, although it would fit well in this PR.

ieaves:
> What do you think of moving the `model.py` and `model_factory.py` to transports as well and renaming them accordingly? e.g. `model.py -> base.py` and `class Model -> class Transport`. @ieaves Could also be done in a follow-up, although it would fit well in this PR.

To be totally honest I was sort of grasping in the dark on ideal nomenclature because I'm not 100% certain about the division between Model, ModelFactory, and transport.

Model Factories are mostly responsible for detecting the class of model on the basis of a user specified transport. That implies a roughly one to one relationship between the concept of a `transport` and the concept of a `model`. Indeed, some "models" like `URL` and `OCI` really can't be anything other than a transport but much of the responsibility defined on models isn't transport specific. 


This is just a long winded way of asking a question though. Are Models just Transports or is there some unintentional coupling between these two classes that needs to be split out further?

ramalama/model.py
ieaves: You had me thinking about it this morning. Should we be defensively coding around potential future engines? Any chance we would be supporting for example [nerdctl](https://github.com/containerd/nerdctl)?

Something like

```python
            if self.engine.use_podman:
                mount_cmd = f"--mount=type=image,src={self.model},destination={MNT_DIR},subpath=/models,rw=false"
            elif self.engine.use_docker:
                output_filename = self._get_entry_model_path(args.container, True, args.dryrun)
                volume = populate_volume_from_image(self, os.path.basename(output_filename))
                mount_cmd = f"--mount=type=volume,src={volume},dst={MNT_DIR},readonly"
            else:
                raise UnsupportedException(f"No compatible oci mount method for engine: {self.engine.args.engine}")
```

ieaves:
Okay, @engelmi take a look at this variant with the proposed renaming convention. There were a few elements I was feeling a little squeamish about, for example, in the (previously) Model object there is a `type` attribute.

```
class Transport(TransportBase):
    """Transport super class"""

    type: str = "Transport"
```
Previously the default value was "Model," I've changed it to Transport which doesn't break any tests. A quick search for `.type == ` doesn't show us ever looking for Model explicitly so I think this is fine but I figured it was worth flagging just in case.

ramalama/model.py
rhatdan: I was asked at a conference this week about us supporting `container` command on a MAC. My answer at this point is that I don't believe `container` has enough functionality to do what docker does, so I am not sure if the basic ramalama commands would work, mounting OCI images into a container for example.

rhatdan:
Lets do the changes to Model versus Transport in a different PR.

Merging.

Issue #1908, opened 2025-09-08
System test failure: run smollm with prompt


engelmi:
Since the latest llama.cpp and whisper update in https://github.com/containers/ramalama/pull/1874, the system test `ramalama run smollm with prompt ` is failing:
```bash
 not ok 18 [030] ramalama run smollm with prompt in 7157ms
# (from function `bail-now' in file test/system/helpers.podman.bash, line 122,
#  from function `die' in file test/system/helpers.podman.bash, line 791,
#  from function `run_ramalama' in file test/system/helpers.bash, line 186,
#  in test file test/system/030-run.bats, line 126)
#   `run_ramalama run --temp 0 ${MODEL} "What is the first line of the declaration of independence?"' failed
#
# [20:28:03.556813208] $ /home/runner/work/ramalama/ramalama/bin/ramalama run --temp 0 smollm:135m What is the first line of the declaration of independence?
# [20:28:09.522170822] Downloading ollama://library/smollm:135m ...
# Trying to pull ollama://library/smollm:135m ...
# Downloading smollm
# 
# user
# What are some of the most interesting or surprising facts about the human brain?Error: list index out of range
# [20:28:09.525441614] [ rc=22 (** EXPECTED 0 **) ]
# #/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
# #| FAIL: exit code is 22; expected 0
# #\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
# # [teardown]
```

engelmi:
I was able to reproduce this in an ubuntu 24.04 (as used in the CI) container and using the `--nocontainer` flag in the ramalama command (which makes sense since it fails only for the bats-nocontainer). 

When experimenting a bit I also found out that this error does not stop the temporarily started llama-server properly:

<img width="1319" height="329" alt="Image" src="https://github.com/user-attachments/assets/24cdcf72-b6f5-4021-9585-693b32073594" />

/cc @rhatdan @olliewalsh

engelmi:
This line seems to be the root cause:
https://github.com/containers/ramalama/blob/main/ramalama/chat.py#L38

I guess the response from llama-server changed in one of the latest releases.

PR #1910, opened 2025-09-08
Bump to latest version of llama.cpp

rhatdan:


engelmi (approved):
LGTM

rhatdan:
Will see if this fixes: https://github.com/containers/ramalama/issues/1908

rhatdan:
@engelmi seems we are only seeing this issue with --nocontainer or is the problem that we have not updated the images with the latest llama.cpp?

PR #1913, opened 2025-09-09
Fix ramalama run with prompt index error

engelmi:
Fixes: https://github.com/containers/ramalama/issues/1908

Fixed an index error when a line of the assistant response does not include a choice field.

## Summary by Sourcery

Bug Fixes:
- Handle missing or empty 'choices' field in assistant response to avoid index errors when parsing streamed data


rhatdan:
LGTM

PR #1915, opened 2025-09-09
Do not set the ctx-size by default

rhatdan:
llama.cpp is defaulting to ctx-size of 4098 and we were hard coding 2049, which means we were not using the default setting. This PR changes to use ctx-size=0 which will not be specified in the command unless the value is > 0, so llama-server default will be used.

Also we hard coded cache_reuse=256 with no way for user to override, this PR adds support for cache_reuse being set in ramalama.conf and on the command line.

## Summary by Sourcery

Enable user-configurable cache reuse and switch to using llama-server’s default context size by default instead of hard-coding a ctx_size

New Features:
- Add --cache-reuse option to the CLI and support cache_reuse configuration in ramalama.conf

Enhancements:
- Set default ctx_size to 0 to defer to llama-server’s built-in default and only emit --ctx-size when explicitly >0

Documentation:
- Add documentation for the cache_reuse option and update ctx-size defaults across man pages and example configs

Tests:
- Update system tests to check for cache-reuse flag and ensure --ctx-size isn’t shown by default

### DIFF
--- a/docs/ramalama-perplexity.1.md
+++ b/docs/ramalama-perplexity.1.md
@@ -29,8 +29,11 @@ URL support means if a model is on a web site or even on your local system, you 
 #### **--authfile**=*password*
 path of the authentication file for OCI registries
 
+#### **--cache-reuse**=256
+Min chunk size to attempt reusing from the cache via KV shifting
+
 #### **--ctx-size**, **-c**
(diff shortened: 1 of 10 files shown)


giuseppe (approved):
LGTM

PR #1928, opened 2025-09-11
--flash-attn requires an option in llama-server now

rhatdan:
## Summary by Sourcery

Enhancements:
- Include "on" argument with --flash-attn flag in llama-server command invocation


rhatdan:
@engelmi PTAL

PR #1930, opened 2025-09-11
s390x: switch to a smaller bigendian model for testing

mikebonnet:
Should improve performance of integration tests.

## Summary by Sourcery

Switch integration tests on big-endian s390x to a smaller test model for improved performance and clean up related configurations.

Enhancements:
- Replace granite-be-3.0:1b with a smaller stories-be:260k model for big-endian tests
- Add stories-be:260k shortname mapping in shortnames.conf
- Remove s390x-specific timeout override in the bench test

Tests:
- Update pull and run system tests to use the new default big-endian model and remove explicit granite model overrides


rhatdan:
LGTM

Issue #1873, opened 2025-08-27
Add Bats test to cover compose support in ramalama serve


abhibongale:
#1839 was merged and during review it was suggested that compose support in ramalama server should be covered in automated tests.

Adding test coverage would help to improve the confidence in the functionality.

# Acceptance Criteria
- Add bats tests to validate compose basic behvaiour.

# Addtional Context
- Test coverage was mentioned in #1839

PR #1839, opened 2025-08-14
Add Docker Compose generator

abhibongale:
This commit introduces the `--generate=compose` option to the `ramalama serve` command, enabling users
to generate a `docker-compose.yaml` file for a given model.

sourcery-ai suggested changes:
1. Add test_genfile_empty_content. This will verify that genfile handles empty input gracefully and avoids unexpected behavior.

2. Supporting condition to handle  'oci://' prefix for RAG source for consistency.

3. updating the parsing logic to support complex port formats, such as those i with IP addresses or protocols, to ensure compatibility with all valid Docker Compose specifications.

4. approach excludes images for other GPU backends like ROCm or custom builds. Please update the check to support a wider range of GPU-enabled images.

fixes #184

PR #1934, opened 2025-09-16
Add bats test to cover docker-compose in serve

abhibongale:
During #1839 review we decided that having bats test will be very helpful. This PR adds docker-compose bats tests.

Also fixes the issue of generating compose file with static name.

Fixes: #1873


ramalama/compose.py
```diff
@@ -183,8 +183,8 @@ def generate(self) -> PlainFile:
 
 
 def genfile(name: str, content: str) -> PlainFile:
-    file_name = "docker-compose.yaml"
-    print(f"Generating Docker Compose file: {file_name}")
+    file_name = f"{name}.yaml"
+    print(f"Generating Docker Compose YAML file: {file_name}")
```
rhatdan: Can we just say Generating Compose YAML?

ramalama/compose.py
```diff
@@ -183,8 +183,8 @@ def generate(self) -> PlainFile:
 
 
 def genfile(name: str, content: str) -> PlainFile:
-    file_name = "docker-compose.yaml"
-    print(f"Generating Docker Compose file: {file_name}")
+    file_name = f"{name}.yaml"
```
engelmi: This broke the unit tests in `test/unit/test_compose.py`. Please update them.

engelmi (commented):
Small comment (see below), otherwise LGTM

rhatdan:
LGTM

PR #1942, opened 2025-09-17
Add e2e pytest test for bench command

telemaco:
- Add a new e2e test for the `ramalama bench` command
- Introduce a `test_model` fixture to support different architectures
- Skip the test if the `llama-bench` command is not available


telemaco:
@engelmi am I doing something wrong with failed pipelines?

engelmi:
> @engelmi am I doing something wrong with failed pipelines?

I don't think so. According to the [status of Testing Farm](https://status.testing-farm.io/) there are some issues since a few days - and packit basically just forwards to TF. So the failures are unrelated and not blocking.
Edit: Yes, restarting the packit pipelines results in the same error. I guess its really a TF issue.

test/e2e/test_bench.py
```diff
@@ -0,0 +1,13 @@
+import re
+from test.conftest import skip_if_no_llama_bench
+from test.e2e.utils import check_output
+
+import pytest
+
+
+@pytest.mark.e2e
+@skip_if_no_llama_bench
+def test_model_and_size_columns(test_model):
```
engelmi, line 10: If I understand this correctly, this is the rewrite for the `002-bench.bats`, right? Do we have an overview which tests are being ported already and which still needs to? If not, lets use a simple checklist in some GH issue for this. 
@telemaco Is there already an issue for the pytest rewrite? If not, I'd create one.

engelmi (approved):
LGTM, just a comment about keeping track of the work done.

test/e2e/test_bench.py
telemaco: @engelmi Makes sense to me. There isn't any issue about it. I could create an issue later to track all the migration if that works for you.

engelmi: That would be great, thanks!

rhatdan: Once we have the tests moved, then we can start to remove the bats test.

rhatdan:
LGTM

Issue #1872, opened 2025-08-27
Add documentation for compose support in `--help` and man pages


abhibongale:
#1839 was closed recently and it was noted that compose functionality which is now supported lacks documentation. It should document the feature in the `ramalama server --help` and in the associated man pages.

This will allow user to understand how to use the compose feature with the serve command.

# Accpetance Criteria:
- Document in the man page detailing how compose is used with serve with relevant flags and scenario.
- Ensure that `ramalama server --help` mentions compose support explicitly.

# Additional Context:
- Lack fo documentation was mentioned in the #1839.
- Feature has poor visibility due to lack of documentation.

PR #1943, opened 2025-09-18
Add Compose docs

abhibongale:
Added the docker-compose.yaml example and updated the type option table with compose option.

fixes: #1872

## Summary by Sourcery

Document Compose file generation support in the CLI man page

Documentation:
- Add 'compose' as a supported output type in the generate option table
- Include a detailed example of generating a docker-compose.yaml file with ramalama serve


engelmi (approved):
LGTM

PR #1946, opened 2025-09-18
feat: added new flag default-template for models to use tools

bmahabirbu:


rhatdan:
Would it make more sense to allow the user to specify a chat template --chat-templat /tmp/chat.template?  And then have --chat-template none or --chat-template default?

engelmi:
> Would it make more sense to allow the user to specify a chat template --chat-templat /tmp/chat.template? And then have --chat-template none or --chat-template default?

Yes, I think having a `--chat-template-file <path>` would be great and aligns with what llama-server does. This RamaLama CLI option would have the highest priority, followed by the extracted chat template and, finally, using the inference engines default template as a fallback. I think we discussed this at one point, but lost track of it.

rhatdan:
When running within the container, the --chat-template option would have to volume mount the path into the container.  This would complicate the use of quadlets and kube.yaml, but for now lets just add this and we would have to point out that this would need ot be handled within an image if a user put the AI into production.  Potentially having the user ship the template within the container.  --chat-template=none would just remove the --chat-template option from the inference engine.

rhatdan:
@bmahabirbu Should this PR be closed or are you still working on it?

Issue #1882, opened 2025-08-29
Incorrect chat template file used for llama3.2 model


booxter:
### Issue Description

This looks similar to https://github.com/containers/ramalama/issues/1855 but it doesn't reproduce for "tiny" (but it still happens for "llama3.2") and I get the error on 0.21.1 that includes the fix.

```
$ ramalama --debug serve llama3.2
[...]
common_init_from_params: added <|end_of_text|> logit bias = -inf
common_init_from_params: added <|eom_id|> logit bias = -inf
common_init_from_params: added <|eot_id|> logit bias = -inf
common_init_from_params: setting dry_penalty_last_n to ctx_size = 2048
common_chat_templates_init: failed to parse chat template (defaulting to chatml): Failed to parse number: '.' ([json.exception.parse_error.101] parse error at line 1, column 1: syntax error while parsing value - invalid literal; last read: '.') at row 5, column 21:

{{ if .System }}{{ .System }}
                    ^
{{- end }}

srv          init: initializing slots, n_slots = 1
slot         init: id  0 | task -1 | new slot n_ctx_slot = 2048
slot        reset: id  0 | task -1 |
main: model loaded
main: chat template, chat_template: {%- for message in messages -%}
  {{- '<|im_start|>' + message.role + '
' + message.content + '<|im_end|>
' -}}
{%- endfor -%}
{%- if add_generation_prompt -%}
  {{- '<|im_start|>assistant
' -}}
{%- endif -%}, example_format: '<|im_start|>system
You are a helpful assistant<|im_end|>
<|im_start|>user
Hello<|im_end|>
<|im_start|>assistant
Hi there<|im_end|>
<|im_start|>user
How are you?<|im_end|>
<|im_start|>assistant
'
```

@olliewalsh FYI

### Steps to reproduce the issue

`ramalama --debug serve llama3.2`

### Describe the results you received

.

### Describe the results you expected

.

### ramalama info output

```yaml
{
    "Accelerator": "none",
    "Config": {},
    "Engine": {
        "Info": {
            "Client": {
                "APIVersion": "5.6.0",
                "Built": 315532800,
                "BuiltTime": "Mon Dec 31 19:00:00 1979",
                "GitCommit": "",
                "GoVersion": "go1.24.5",
                "Os": "darwin",
                "OsArch": "darwin/arm64",
                "Version": "5.6.0"
            },
            "host": {
                "arch": "arm64",
                "buildahVersion": "1.40.1",
[...]

PR #1947, opened 2025-09-18
Template conversion system and messages template wrapping for Go templates

ieaves:
## Summary by Sourcery

Implement a template conversion subsystem with typed JSON DTO definitions and integrate it into the model store

New Features:
- Add a template_conversion module to convert and normalize chat templates and wrap them in an OpenAI messages loop
- Integrate convert_template into the model store to automatically transform legacy Go templates into Jinja format
- Define TypedDict classes for request/response DTOs and update to_dict methods to return typed dicts

Enhancements:
- Add type hints and unify list formatting in model, chat, amdkfd, and daemon handler modules

Tests:
- Add unit tests for wrap_template_with_messages_loop covering various template structures

### DIFF
--- a/ramalama/model_store/store.py
+++ b/ramalama/model_store/store.py
@@ -6,11 +6,11 @@ from http import HTTPStatus
 from pathlib import Path
 from typing import Optional, Sequence, Tuple
 
-from ramalama.model_store import go2jinja
 from ramalama.common import perror, sanitize_filename, verify_checksum
 from ramalama.endian import EndianMismatchError, get_system_endianness
 from ramalama.logger import logger
 from ramalama.model_inspect.gguf_parser import GGUFInfoParser, GGUFModelInfo
+from ramalama.model_store import go2jinja
 from ramalama.model_store.constants import DIRECTORY_NAME_BLOBS, DIRECTORY_NAME_REFS, DIRECTORY_NAME_SNAPSHOTS
 from ramalama.model_store.global_store import GlobalModelStore
 from ramalama.model_store.reffile import RefJSONFile, StoreFile, StoreFileType, migrate_reffile_to_refjsonfile
--- a/ramalama/model_store/template_conversion.py
+++ b/ramalama/model_store/template_conversion.py
@@ -1,6 +1,8 @@
-from jinja2 import Environment, meta
 import re
 from functools import lru_cache
+
+from jinja2 import Environment, meta
+
 from ramalama.model_store import go2jinja
 
 


ieaves:
@rhatdan @engelmi @olliewalsh This PR is a follow-on proposal from the discussion in #1882. It does two things

1. Introduces a framework for converting between template structures
2. Adds a simple for loop wrapper around go converted templates when the output template is missing a messages loop.

This is just a direction proposal we can play with. In my local testing it seemed to resolve the chat issues I was seeing with Ollama models. I haven't run it through an exhaustive set of tests across all of the model variants though.

rhatdan:
@engelmi @olliewalsh PTAL

ieaves:
Hey @engelmi and @olliewalsh any thoughts on this?

ramalama/model_store/template_conversion.py
```diff
@@ -0,0 +1,101 @@
...
+
+class OllamaTemplateStyle(TemplateStyle):
+    @singledispatchmethod
+    def convert(self, target_style: BaseStyle):
+        super().convert(target_style)
+
+    @convert.register
+    def _(self, target_style: OpenAIStyle) -> str:
+        template = go2jinja.go_to_jinja(self.template)
+        if "messages" not in template or "for message in messages" not in template:
```
olliewalsh: since https://github.com/containers/ramalama/pull/1959 we have jinja2 so could now use https://jinja.palletsprojects.com/en/stable/api/#jinja2.meta.find_undeclared_variables to check if messages is referenced in the template

ramalama/daemon/dto/model.py
```diff
@@ -1,5 +1,15 @@
 import json
 from dataclasses import dataclass
+from typing import TypedDict
```
olliewalsh: are all of these typing change unrelated?

ieaves: Yeah unrelated. I accidentally started on a separate branch begun while on an airplane working on things I could poke at without internet.

ramalama/model_store/template_conversion.py
ieaves: Great idea. Implemented!

ieaves:
What other template styles have yall been seeing need for? Harmony for the gpt-oss release?

olliewalsh:
> What other template styles have yall been seeing need for? Harmony for the gpt-oss release?

I'm not sure if we need to do anything special for harmony. IIUC for the Hugging Face gpt-oss llama.cpp should just use the embedded jinja template.
For the ollama gpt-oss with it's [go template](https://ollama.com/library/gpt-oss:latest/blobs/fa6710a93d78) ... I'm not 100% sure yet. May "just work" if we can get go template conversion to handle it (I expect it's too complex for the current logic, but I'm working on this).

ramalama/model_store/template_conversion.py
```diff
@@ -0,0 +1,99 @@
...
+            template = wrap_template_with_messages_loop(template)
+        return template
```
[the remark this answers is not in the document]
ieaves: If template conversion fails then the resulting template is either definitionally invalid, in which case it would break the users ability to interact with the llm, or a legitimate bug in which case it needs to be identified and fixed. Reciprocally, if the caller has alternative means of dealing with an invalid template they can implement exception handling as needed.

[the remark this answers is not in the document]
olliewalsh: could maybe do something like
```
class TemplateConversionError(Exception):
    pass
...
...
try:
   ...
except Exception as e:
    raise TemplateConversionError("template conversion failed") from e
```

So the caller would just need to handle TemplateConversionError

ramalama/model_store/store.py
```diff
@@ -230,22 +231,15 @@ def _try_convert_existing_chat_template(self, model_tag: str, snapshot_hash: str
...
 
-            if not go2jinja.is_go_template(chat_template):
-                return True
-
-            try:
-                jinja_template = go2jinja.go_to_jinja(chat_template)
-            except Exception as ex:
-                logger.debug(f"Failed to convert Go Template to Jinja: {ex}")
+            normalized_template = convert_template(chat_template)
+            if normalized_template == chat_template:
```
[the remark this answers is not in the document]
ieaves: Updated to include an explicit `needs_conversion` function rather than converting returning a bool

test/unit/test_template_conversion.py
```diff
@@ -0,0 +1,169 @@
...
+{{ response }}<|end|>"""
+
+        assert result == expected
+
+    def test_empty_template(self):
+        """Test empty template."""
+        input_template = ""
+        result = wrap_template_with_messages_loop(input_template)
+        expected = "{% for message in messages %}{% endfor %}"
+        assert result == expected
```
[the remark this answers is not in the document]
ieaves: I don't see what point whitespace stripping or comment stripping would provide. These templates wouldn't produce anything anyways.

ramalama/model_store/store.py
```diff
@@ -230,22 +231,15 @@ def _try_convert_existing_chat_template(self, model_tag: str, snapshot_hash: str
...
                 return False
```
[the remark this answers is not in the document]
ieaves: I believe this is resolved

ramalama/model_store/template_conversion.py
```diff
@@ -0,0 +1,109 @@
...
+            )
+        wrapped = wrapped.replace(f"{{{{ {role} }}}}", "{{ message.content }}")
+
+    # recombine wrapped template
+    return f"{{% for message in messages %}}{wrapped}{{% endfor %}}{final_assistant_output}"
+
+
+def get_jinja_variables(template: str) -> set[str]:
+    """Returns all variables associated with a jinja template except those explicitly set in the template"""
+    env = Environment()
```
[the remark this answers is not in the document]
ieaves: These templates aren't going to be rendered in a browser. They are simple text

ramalama/model_store/template_conversion.py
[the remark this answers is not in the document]
ieaves: Okay, there are two error types now: `TemplateConversionError` and `TemplateIdentificationError`. These are used by the store [here](https://github.com/containers/ramalama/pull/1947/files#diff-f8f362bdc3adba7f1a41869e7f651afc1d010746db4a5220f4494d4d1cba6a05R237) to continue iteration over potential ref_file.chat_templates. However, I've specifically not included any error handling when these functions are used in `_ensure_chat_template` where the converted output is our last available template strategy.

ieaves:
Okay, sounds good. I'll leave things as they are with these two template styles for now and we can add more as needed in the future!

ieaves:
@olliewalsh what do you think? Is there anything else we should consider adding on this one or is it good to go?

ramalama/model_store/template_conversion.py
```diff
@@ -0,0 +1,146 @@
...
+
+    @convert.register
+    def _(self, target_style: OpenAIStyle) -> str:
+        return self.template
+
+
+class OllamaTemplateStyle(TemplateStyle):
+    style = OllamaStyle()
+
+    @singledispatchmethod
```
olliewalsh: This seems unnecessarily complex. I would have just overridden the convert method in OllamaTemplateStyle, maybe named convert_to_jinja with no target_type.

ramalama/chat.py
```diff
@@ -300,10 +301,7 @@ def default(self, user_content):
         self.request_in_process = False
 
     def _make_request_data(self):
-        data = {
-            "stream": True,
-            "messages": self.conversation_history,
-        }
+        data = {"stream": True, "messages": self.conversation_history}
```
olliewalsh: This is causing a merge conflict now. Maybe better to drop all of these type annotation changes, do them in a follow up PR?

olliewalsh:
> @olliewalsh what do you think? Is there anything else we should consider adding on this one or is it good to go?

Works for me. Needs a rebase though.

Users will still need to remove and re-pull any broken models to rerun the template conversion though. Maybe we can try to resolve that issue in a follow up PR.

ramalama/model_store/template_conversion.py
ieaves: I wrote it this way with the expectation that there are likely to be more than two template styles in the future. This way the API stays clean and there's a single easily readable method converting from one style to each alternative. If it feels overengineered I can switch it out but from a stylistic perspective I find dispatch based approaches produce extremely clean APIs rather than monolithic methods with tons of nested conditionals.

ieaves:
Alright, pulled out the typing and left only the template conversation in. Since this PR is so old I would definitely like to get it in separately from a future template migration though. If you are good to merge this when you see it I'll take a look at a migration PR tomorrow morning.

rhatdan:
@olliewalsh Is this good to go in your opinion?

ramalama/model_store/template_conversion.py
olliewalsh: I do understand and really like the approach, just seems unnecessary until/unless there are multiple template conversions to handle. Fine to leave it in place though.

ramalama/model_store/template_conversion.py
```diff
@@ -0,0 +1,146 @@
...
+
+class StyleHandler:
+    def __init__(self, target_template_style: OpenAITemplateStyle):
+        self.target_template_style = target_template_style
+
+    def get_template_style(self, template: str) -> TemplateStyle:
+        return identify_template_style(template)
+
+    def needs_conversion(self, template_style: TemplateStyle) -> bool:
+        return isinstance(template_style, self.target_template_style)
```
olliewalsh: logic appears to be inverted, missing `not`?

ramalama/model_store/template_conversion.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def convert_template(self, template_style: TemplateStyle) -> str:
+        return template_style.convert(self.target_style)
```
olliewalsh: should be `self.target_template_style`?

ramalama/model_store/template_conversion.py
```diff
@@ -0,0 +1,146 @@
...
+    def _(self, target_style: OpenAIStyle) -> str:
+        return self.template
+
+
+class OllamaTemplateStyle(TemplateStyle):
+    style = OllamaStyle()
+
+    @singledispatchmethod
+    def convert(self, target_style: BaseStyle):
+        super().convert(target_style)
```
olliewalsh: dispatch doesn't seem to be working? When I fix the other issues this is raising `Exception: No supported conversion method OllamaTemplateStyle and type templates`

olliewalsh:
> Works for me.

I must have been mistaken - I guess the conversion was failing, then the embedded chat template was used instead

olliewalsh:
@sourcery-ai review

SourceryAI:
Hi @olliewalsh! 👋

Only authors and team members can run @sourcery-ai commands on public repos.

If you are a team member, install the [@sourcery-ai bot](https://github.com/apps/sourcery-ai/installations/new) to get access ✨

ramalama/model_store/template_conversion.py
olliewalsh: or `self.target_template_style.style`, for the method dispatch to work

ieaves:
@olliewalsh since it was causing issues I simplified to just Go and Jinja per your advice. This only adds a `convert_go_to_jinja` function with the new message wrapping.


I went ahead and switched out the looped replace implementation in the wrapper for a compiled regex as well.

ieaves:
I think this should also resolve loading older template version as well. I've added an additional check in `_try_convert_existing_chat_template` which checks whether we consider the template to be a valid openai template (i.e. has a `messages` variable).

Not 100% sure this is totally foolproof but it seemed better than trying to add migrations.

olliewalsh:
> @olliewalsh since it was causing issues I simplified to just Go and Jinja per your advice. This only adds a `convert_go_to_jinja` function with the new message wrapping.
> 
> I went ahead and switched out the looped replace implementation in the wrapper for a compiled regex as well.

FWIW it was working for me for the ollama tiny model with the fixes that I suggested. Latest commit looks like is should be good too - I'll try with a number of different models tomorrow...

ieaves:
I don't have a strong preference either way and can switch it back. I agree it felt complex for how we are using it right now. Obviously I liked it conceptually but understand both perspectives :).

olliewalsh:
> FWIW it was working for me for the ollama tiny model with the fixes that I suggested. Latest commit looks like is should be good too - I'll try with a number of different models tomorrow...

Sorry about the delay.
Fixes https://github.com/containers/ramalama/issues/2025. `ollama://library/deepseek-r1:1.5b` and `ollama://library/tinyllama` work well.
Still see chat template issues with `ollama://library/smollm:135m` but we can investigate this in a follow-up PR.

@rhatdan looks good to me if you want to merge

rhatdan:
Ok will merge and move to 0.13.0

PR #1955, opened 2025-09-21
Add NVIDIA GPU support to quadlet generation for `ramalama serve`

wang7x:
**Summary:**  
When running `ramalama serve` the tool automatically detects NVIDIA GPUs and mounts the required devices (`/dev/nvidia*`) into the podman container. However, the same command used with the `--generate=quadlet` option produces a `.container` file that lacks the necessary `AddDevice` entries for the GPU devices. As a result, the generated systemd service cannot utilize the GPU, leading to CPU‑only execution or a failure to start the container on NVIDIA hardware.

**What was changed:**  
* Updated the quadlet generation logic in `ramalama-serve` to:
  1. Add the appropriate `AddDevice` entries for all required NVIDIA device nodes.
* Updated unit tests to verify that the generated quadlet file contains the expected GPU device entries.

**Impact:**  
* Users can now generate quadlet services that correctly use NVIDIA GPUs out of the box, without manual editing of the `.container` file.
* The `ramalama serve --generate=quadlet` command behaves consistently with the normal `ramalama serve` command regarding GPU usage.

**Testing:**  
* All unit tests passed.
* Run `ramalama serve --generate=quadlet <model>` on a machine with NVIDIA GPU – the resulting `.container` file contains `AddDevice=nvidia.com/gpu=all` lines.
* Verify that the container starts and the GPU is visible inside (e.g., via `nvidia-smi` inside the container or by checking CUDA runtime success).

### DIFF
--- a/ramalama/quadlet.py
+++ b/ramalama/quadlet.py
@@ -56,6 +56,7 @@ class Quadlet:
         quadlet_file.add("Container", "AddDevice", "-/dev/accel")
         quadlet_file.add("Container", "AddDevice", "-/dev/dri")
         quadlet_file.add("Container", "AddDevice", "-/dev/kfd")
+        quadlet_file.add("Container", "AddDevice", "nvidia.com/gpu=all")
         quadlet_file.add("Container", "Image", f"{self.image}")
         quadlet_file.add("Container", "RunInit", "true")
         quadlet_file.add("Container", "Environment", "HOME=/tmp")
--- a/test/unit/data/test_quadlet/basic/tinyllama.container
+++ b/test/unit/data/test_quadlet/basic/tinyllama.container
@@ -6,6 +6,7 @@ After=local-fs.target
 AddDevice=-/dev/accel
 AddDevice=-/dev/dri
 AddDevice=-/dev/kfd
+AddDevice=nvidia.com/gpu=all
 Image=testimage
 RunInit=true
 Environment=HOME=/tmp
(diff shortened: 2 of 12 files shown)


rhatdan:
I would like to see what happens on a system without Nvidia GPUs.  I believe quadlets would handle a missing /dev/nvidia GPU when not running on an NVIDIA system with the -/dev/nividia option.

wang7x:
> I would like to see what happens on a system without Nvidia GPUs. I believe quadlets would handle a missing /dev/nvidia GPU when not running on an NVIDIA system with the -/dev/nividia option.

On a system without NVIDIA GPUs, the container started and fell back to CPU execution. Performance was noticeably slower.

rhatdan:
Ok so Podman does the right thing and ignores the flag if Nvidia is not available.

You need to sign your commits.

rhatdan:
git commit -a --amend -s
git push --force

rhatdan:
I also think adding the Nvidia check as suggested by the AI makes sense.

wang7x:
> Ok so Podman does the right thing and ignores the flag if Nvidia is not available.
> 
> You need to sign your commits.

I previously misunderstood — running on a machine without an NVIDIA GPU would cause Podman to fail at startup. I’ve now modified it to detect whether a GPU is present, thanks for the reminder.

rhatdan:
For some reason the DCO check does not like your signature.  Could you make sure there is a blank line between the title and your signature, perhaps that is what it is complaining about.

rhatdan:
@wang7x Are you still working on this?

wang7x:
> @wang7x Are you still working on this?

Hey @rhatdan, sorry for the gap — I was on holidays. Back now and will work through the feedback. Thanks for waiting!

rhatdan:
LGTM

PR #1958, opened 2025-09-22
Fix handling of API_KEY in ramalama chat

rhatdan:
First all RAMALAMA environment variables should be prefixed with RAMALAMA, so changing API_KEY to RAMALAMA_API_KEY.

Also add api_key to ramalama.conf config files.


rhatdan:
@bmahabirbu PTAL

rhatdan:
@olliewalsh @mikebonnet @engelmi @bmahabirbu PTAL

engelmi (commented):
We should probably add the new test also to the [test/e2e/test_help.py](https://github.com/containers/ramalama/blob/main/test/e2e/test_help.py) so that we don't miss it when switching to pytest completely.

Otherwise the PR LGTM

bmahabirbu:
Sorry didnt get to check this will test using my demo script

PR #1961, opened 2025-09-23
konflux: fix creation of the PipelineRun when a tag is pushed

mikebonnet:
The `event-type` label isn't created automatically, need to specify it manually.
Use the newer `konflux-integration-runner` service account.

## Summary by Sourcery

Fix the PipelineRun configuration for tag push events by adding the required event-type label and updating the service account

Bug Fixes:
- Add the missing pac.test.appstudio.io/event-type: push label to the release PipelineRun

Enhancements:
- Switch the taskRunTemplate service account to konflux-integration-runner


rhatdan:
LGTM

PR #1976, opened 2025-09-28
feat: support safetensors-only repos across runtimes

kush-gupt:
Fixes #1883 

- Centralize model argument resolution in ModelStore.resolve_model_argument for safetensor only repos
- For repos with .safetensors and no GGUF, pass the repository directory as the model argument (MNT_DIR in container; snapshot dir on host). Otherwise fall back to the GGUF entry path. Honor --dryrun. No change to GGUF flow.
- MLX: fix detection and omit explicit model arg for OpenAI-compatible /v1 endpoints to use the server default ("default_model").
- CI: align mlx-lm installation steps with the real install script.
- Tests: fix runtime argument usage and clarify MLX /v1 endpoint comment.
- Guardrail: limit output for very small models to avoid runaway responses.

Allows for the following to work again:
```
ramalama --runtime=mlx run hf://mlx-community/Unsloth-Phi-4-4bit
```

## Summary by Sourcery

Centralize model argument resolution to support safetensors-only repositories and adapt runtimes accordingly, update CI installation for mlx-lm, and align tests with the new behavior.

New Features:
- Add ModelStore.resolve_model_argument to return directory paths for safetensors-only repos.
- Allow passing repository directory as the model argument when no GGUF files are present.

Enhancements:
- Refactor model.py and transports/base.py to use the new resolution logic for model paths.
- Omit explicit model flag for MLX /v1 endpoints to leverage the server default_model.
- Guardrail small-model outputs to prevent runaway responses.

CI:
- Update CI workflow to install mlx-lm via ‘uv tool install’ instead of pip.

Tests:
- Adjust system tests for runtime-args usage and MLX installation checks.
- Clarify unit test comments for MLX runtime endpoint behavior.

PR #1977, opened 2025-09-29
[CI] smollm:135 is hanging on macos, switch to tiny for now

olliewalsh:
macos CI job has been timing out since https://github.com/containers/ramalama/pull/1916. Switch to tiny model for the prompt test, while investigating the macos issue

## Summary by Sourcery

Switch the prompt test to use the tiny model to avoid CI timeouts on macOS.

Bug Fixes:
- Prevent macOS CI hang by using the tiny model in the prompt test

CI:
- Work around macOS CI job timing out by switching to the tiny model for the prompt test

Tests:
- Update the 'ramalama run with prompt' test to invoke the tiny model instead of the default model

### DIFF
--- a/test/system/030-run.bats
+++ b/test/system/030-run.bats
@@ -126,7 +126,7 @@ EOF
 }
 
 @test "ramalama run with prompt" {
-    run_ramalama run --temp 0 $(test_model ${MODEL}) "What is the first line of the declaration of independence?"
+    run_ramalama run --temp 0 $(test_model tiny) "What is the first line of the declaration of independence?"
 }
 
 @test "ramalama run --keepalive" {


test/system/030-run.bats
```diff
@@ -126,7 +126,7 @@ EOF
 }
 
 @test "ramalama run with prompt" {
-    run_ramalama run --temp 0 $(test_model ${MODEL}) "What is the first line of the declaration of independence?"
+    run_ramalama run --temp 0 $(test_model tiny) "What is the first line of the declaration of independence?"
```
olliewalsh, line 129: @kush-gupt I see this is also fixed this in https://github.com/containers/ramalama/pull/1976 by using ` --runtime-args='-n 128'`. Any preference?

I suspect the root cause is that ramalama run does not use the same setting as the web chat front-end (e.g repeat-penalty).

kush-gupt: This -n limits the output tokens to 128 (and not letting the model go on forever)

The root cause (from what I've seen) is that something has happened to the mac runner to make running models SUPER slow on it (likely on the Github side), but if you restrict the total output allowed it completes rather than just timing out!

rhatdan: Should we expose the `-n` option to RamaLama?

I will take either solution.

engelmi: Are there options in other inference engines similar to the `-n` in llama-server? For example, in vllm `--max-num-batched-tokens` seems related. If yes, then exposing it in RamaLama makes sense. Otherwise it might be too specific.

kush-gupt: There is `max_tokens` in [mlx_lm.server](https://github.com/ml-explore/mlx-lm/blob/0c0b72221f659f1e8b6f10bd01952dfc646c2b29/mlx_lm/SERVER.md#:~:text=generation%20should%20stop.-,max_tokens,-%3A%20(Optional)%20An%20integer) that effectively does the same too

engelmi: Ah ok, then there would be two engines at least benefiting from it. 
@kush-gupt Could you create a GH issue to add it so we can keep track? Would probably make sense to add this after the inference engine spec https://github.com/containers/ramalama/pull/1959

olliewalsh:
Closing as an alternative workaround merged in #1981 
Note: I expect we can revert all of these workarounds once https://github.com/containers/ramalama/pull/1947 merges

PR #1982, opened 2025-09-30
Add unified --max-tokens CLI argument for output token limiting

### LINKED ISSUE #1981: Abstract the maximum output for runtimes into the Ramalama cli

- llama.cpp has `-n`
- MLX has `max_tokens`
- vLLM has `max-num-batched-tokens`

As discussed in #1977 [here](https://github.com/containers/ramalama/pull/1977#discussion_r2387180389), exposing these options in Ramalama CLI makes sense!

Have some cli arg that accepts some form of max token length, and is able to abstract the right param to the right inference runtime

can be added through --runtime-args, but is different for each and every runtime

### Additional context

_No response_

rhatdan:
Implements GitHub issue #1981 by adding a unified --max-tokens argument that works across different AI runtimes with appropriate parameter mapping.

Changes:
- Add max_tokens field to BaseConfig with default value 0 (unlimited)
- Add --max-tokens CLI argument to run, serve, and perplexity commands
- Implement runtime-specific parameter mapping:
  * llama.cpp: maps to -n parameter
  * MLX: maps to --max-tokens parameter
  * vLLM: maps to --max-tokens parameter
- Add daemon service support for max_tokens parameter
- Include comprehensive unit and e2e tests

The --max-tokens argument accepts integer values where 0 means unlimited output tokens, providing a consistent interface across all supported AI runtimes while maintaining compatibility with existing runtime_args.

Update man pages and config documentation for --max-tokens

Fixes #1981

## Summary by Sourcery

Implement a unified --max-tokens option for limiting output tokens across all supported AI runtimes by adding a config field, CLI arguments, runtime-specific flag mappings, daemon support, and documentation updates.

New Features:
- Add global --max-tokens CLI argument (run, serve, perplexity) to limit model output tokens
- Introduce max_tokens field in BaseConfig with a default of 0 (unlimited)

Enhancements:
- Map the unified max_tokens parameter to runtime-specific flags (-n for llama.cpp, --max-tokens for MLX and vLLM)
- Extend the daemon service to include max_tokens in generated serve commands

Documentation:
- Update man pages and configuration docs to document the --max-tokens option and max_tokens setting

Tests:
- Add unit tests verifying max_tokens mapping and behavior for llama.cpp, MLX, and vLLM
- Add e2e tests to validate CLI help text, argument parsing, and handling of valid/invalid --max-tokens values

### DIFF
--- a/ramalama/daemon/service/command_factory.py
+++ b/ramalama/daemon/service/command_factory.py
@@ -38,6 +38,9 @@ class CommandFactory:
         if "temp" not in self.request_args:
             self.request_args["temp"] = CONFIG.temp
 
+        if "max_tokens" not in self.request_args:
+            self.request_args["max_tokens"] = CONFIG.max_tokens
+
         if "ngl" not in self.request_args:
             self.request_args["ngl"] = CONFIG.ngl
 
@@ -104,7 +107,7 @@ class CommandFactory:
         if self.request_args.get("webui") == "off":
             cmd.extend(["--no-webui"])
 
-        if check_nvidia() or check_metal(SimpleNamespace({"container": False})):
+        if check_nvidia() or check_metal(SimpleNamespace(container=False)):
             cmd.extend(["--flash-attn", "on"])
 
         # gpu arguments
@@ -114,5 +117,10 @@ class CommandFactory:
         cmd.extend(["-ngl", f"{ngl}"])
         threads = self.request_args.get("threads")
         cmd.extend(["--threads", str(threads)])
+
+        # Add max tokens parameter for llama.cpp
+        max_tokens = self.request_args.get("max_tokens", 0)
+        if max_tokens > 0:
+            cmd.extend(["-n", str(max_tokens)])
 
         return cmd
--- a/docs/ramalama-perplexity.1.md
+++ b/docs/ramalama-perplexity.1.md
@@ -84,6 +84,13 @@ Accelerated images:
 pass --group-add keep-groups to podman (default: False)
 If GPU device on host system is accessible to user via group access, this option leaks the groups into the container.
 
+#### **--max-tokens**=*integer*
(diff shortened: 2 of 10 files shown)


rhatdan:
This PR was generated with the helper of Cursor.

giuseppe (approved):
LGTM

ramalama/model.py
```diff
@@ -0,0 +1,960 @@
+import os
```
engelmi: This was was moved to [ramalama/transports/base.py](https://github.com/containers/ramalama/blob/main/ramalama/transports/base.py) and can be deleted.

ramalama/daemon/service/command_factory.py
```diff
@@ -59,6 +59,9 @@ def _set_defaults(self):
         if "seed" not in self.request_args:
             self.request_args["seed"] = ""
 
+        if "max_tokens" not in self.request_args:
```
engelmi: After merging the [inference spec PR](https://github.com/containers/ramalama/pull/1959), no changes are required here (and apparently I forgot to remove this part in the daemon, will create a PR to do so soon).  

To make the max tokens accessible, it needs to be added to the context here:
https://github.com/containers/ramalama/blob/main/ramalama/command/context.py#L10
The new variable can then be used in the specs, for example llama.cpp:
https://github.com/containers/ramalama/blob/main/inference-spec/engines/llama.cpp.yaml
The new entry should be something like (based on the logic below):
```yaml
- name: "-n"
  description: "Maximum tokens to generate"
  value: "{{ args.max_tokens }}"
  if: "{{ args.max_tokens > 0 }}"
```

engelmi: ~The `perplexity` operation was not (yet) refactored to use the spec. So it either needs to be added and refactored or temporarily added to the perplexity code for now.~
**Update:** Added `perplexity` and `bench` to the spec here: https://github.com/containers/ramalama/pull/1986

PR #1998, opened 2025-10-02
Fix typo in llama.cpp engine spec

olliewalsh:
## Summary by Sourcery

Bug Fixes:
- Remove erroneous parentheses from the host.uses_metal check in the flash-attn conditional


olliewalsh:
@engelmi @rhatdan FYI, `ramalama run` failing with `TypeError: 'bool' object is not callable` when `or` doesn't short-circuit

rhatdan:
LGTM

engelmi:
Good catch!

Issue #1989, opened 2025-10-01
Multi-part GGUF model support


tangentsoft:
### Feature request description

This currently fails:

```
$ ramalama run gpt-oss:120b
Downloading hf://ggml-org/gpt-oss-120b-GGUF ...
Trying to pull hf://ggml-org/gpt-oss-120b-GGUF ...
Error: Command '['podman', 'logs', 'ramalama_BomxHKti1w']' returned non-zero exit status 125.
```

I believe it is due to the fact that this is split into 2 or 3 parts, depending on where you go looking.

I originally stumbled across this by trying to avoid a redundant download through reuse of the GGUF files that LM Studio previously pulled successfully:

```
$ ls -lhn ~/.lmstudio/models/lmstudio-community/gpt-oss-120b-GGUF
total 60G
-rw-r--r--. 1 1000 1000 38G Sep 28 12:40 gpt-oss-120b-MXFP4-00001-of-00002.gguf
-rw-r--r--. 1 1000 1000 22G Sep 28 12:40 gpt-oss-120b-MXFP4-00002-of-00002.gguf

$ time ramalama run file://.lmstudio/models/lmstudio-community/gpt-oss-120b-MXFP4-00001-of-00002.gguf
Error: Command '['podman', 'logs', 'ramalama_a4iJXQ1MUi']' returned non-zero exit status 125.

real    0m21.629s
user    0m0.427s
sys     0m0.160s
```

I added the `time` call to show that it does _try_ to load it, then presumably fails when it hits EOF and doesn't know how to chain to the 00002 file.

I'm filing this as a feature request instead of a bug because I expect the use case hasn't come up yet. This isn't a matter of broken functionality, it's outright missing.

### Suggest potential solution

Recognize the "NNNNN-of-MMMMM" naming scheme and chain appropriately.

### Have you considered any alternatives?

I tried a web search for concatenation methods but didn't find any solution I liked.

Even if it worked, it would be a suboptimal solution because it would require creating a duplicate copy of 60G of data purely to make Ramalama happy. It should be possible to share the existing GGUF files.

fozzee:
ramalama does support the split model files, but it looks like it doesn't support running a split model from a `file://` url.

Running `ramalama run gpt-oss:120b` command is not working in the current release version of ramalama because of a parsing bug when the model is in the top level directory of the model repo. That bug will be fixed once ramalama version `0.12.4` is released. See issues #1878 #1974.

[...]

PR #2001, opened 2025-10-04
add support for split model file url

fozzee:
fix local model file url returning early before assembling split model file list

## Summary by Sourcery

Add support for split model file URLs by assembling and handling individual split files for both local and remote models, fix premature return when processing local model URLs, and improve file copying behavior.

New Features:
- Add support in pull() for split model file URLs by generating a list of individual split files

Bug Fixes:
- Fix premature return for local model file URLs that prevented assembling split model file list

Enhancements:
- Use LocalModelFile with generated SHA256 hashes for local split model files

## Summary by Sourcery

Support split model file URLs by properly assembling split files for both local and remote models, fix local URL handling order, and improve hash generation for split files

New Features:
- Support pulling split model file URLs by assembling individual split segments

Bug Fixes:
- Fix premature return that prevented assembling split model files for local file URLs

Enhancements:
- Use LocalModelFile with generated SHA256 hashes for local split model files and apply SHA256 hashing for remote segments

### DIFF
--- a/ramalama/transports/url.py
+++ b/ramalama/transports/url.py
@@ -92,17 +92,28 @@ class URL(Transport):
         for i in range(1, total_parts + 1):
             file_name = f"{filename_base}-{i:05d}-of-{total_parts:05d}.gguf"
             url = f"{self.type}://{path_part}{file_name}"
-            files.append(
-                SnapshotFile(
-                    url=url,
-                    header={},
-                    hash=snapshot_hash,
-                    type=SnapshotFileType.GGUFModel,
-                    name=file_name,
-                    should_show_progress=True,
-                    required=True,
+            if self.type == "file":
+                files.append(
+                    LocalModelFile(
+                        url=f"{path_part}{file_name}",
+                        header={},
+                        model_file_hash=generate_sha256(file_name),
+                        name=file_name,
+                        required=True,
+                    )
                 )
-            )
+            else:
+                files.append(
+                    SnapshotFile(
+                        url=url,
+                        header={},
+                        hash=generate_sha256(file_name),
+                        type=SnapshotFileType.GGUFModel,
+                        name=file_name,
+                        should_show_progress=True,
+                        required=True,
+                    )
+                )
 
         return files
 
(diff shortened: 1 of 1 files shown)


rhatdan:
@olliewalsh @engelmi PTAL

ramalama/transports/url.py
```diff
@@ -35,7 +35,7 @@ def download(self, blob_file_path, snapshot_dir):
         if not os.path.exists(self.url):
             raise FileNotFoundError(f"No such file: '{self.url}'")
         # moving from the local location to blob directory so the model store "owns" the data
-        shutil.copy(self.url, blob_file_path)
+        shutil.copy(self.url, blob_file_path, follow_symlinks=False)
```
engelmi: If I understand it correctly, this means we only copying the symlink. So on the one side we don't use more disk space and on the other we can't guarantee that the file the symlink is pointing to gets removed. I'd prefer copying the actual file to be in control and avoid split situations.

olliewalsh:
This used to work (https://github.com/containers/ramalama/pull/1426), has there been a regression?

ramalama/transports/url.py
fozzee: Yes, this will copy the model file without following or resolving symlinks. This only changes the current behavior of ramalama if the user uses a symlink as the model url.

I agree it isn't ideal. However, I also think it is unexpected behavior from the user perspective to follow a symlink and make a copy of the target model file if the user is providing a symlink to the model file.

I will remove this for now since this implementation is naive and breaks if the symlink is targeting a relative path. This PR can just focus on fixing the split model file:// url.

fozzee:
> This used to work (https://github.com/containers/ramalama/pull/1426), has there been a regression?

Yes. It looks like the split model logic was refactored into the URL tranport and it didn't account for a code path including both split model files and the file:// url usage. Also, the url split model unit test only tests https:// urls.

There seem to be some other corner case issues like using an absolute path instead of a relative path. And running `ramalama ls` shows a model url for file:// urls that assumes it is an absolute path with an extra prefixed `/`. So there is probably some further review of this area of the code needed.

ramalama/transports/url.py
fozzee: This is removed now.

fozzee:
Removed the `follow_symlinks=False` from the commit and removed `Fixes: #1989` from prior comment to continue discussion about how to handle use of models with model files external to model store.

I will leave this PR as just the fix for split model file urls.

ramalama/transports/url.py
```diff
@@ -92,17 +92,28 @@ def _assemble_split_file_list(self, snapshot_hash: str) -> list[SnapshotFile]:
         for i in range(1, total_parts + 1):
```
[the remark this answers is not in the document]
engelmi: Not the goal of this PR to refactor this part. Could be done in a separate one.

PR #2024, opened 2025-10-13
docs: Update README with correct MLX install instructions

kush-gupt:
The MLX backend has been preferring to use the mlx.lm server libraries under path, which `uv tool install mlx-lm` handles quite well. It's also how the Ramalama install script sets up MLX, so we should align the docs to that instead of the first implementation with the python libraries.

## Summary by Sourcery

Update README to instruct installing the mlx-lm package via 'uv tool' commands instead of pip and provide an upgrade option

Documentation:
- Replace pip-based mlx-lm installation instructions with 'uv tool install mlx-lm' commands
- Add 'uv tool upgrade mlx-lm' step for users to update to the latest version


rhatdan:
LGTM

PR #2050, opened 2025-10-21
Set gguf quantization default to Q4_K_M

engelmi:
Relates to: https://github.com/containers/ramalama/issues/642

## Summary by Sourcery

Enhancements:
- Change the CLI convert command to default to Q4_K_M quantization format

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -28,6 +28,7 @@ from ramalama.common import accel_image, get_accel, perror
 from ramalama.config import (
     COLOR_OPTIONS,
     CONFIG,
+    GGUF_QUANTIZATION_MODES,
     SUPPORTED_ENGINES,
     SUPPORTED_RUNTIMES,
     coerce_to_bool,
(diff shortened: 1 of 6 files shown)


ramalama/cli.py
```diff
@@ -674,6 +674,7 @@ def convert_parser(subparsers):
             "Q6_K",
             "Q8_0",
         ],
+        default="Q4_K_M",
```
rhatdan: Can you put this in ramalama.conf

engelmi: Done. Also added a system test for this (no pytest e2e yet for convert). 
PTAL @rhatdan

rhatdan:
LGTM once you get tests to pass.

engelmi:
The two CI jobs with macos are failing due to the mlx version mismatch. This should be solved by https://github.com/containers/ramalama/pull/2051.
Not sure why the two konflux jobs are failing, but it seems to be unrelated to the PR changes.

PR #2053, opened 2025-10-22
run RAG operations in a separate container

mikebonnet:
This is a major rearchitecture of container management for operations using binaries other than `llama-server`, including `ramalama rag`, `ramalama convert`, `ramalama run --rag`, and `ramalama serve --rag`.

The `-rag` images are now independent images which don't contain `llama-server`, and are run in separate containers for operations involving `doc2rag`, `convert_hf_to_gguf.py`, and `rag_framework`. Build times for `-rag` images have been dramatically reduced and builds are now fully parallelized.

`-rag` containers are built for each backend supported by `torch`: `cpu`, `cuda`, `rocm`, and `xpu`. The appropriate `-rag` image for the available accelerator is selected automatically in the same way the `llama-server` image is selected, falling back to `cpu` if no accelerators are available. All other `-rag` images are obsolete and are no longer being built.

When running `run --rag` and `serve --rag`, a separate container is created which acts as a `RAG` proxy, enriching the requests from the client with context from the RAG database, passing the requests to the LLM running in a separate container, and returning the results to the client. The `RAG` database is mounted into the `rag_framework` container.

`ramalama rag` runs `doc2rag` in the `-rag` container using `podman run` with the relevant env vars set for the available accelerator. If a OCI image is requested, the image build is run as a separate step using `podman build`.

`ramalama convert` runs `convert_hf_to_gguf.py` in the `-rag` container using `podman run` with the relevant env vars set for the available accelerator. `llama-quantize` is then run using the appropriate `llama-server` image for the available accelerator. If a OCI image is requested, the image build is run as a separate step using `podman build`.

The inference interface is used to generate commands whenever possible. Several new `Engine` subclasses have been created to handle execution of the container manager for various container run and image build operations.

## Summary by Sourcery

Rearchitect container management for Retrieval-Augmented Generation by decoupling RAG processing into standalone OCI images per accelerator and running RAG operations in dedicated containers

New Features:
- Introduce standalone `-rag` images for `doc2rag`, model conversion and RAG proxy operations built per backend (cpu, cuda, rocm, xpu)
- Add `ramalama rag` command to generate vector databases in a separate container
- Add `ramalama convert` command to convert models to GGUF in a dedicated container
- Support `ramalama run --rag` and `ramalama serve --rag` to launch a RAG proxy container that enriches requests and forwards them to the LLM container

Enhancements:
- Refactor container engine logic into BaseEngine, Engine and BuildEngine classes to streamline run and build workflows
- Generalize image selection via `accel_image` and `rag_image` functions to automatically pick appropriate images based on accelerator and policy
- Parallelize `-rag` image builds and simplify requirement file generation with new pip compile targets

Documentation:
- Document new `--rag-image`, `--gguf-image` and `--pull` CLI flags and update man pages and inference-specs with RAG commands and options

Tests:
- Add and update unit and system tests covering RAG dryrun, run/serve with RAG proxy, port exclusion logic and container mount behaviors


inference-spec/engines/llama.cpp.yaml
```diff
@@ -114,3 +114,77 @@ commands:
       name: "llama-bench"
       binary: "llama-bench"
       options: *bench_perplexity_options
+  - name: rag
```
mikebonnet, line 121: @engelmi I'm using the inference spec to generate the commands for `rag`, `run--rag`, `serve --rag`, and `convert`. I might be abusing it a bit, let me know if you think there's a better way to do this.

container-images/scripts/rag_framework
```diff
@@ -122,117 +197,287 @@ class milvus:
...
+        rag_context = self._perform_rag_lookup(latest_message.content)
+        conversation_history = self._extract_conversation_context(messages[:-1])
 
-            ### Context:
-            {reranked_context.strip()}
+        # Enhanced system prompt with RAG context
+        system_prompt = (
+            textwrap.dedent(
+                """
+            You are an expert software architect assistant.
```
mikebonnet, line 255: @bmahabirbu This is the RAG prompt suggested by Claude. Let me know if you think it makes sense, or if we should revert it to something more similar to what was used before.

bmahabirbu: Sounds good! Ill give it a test but this prompt looks good! I tested a few different prompts and they all work similarly

ramalama/transports/oci.py
```diff
@@ -195,22 +216,35 @@ def _create_manifest(self, target, imageid, args):
...
-            f"{annotations.AnnotationTitle}=args.SOURCE",
+            f"{annotations.AnnotationTitle}={args.SOURCE}",
             target,
             imageid,
         ]
-        run_cmd(cmd_args, stdout=None)
+        if args.dryrun:
+            dry_run(cmd_args)
+        else:
+            run_cmd(cmd_args, stdout=None)
```
rhatdan, line 228: We should probably add an option to run_cmd called dryrun=False, and consolidate all of these calls.

ramalama/cli.py
```diff
@@ -688,6 +692,28 @@ def convert_parser(subparsers):
         help="GGUF quantization format",
     )
     add_network_argument(parser)
+    parser.add_argument(
```
rhatdan, line 707: Would a command like

ramalama --image foobar --gguf-image goobar ...

Make sense?

Do we use multiple images within the same command?

rhatdan:
I find having a --image, --gguf-image and --rag-image is going to be very confusing for users.

rhatdan:
@bmahabirbu @olliewalsh @engelmi PTAL

ramalama/cli.py
mikebonnet: Yes, multiple images are used in a single `convert` command. `convert_hf_to_gguf.py` is run from the `-rag` image (because it has a bunch of heavyweight python dependencies, including `torch`) and then `llama-quantize` is run from the `llama-server` image.

`--gguf-image` is the image that contains the `convert_hf_to_gguf.py` script. It's just the `-rag` image, so we could rename the option to `--rag-image` for consistency with usage elsewhere.

inference-spec/engines/llama.cpp.yaml
engelmi: The name of the command was intended to match the subcommand of RamaLama, e.g. `ramalama serve` matches the `serve` entry here. I am not sure how this could/should look for command options such as `serve --rag`, but I haven't looked through the code yet, though. 
Otherwise, I think its a good use of inference spec.

ramalama/cli.py
mikebonnet: I changed the name of the option to `--rag-image` for consistency, and it is auto-selected the same way the `llama-server` image is. In normal operation the user won't have to specify either `--image` or `--rag-image` for `ramalama convert` to run successfully.

mikebonnet:
> I find having a --image, --gguf-image and --rag-image is going to be very confusing for users.

I removed `--gguf-image`, and users won't need to set the other options during normal operation.

rhatdan:
LGTM

bmahabirbu:
Ah was hoping to get on the review today apologies for being late but I like the idea of just getting this in and if anything can make follow up prs regarding changes!

PR #2069, opened 2025-10-27
konflux: merge before building/testing PRs

mikebonnet:
Prevents accepting out-of-date/incompatible code and increases the value of Konflux testing.

## Summary by Sourcery

Enhancements:
- Add mergeTargetBranch parameter to the pull-request Tekton pipeline to enable merging the base branch prior to task execution

### DIFF
--- a/.tekton/pipelines/pull-request-pipeline.yaml
+++ b/.tekton/pipelines/pull-request-pipeline.yaml
@@ -132,6 +132,10 @@ spec:
       value: $(params.output-image).git
     - name: ociArtifactExpiresAfter
       value: $(params.image-expires-after)
+    - name: depth
+      value: "50"
+    - name: mergeTargetBranch
+      value: "true"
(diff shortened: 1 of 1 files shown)


mikebonnet:
@olliewalsh I believe this is what you were asking for.

mikebonnet:
Question: has anyone ever submitted a PR with more than 20 new commits in it? If so, we should increase the value of the `depth` param.

rhatdan:
I have never seen one greater the 20, but if there is not a huge reason to not expand it, I would be fine to expand.

mikebonnet:
Increased to 50, which I think is the value GitLab uses.

rhatdan:
LGTM

olliewalsh:
> Increased to 50, which I think is the value GitLab uses.

I really hope there are no PRs with 50+ commits!

PR #2083, opened 2025-10-30
Fix the rpm name in docs

olliewalsh:
The rpm name changed from python3-ramalama to ramalama in https://github.com/containers/ramalama/pull/1693.

The rpm name in README.md was updated at that time but the change seems to have been lost in a subsequent commit.

## Summary by Sourcery

Correct the RPM package name in documentation to reflect the change from python3-ramalama to ramalama.

Documentation:
- Update Fedora install instructions in README.md to use ramalama instead of python3-ramalama
- Update Fedora install instructions in installation.mdx to use ramalama instead of python3-ramalama


rhatdan:
The old name should still work, but this is better.

PR #2091, opened 2025-10-31
Switch DEFAULT_PORT_RANGE to use 100 ports

rhatdan:
Currently we limit the number of ports that ramalama serve will attempt to 10, expand this to 100 so that we are more likely to get an open port.

## Summary by Sourcery

Extend the default port selection window to 100 ports and update documentation and tests to reflect the new configuration

New Features:
- Expand the default service port range from 10 to 100 ports

Documentation:
- Revise ramalama-serve and ramalama-run man pages to document the expanded port range, format config file references with backticks, and include ramalama.conf in SEE ALSO sections

Tests:
- Update test_compute_ports to expect 101 ports in the default range

### DIFF
--- a/docs/ramalama-run.1.md
+++ b/docs/ramalama-run.1.md
@@ -29,7 +29,7 @@ URL support means if a model is on a web site or even on your local system, you 
 
 #### **--api**=**llama-stack** | none**
 unified API layer for Inference, RAG, Agents, Tools, Safety, Evals, and Telemetry.(default: none)
-The default can be overridden in the ramalama.conf file.
+The default can be overridden in the `ramalama.conf` file.
 
 #### **--authfile**=*password*
 path of the authentication file for OCI registries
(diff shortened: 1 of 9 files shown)


rhatdan:
@engelmi @olliewalsh @mikebonnet @bmahabirbu @ieaves PTAL

rhatdan:
I was concerned 10 was too small, and agree 100 would be ridiculous. but not much overhead and less likely to bite us in the butt at some point.

olliewalsh:
+1 for a range of 100 ports. It tends to be pretty busy around the 8080 range.

ieaves (commented):
LGTM. 

If you wanted to make the port range configurable you could add

```python
@dataclass
class BaseConfig:
    ...
    port_width: int = 100
```

Then add `port_width` port_width to the int  validators in `load_env_config`

```python
def load_env_config(env: Mapping[str, str] | None = None) -> dict[str, Any]:
    ...
    for key in ['threads', 'ctx_size', 'ngl', 'port_width']:
```


And then

```python
DEFAULT_PORT_RANGE: tuple[int, int] = (DEFAULT_PORT, DEFAULT_PORT + config.port_width)
```

Or, add the default_port_range as a property on the config so we don't have to carry around special keywords everywhere.

PR #2102, opened 2025-11-03
konflux: re-enable s390x builds

mikebonnet:
Removing `-DGGML_NATIVE=ON` disables host detection, so the target architecture needs to be set to `z15` manually.

## Summary by Sourcery

Re-enable s390x builds by explicitly setting the target architecture flag and adding s390x to the CI pipeline matrices

Build:
- Set -DARCH_FLAGS=-march=z15 for s390x in the llama and whisper build script

CI:
- Include linux/s390x in bats and ramalama pull-request, push, and integration Tekton pipelines


rhatdan:
LGTM

PR #2105, opened 2025-11-04
Integrate daemon in ramalama cli

engelmi:
## Testing this PR

First, changes in this PR need to be built and embedded into a locally built ramalama container:
```bash
make pypi-build
podman build -t quay.io/ramalama/myramalama:latest -f container-images/myramalama .
```
This assumes that the container file `container-images/myramalama` contains something like this:
```
FROM quay.io/ramalama/rocm:latest

COPY dist/ramalama-0.14.0-py3-none-any.whl ./
RUN pip install --force ./ramalama-0.14.0-py3-none-any.whl
```

Now the integration can be tested. For example:
```bash
# serving a model
$ ./bin/ramalama --use-daemon serve smollm:135m --pull never --image quay.io/ramalama/myramalama:latest
smollm:135m available at: http://127.0.0.1:8080/models/library/smollm

# chatting with a model directly
$ ./bin/ramalama --use-daemon run smollm:135m --pull never --image quay.io/ramalama/myramalama:latest
smollm:135m available at: http://127.0.0.1:8080/models/library/smollm
🦭 > hi, how are you?

# serving multiple models in the same daemon
$ ./bin/ramalama --use-daemon --daemon-name=ramalama-daemon-multiple serve smollm:135m --pull never --image quay.io/ramalama/rocm:latest
Error: no such object: "ramalama-daemon-multiple"
smollm:135m available at: http://127.0.0.1:8080/models/library/smollm
$ ./bin/ramalama --use-daemon --daemon-name=ramalama-daemon-multiple serve tinyllama --pull never --image quay.io/ramalama/rocm:latest
tinyllama available at: http://127.0.0.1:8080/models/library/tinyllama
# The models served in the container can be viewed at
# http://127.0.0.1:8080/models
# or http://127.0.0.1:8080/api/ps

# listing all running daemon container
$ ./bin/ramalama --use-daemon ps
CONTAINER ID  IMAGE                         COMMAND               CREATED             STATUS             PORTS                   NAMES
d91bec54198a  quay.io/ramalama/rocm:latest  ramalama --store ...  About a minute ago  Up About a minute  0.0.0.0:8080->8080/tcp  ramalama-daemon-multiple

```
**$${\color{red}Important:}$$**
The webui of llama.cpp is available under `http://127.0.0.1:8080/models/library/smollm` directly, but will continuously try to redirect. This is due to an issue in the llama.cpp webui as it expects to be served at `http://127.0.0.1:8080` and not a subpath.

## Summary by Sourcery

Integrate support for using the RamaLama daemon as the backend for CLI commands by adding feature flags, client and stub logic, and updating handlers, engine, DTOs, and documentation to enable starting, serving, and chatting through a long-running daemon process.

New Features:
- Add --use-daemon and --daemon-name flags to the CLI to route run and serve commands through the daemon
- Introduce DaemonStub and DaemonClient classes to manage daemon lifecycle and interact with daemon HTTP API for serve, stop, and health operations

Enhancements:
- Refactor container listing to accept a dynamic label filter for distinguishing daemon-managed containers
- Simplify and extend daemon HTTP handlers: add a health endpoint, streamline proxy request forwarding, and improve expiration and shutdown logging
- Update ServeRequest DTO to use a list for exec_args and adjust serialization and parsing accordingly
- Extend common genname function to accept a prefix parameter
- Rename and clean up logger configuration to prevent duplicate handlers

Documentation:
- Add documentation for the --use-daemon feature flag in CLI man pages and MD reference guides

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -38,6 +38,7 @@ from ramalama.config import (
     get_inference_spec_files,
     load_file_config,
 )
+from ramalama.daemon_stub import DaemonStub, run_daemon
 from ramalama.endian import EndianMismatchError
 from ramalama.log_levels import LogLevel
 from ramalama.logger import configure_logger, logger
@@ -216,6 +217,7 @@ def create_argument_parser(description: str):
         prog="ramalama",
         description=description,
         formatter_class=argparse.RawTextHelpFormatter,
+        exit_on_error=False,
     )
     configure_arguments(parser)
     return parser
@@ -275,6 +277,21 @@ The RAMALAMA_IN_CONTAINER environment variable modifies default behaviour.""",
     parser.add_argument(
         "--noout",
         help=argparse.SUPPRESS,
+    )
+
+    # Feature flags
+    parser.add_argument(
+        "--use-daemon",
+        dest="use_daemon",
+        default=False,
+        action="store_true",
+        help="Feature Flag: Enable using the daemon as backend by default",
+    )
+    parser.add_argument(
+        "--daemon-name",
+        dest="daemon_name",
+        help="Requires --use-daemon. Specifies the name of the daemon either for creation or further use.",
+        completer=suppressCompleter,
     )
 
 
@@ -506,7 +523,14 @@ def containers_parser(subparsers):
 
 
 def list_containers(args):
-    containers = engine.containers(args)
+    label = engine.LABEL_CONTAINER_RAMALAMA
+
+    # Feature Flag:
+    # Use daemon backend when feature flag is given
+    if args.use_daemon:
+        label = engine.LABEL_CONTAINER_RAMALAMA_DAEMON
+
+    containers = engine.containers(args, label)
     if len(containers) == 0:
         return
     print("\n".join(containers))
@@ -1133,6 +1157,14 @@ def run_cli(args):
         except Exception as exc:
             raise e from exc
 
+    # Feature Flag:
+    # Use daemon backend when feature flag is given
+    if args.use_daemon:
+        stub = DaemonStub(shortnames, args, sys.argv)
+        serve_path = stub.serve_model(getattr(args, "daemon_name"))
+        stub.chat(serve_path)
+        return
+
     if args.rag:
         if not args.container:
             raise ValueError("ramalama run --rag cannot be run with the --nocontainer option.")
@@ -1151,6 +1183,15 @@ def serve_parser(subparsers):
 
 
 def serve_cli(args):
+
+    # Feature Flag:
+    # Use daemon backend when feature flag is given
+    if args.use_daemon:
+        stub = DaemonStub(shortnames, args, sys.argv)
+        serve_path = stub.serve_model(getattr(args, "daemon_name"))
+        stub.wait_for_model(serve_path)
+        return
+
     if not args.container:
         args.detach = False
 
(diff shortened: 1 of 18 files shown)


ramalama/daemon/handler/daemon.py
```diff
@@ -114,6 +123,11 @@ def _handle_post_serve(self, handler: http.server.SimpleHTTPRequestHandler):
         logger.debug(f"Received serve request: {serve_request.serialize()}")
 
         port = self.model_runner.next_available_port()
+        resolved_model = self.shortnames.resolve(serve_request.model_name)
```
rhatdan: This is going to be a problem, if this is resolved within the container, since the shortnames.conf file file the host is not exposed to the container.  We need a way to build a shortnames.conf file from host and inject into the container, or some kind of reference file to models stored locally.

ramalama/daemon/handler/proxy.py
```diff
@@ -43,31 +43,43 @@ def handle_delete(self, handler: http.server.SimpleHTTPRequestHandler, is_referr
...
-            msg = f"No model for path '{proxy_path}' found"
+        for path, model in self.model_runner.served_models.items():
+            if path in handler.path:
+                requested_model = model
+                forward_path = handler.path.replace(path, "")
+                if forward_path == "":
+                    forward_path = "/"
+                continue
+
+        # forward_path = ""
```
rhatdan: Do not like having commented out lines in PR. Do you plan on turning this on, when the PR is out of draft?

ramalama/cli.py
```diff
@@ -279,6 +280,21 @@ def configure_arguments(parser):
         help=argparse.SUPPRESS,
     )
 
+    # Feature flags
+    parser.add_argument(
+        "--use-daemon",
+        dest="use_daemon",
+        default=False,
```
rhatdan, line 286: Should come from ramalama.conf and we want this to default to TRUE.

ramalama/cli.py
```diff
@@ -279,6 +280,21 @@ def configure_arguments(parser):
...
+        action="store_true",
+        help="Feature Flag: Enable using the daemon as backend by default",
+    )
+    parser.add_argument(
```
rhatdan, line 290: Is this for debugging purposes?

ramalama/daemon_stub.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def __init__(
+        self, shortnames: Shortnames, args: Namespace, raw_args: list[str], wait_for_healthy_timeout: int = 10
+    ):
+        self.shortnames = shortnames
+        self.cli_args = args
+        self.raw_args = raw_args
+        self.wait_for_healthy_timeout = wait_for_healthy_timeout
+
+        self.client = DaemonClient("127.0.0.1", self.cli_args.port)
```
rhatdan, line 24: Should 127.0.0.1 be a constant?  Should it be configurable?

ramalama/daemon/handler/proxy.py
engelmi: Ups, forgot to remove the old, commented code. The new one doesn't rely on the referrer anymore. This doesn't solve the path mismatch error in the webui of the recent llama.cpp, though.

ramalama/cli.py
engelmi: I thought of it was a temporary feature flag to keep the behavior as is at the moment by default. I can add it to the configuration.

ramalama/cli.py
engelmi: The `--daemon-name` is intended to be used to identify the daemon container that should be used (or create it with that name if given). For example, you could serve two models in one daemon container (e.g. embedded and llm) and another one in a second container.
I'll update the description of the PR with examples on how to use soon-ish.

ramalama/daemon_stub.py
engelmi: At the moment it should be a constant since the daemon should run locally based on how I set everything up. Later on we can easily make this configurable to be anywhere - but we have to figure out how the CLI should look like.

ramalama/cli.py
rhatdan: Ok maybe we should meet to discuss the design, I am not sure I understand what  you are after here.

ramalama/daemon/handler/daemon.py
engelmi: I agree. 
Would it be fine if I'd mount the configuration into the container (similar to the model store directory) for now?

ramalama/cli.py
rhatdan: Well if we plan on yanking the option eventually then fine. But it would complicate the code to continue to support the previous behavior and the new. I guess I was thinking we would just move forward with always using the daemon and remove the previous behavior.

engelmi: Yes, that would be the goal - thats why I was a bit bumped with the llama.cpp webui issue. The daemon breaks this at the moment, unfortunately. The `run` command with the chat works (since the plain REST API from llama.cpp is used there).
With the feature flag I intended to work on and fix this over the next weeks and don't block new releases and/or avoid annoying users.

ramalama/daemon/handler/daemon.py
engelmi: Ah, found a different approach: The original cli input for the model is preserved in the `args.UNRESOLVED_MODEL` field. This simplifies all of this.

ramalama/daemon/handler/proxy.py
engelmi: Removed.

Issue #2145, opened 2025-11-13
ramalama version 0.14.0 stopped working on Fedora 42


dwrobel:
```
$ramalama run llama3.2 "Write a hello world application in python"
```
works on version 0.13.0 but stopped working on 0.14.0 on Fedora 42.

Steps to reproduce:

```
$ ramalama version
ramalama version 0.13.0
$ ramalama run llama3.2 "Write a hello world application in python"
**Hello World Application in Python**

Here is a simple "Hello World" application written in Python.

# hello_world.py

def main():
    """
    Prints a hello world message to the console.
    """
    print("Hello, World!")

if __name__ == "__main__":
    main()

**Explanation:**

* The `main` function is where our application's main logic lives. In this case, it simply prints a "Hello, World!" message to the console.
* The `if __name__ == "__main__":` line is used to ensure that the `main` function is only executed when the script is run directly (i.e., not when it's imported as a module by another script).
* The `print` function is used to output the "Hello, World!" message.

**Running the Application:**

1. Save this code in a file called `hello_world.py`.
2. Open a terminal or command prompt and navigate to the directory where you saved the file.
3. Type `python hello_world.py` to run the application.
4. You should see the "Hello, World!" message printed to the console.

**Tips and Variations:**

* You can modify the `main` function to perform any other actions, such as reading input from the user or writing output to a file.
* To add more complexity to your application, you could use Python's built-in `input` function to read user input and respond with a greeting.
* To make your application more reusable, you could break it out into separate functions or modules and import them into other scripts.

I hope this helps! Let me know if you have any questions or if you'd like to see more examples of Python programming.
$ sudo dnf update -y ramalama
Updating and loading repositories:
Repositories loaded.
Package                                                                      Arch              Version                                                                       Repository                                       Size
Upgrading:
 ramalama                                                                    noarch            0.14.0-1.fc42                                                                 updates                                       1.2 MiB
   replacing ramalama                                                        noarch            0.13.0-1.fc42                                                                 <unknown>                                     1.1 MiB

Transaction Summary:
 Upgrading:          1 package
 Replacing:          1 package

Total size of inbound packages is 436 KiB. Need to download 436 KiB.
After this operation, 82 KiB extra will be used (install 1 MiB, remove 1 MiB).
[1/1] ramalama-0:0.14.0-1.fc42.noarch                                                                                                                                                     100% |   1.2 MiB/s | 436.0 KiB |  00m00s
[...]

PR #2146, opened 2025-11-13
Fix-up ref file path field when it has been moved

olliewalsh:
The ref file includes a path field.
If this does not match the actual ref file path being read then update the field.

Fixes: https://github.com/containers/ramalama/issues/2145

## Summary by Sourcery

Update ref JSON files to correct their path field upon relocation and enable full exception logging under the CLI's debug flag

Bug Fixes:
- Ensure the internal path field of a ref JSON file is updated to match its actual location when the file has been moved

Enhancements:
- Add exception stack trace logging in the CLI when debug mode is enabled

### DIFF
--- a/ramalama/model_store/reffile.py
+++ b/ramalama/model_store/reffile.py
@@ -220,6 +220,6 @@ class RefJSONFile:
             )
             # ref file has moved
             if ref_file.path != data["path"]:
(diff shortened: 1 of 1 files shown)


olliewalsh:
@rhatdan PTAL

rhatdan (approved):
LGTM

PR #2169, opened 2025-11-18
don't install ramalama (or anything python-related) into the inference engine images

mikebonnet:
Since `RAG` has been moved into its own images, no python runtime environment is required in the inference engine images.

## Summary by Sourcery

Remove Python runtime and ramalama installation steps from inference engine images, simplify build scripts and Containerfiles, and update CI pipelines to use bare script entrypoint names.

Bug Fixes:
- Fix typo in bind mount comment in rag.py (behaviour).

Enhancements:
- Eliminate Python dependencies and ramalama installation in stable diffusion, llama, and whisper build scripts.
- Consolidate and simplify dnf installs to minimal build tools and disable weak dependencies.
- Streamline Containerfiles to only copy and run unified build scripts from /src.
- Simplify container_build.sh to pass entrypoint script names without absolute paths.

CI:
- Update Tekton pipeline YAMLs to reference entrypoint scripts by name instead of absolute paths.

### DIFF
--- a/ramalama/rag.py
+++ b/ramalama/rag.py
@@ -132,7 +132,7 @@ class RagEngine(Engine):
         if self.sourcetype is RagSource.DB:
             # Convert to container-friendly path format (handles Windows path conversion)
             rag = get_container_mount_path(self.args.rag)
-            # Read-write is the default behavior for bind mounts in both Docker and Podman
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

PR #2174, opened 2025-11-19
fix: now summarizer clears cache

bmahabirbu:
## Summary by Sourcery

Improve chat context management and recovery when llama.cpp server context limits are exceeded, adding pluggable strategies and cache clearing with automatic retry.

New Features:
- Introduce pluggable context management strategies, including observation masking and an LLM-based summarizer, to control conversation history size.
- Add a llama.cpp server API helper module to manage slot cache, query context size, and perform health checks.
- Expose a CLI option to choose the context management strategy and configure slot state persistence via slot-save-path in the llama.cpp engine config.

Bug Fixes:
- Handle context-size-related errors from the server by clearing the slot cache, reducing conversation history, and retrying the request instead of failing silently or hanging.
- Avoid accumulating consecutive user-only messages in conversation history when no assistant response is returned, preventing pathological dialogue states.

Enhancements:
- Improve response streaming to detect and log empty or error-only responses, including context-related server errors.
- Refine chat failure recovery by attempting context reduction and cache clearing when connections fail with long histories before terminating the session.

Tests:
- Update llama.cpp engine command construction tests to assert the new slot-save-path argument is passed to the server.


bmahabirbu:
aided by cursor

ramalama/chat.py
```diff
@@ -125,72 +138,171 @@ def prep_rag_message(self):
...
-
-    def _check_and_summarize(self):
-        """Check if conversation needs summarization and trigger it."""
-        summarize_after = getattr(self.args, "summarize_after", 0)
-        if summarize_after > 0:
-            self.message_count += 2  # user + assistant messages
-            if self.message_count >= summarize_after:
-                self._summarize_conversation()
-                self.message_count = 0  # Reset counter after summarization
+            print(f"Failed to get AI summary, using basic extraction...", flush=True)
```
rhatdan: This should be in the logger.warning.

ramalama/chat.py
```diff
@@ -125,72 +138,171 @@ def prep_rag_message(self):
...
+            
+            # Add the basic summary
+            new_history.append({"role": "system", "content": f"Previous conversation: {basic_summary}"})
+            
+            # Keep only recent messages
+            new_history.extend(recent_msgs)
+            
+            self.conversation_history = new_history
+            logger.debug(f"Created basic summary: {len(messages_to_summarize)} messages -> summary")
+            print("Basic summary created.", flush=True)
```
rhatdan: Looks like Debug message

ramalama/chat.py
```diff
@@ -429,12 +539,63 @@ def _make_request_data(self):
...
+            # Note: llama.cpp slot endpoint is at /slots (not /v1/slots)
+            base_url = self.args.url
+            if base_url.endswith('/v1'):
+                base_url = base_url[:-3]  # Remove '/v1' suffix
+            elif '/v1/' in base_url:
+                base_url = base_url.replace('/v1/', '/')
+            slot_url = f"{base_url}/slots/0?action=erase"
+            
+            logger.debug(f"Clearing slot cache at: {slot_url}")
+            print("Clearing server cache first...", flush=True)
```
rhatdan: Debug msg?

ramalama/chat.py
```diff
@@ -429,12 +539,63 @@ def _make_request_data(self):
...
+            
+            with urllib.request.urlopen(request, timeout=5) as response:
+                result = response.read()
+                logger.debug(f"Slot cache cleared: {result}")
+            
+            print("Cache cleared. Waiting for server to stabilize...", flush=True)
+            time.sleep(1)  # Give server time to free memory
+            
+            # Step 2: Now summarize the conversation (server has capacity now)
+            print("Summarizing conversation...", flush=True)
```
rhatdan: Debug?

ramalama/chat.py
```diff
@@ -429,12 +539,63 @@ def _make_request_data(self):
...
+            summarized = self._summarize_conversation()
+            if not summarized:
+                logger.warning("Cannot summarize - using truncation instead")
+                # Fallback: keep only system messages or clear completely
+                if self.conversation_history and self.conversation_history[0].get("role") == "system":
+                    self.conversation_history = [self.conversation_history[0]]
+                else:
+                    self.conversation_history = []
+                print("Conversation history cleared.", flush=True)
```
rhatdan: Debug?

ramalama/chat.py
```diff
@@ -429,12 +539,63 @@ def _make_request_data(self):
...
+            
+            print("Ready to retry...", flush=True)
```
rhatdan: Debug

ramalama/chat.py
```diff
@@ -429,12 +539,63 @@ def _make_request_data(self):
...
+            return True
+            
+        except Exception as e:
+            logger.warning(f"Failed to clear slot cache: {e}")
+            print(f"Warning: Failed to clear cache: {e}", flush=True)
```
rhatdan: Logger should have handled this.

ramalama/chat.py
```diff
@@ -455,15 +638,99 @@ def _req(self):
...
+                retry_request = self._make_request_data()
+                try:
+                    retry_response = urllib.request.urlopen(retry_request, timeout=30)
+                    return res(retry_response, self.args.color)
+                except Exception as retry_error:
+                    logger.warning(f"Retry failed after cache clear: {retry_error}")
+                    perror(f"\rError: Failed to complete request after clearing cache: {retry_error}")
+                    return None
+            else:
+                perror("\rError: Context size exceeded and failed to clear cache")
```
rhatdan, line 508: Should this be raised?

ramalama/chat.py
```diff
@@ -455,15 +638,99 @@ def _req(self):
...
```
rhatdan, line 530: Should this be raised?

ramalama/chat.py
```diff
@@ -455,15 +638,99 @@ def _req(self):
...
+                return None
+
+        # If we couldn't connect and have a long conversation history,
+        # it might be a context size issue that caused the server to hang/crash
+        if not response:
+            # Check if we have a long conversation that might have caused issues
+            if len(self.conversation_history) > 3:
+                logger.debug("Connection failed with long conversation history - attempting recovery")
+                print("\rConnection failed. Conversation history may be too large. Attempting recovery...", flush=True)
```
rhatdan, line 539: logger.warn

ramalama/chat.py
```diff
@@ -455,15 +638,99 @@ def _req(self):
...
+                        # Wait a moment for server to stabilize
+                        time.sleep(3)
+                    
+                    # Retry the request with summarized history
+                    logger.debug("Retrying request after recovery")
+                    retry_request = self._make_request_data()
+                    try:
+                        retry_response = urllib.request.urlopen(retry_request, timeout=30)
+                        result = res(retry_response, self.args.color)
+                        print("Recovery successful!", flush=True)
```
rhatdan, line 561: logger.debug

ramalama/chat.py
```diff
@@ -455,15 +638,99 @@ def _req(self):
...
+                        return result
+                    except Exception as retry_error:
+                        logger.warning(f"Retry failed after recovery: {retry_error}")
+                        perror(f"\rError: Failed to complete request after recovery")
```
rhatdan: Should this be raised or at least logger.warn

ramalama/chat.py
```diff
@@ -455,15 +638,99 @@ def _req(self):
...
+            
+            # Only show error and kill if not in initial connection phase and we didn't recover
+            if not getattr(self.args, "initial_connection", False):
+                perror(f"\rError: could not connect to: {self.url}")
```
rhatdan, line 569: Warn or debug?

bmahabirbu:
@rhatdan I overhauled this pr to make it more modular and added another technique based on https://medium.com/@balajibal/agent-context-management-why-simple-observation-masking-beats-llm-summarisation-4961cb67be89

This would a nice break off point to add sort of a ramalama plugin for different context strategies

ramalama/chat.py
```diff
@@ -40,12 +42,31 @@ def res(response, color):
...
             json_line = json.loads(line[len("data: ") :])
+
+            # Check for context size error or server errors that indicate context issues
+            if "error" in json_line:
+                error = json_line["error"]
+                error_type = error.get("type", "")
+                error_msg = error.get("message", "").lower()
+
+                # Treat these as context size errors - they indicate we need to summarize more aggressively
+                if error_type == "exceed_context_size_error":
```
olliewalsh, line 61: is this error type specific to llama.cpp?

ramalama/cli.py
```diff
@@ -1053,6 +1053,14 @@ def chat_run_options(parser):
         metavar="N",
         help="automatically summarize conversation history after N messages to prevent context growth (0=disabled)",
     )
+    parser.add_argument(
+        "--context-strategy",
+        choices=["observation_masking", "llm_summarizer"],
```
olliewalsh: Should `none` be an option too?

ramalama/chat_api_llama_server.py
```diff
@@ -0,0 +1,216 @@
...
+
+            with urllib.request.urlopen(health_url, timeout=timeout) as response:
+                if response.status == 200:
+                    return True
+        except Exception as e:
+            logger.debug(f"Health check failed: {e}")
+
+        return False
+
+    def wait_for_ready(self, timeout: float = 30.0, check_interval: float = 0.5) -> bool:
```
olliewalsh: There are a lot of hard-coded timeouts. Do they need to be configurable?

ramalama/chat.py
bmahabirbu: Yes it is! its a crude work around to catch the error maybe i can use the --metrics tab instead to get a handle for context usage

ramalama/cli.py
bmahabirbu: My idea is if a user doesnt specify --context-strategy then it wouldnt use the new code. 

hmm maybe a rename is needed. we can rename this to --smart-context-strategy or scs something similar to be more clear

ramalama/chat_api_llama_server.py
bmahabirbu: yes a good idea!

bmahabirbu:
https://block.github.io/goose/docs/guides/sessions/smart-context-management/

based on this I can also add a clear mode a truncation mode (which should just enable llama.cpp context-shifting)

bmahabirbu:
sorry its been awhile I want to get this one in hopefully

rhatdan:
@olliewalsh @mikebonnet @engelmi PTAL
/gemini review

rhatdan:
@ieaves PTAL

ramalama/summarizer/base.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def estimate_tokens(self, text: str) -> int:
+        """Rough estimate of token count from text (approx 4 chars per token)."""
+        return len(text) // 4
+
+    def estimate_history_tokens(self, history: list[dict]) -> int:
+        """Estimate total tokens in conversation history."""
+        total = 0
+        for msg in history:
+            total += self.estimate_tokens(msg.get("content", ""))
```
ieaves, line 43: This is tough because depending on the API / supported API capabilities I don't think we can either rely on the message content as the only source of token usage or even that the root `content` field will always contain the relevant context (e.g. tool calling).

That being said, unless I'm misremembering the API response should contain total token usage in which case we only ever need to estimate token usage of the next API response.

ramalama/summarizer/llm_summarizer.py
```diff
@@ -0,0 +1,292 @@
...
+
+        # Separate messages into categories
+        original_system_prompt = None
+        previous_summary = None
+        conversation_messages = []
+
+        for msg in history:
+            if self._is_summary_message(msg):
+                previous_summary = self._extract_summary_content(msg.get("content", ""))
+            elif msg.get("role") == "system" and original_system_prompt is None:
```
ieaves, line 80: I could be totally off on this but my understanding is that system messages are sequentially appended to each other before processing other message tokens. 

This makes it difficult to fully reproduce for the purposes of summarization since model behavior varies over the course of the conversation but given that this is an atypical workflow I imagine that distinction is somewhat unimportant. That being said, if the conversation involves multiple system prompts I think we would need the new prompt to respect subsequent system instructions. Perhaps something like this?

```python
system_prompt = "\n".join(system_messages)
```

This same logic likely applies to the `developer` role if you were interested in adding logic for it.

ramalama/summarizer/llm_summarizer.py
```diff
@@ -0,0 +1,292 @@
...
+                original_system_prompt = msg
+            elif msg.get("role") in ("user", "assistant"):
```
ieaves, line 82: Since we have tool calling supported in various places (like via mcp) it'd be nice to have those included in the summary.

ramalama/summarizer/base.py
```diff
@@ -0,0 +1,109 @@
...
+        Returns:
+            Consolidated history
+        """
+        if len(history) < 2:
+            return history
+
+        consolidated = []
+        pending_user_contents = []
+
+        for msg in history:
```
ieaves: As written I think this modifies the context of the conversation in an unintended way. Take the following sequence.

```
system
user-1
user-2
assistant-1 <- (response to 2, with context 1)
user-3
user-4
```

This function would rewrite the sequence to

```
system
user-1,2
assistant-1 <- (implies this is a response to message `user-1,2` not to `user-2`)
user-3,4
```

Is that what you were going for? I think all we actually care about is consolidating the last last messages without a response. i.e. something like

```
for i, msg in enumerate(history[::-1]):
    if msg.get("role") != "user":
        break

if i > 0:
    consolidated = history[0:-i]
    merged_content = "\n\n".join(msg.get("content", "") for msg in history[-i::])
    consolidated.append({"role": "user", "content": merged_content})
else:
   consolidated = history
```

ramalama/summarizer/llm_summarizer.py
```diff
@@ -0,0 +1,292 @@
...
+                headers={"Content-Type": "application/json"},
+                method="POST",
+            )
+
+            with urllib.request.urlopen(request, timeout=self.timeout) as response:
+                result = json.loads(response.read())
+                summary = result['choices'][0]['message']['content'].strip()
+
+                # Validate summary is useful
+                confusion_phrases = [
```
ieaves: I'm honestly not sure how much value this logic really adds. Some of these stop conditions would probably never trigger. For example,

`there is no` I'm assuming is short for `There is no content to work with` but the full string is 32 characters which would bypass the 20 character check. Even `I can’t assist with that` would exceed the 20 character limit.

One option is to increase the len(summary) check but that starts forcing a tradeoff in true positive vs. false positive rates and I think your decision to bias against false positives with a low character limit was wise. 

All of that is to say, was this added as a prophylactic or did you see an issue in your testing?  If a prophylactic I'd drop it but if systematic (and not a logic bug) you could potentially look at something like TFIDF as a more robust heuristic.

An alternative is to just call the LLM again with the summary (or a portion of the summary) asking if the response looks like a refusal.

ramalama/summarizer/observation_masking.py
```diff
@@ -0,0 +1,173 @@
...
+            max_history: Optional override for max messages to retain
+
+        Returns:
+            Reduced conversation history with recent messages preserved
+        """
+        # Use provided max_history or fall back to configured value
+        if max_history is not None:
+            old_max = self.max_context_messages
+            self.max_context_messages = max_history
+            result = self.reduce_context(history, n_ctx=8192)  # Default large context
```
ieaves: Should n_ctx be a configurable attribute? With the proliferation of ROPE and the like this would be a pretty small context for modern models.

ramalama/summarizer/observation_masking.py
```diff
@@ -0,0 +1,173 @@
...
+            self.max_context_messages = old_max
+            return result
+        return self.reduce_context(history, n_ctx=8192)
+
+    def reduce_context(self, conversation_history: list[dict], n_ctx: int) -> list[dict]:
```
ieaves: Please tell me if I'm misunderstanding something but from the medium post I understood the principle of observation masking to be that given a sequence of identical tool calls (say netstat, or reading a file) that only the most recent <n> truly remain relevant to the conversation and the others were functionally token bloat. This implementation seems to instead limit the models context to the most recent <n> messages. Are these actually the same thing? Is what you really want something that indexes invocations of a tool call and masks older output?


I can see the merit of raw message truncation as well but I'd be hesitant to make it the default strategy as it's a much stronger assumption.

ieaves (commented):
I mostly focused on the summarizer because it looked like Dan and Ollie were looking at the chat

ramalama/summarizer/base.py
bmahabirbu: Yaa this one is tough because when enabling metrics in llama.cpp i still wasnt able to get the estimated token usage properly and had to do this.

will recheck the latest llama.cpp version and update

ieaves: oh booo that's a bummer. Would make everything so much easier.

ramalama/summarizer/llm_summarizer.py
bmahabirbu: its a good call i think that would be a seperate pr that i can do!

ramalama/summarizer/base.py
bmahabirbu: Thank you this is a good catch

ramalama/summarizer/observation_masking.py
bmahabirbu: good catch it should be or at least be a percentage of the given ctx size

bmahabirbu: well actually was vibe coding this as it seems its not needed so removed since ctx is quered else where

ramalama/summarizer/observation_masking.py
bmahabirbu: Your right because observation masking relies on tool calling knowledge to be effective maybe it makes sense to remove this for now since tool calling would be a seperate pr is any

I realize goose cli can do alot for mcp frontend and i think ramalama shouldnt compete with that but thats a seperate discussion!

bmahabirbu:
@ieaves updated i removed some of the features and made the code a tad cleaner

ramalama/summarizer/llm_summarizer.py
ieaves: That's fine by me. At the moment tool calls are simply thrown away though. Could we create an issue for it just so we don't lose track of things?

rhatdan:
@bmahabirbu needs a rebase.

rhatdan:
I think this is still something we want.  Needs a big rebase though.

PR #2192, opened 2025-11-25
Add Provider Abstraction with support for Hosted API Calls

ieaves:
## Summary by Sourcery

Introduce a provider-agnostic chat abstraction and hosted API transport support while updating file loading and testing to use structured chat messages.

New Features:
- Add generic chat provider interfaces and OpenAI-specific implementations for both local-compatible and hosted APIs.
- Introduce an API transport for routing chat calls directly to hosted providers based on model URL schemes (e.g., openai://).

Enhancements:
- Refactor chat handling to use structured ChatMessage/parts, provider-based request creation, and unified streaming response parsing.
- Extend configuration with provider-specific settings (such as OpenAI API key resolution) and wire them into API transports.
- Adapt file loader and chat integration logic to emit and consume structured message parts for text and images.
- Update transports and CLI to recognize API-based models, disallow incompatible rag/serve modes, and route run commands through the new API transport.

Tests:
- Add and update unit tests for file loaders, chat integration, transport factory behavior, and the new API transport and provider abstractions.

### DIFF
--- a/docsite/package-lock.json
+++ b/docsite/package-lock.json
@@ -27,76 +27,6 @@
         "node": ">=18.0"
       }
     },
-    "node_modules/@ai-sdk/gateway": {
-      "version": "2.0.24",
-      "resolved": "https://registry.npmjs.org/@ai-sdk/gateway/-/gateway-2.0.24.tgz",
-      "integrity": "sha512-mflk80YF8hj8vrF9e1IHhovGKC1ubX+sY88pesSk3pUiXfH5VPO8dgzNnxjwsqsCZrnkHcztxS5cSl4TzSiEuA==",
-      "license": "Apache-2.0",
-      "dependencies": {
-        "@ai-sdk/provider": "2.0.1",
-        "@ai-sdk/provider-utils": "3.0.20",
-        "@vercel/oidc": "3.0.5"
-      },
-      "engines": {
-        "node": ">=18"
-      },
-      "peerDependencies": {
-        "zod": "^3.25.76 || ^4.1.8"
-      }
-    },
-    "node_modules/@ai-sdk/provider": {
-      "version": "2.0.1",
-      "resolved": "https://registry.npmjs.org/@ai-sdk/provider/-/provider-2.0.1.tgz",
-      "integrity": "sha512-KCUwswvsC5VsW2PWFqF8eJgSCu5Ysj7m1TxiHTVA6g7k360bk0RNQENT8KTMAYEs+8fWPD3Uu4dEmzGHc+jGng==",
-      "license": "Apache-2.0",
-      "dependencies": {
-        "json-schema": "^0.4.0"
-      },
-      "engines": {
-        "node": ">=18"
-      }
-    },
-    "node_modules/@ai-sdk/provider-utils": {
-      "version": "3.0.20",
-      "resolved": "https://registry.npmjs.org/@ai-sdk/provider-utils/-/provider-utils-3.0.20.tgz",
-      "integrity": "sha512-iXHVe0apM2zUEzauqJwqmpC37A5rihrStAih5Ks+JE32iTe4LZ58y17UGBjpQQTCRw9YxMeo2UFLxLpBluyvLQ==",
-      "license": "Apache-2.0",
-      "dependencies": {
-        "@ai-sdk/provider": "2.0.1",
-        "@standard-schema/spec": "^1.0.0",
-        "eventsource-parser": "^3.0.6"
-      },
-      "engines": {
-        "node": ">=18"
-      },
-      "peerDependencies": {
-        "zod": "^3.25.76 || ^4.1.8"
-      }
-    },
-    "node_modules/@ai-sdk/react": {
-      "version": "2.0.119",
-      "resolved": "https://registry.npmjs.org/@ai-sdk/react/-/react-2.0.119.tgz",
-      "integrity": "sha512-kl4CDAnKJ1z+Fc9cjwMQXLRqH5/gHhg8Jn9qW7sZ0LgL8VpiDmW+x+s8e588nE3eC88aL1OxOVyOE6lFYfWprw==",
-      "license": "Apache-2.0",
-      "dependencies": {
-        "@ai-sdk/provider-utils": "3.0.20",
-        "ai": "5.0.117",
-        "swr": "^2.2.5",
-        "throttleit": "2.1.0"
-      },
-      "engines": {
-        "node": ">=18"
-      },
-      "peerDependencies": {
-        "react": "^18 || ~19.0.1 || ~19.1.2 || ^19.2.1",
-        "zod": "^3.25.76 || ^4.1.8"
-      },
-      "peerDependenciesMeta": {
-        "zod": {
-          "optional": true
-        }
-      }
-    },
     "node_modules/@algolia/abtesting": {
       "version": "1.13.0",
       "resolved": "https://registry.npmjs.org/@algolia/abtesting/-/abtesting-1.13.0.tgz",
@@ -2902,9 +2832,9 @@
       }
     },
     "node_modules/@csstools/postcss-normalize-display-values": {
-      "version": "4.0.0",
-      "resolved": "https://registry.npmjs.org/@csstools/postcss-normalize-display-values/-/postcss-normalize-display-values-4.0.0.tgz",
-      "integrity": "sha512-HlEoG0IDRoHXzXnkV4in47dzsxdsjdz6+j7MLjaACABX2NfvjFS6XVAnpaDyGesz9gK2SC7MbNwdCHusObKJ9Q==",
+      "version": "4.0.1",
+      "resolved": "https://registry.npmjs.org/@csstools/postcss-normalize-display-values/-/postcss-normalize-display-values-4.0.1.tgz",
+      "integrity": "sha512-TQUGBuRvxdc7TgNSTevYqrL8oItxiwPDixk20qCB5me/W8uF7BPbhRrAvFuhEoywQp/woRsUZ6SJ+sU5idZAIA==",
       "funding": [
         {
           "type": "github",
@@ -5344,9 +5274,9 @@
       }
     },
     "node_modules/@types/express-serve-static-core": {
-      "version": "4.19.7",
-      "resolved": "https://registry.npmjs.org/@types/express-serve-static-core/-/express-serve-static-core-4.19.7.tgz",
-      "integrity": "sha512-FvPtiIf1LfhzsaIXhv/PHan/2FeQBbtBDtfX2QfvPxdUelMDEckK08SM6nqo1MIZY3RUlfA+HV8+hFUSio78qg==",
+      "version": "4.19.8",
+      "resolved": "https://registry.npmjs.org/@types/express-serve-static-core/-/express-serve-static-core-4.19.8.tgz",
+      "integrity": "sha512-02S5fmqeoKzVZCHPZid4b8JH2eM5HzQLZWN2FohQEy/0eXTq8VXZfSN6Pcr3F6N9R/vNrj7cpgbhjie6m/1tCA==",
       "license": "MIT",
       "dependencies": {
         "@types/node": "*",
@@ -5461,15 +5391,9 @@
       "license": "MIT"
     },
     "node_modules/@types/node": {
-<<<<<<< HEAD
       "version": "25.0.10",
       "resolved": "https://registry.npmjs.org/@types/node/-/node-25.0.10.tgz",
       "integrity": "sha512-zWW5KPngR/yvakJgGOmZ5vTBemDoSqF3AcV/LrO5u5wTWyEAVVh+IT39G4gtyAkh3CtTZs8aX/yRM82OfzHJRg==",
-=======
-      "version": "25.0.3",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-25.0.3.tgz",
-      "integrity": "sha512-W609buLVRVmeW693xKfzHeIV6nJGGz98uCPfeXI1ELMLXVeKYZ9m15fAMSaUPBHYLGFsVRcMmSCksQOrZV9BYA==",
->>>>>>> ca824d98 (adds support for hosted chat providers)
       "license": "MIT",
       "dependencies": {
         "undici-types": "~7.16.0"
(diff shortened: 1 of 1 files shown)


ramalama/chat_providers/base.py
```diff
@@ -0,0 +1,161 @@
...
+    model: str | None = None
+    temperature: float | None = None
+    max_tokens: int | None = None
+    stream: bool = True
+    extra: Mapping[str, Any] | None = None
+
+    def to_dict(self) -> dict[str, Any]:
+        keys = ["model", "temperature", "max_tokens", "stream"]
+        result = {k: v for k in keys if (v := getattr(self, k)) is not None}
+        result |= {} if self.extra is None else dict(self.extra)
```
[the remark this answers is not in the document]
ieaves: I'm already using a union operator. Your proposed change has identical semantics but requires a reader to parse four operations in a single line. It's much harder to understand.

ieaves:
@rhatdan would you be willing to take a look at this? The primary goal was supporting multiple API backends with different chat implementations but core improvements include:

* Generic chat provider interfaces to provide compatibility with different API interfaces beyond static /v1/chat/completions
* Implementation of the [Responses](https://platform.openai.com/docs/api-reference/responses) interface with an example implementation of model calling to a remote API (openai, in this case).
* A new `Spinner` class which spins GUI updates off into a separate thread. The previous implementation coupled the loading animations to api calls
* Updated documentation and associated tests

rhatdan:
@olliewalsh PTAL

ieaves:
Hey @olliewalsh is there any chance you had an opportunity to look at this?

docs/ramalama.conf
```diff
@@ -197,6 +197,15 @@
...
 #max_retry_delay = 30
+
+# Provider-specific hosted API configuration
+#
+#[ramalama.provider]
+#openai_api_key = ""
+#
+# Optional provider-specific API key used when contacting OpenAI-hosted
+# transports. If unset, RamaLama falls back to the RAMALAMA_API_KEY value
+# or environment variables expected by the provider.
```
olliewalsh, line 218: Description should be above the conf option. Missing an empty line after

docs/ramalama.conf.5.md
```diff
@@ -238,6 +238,15 @@ The maximum number of times to retry a failed download
 
 The maximum delay between retry attempts in seconds
 
+## RAMALAMA.PROVIDER TABLE
+The `ramalama.provider` table configures hosted API providers that RamaLama can proxy to.
+
+`[[ramalama.provider]]`
```
olliewalsh, line 259: Might be cleaner to have a config section per provider e.g:
```
[[ramalama.provider.openai]]
api_key=...
```

ramalama/cli.py
```diff
@@ -1133,14 +1147,21 @@ def run_cli(args):
         except Exception as exc:
             raise e from exc
 
+    is_api_transport = isinstance(model, APITransport)
```
olliewalsh, line 1194: Not sure this should be exposed. `ramalama run` is essentially `ramalama serve` + `ramalama chat`. I only expected this to be an option for chat.

shortnames/shortnames.conf
```diff
@@ -61,3 +61,4 @@
   "smolvlm:500m" = "hf://ggml-org/SmolVLM-500M-Instruct-GGUF"
   "tiny" = "hf://TheBloke/TinyLlama-1.1B-Chat-v1.0-GGUF"
   "tinyllama" = "hf://TheBloke/TinyLlama-1.1B-Chat-v1.0-GGUF"
+  "gpt-5.1" = "openai://gpt-5.1-2025-11-13"
```
olliewalsh, line 64: what happens if I try to `ramalama pull` this?

ramalama/transports/api.py
```diff
@@ -0,0 +1,95 @@
+from __future__ import annotations
```
olliewalsh: Is this necessary? I don't see anything obvious that would require it.

docs/ramalama.conf.5.md
ieaves: Love it.

shortnames/shortnames.conf
ieaves: It'll return the default not implemented error currently, e.g.

```
ramalama pull openai://gpt-5.1-2025-11-13
Error: ramalama pull for 'APITransport' not implemented
```

I can add a custom error message to make this more explicit though.

ramalama/cli.py
ieaves: From a purity perspective I understand / agree with this but in practice I don't think most users (particularly new users) really understand this distinction. In practice `run` is just the most convenient way to immediately drop into a chat interface with a running model otherwise they would go straight to `ramalama serve`.

shortnames/shortnames.conf
ieaves: What do you think about something like this?

```
ramalama pull gpt-5.1                    
gpt-5.1-2025-11-13 is provided over a hosted API preventing direct pulling of the model file.
```

ieaves:
Thanks so much @olliewalsh! I've pushed changes for all of the comments except `run`. My instinct is that from a user journey perspective allowing users to "run" a hosted api is still the right path but I understand the potential risk of confusion depending on how we read the term. I'm not adamant at all, I just wanted to raise the point and get your thoughts.

ieaves:
@mikebonnet Is this okay to merge? If not, can you share any logs related to the failing checks?

ieaves:
@olliewalsh I'd really appreciate some help getting this merged if you didn't mind giving a sign off.

rhatdan:
@ieaves Linter is failing

ieaves:
@rhatdan should be fixed, I've also squashed.

.github/workflows/pypi.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       artifact-metadata: write
     steps:
       - name: Harden Runner
-        uses: step-security/harden-runner@20cf305ff2072d973412fa9b1e3a4f227bda3c76 # v2.14.0
+        uses: step-security/harden-runner@df199fb7be9f65074067a9eb93f12bb4c5547cf2 # v2.13.3
```
olliewalsh: I assume these dependency changes were committed by accident

.tekton/pipelines/pull-request-pipeline.yaml
```diff
@@ -151,7 +151,7 @@ spec:
       - name: name
         value: git-clone-oci-ta
       - name: bundle
-        value: quay.io/konflux-ci/tekton-catalog/task-git-clone-oci-ta:0.1@sha256:56f65a16d3d0485c64ad85af2c1f3e9b0bb4d02d63f2fd0ebb9498d219ca723d
+        value: quay.io/konflux-ci/tekton-catalog/task-git-clone-oci-ta:0.1@sha256:0a89e1a6304076525e9766f63a4cd006763d21d5aca6863281fc427537a23c6f
```
olliewalsh: I assume these dependency changes were committed by accident

.tekton/pipelines/push-pipeline.yaml
```diff
@@ -139,7 +139,7 @@ spec:
       - name: name
         value: git-clone-oci-ta
       - name: bundle
-        value: quay.io/konflux-ci/tekton-catalog/task-git-clone-oci-ta:0.1@sha256:56f65a16d3d0485c64ad85af2c1f3e9b0bb4d02d63f2fd0ebb9498d219ca723d
+        value: quay.io/konflux-ci/tekton-catalog/task-git-clone-oci-ta:0.1@sha256:0a89e1a6304076525e9766f63a4cd006763d21d5aca6863281fc427537a23c6f
```
olliewalsh: I assume these dependency changes were committed by accident

ramalama/chat_providers/base.py
```diff
@@ -0,0 +1,182 @@
...
+@dataclass(slots=True)
+class ChatStreamEvent:
+    """A provider-agnostic representation of a streamed delta."""
+
+    text: str | None = None
+    raw: dict[str, Any] | None = None
+    done: bool = False
+
+
+class ChatProviderError(RuntimeError):
```
olliewalsh: Seems unusual to base it on RuntimeError, why not Exception?

ramalama/chat_providers/openai.py
```diff
@@ -0,0 +1,333 @@
...
+    SystemMessage,
+    ToolMessage,
+    UserMessage,
+    serialize_part,
+)
+
+
+@singledispatch
+def message_to_completions_dict(message: Any) -> dict[str, Any]:
+    raise ValueError(f"Undefined message type {type(message)}")
```
olliewalsh: might be better to create a specific exception type that we can catch (I know KeyError/ValueError are currently abused throughout the codebase, but hoping to resolve that soon)

Makefile
```diff
@@ -4,7 +4,11 @@ SELINUXOPT ?= $(shell test -x /usr/sbin/selinuxenabled && selinuxenabled && echo
 PREFIX ?= /usr/local
 BINDIR ?= ${PREFIX}/bin
 SHAREDIR ?= ${PREFIX}/share
-PYTHON ?= $(shell command -v python3 python|head -n1)
+# PYTHON ?= $(shell command -v python3 python|head -n1)
```
olliewalsh: what is the purpose of this change?

test/conftest.py
```diff
@@ -7,6 +7,8 @@
 
 import pytest
 
+os.environ.setdefault("RAMALAMA_CONFIG", "/dev/null")
```
olliewalsh: what is the purpose of this change?

ramalama/cli.py
olliewalsh: ok, I guess the alternative would be to return an error telling the user to run `ramalama chat ...`

Makefile
ieaves: Because we otherwise don't respect venv's.

test/conftest.py
ieaves: Because test behavior is modified by the presence or absence of user specific configuration files. This breaks test assumptions in a variety of places and in particular tests of configuration file specific behavior.

ramalama/chat_providers/base.py
ieaves: The only time it's raised is when the payload structure was undefined. That's not generally a recoverable error. I can subclass to anything though.

.github/workflows/pypi.yml
ieaves: Yes! I'll fix this.

ramalama/cli.py
ieaves: If having functionality on `run` is a sticking point I'll take it out and only expose through chat.

test/conftest.py
olliewalsh: @telemaco FYI

ramalama/chat_providers/base.py
olliewalsh: Generally I have not seen any of the built-in exceptions by sub-classed by application exception types, apart from the Exception type. Can wait until I work on fixing up all of the exception handling logic though.

test/conftest.py
olliewalsh: I think it would be best to create a separate PR for this and the venv fix, the PR is already doing a lot

ramalama/cli.py
olliewalsh: nah, I'm not 100% sure about it but it's ok

test/unit/test_file_loader.py
```diff
@@ -269,9 +278,10 @@ def test_builder_load_text_files_only(self):
             messages = builder.load(tmp_file.name)
 
             assert len(messages) == 1
-            assert messages[0]["role"] == "system"
-            assert "Test content" in messages[0]["content"]
-            assert f"<!--start_document {tmp_file.name}-->" in messages[0]["content"]
+            assert messages[0].role == "user"
```
olliewalsh, line 281: why was this previously system role and now user role?

test/unit/test_file_loader_integration.py
```diff
@@ -32,9 +37,10 @@ def test_chat_with_file_input_single_file(self, mock_urlopen):
             # Check that the system message was added to conversation history
             assert len(shell.conversation_history) == 1
             system_message = shell.conversation_history[0]
-            assert system_message["role"] == "system"
-            assert "This is test content for chat input" in system_message["content"]
-            assert f"<!--start_document {tmp_file.name}-->" in system_message["content"]
+            assert system_message.role == "user"
```
olliewalsh: nit: i would remove "system" from the var names and comment if it's a user role message

test/conftest.py
ieaves: Okay, I can take it out. I've previously added `conftest` autouse fixtures to mange user configuration settings provisioned through environment variables so that we weren't having to monkeypatch the CONFIG file everywhere. 

On my local I make a point to regularly run using both docker and podman and it ends up being easier to modify the `ramalama.conf` to set the engine in a single place rather than setting env vars in multiple terminals.

ramalama/chat_providers/base.py
ieaves: Oh really? It's not a particularly rare pattern in my experience (e.g. [click](https://github.com/pallets/click/blob/cdab890e57a30a9f437b88ce9652f7bfce980c1f/src/click/exceptions.py#L294), [trio](https://github.com/python-trio/trio/blob/01243945fc3d46d4483acf486177389e59fdc6b6/src/trio/_core/_exceptions.py#L39), [pandas](https://github.com/pandas-dev/pandas/blob/a1fba1b9454a469a18bde73ac62bb5da9ee30727/pandas/errors/__init__.py#L7890))

test/unit/test_file_loader.py
ieaves: It's been awhile since the original file loading PR but I think the original thinking was that the user directed the chat to begin with the document context pre-populated (e.g. `ramalama chat --rag ./path/to/files`) so it's not exactly a user message. In reality that doesn't make sense though, the context isn't itself a system directive. It's still user directed context more akin to the user intentionally uploading a file in the chat GUI (just through the CLI).

olliewalsh: ack, agree it should be user role

ieaves:
@olliewalsh HUGE thanks for taking the time to look at this today. Seriously, I know it's a big PR and going through this is time consuming. 

The latest push should resolve the items you identified. I've reverted the make file and conftest changes, introduced custom error types for the APIs (for now they live near where they are used but it sounds like you're planning to PR a central home for us to define custom errors in the future which would be great), fixed the test naming issues, and reverted the bad rebase changes to .tekton and workflows.


The only thing I left was related to `ramalama run` since we didn't land on a firm position one way or another. For what it's worth, even if we merge this as is if it becomes a problem I'm 100% onboard removing it from run in the future.

olliewalsh:
@ieaves https://github.com/containers/ramalama/pull/2313 should fix the windows e2e tests

test/conftest.py
telemaco: yeah, I'd separate this and the venv fix in another PR.

ieaves:
@olliewalsh awesome, thank you! I saw the sigkill 137 in the error and assumed it was getting OOM killed (which also happens on mac with smollm:135m).

olliewalsh:
> @olliewalsh awesome, thank you! I saw the sigkill 137 in the error and assumed it was getting OOM killed (which also happens on mac with smollm:135m).

It's merged, so should pass CI now if you revert the most recent commit

.tekton/integration/tasks/test-vm-cmd.yaml
```diff
@@ -111,16 +111,12 @@ spec:
         EXTRA_ARGS+=("-v" "/tmp/nvidia.yaml:/etc/cdi/nvidia.yaml")
       fi
       podman pull $TEST_IMAGE
-
```
olliewalsh, line 127: Was a merge from main accidentally included when squashing?

ieaves: I rebased to main, I might have forgotten to pull the latest first?

docsite/package-lock.json
```diff
@@ -5391,9 +5461,15 @@
       "license": "MIT"
     },
     "node_modules/@types/node": {
+<<<<<<< HEAD
```
olliewalsh: relic from a merge conflict, also seems to be downgrading packages

olliewalsh:
automerge-- just merged this before CI ran successfully. @ieaves if the CI jobs are failing can you push a follow-up PR?

Issue #2089, opened 2025-10-31
RAG mode is using cuda in a ROCm system


jalberto:
### Issue Description

when running ramalama in fedora silverblue on AMD (strix halo) it tries to use CUDA and crash

### Steps to reproduce the issue

```sh
ramalama rag ./foo.pdf local/myrag

➜ ramalama rag ./foo.pdf local/myrag
Converting foo.pdf .   2025-10-31 12:24:24,154 - INFO - detected formats: [<InputFormat.PDF: 'pdf'>]
2025-10-31 12:24:24,162 - INFO - Going to convert document batch...
2025-10-31 12:24:24,162 - INFO - Initializing pipeline for StandardPdfPipeline with options hash 75463f421d05cb4304e1f714cf00d35d
2025-10-31 12:24:24,166 - INFO - Loading plugin 'docling_defaults'
2025-10-31 12:24:24,166 - INFO - Registered picture descriptions: ['vlm', 'api']
2025-10-31 12:24:24,170 - INFO - Loading plugin 'docling_defaults'
2025-10-31 12:24:24,170 - INFO - Registered ocr engines: ['auto', 'easyocr', 'ocrmac', 'rapidocr', 'tesserocr', 'tesseract']
Converting foo.pdf ..   2025-10-31 12:24:27,612 - INFO - Accelerator device: 'cuda:0'
Converting foo.pdf ...   2025-10-31 12:24:41,667 - INFO - Accelerator device: 'cuda:0'
Converting foo.pdf .   2025-10-31 12:24:42,157 - INFO - Processing document foo.pdf
2025-10-31 12:24:42,317 - WARNING - Encountered an error during conversion of document c33927cdfcd4b22717a49932168d476cda8d350d6348494cff58ac69822272b7:
Traceback (most recent call last)
```

### Describe the results you received

it crash with a python stack trace

### Describe the results you expected

to use ROCm to access the GPU for processing

### ramalama info output

```yaml
"mistral": "hf://lmstudio-community/Mistral-7B-Instruct-v0.3-GGUF/Mistral-7B-Instruct-v0.3-Q4_K_M.gguf",
            "mistral-small3.1": "hf://bartowski/mistralai_Mistral-Small-3.1-24B-Instruct-2503-GGUF/mistralai_Mistral-Small-3.1-24B-Instruct-2503-IQ2_M.gguf",
[...]

PR #2211, opened 2025-12-04
doc2rag: if not using cuda, explicitly set the accelerator to cpu

mikebonnet:
docling doesn't support intel or rocm accelerators, and is incorrectly selecting cuda in some cases. Only enable cuda if ramalama has already detected it.

Fixes: #2089

## Summary by Sourcery

Adjust accelerator selection for doc2rag to explicitly use CPU when CUDA is not detected, avoiding incorrect CUDA selection by docling.

Bug Fixes:
- Prevent docling from incorrectly selecting CUDA accelerators on unsupported hardware by only enabling CUDA when it has been pre-detected.

Build:
- Update doc2rag container script configuration to default to CPU accelerator when CUDA is unavailable.

### DIFF
--- a/container-images/scripts/doc2rag
+++ b/container-images/scripts/doc2rag
@@ -18,6 +18,7 @@ from pathlib import Path
 import docling
 import qdrant_client
 from docling.chunking import HybridChunker
+from docling.datamodel.accelerator_options import AcceleratorDevice, AcceleratorOptions
 from docling.datamodel.base_models import InputFormat
 from docling.datamodel.pipeline_options import PdfPipelineOptions
 from docling.document_converter import DocumentConverter, PdfFormatOption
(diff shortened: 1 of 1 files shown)


container-images/scripts/doc2rag
```diff
@@ -40,8 +41,14 @@ class Converter:
 
     def __init__(self, args):
         # Docling Setup (Turn off OCR (image processing) for drastically reduced RAM usage and big speed increase)
-        pipeline_options = PdfPipelineOptions()
-        pipeline_options.do_ocr = args.ocr
+        if os.environ.get("CUDA_VISIBLE_DEVICES"):
+            dev = AcceleratorDevice.CUDA
+        else:
+            dev = AcceleratorDevice.CPU
```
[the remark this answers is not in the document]
mikebonnet: Keeping the more verbose format for readability, and in case we need to add more clauses in the future.

container-images/scripts/doc2rag
```diff
@@ -40,8 +41,14 @@ class Converter:
 
     def __init__(self, args):
         # Docling Setup (Turn off OCR (image processing) for drastically reduced RAM usage and big speed increase)
-        pipeline_options = PdfPipelineOptions()
-        pipeline_options.do_ocr = args.ocr
+        if os.environ.get("CUDA_VISIBLE_DEVICES"):
```
olliewalsh: IIRC this can be set to "","None", "-1" to disable CUDA

olliewalsh (changes requested):

container-images/scripts/doc2rag
mikebonnet: Updated to check for those values.

olliewalsh:
LGTM

PR #2247, opened 2025-12-16
Revert "Workaround for CUDA image not pointing to libcuda.so.1 in ld.so.conf"

olliewalsh:
This reverts commit a22e209551cd1e93de9ad830dac50147168a06d7 which breaks nvidia-container-toolkit 1.18.1-1 on Fedora43: `ggml_cuda_init: failed to initialize CUDA: system has unsupported display driver / cuda driver combination`

## Summary by Sourcery

Bug Fixes:
- Remove the CUDA ld.so.conf compat path hack that was causing issues with nvidia-container-toolkit on Fedora 43.


rhatdan:
LGTM

PR #2248, opened 2025-12-17
Fix quadlet generation for multi-part models

### LINKED ISSUE #2017: Quadlet generator should understand multi-part models

If you say this:

```
$ ramalama serve --generate quadlet gpt-oss:120b
$ grep -c ^Mount gpt-oss-120b-GGUF.container
2
```

…the output _should_ be 4 because this is a 3-part model, plus a small (17k) metadata file, all of which need to be bind-mounted into the container.

Since #1989, Ramalama can do this for `file://` paths of the pattern `$MODEL-0000N-of-0000M.gguf`, but it is failing in this instance, where the parts are in OCI artifact form, thus named with random SHA256 hashes.

Parse the small metadata file and generate `Mount=` declarations for each model part found, named so that the #1989 feature recognizes them by glob pattern.

The only such alternative I can see is hand-hacking the generated Quadlet file, and my attempts to do that resulted in startup errors. Ramalama is in a better position to know the right thing to do here than the admin.

### Additional context

_No response_

rhatdan:
Add support for generating multiple Mount= declarations in quadlet files for multi-part models (e.g., models split into multiple GGUF files).

Changes:
- Add _get_all_model_part_paths() method to Transport class to retrieve all parts of a multi-part model from the ref file
- Update Quadlet class to accept and store model_parts list
- Modify _gen_model_volume() to generate Mount= entries for each part
- Update generate_container_config() to pass model parts to quadlet
- Add test case for multi-part model quadlet generation

The fix ensures that when generating quadlet files for models like gpt-oss-120b that are split across multiple files, all parts are properly mounted into the container with correct src/dest paths.

Fixes #2017

## Summary by Sourcery

Ensure quadlet generation correctly mounts all parts of multi-part models instead of only a single model file.

Bug Fixes:
- Fix quadlet model volume generation so all GGUF parts of a multi-part model are mounted into the container, not just the entry file.
- Handle safetensor and OCI models consistently when determining model mount paths for quadlet and kube generation.

Tests:
- Add unit coverage for quadlet generation with a multi-part GGUF model, including expected container file output.

### DIFF
--- a/ramalama/quadlet.py
+++ b/ramalama/quadlet.py
@@ -16,12 +16,16 @@ class Quadlet:
         args,
         exec_args,
         artifact: bool,
+        model_parts: Optional[list[Tuple[str, str]]] = None,
     ):
         self.src_model_path, self.dest_model_path = model_paths
         self.src_chat_template_path, self.dest_chat_template_path = (
             chat_template_paths if chat_template_paths is not None else ("", "")
         )
         self.src_mmproj_path, self.dest_mmproj_path = mmproj_path if mmproj_path is not None else ("", "")
+
+        # Store all model parts for multi-part models
+        self.model_parts = model_parts if model_parts is not None else [(self.src_model_path, self.dest_model_path)]
 
         if self.src_model_path.startswith("oci://"):
             self.src_model_path = self.src_model_path.removeprefix("oci://")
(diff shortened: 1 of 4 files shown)


rhatdan:
@tangentsoft PTAL

PR #2256, opened 2025-12-18
Fix llama-stack oci runtime on CUDA

olliewalsh:
Typically the container runtime for CUDA is specified using the `podman --runtime` arg.
However lama-stack uses `podman kube play` which has no way to override the runtime. Use a temporary containers.conf file to do this instead.

## Summary by Sourcery

Ensure llamma-stack Podman integration supports configurable OCI runtimes, including CUDA setups, when using `podman kube play`.

Enhancements:
- Allow non-default Podman engine binaries whose names start with `podman` instead of requiring an exact `podman` basename.
- Introduce runtime selection via a temporary `containers.conf` override so CUDA/NVIDIA runtimes can be used with `podman kube play`.

### DIFF
--- a/ramalama/stack.py
+++ b/ramalama/stack.py
@@ -245,9 +245,9 @@ spec:
         if not oci_runtime:
             return exec_cmd(exec_args)
 
-        oci_runtime_name = (
-            run_cmd([self.args.engine, "info", "--format", "{{.Host.OCIRuntime.Name}}"]).stdout.decode("utf-8").strip()
-        )
+        oci_runtime_name = run_cmd(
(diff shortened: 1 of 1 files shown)


olliewalsh:
@rhatdan @bmahabirbu @engelmi @mikebonnet PTAL, would be good to merge this before cutting the 0.16 release

olliewalsh:
Previously this happened to work because of the workaround that mangled ldconfig within the container. Since https://github.com/containers/ramalama/pull/2247 we need to do this correctly using the nvidia-container-runtime

rhatdan:
LGTM

PR #2257, opened 2025-12-19
Use "with pytest.raises" in tests checking for expected exceptions.

jwieleRH:
Using the "with pytest.raises" context manager explicitly guarantees that the test will fail unless the expected exception is raised. "pytest.mark.xfail" without the strict parameter will merely log the unexpected behavior but will not necessarily fail the test.

Collapse the two invalid_log_level tests into one parametrized test.

## Summary by Sourcery

Tests:
- Combine separate invalid log level configuration tests into a single parametrized test using pytest.raises for the expected ValueError.

### DIFF
--- a/test/unit/test_config.py
+++ b/test/unit/test_config.py
@@ -181,14 +181,10 @@ def test_load_env_config_coerces_log_level():
     assert cfg["log_level"] == LogLevel.DEBUG
 
 
-@pytest.mark.xfail(raises=ValueError)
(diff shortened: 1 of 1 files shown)


rhatdan:
@telemaco PTAL

telemaco:
LGTM. I'd add a description about the test use cases. For instance `@pytest.mark.parametrize("invalid_level", ["notalevel", "InVaLiD"], id=["not existing level", "invalid case"])`

Issue #2277, opened 2026-01-06
Add support for Devstral 2 Small


Robsteranium:
### Feature request description

[Mistral released Devstral 2](https://mistral.ai/news/devstral-2-vibe-cli) last month. I'd like to run it with ramalama but I can't get it working with ramalama version 0.16.0.

Attempting to use a quantised version in GGUF format gives:

```
$ ramalama serve hf://unsloth/Devstral-Small-2-24B-Instruct-2512-GGUF
...
llama_model_load: error loading model: error loading model architecture: unknown model architecture: 'mistral3'
llama_model_load_from_file_impl: failed to load model
```

### Suggest potential solution

It looks to me like we would need to update `llama.cpp` to at least release [b7371](https://github.com/ggml-org/llama.cpp/releases/tag/b7371) which appears to merge [the PR adding support for Mistral](https://github.com/ggml-org/llama.cpp/pull/17945).

### Have you considered any alternatives?

I initially tried running Mistral's (unquantised/ non-gguf) release on hugging face but this fails:

```
$ ramalama serve hf://mistralai/Devstral-Small-2-24B-Instruct-2512

...
main: loading model
srv    load_model: loading model '/mnt/models'
llama_model_load_from_file_impl: using device CUDA0 (NVIDIA RTX PRO 4500 Blackwell) (0000:e1:00.0) - 29042 MiB free
gguf_init_from_file_impl: failed to read magic
llama_model_load: error loading model: llama_model_loader: failed to load model from /mnt/models
llama_model_load_from_file_impl: failed to load model
common_init_from_params: failed to load model '/mnt/models', try reducing --n-gpu-layers if you're running out of VRAM
srv    load_model: failed to load model, '/mnt/models'
srv    operator(): operator(): cleaning up before exit...
main: exiting due to model loading error
```

I presume that this is expected if `llama.cpp` requires the GGUF format, but it wasn't apparent from the ramalama documentation (the section on transports gives the impression that any HF URI would work).

Mistral recommends using vLLM. I tried this with ramalama but I get a different error (also for the GGUF URI above):

```
$ ramalama --runtime=vllm --debug serve hf://mistralai/Devstral-Small-2-24B-Instruct-2512
2026-01-06 14:12:35 - DEBUG - Checking if 8080 is available
2026-01-06 14:12:35 - DEBUG - run_cmd: nvidia-smi
2026-01-06 14:12:35 - DEBUG - Working directory: None
[...]

PR #2278, opened 2026-01-06
Bump the version of whisper.cpp and llama.cpp

rhatdan:
Fixes: https://github.com/containers/ramalama/issues/2277

## Summary by Sourcery

Build:
- Bump the default llama.cpp and whisper.cpp commit hashes in the container image build script to newer revisions.

### DIFF
--- a/container-images/scripts/build_llama_and_whisper.sh
+++ b/container-images/scripts/build_llama_and_whisper.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-DEFAULT_LLAMA_CPP_COMMIT="b52edd25586fabb70f0c21b274473b307cf14499"
-DEFAULT_WHISPER_COMMIT="c62adfbd1ecdaea9e295c72d672992514a2d887c"
+DEFAULT_LLAMA_CPP_COMMIT="968929528c6a05e10249366fbe5f0330ad9af678"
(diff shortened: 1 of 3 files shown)


container-images/scripts/build_llama_and_whisper.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-DEFAULT_LLAMA_CPP_COMMIT="b52edd25586fabb70f0c21b274473b307cf14499"
-DEFAULT_WHISPER_COMMIT="c62adfbd1ecdaea9e295c72d672992514a2d887c"
+DEFAULT_LLAMA_CPP_COMMIT="968929528c6a05e10249366fbe5f0330ad9af678"
```
olliewalsh, line 3: llama.cpp and whisper cuda builds are failing with cuda 12.8.1. Upgrading to cuda 12.9.1 should avoid the issue

olliewalsh (changes requested):

container-images/scripts/build_llama_and_whisper.sh
rhatdan: Is it safe now to make that change?

olliewalsh: Moving to 13.x was the concern IIRC

PR #2294, opened 2026-01-09
CI: fix race in test_pull.py::test_pull_with_registry

olliewalsh:
Previously this would sort the models on mtime and choose the first image as the oci image.
This was not correct, the oci image must be newer than the file:// used to create it.

However it works because the oci image mtime has seconds precision while file has ms.
If they are modified in the same second then the oci model will be "older" than the file model.
If not in the same second then oci model will be "newer".

The order isn't significant to the test assert, it just checks that oci model exists, so rewrite the test to avoid this issue completely.

## Summary by Sourcery

Tests:
- Change the pull-with-registry test to assert the presence of the fake model in the full model list rather than assuming it is the first entry.


telemaco:
LGTM

PR #2296, opened 2026-01-09
Enable GGML_HIP_ROCWMMA_FATTN in llamacpp ROCm build

olliewalsh:
Upstream llamacpp image now defaults to GGML_HIP_ROCWMMA_FATTN.

Requires ROCm 7.x (via f44) to get a sufficient version of rocwmma.

Also update AMDGPU_TARGETS targets to be consitent with upstream.

## Summary by Sourcery

Update the ROCm-based llamacpp container build to enable ROCWMMA fast attention and align GPU target architecture defaults with upstream.

New Features:
- Enable GGML_HIP_ROCWMMA_FATTN in the ROCm llamacpp build configuration.

Enhancements:
- Require newer ROCm toolchain packages, including rocwmma, and update default AMDGPU_TARGETS to match upstream settings.


container-images/rocm/Containerfile
```diff
@@ -1,4 +1,4 @@
-FROM quay.io/fedora/fedora:43
+FROM quay.io/fedora/fedora:44
```
rhatdan: Since 44 is not released yet, I think this is dangerous to do.

rhatdan (approved):

container-images/rocm/Containerfile
olliewalsh: F43 is still on ROCm 6.x though. Could maybe build the latest upstream from AMD but that would be pretty slow I expect

olliewalsh: could try just pulling the rocm rpms from 44/rawhide and keep everything else on 43

rhatdan:
LGTM

olliewalsh:
Taking this out of draft now that 0.17 is released

rhatdan:
@olliewalsh this one seems to have fallen through the cracks.

olliewalsh:
> @olliewalsh this one seems to have fallen through the cracks.

yeah, wasn't very urgent. Now that we have f44 I will ressurect it.

mikebonnet (approved):
Looks good!

PR #2309, opened 2026-01-13
Improving cold start time on cli invocation.

ieaves:
I noticed that basic ramalama commands were starting to feel sluggish so the goal with this PR was to reduce cold start time for actions taken in the cli. From my local testing commands like `ramalama --help`, `ramalama list`, and `ramalama info` were all ~70% faster. Additionally, on my laptop tests ran ~50% faster.

## Summary by Sourcery

Improve CLI startup performance and configuration handling by lazily loading heavy dependencies and centralizing access to runtime configuration.

New Features:
- Add cached helpers for default images, RAG images, and shortname resolution to avoid repeated computation.
- Introduce a configuration accessor that exposes a default port range derived from the configured port.

Enhancements:
- Refactor CLI and transport code to use lazy, per-function imports and cached helpers instead of importing heavy modules at startup.
- Replace global CONFIG usage with a cached get_config() accessor throughout the codebase for more consistent configuration access.
- Update runtime option defaults and daemon/stack behavior to be driven by the central configuration accessor rather than module-level constants.
- Cache accelerator detection and CUDA version checks to avoid repeated probing across commands.
- Adjust port computation to use the configurable default port range instead of a fixed module-level constant.
- Move shared type and option definitions into a dedicated config_types module for clearer separation of concerns.
- Improve MCP agent logging setup to respect the configured log level without globally reconfiguring logging unnecessarily.

Tests:
- Update unit tests to work with the new get_config() accessor, cached helpers, and configuration-driven defaults, including port range expectations and accelerator caching behavior.

### DIFF
--- a/ramalama/benchmarks/manager.py
+++ b/ramalama/benchmarks/manager.py
@@ -1,5 +1,4 @@
 import json
-import logging
 from dataclasses import asdict
 from functools import cached_property
 from pathlib import Path


ramalama/cli.py
```diff
@@ -829,7 +896,10 @@ def push_cli(args):
     source_model = _get_source_model(args, transport=transport)
 
     if args.TARGET:
+        shortnames = get_shortnames()
         target = shortnames.resolve(args.TARGET)
+    from ramalama.transports.transport_factory import New, TransportFactory
+
     target_model = New(target, args)
```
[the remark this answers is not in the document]
ieaves: Isn't target set by default on line 892 as args.SOURCE?

ramalama/transports/base.py
```diff
@@ -194,8 +184,11 @@ def model_type(self) -> str:
         return self._model_type
 
     @property
-    def model_store(self) -> ModelStore:
+    def model_store(self) -> "ModelStore":
         if self._model_store is None:
+            from ramalama.model_store.global_store import GlobalModelStore
```
olliewalsh: https://wiki.python.org/moin/PythonSpeed/PerformanceTips#Import_Statement_Overhead recommends this approach:

```
email = None

def parse_email():
    global email
    if email is None:
        import email
```

olliewalsh (commented):
Did you profile this? I wonder how much was due to global imports vs e.g calling out to nvidia-smi to check the cuda version

ramalama/transports/base.py
ieaves: IIRC there were four rounds of changes I made. The first included caching the startup checks like nvidia-smi which gave us ~60% faster cold start. The others each incrementally gave 3 to 8% improvement depending on the route (I was only testing help, info, and list). 

Want to move all of the imports back to the top level? It's not the most important improvement. I'd _strongly_ prefer not to introduce a bunch of global state though.

olliewalsh: ack, not sure if there is a more elegant way.... maybe best to split this up, and just land the nvidia-smi etc.. caching at first

ieaves:
Okay! This should be good to go!

ramalama/cli.py
```diff
@@ -1346,14 +1395,12 @@ def daemon_start_cli(args):
...
         "8080" if is_daemon_in_container else args.port,
         "--host",
-        CONFIG.host if is_daemon_in_container else args.host,
+        get_config().host if is_daemon_in_container else args.host,
     ]
     exec_cmd(daemon_cmd)
 
 
 def daemon_run_cli(args):
-    from ramalama.daemon.daemon import run
```
olliewalsh, line 1443: Looks like moving this created the circular import that now needs to be handled in daemon.py, so might as well leave it here

ramalama/daemon/handler/daemon.py
```diff
@@ -107,6 +106,8 @@ def _handle_get_tags(self, handler: http.server.SimpleHTTPRequestHandler):
         handler.wfile.flush()
 
     def _handle_post_serve(self, handler: http.server.SimpleHTTPRequestHandler):
+        from ramalama.cli import parse_args_from_cmd  # resolve circular import
```
olliewalsh: see other comment

ieaves:
Hey @olliewalsh could you approve this? We are all green

ramalama/cli.py
```diff
@@ -182,17 +200,16 @@ def init_cli():
...
     if any(arg in ("--dryrun", "--dry-run", "--generate") or arg.startswith("--generate=") for arg in sys.argv[1:]):
-        CONFIG.dryrun = True
+        config.dryrun = True
     # Phase 1: Parse the initial arguments to set CONFIG.runtime etc... as this can affect the subcommands
     initial_parser = get_initial_parser()
     initial_args, _ = initial_parser.parse_known_args(cmd)
     for arg in initial_args.__dict__.keys():
-        if hasattr(CONFIG, arg):
-            setattr(CONFIG, arg, getattr(initial_args, arg))
-    # Phase 2: Re-parse the arguments with the subcommands enabled
```
olliewalsh, line 195: Will merge this as-is, but need to restore the comment in a follow-up

PR #2339, opened 2026-01-22
Add benchmark metrics persistence

ieaves:
## Summary by Sourcery

Introduce persistent benchmark result storage and a new CLI for viewing historical benchmarks, while enriching bench output formatting and configuration support.

New Features:
- Add a `benchmarks` CLI command with subcommands (currently `list`) to view stored benchmark results in table or JSON formats with pagination.
- Persist `bench` command outputs as structured benchmark records, capturing device, configuration, and result metadata for later inspection.
- Add configuration support for benchmarks, including a configurable storage folder with sensible defaults.

Enhancements:
- Extend the `bench` command to support selectable output formats (table or JSON) and render richer tabular benchmark summaries including model parameters and throughput.
- Refine configuration file discovery via a cached helper and wire benchmark settings into the global config.
- Adjust CLI help/tests and llama.cpp benchmark engine spec so llama-bench emits JSON suitable for structured parsing.

Documentation:
- Document the new benchmarks configuration section and add a dedicated `ramalama-benchmarks(1)` man page, while updating existing man pages and examples to reference the new commands and options.

Tests:
- Add unit tests for the benchmarks manager and config documentation expectations, and update existing bench-related e2e/system tests for the new table headers and help behavior.

### DIFF
--- a/ramalama/command/context.py
+++ b/ramalama/command/context.py
@@ -2,13 +2,12 @@ import argparse
 import os
 from typing import Optional
 
-from ramalama.common import check_metal, check_nvidia
+from ramalama.common import check_metal, check_nvidia, get_accel_env_vars
 from ramalama.console import should_colorize
 from ramalama.transports.transport_factory import CLASS_MODEL_TYPES, New
 
 
 class RamalamaArgsContext:
-
     def __init__(self) -> None:
         self.cache_reuse: Optional[int] = None
         self.container: Optional[bool] = None
@@ -52,7 +51,6 @@ class RamalamaArgsContext:
 
 
 class RamalamaRagGenArgsContext:
-
     def __init__(self) -> None:
         self.debug: bool | None = None
         self.format: str | None = None
@@ -74,7 +72,6 @@ class RamalamaRagGenArgsContext:
 
 
 class RamalamaRagArgsContext:
-
     def __init__(self) -> None:
         self.debug: bool | None = None
         self.port: str | None = None
@@ -92,7 +89,6 @@ class RamalamaRagArgsContext:
 
 
 class RamalamaModelContext:
-
     def __init__(self, model: CLASS_MODEL_TYPES, is_container: bool, should_generate: bool, dry_run: bool):
         self.model = model
         self.is_container = is_container
@@ -128,7 +124,6 @@ class RamalamaModelContext:
 
 
 class RamalamaHostContext:
-
     def __init__(
         self, is_container: bool, uses_nvidia: bool, uses_metal: bool, should_colorize: bool, rpc_nodes: Optional[str]
     ):
@@ -140,7 +135,6 @@ class RamalamaHostContext:
 
 
 class RamalamaCommandContext:
-
     def __init__(
         self,
         args: RamalamaArgsContext | RamalamaRagGenArgsContext | RamalamaRagArgsContext,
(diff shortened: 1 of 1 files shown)


ieaves:
@olliewalsh something went wonky with #2237 and I had to open a new PR. Sorry! This contains the suggested documentation changes and refactor from sqlite to jsonl.

ramalama/transports/base.py
```diff
@@ -452,7 +460,54 @@ def setup_mounts(self, args):
...
+
+        output_format = getattr(args, "format", "table")
+
+        if args.dryrun:
+            if args.container:
+                self.engine.dryrun()
+            else:
+                dry_run(cmd)
+            escaped_cmd = [shlex.quote(arg) for arg in cmd]
+            result = subprocess.CompletedProcess(args=escaped_cmd, returncode=0, stdout="", stderr="")
```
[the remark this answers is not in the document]
ieaves: This is a false positive. 

subprocess.CompletedProcess(...) is just a data container that represents the result of a process. It does not spawn anything, so there is no command injection vector in that constructor call.

ramalama/config.py
```diff
@@ -136,6 +136,25 @@ def coerce_to_bool(value: Any) -> bool:
     raise ValueError(f"Cannot coerce {value!r} to bool")
 
 
+def get_default_benchmarks_storage_folder() -> Path:
+    conf_dir = None
+    for dir in DEFAULT_CONFIG_DIRS:
```
olliewalsh: this should use the store dir. The config dir may not be writable

ieaves: 👍. Ideally we'd be able to default the benchmarks storage directory to the user specified store path in this case. Unfortunately, `is_set` doesn't currently support checking nested subfields so I put it off and instead had it use the default store path as the default benchmarks base path.

The config is getting pretty complicated at this point. Do you think it'd be worth bringing something like pydantic in and managing this through the likes of [BaseSettings](https://docs.pydantic.dev/latest/concepts/pydantic_settings/)?

ieaves:
@olliewalsh This should be good to go if you're ready to merge.

ramalama/cli.py
```diff
@@ -509,10 +517,77 @@ def add_network_argument(parser, dflt: str | None = "none"):
...
+
+        if args.format == "json":
+            output = [asdict(item) for item in results]
+            print(json.dumps(output, indent=2, sort_keys=True))
+        else:
+            print_bench_results(results)
+
+    except MissingStorageFolderError:
+        print("Error: RAMALAMA__BENCHMARKS_STORAGE_FOLDER not configured")
+        sys.exit(1)
```
olliewalsh: Better to raise an exception and let main() handle the exit code

ieaves: Might make sense to pull that code out of the try/except altogether in that case.

ramalama/transports/base.py
```diff
@@ -463,7 +470,52 @@ def setup_mounts(self, args):
...
+
+        output_format = getattr(args, "format", "table")
+
+        if args.dryrun:
+            if args.container:
+                self.engine.dryrun()
+            else:
+                dry_run(cmd)
+
+            result = subprocess.CompletedProcess(args=cmd, returncode=0, stdout="", stderr="")
```
olliewalsh: Is this necessary? Everything else just returns None

ieaves: It's not necessary, it just keeps the typing consistent without the early return. I'll switch out to an early return instead though since it's cleaner.

PR #2345, opened 2026-01-25
cuda: reduce image size by only copying required binaries from the installation directory

mikebonnet:
The installation directory contains unused binaries and compilation objects that are significant in size. Only copy binaries that are required for the runtime operation of the image.

## Summary by Sourcery

Build:
- Update the CUDA container image build to copy only required executables and shared libraries from the builder image into the runtime image to reduce image size.

### DIFF
--- a/container-images/cuda/Containerfile
+++ b/container-images/cuda/Containerfile
@@ -11,8 +11,16 @@ RUN ./build_llama_and_whisper.sh cuda
 # Final runtime image
 FROM docker.io/nvidia/cuda:${CUDA_VERSION}-runtime-ubi9
 
-# Copy the entire installation directory from the builder
-COPY --from=builder /tmp/install /usr
+RUN --mount=type=bind,from=builder,source=/tmp/install,target=/tmp/install \
+    cp -a /tmp/install/bin/llama-bench \
+          /tmp/install/bin/llama-perplexity \
+          /tmp/install/bin/llama-quantize \
+          /tmp/install/bin/llama-server \
+          /tmp/install/bin/rpc-server \
+          /tmp/install/bin/whisper-server \
+          /tmp/install/bin/*.so* \
+          /usr/bin/ && \
+    cp -a /tmp/install/lib64/*.so* /usr/lib64/
 
 # Install python3.12 and ramalama to support a non-standard use-case
 RUN dnf -y install python3.12 && dnf -y clean all && ln -sf python3.12 /usr/bin/python3


container-images/cuda/Containerfile
```diff
@@ -11,8 +11,16 @@ RUN ./build_llama_and_whisper.sh cuda
...
+RUN --mount=type=bind,from=builder,source=/tmp/install,target=/tmp/install \
+    cp -a /tmp/install/bin/llama-bench \
+          /tmp/install/bin/llama-perplexity \
+          /tmp/install/bin/llama-quantize \
+          /tmp/install/bin/llama-server \
+          /tmp/install/bin/rpc-server \
+          /tmp/install/bin/whisper-server \
+          /tmp/install/bin/*.so* \
+          /usr/bin/ && \
+    cp -a /tmp/install/lib64/*.so* /usr/lib64/
```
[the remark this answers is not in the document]
mikebonnet: Unnecessary

container-images/cuda/Containerfile
```diff
@@ -11,8 +11,16 @@ RUN ./build_llama_and_whisper.sh cuda
 # Final runtime image
 FROM docker.io/nvidia/cuda:${CUDA_VERSION}-runtime-ubi9
 
-# Copy the entire installation directory from the builder
-COPY --from=builder /tmp/install /usr
+RUN --mount=type=bind,from=builder,source=/tmp/install,target=/tmp/install \
+    cp -a /tmp/install/bin/llama-bench \
```
olliewalsh, line 15: How much does this reduce the image size?

mikebonnet: It saves about 958M. 
```
$ echo $(( $(podman inspect --format '{{.Size}}' quay.io/redhat-user-workloads/ramalama-tenant/cuda:face0f30faf19b71e3c1c18e3bb38bc69bd1af53) - $(podman inspect --format '{{.Size}}' quay.io/redhat-user-workloads/ramalama-tenant/cuda:on-pr-8db0f4b5cb5d4b3154baa26bcc8fce6ec13eba67) ))
958469241
```

olliewalsh: Can we drop whisper from the build? It's statically linking the ggml cuda backend, so all of it's binaries are huge.

mikebonnet: I'm going to do that in a separate PR later today.

olliewalsh: ok, better to land that first? then should be ok to continue copying everything from the bin dir. For the lib dir still better to use the glob to omit the large ggml-cuda static lib

mikebonnet: I'd rather land this first. Do we want all the binaries compiled by `llama.cpp` to be present in the image? I feel like that increases the likelihood of people using the images in unexpected and unsupported ways.

olliewalsh: I would include them all, at least for now, to be consistent with the the upstream "full" image.

mikebonnet: Ok, that makes sense. I'll submit the whisper PR in a bit.

PR #2355, opened 2026-01-27
Adds ability to `inspect` shortnames

ieaves:
## Summary by Sourcery

Add support for inspecting and listing model shortnames via the CLI and extend inspect to handle unresolved shortname references.

New Features:
- Allow `ramalama inspect` to list available model shortnames via a new `--shortnames` flag that works without a model argument.
- Support inspecting models referenced by shortnames by resolving them when no ref file is found and formatting the result via a new ShortNameInfo helper.

Enhancements:
- Refactor bench command execution to reuse a shared execute_command helper.
- Update inspect metadata typing to use Dict for consistency.

Documentation:
- Document the optional model argument and the new `--shortnames` flag for `ramalama inspect`, including example usage.

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -685,11 +685,13 @@ def _list_models_from_store(args):
             size_sum += file.size
             last_modified = max(file.modified, last_modified)
 
-        ret.append({
-            "name": f"{model} (partial)" if is_partially_downloaded else model,
-            "modified": datetime.fromtimestamp(last_modified, tz=local_timezone).isoformat(),
-            "size": size_sum,
-        })
+        ret.append(
+            {
+                "name": f"{model} (partial)" if is_partially_downloaded else model,
+                "modified": datetime.fromtimestamp(last_modified, tz=local_timezone).isoformat(),
+                "size": size_sum,
+            }
+        )
 
     # sort the listed models according to the desired order
     ret.sort(key=lambda entry: entry[args.sort], reverse=args.order == "desc")


engelmi (commented):
The `info` command already provides the available shortnames, incl. the files where these are defined: 
```bash
$ ramalama info
...
"Shortnames": {
        "Files": [
            "/home/mengel/projects/containers/ramalama/shortnames/shortnames.conf",
            "/home/mengel/.local/share/ramalama/shortnames.conf",
            "/usr/share/ramalama/shortnames.conf"
        ],
        "Names": {
            "cerebrum": "huggingface://froggeric/Cerebrum-1.0-7b-GGUF/Cerebrum-1.0-7b-Q4_KS.gguf",
            "deepseek": "ollama://deepseek-r1",
...
```
If we want to list shortnames only, I'd suggest adding the `--shortnames` options to the `info` command.

ieaves:
@engelmi I was trying to make accessing shortname information follow the same pattern we use for getting information about other models. I can close this if we think that's out of scope though.

engelmi:
> @engelmi I was trying to make accessing shortname information follow the same pattern we use for getting information about other models. I can close this if we think that's out of scope though.

I'd say `inspect` looks at the specific AI model(s) and `info` is about RamaLama, its config and the system. Since the shortnames are configured by RamaLama, I'd put this into the `info` command. 
@ieaves WDYT about moving this to `ramalama info --shortnames`? We could even show from which config file the shortname is coming from (which is not visible at the moment)

ieaves:
@engelmi sure! We can do that. I updated the PR with custom printing for shortnames specifically. Take a look and see if it matches what you had in mind :).

engelmi:
@ieaves Sorry for the late reply. Looks great! :) 

Could you stash the commits? Then we can merge, in my point of view.

rhatdan:
I would prefer the 
ramalama info --shortnames option, to show all info about a shortname on a single line, that way making it more grep-able.
Something like:

$ ramalama info --shortnames		
smollm:135m = ollama://smollm:135m (/usr/share/ramalama/shortnames.conf)
tinyllama:latest = ollama://tinyllama:latest (/home/dwalsh/.config/ramalama/shortnames.conf)

Or
$ ramalama info --shortnames		
smollm:135m=ollama://smollm:135m (/usr/share/ramalama/shortnames.conf)
tinyllama:latest=ollama://tinyllama:latest (/home/dwalsh/.config/ramalama/shortnames.conf)

Or output in a json format, that would be easily parsible from command line.

ieaves:
@rhatdan switched it over to a flat structure

```
$ ramalama info --shortnames
smollm:135m=ollama://smollm:135m (/usr/share/ramalama/shortnames.conf)
tinyllama:latest=ollama://tinyllama:latest (/home/dwalsh/.config/ramalama/shortnames.conf)
```

rhatdan:
I don't see the PR as being updated?

ieaves:
Blech, sorry. I must have pushed without --force. Check again.

docs/ramalama-info.1.md
```diff
@@ -120,6 +124,15 @@ $ ramalama info
 }
 ```
 
+List available shortnames
+```
+$ ramalama info --shortnames
```
rhatdan, line 128: This man pages shows the old format.

ramalama/cli.py
```diff
@@ -699,6 +705,15 @@ def _list_models(args):
 
 def info_cli(args: DefaultArgsType) -> None:
     shortnames = get_shortnames()
+    if getattr(args, 'shortnames', None):
+        for name in sorted(shortnames.shortnames):
+            config_source = shortnames.config_sources.get(name)
+            source = shortnames.shortnames[name]
+
+            message = f"{name} = {source} ({config_source})"
```
rhatdan: Please change to:
```
            message = f"{name}={source} ({config_source})"
```

ieaves: I went ahead and made this change but could I ask what changed your mind from last week about spacing around the `=`? My take between the two options was that the spaced option was more readable for humans actually inspecting the output.

rhatdan: I am just looking at bash script-able again. Might not be a big issue.
```
$ foo = bar
bash: foo: command not found...
$ foo=bar
```
Also it becomes easier to handle this output from commands like awk and cut to only grab the first object.

```
$ ramalama info --shortnames | awk '/granite { print $1 }'
```
Not sure I did the syntax correctly.

PR #2358, opened 2026-01-28
Reduce CI load and fix unreliable tests

olliewalsh:
Limit CI workflow to a single concurrent run per github ref (branch or PR) - for PRs cancel the running job if the PR is updated.

Stop wasting resources running the expensive test jobs if lint or unit-tests fail.

Drop the bats tests, the e2e tests now cover this.
Revert a workaround that was merged for a bats test timeout.

Report top 10 test duration for unit and e2e tests.

Stop using pull time to determine if test_pull_using_ollama_cache used cached model as this often fails.

Use --pull never for generate tests.

## Summary by Sourcery

Limit CI workflows to a single concurrent run per ref, gate expensive jobs on fast checks, and clean up test and context logic to improve reliability and observability.

Enhancements:
- Simplify command context host initialization by always probing NVIDIA support instead of conditionally skipping GPU checks based on environment.

CI:
- Restrict CI concurrency to a single run per ref and auto-cancel in-progress workflow runs on updated pull requests.
- Make container build and all e2e test jobs depend on successful lint and unit-test jobs to avoid running heavy jobs on failing changes.
- Remove bats-based CI jobs across Linux, Docker, and macOS in favor of existing e2e coverage while retaining the container image build job.

Tests:
- Add top-10 slow test duration reporting for both unit and e2e pytest runs to aid performance troubleshooting.
- Stabilize e2e pull tests by asserting on explicit cached-output text instead of comparing pull timings.
- Ensure serve generation tests (e2e and bats) use `--pull never` to avoid unintended network pulls and flakiness.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,47 +64,6 @@ jobs:
         run: |
           uv run -- make unit-tests
 
-  build-image:
-    needs:
-      - lint
-      - unit-test
-    runs-on: ubuntu-24.04
-    timeout-minutes: 60
-    steps:
-      - uses: actions/checkout@v6
-
-      - name: Install the latest version of uv
-        uses: astral-sh/setup-uv@v7
-
-      - name: Install dependencies
-        shell: bash
-        run: |
-          df -h
-          sudo apt-get update
-          sudo apt-get install -y bash codespell pipx podman
-          uv run -- make install-requirements
-
-      - name: Upgrade to podman 5
-        run: |
-           set -e
-           # /mnt has ~ 65 GB free disk space. / is too small.
-           sudo mkdir -m a=rwx -p /mnt/tmp /mnt/runner
-           sudo mkdir -m o=rwx -p /home/runner/.local
-           sudo chown runner:runner /mnt/runner /home/runner/.local
-           sudo mount --bind /mnt/runner /home/runner/.local
-           # Enable universe repository which contains podman
-           sudo add-apt-repository "deb http://old-releases.ubuntu.com/ubuntu oracular universe"
-           # Update package lists
-           sudo apt-get update
-           sudo apt-get purge firefox
-           # Install specific podman version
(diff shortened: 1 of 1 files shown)


ramalama/command/context.py
```diff
@@ -163,13 +169,9 @@ def from_argparse(cli_args: argparse.Namespace) -> "RamalamaCommandContext":
...
         else:
             model = None
-
-        skip_gpu_probe = should_generate or bool(get_accel_env_vars())
-        uses_nvidia = True if skip_gpu_probe else (check_nvidia() is None)
-
         host = RamalamaHostContext(
             is_container,
-            uses_nvidia,
+            check_nvidia() is None,
```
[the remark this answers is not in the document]
olliewalsh: Indeed, but that came from a revert commit so I'll address it in a follow-up PR

olliewalsh:
@mikebonnet PTAL

.github/workflows/ci.yml
```diff
@@ -98,7 +64,10 @@ jobs:
         run: |
           uv run -- make unit-tests
 
-  bats:
+  build-image:
```
mikebonnet: Do we need to build the image in Github Actions at all? Are the e2e tests using this image, or just pulling the `:<version>` image from quay.io anyway?

mikebonnet (approved):
One question, but otherwise LGTM!

.github/workflows/ci.yml
mikebonnet: Another option would be to wait for the Konflux image build to complete and pull that image before starting the e2e tests.

olliewalsh: It's not used in the e2e test. Since the konflux test should catch any regressions with image building it should be ok to drop it

olliewalsh: pushed a commit to drop the build-image job too

PR #2359, opened 2026-01-28
Use default (auto) value for llama.cpp flash-attn 

olliewalsh:
llama.cpp --flash-attn now defaults to auto so no need to set this explicitly.

Also fix inverted logic for uses_nvidia in inference spec context.

## Summary by Sourcery

Rely on llama.cpp's default flash attention behavior and correct host GPU detection logic.

Bug Fixes:
- Fix inverted uses_nvidia flag initialization in host context creation.

Enhancements:
- Stop explicitly passing --flash-attn to llama.cpp and update specs and tests to match the new default behavior.

Tests:
- Update unit, e2e, and system tests to remove explicit expectations around --flash-attn flags.


olliewalsh:
@mikebonnet PTAL

PR #2367, opened 2026-01-29
Adds python sdk for ramalama

ieaves:
## Summary by Sourcery

Introduce a Python SDK package for Ramalama that enables programmatic model management and local chat-based inference, including sync and async helpers around the existing CLI and transport system.

New Features:
- Add a high-level synchronous RamalamaModel interface for starting a model server, managing its lifecycle, and performing chat completions over an OpenAI-compatible API.
- Add an AsyncRamalamaModel variant with asyncio-compatible server lifecycle management, download helpers, and chat calls.
- Expose a ModelStore helper to query locally available models from the global model store, filtered and sorted by recency.
- Provide utility functions and schemas for health checks, chat requests, and model metadata, plus custom error types for missing container runtimes and server timeouts.
- Package the Python SDK as a distributable project with pyproject metadata, documentation stubs, examples, and a publish Makefile, and exclude it from the top-level Makefile script scanning.

Build:
- Add a dedicated pyproject.toml and packaging configuration for the ramalama-sdk Python package, including dev dependency groups and tooling config.

Documentation:
- Add top-level and Python-SDK-specific READMEs documenting installation, capabilities, usage examples, and supported model transports.
- Add MkDocs configuration and API docs scaffolding for the Python SDK.

Tests:
- Add unit and integration tests for server attribute tracking, chat behavior (including failure modes and history), and model listing logic.
- Add container-backed end-to-end tests covering serving, stopping, downloading models, chatting with history, and listing models for both sync and async SDK interfaces.

### DIFF
--- a/ramalama-sdk/sdk/ramalama-python/src/ramalama_sdk/main.py
+++ b/ramalama-sdk/sdk/ramalama-python/src/ramalama_sdk/main.py
@@ -11,16 +11,16 @@ from functools import cached_property
 from http.client import HTTPException
 from types import SimpleNamespace
 
+from ramalama_sdk.errors import RamalamaNoContainerManagerError, RamalamaServerTimeoutError
+from ramalama_sdk.schemas import ChatMessage, ModelRecord
+from ramalama_sdk.utils import is_server_healthy, list_models, make_chat_request
+
 from ramalama.command.factory import assemble_command
 from ramalama.config import CONFIG as ramalama_conf
 from ramalama.engine import stop_container
 from ramalama.model_store.global_store import GlobalModelStore
 from ramalama.transports.base import Transport, compute_serving_port
 from ramalama.transports.transport_factory import New
-
-from ramalama_sdk.errors import RamalamaNoContainerManagerError, RamalamaServerTimeoutError
-from ramalama_sdk.schemas import ChatMessage, ModelRecord
-from ramalama_sdk.utils import is_server_healthy, list_models, make_chat_request
 
 logger = logging.getLogger(__name__)
 
--- a/ramalama-sdk/sdk/ramalama-python/src/ramalama_sdk/utils.py
+++ b/ramalama-sdk/sdk/ramalama-python/src/ramalama_sdk/utils.py
@@ -3,9 +3,9 @@ import urllib.request
 from datetime import datetime
 from typing import TYPE_CHECKING
 
-from ramalama.model_store.global_store import GlobalModelStore
-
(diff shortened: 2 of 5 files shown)


ramalama-sdk/sdk/ramalama-python/src/ramalama_sdk/main.py
```diff
@@ -0,0 +1,370 @@
...
+from ramalama_sdk.errors import RamalamaNoContainerManagerError, RamalamaServerTimeoutError
+from ramalama_sdk.schemas import ChatMessage, ModelRecord
+from ramalama_sdk.utils import is_server_healthy, list_models, make_chat_request
+
+from ramalama.command.factory import assemble_command
+from ramalama.config import CONFIG as ramalama_conf
+from ramalama.engine import stop_container
+from ramalama.model_store.global_store import GlobalModelStore
+from ramalama.transports.base import Transport, compute_serving_port
+from ramalama.transports.transport_factory import New
```
olliewalsh, line 23: @ieaves hey, since this is wrapping unstable/internal apis that would mean when changing the internals we would need to ensure it does not break this interface?
This seems like a significant maintenance overhead, as significant as a new CLI interface perhaps.

olliewalsh: I would approach this from the other direction - refactor the internal APIs into a more well-structured abstraction layer.

This could then be used by the cli handlers, targeted functional tests  etc... perhaps some other use-cases in future (http api maybe).
The key criteria for me is that there would be one and only one interface that everything uses, one and only one interface to maintain - zero/minimal additional maintenance cost.
This could provide a potential integration point for external vertical applications etc... but that would need to be with the understanding that the abstraction layer is unstable, and not expected to be stable for the foreseeable future.

ieaves: > This seems like a significant maintenance overhead, as significant as a new CLI interface perhaps.


Ultimately, yes. That's one reason I checked in with you / the team on matrix before opening this PR because I wasn't sure if you would be open to having this upstreamed. I think an SDK like this makes the project substantially more powerful but it, of course, requires maintenance.


> but that would need to be with the understanding that the abstraction layer is unstable, and not expected to be stable for the foreseeable future.

Hoooo boy do I agree with you here. I think this has to be considered very much unstable. Working on the SDK actually helped expose some of the structural friction in the current design which we can improve.

ieaves:
Insufficient interest from maintainers.

PR #2374, opened 2026-01-30
Adds anthropic provider and updates mcp to use provider implementations

ieaves:
## Summary by Sourcery

Add an Anthropic chat provider and refactor MCP and transport flows to use provider implementations for hosted API models.

I ended up including gemini and bedrock into this as well because I needed the MCP changes to verify tool calling worked correctly with each 

New Features:
- Introduce an Anthropic chat provider implementing the Messages API, including streaming, tools, and model listing.
- Support Anthropic-hosted models via the existing API transport and short-name configuration.
- Allow MCP agents to use pluggable chat providers instead of direct HTTP calls.

Bug Fixes:
- Ensure API transports do not assemble local server commands, preventing errors for hosted providers.
- Propagate default provider URL and API key into CLI args when using API transports.

Enhancements:
- Unify error handling for unsupported message types across chat providers.
- Adjust CLI/MCP integration to pass chat providers rather than raw URLs and args.
- Improve manual tool execution logic to fall back to user input when no question is provided.

Tests:
- Add unit tests for the Anthropic provider covering auth, payload building, streaming parsing, and model listing.
- Extend provider selection tests to cover Anthropic and update unknown-provider handling.
- Add regression tests to ensure CLI with API transports skips server command assembly and runs cleanly with hosted providers.

### DIFF
--- a/docs/ramalama-run.1.md
+++ b/docs/ramalama-run.1.md
@@ -17,13 +17,13 @@ ramalama\-run - run specified AI Model as a chatbot
 | rlcr          | rlcr://   | [`ramalama.com`](https://registry.ramalama.com) |
 | OCI Container Registries | oci:// | [`opencontainers.org`](https://opencontainers.org)|
 |||Examples: [`quay.io`](https://quay.io),  [`Docker Hub`](https://docker.io),[`Artifactory`](https://artifactory.com)|
-| Hosted API Providers | openai:// | [`api.openai.com`](https://api.openai.com)|
+| Hosted API Providers | openai://, gemini://, bedrock:// | [`api.openai.com`](https://api.openai.com), [`generativelanguage.googleapis.com`](https://generativelanguage.googleapis.com), [`bedrock-runtime.<region>.amazonaws.com`](https://docs.aws.amazon.com/bedrock/latest/userguide/bedrock-runtime.html)|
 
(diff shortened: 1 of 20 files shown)


olliewalsh:
Hi @ieaves,
I have concerns about the hosted provider functionality in general as it is significant scope creep from the project goal of simplifying the local serving of AI model using containers.
It may be better suited as a separate project downstream.

rhatdan:
@olliewalsh @mikebonnet @engelmi PTAL

ieaves:
Let's talk about it. Is the concern with the number of supported providers or with the concept of additional providers at all?

Here is my thought process:

I think supporting different API structures beyond the /completions interface is generically useful as we intrinsically need that functionality to support model runtimes or image serving over other APIs. It seems unlikely to me that /completions will be eternal. OpenAI has already moved on in part because there are capabilities you can't easily express using it. Having the ability to interface with multiple API structures is a net benefit to all users. This seems like the sort of functionality we will need at some point and which provides value to all of our potential users.

The fact that we get to support hosted providers for free as a byproduct of supporting multiple API's is a pure benefit. 

I think the question about scope creep is fair though. I'd been thinking about the project as having the goal of moving people _to_ self-hosting and demonstrating how easy that was to do with containers. From that perspective the ability to invoke both hosted and containerized applications through the same command is not just really cool, but also really useful. I think if the project is restricted to being a light wrapper around llama.cpp and either docker or podman it's going to be difficult to achieve the higher goal.

ieaves:
Insufficient interest from maintainers.

PR #2399, opened 2026-02-05
Upgrade to new license specifications.

jwieleRH:
As of setuptools version 77.0.0, project.license as a TOML table is deprecated. The new license representation is project.license, a valid SPDX license expression consisting of one or more license identifiers and project.license-files, a list of license files.  See: https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license.

Remove License classifier: License classifiers have been superseded by license expressions (see https://peps.python.org/pep-0639/). PyPI.org still recognizes license classifiers as search terms, but tools that enforce compliance with PEP-639 balk.

Remove [tool.setuptools] license-files. It is redundant.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools"]
+requires = ["setuptools>=69.0.0"]
 build-backend = "setuptools.build_meta"
 
 [project]
@@ -32,7 +32,6 @@ maintainers = [
    { name = "Mike Bonnet", email = "mikeb@redhat.com" },
 ]
 classifiers = [
-  "License :: OSI Approved :: MIT License",
   "Environment :: Console",
   "Intended Audience :: Developers",
   "Intended Audience :: Science/Research",
@@ -50,9 +49,8 @@ classifiers = [
   "Topic :: Scientific/Engineering :: Artificial Intelligence",
   "Topic :: Utilities",
 ]
-
-[project.license]
-text = "MIT"
+license = "MIT"
+license-files = ["LICENSE"]
 
 [project.optional-dependencies]
 dev = [
@@ -102,7 +100,6 @@ ramalama = "ramalama.cli:main"
 
 [tool.setuptools]
 include-package-data = true
-license-files = ["LICENSE"]
 
 [tool.setuptools.package-data]
 "ramalama" = ["py.typed"]


pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools"]
+requires = ["setuptools>=77.0.0"]
```
rhatdan: Will this cause issues for installing on RHEL9?

olliewalsh: Was just about to ask that...

There are other issues to address for el9 though. If this works on el10 I think we could patch it for rpm builds on el9.

Is this deprecated and scheduled for removal, or just a deprecated warning?

olliewalsh: Does not work on el10:
```
No matching package to install: 'python3dist(setuptools) >= 77'
Not all dependencies satisfied
Error: Some packages could not be found.
```

jwieleRH: PRETTY_NAME="Red Hat Enterprise Linux 9.7 (Plow)"
ERROR: Package 'ramalama' requires a different Python: 3.9.25 not in '>=3.10'
RHEL9.7 appears to have Python 3.9.25 while RamaLama requires Python 3.10 so that ship may have sailed.

rhatdan: You can install Python3.11 I believe on RHEL.

olliewalsh: > You can install Python3.11 I believe on RHEL.

on el9? yes, 3.12 rpm is available too. However in both cases the dependencies we require are not packaged (nor are they in EPEL)

rhatdan: Well Red Hat really needs RamaLama to work on RHEL10, and would be nice to work on RHEL9, if there is a way to hack around this in the Spec file, then I am fine with the change.

jwieleRH:
From the log of "make pypi-build":
        Please use a simple string containing a SPDX expression for `project.license`. You can also use `project.license-files`. (Both options available on setuptools>=77.0.0).

        By 2027-Feb-18, you need to update your project and remove deprecated calls
        or your builds will no longer be supported.

        See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.

olliewalsh:
That's next year

olliewalsh:
In the meantime let's just watch https://github.com/pypa/setuptools/issues/4903

Issue #2387, opened 2026-01-31
Install problems on LMDE 7 / Debian 13


netzwanze:
### Issue Description

Hello,

I tried to install ramalama using the installation script and got this message :

```
~/ramalama$ curl -fsSL https://ramalama.ai/install.sh | bash
  _____                       _
 |  __ \                     | |
 | |__) |__ _ _ __ ___   __ _| |     __ _ _ __ ___   __ _
 |  _  // _` | '_ ` _ \ / _` | |    / _` | '_ ` _ \ / _` |
 | | \ \ (_| | | | | | | (_| | |___| (_| | | | | | | (_| |
 |_|  \_\__,_|_| |_| |_|\__,_|______\__,_|_| |_| |_|\__,_|

Hit:1 http://de.mirrors.clouvider.net/debian trixie InRelease
Hit:2 http://deb.debian.org/debian trixie-backports InRelease                                                                                                                                
Get:3 http://security.debian.org trixie-security InRelease [43.4 kB]                                                                                                                         
Hit:4 http://de.mirrors.clouvider.net/debian trixie-updates InRelease                                                                                                                        
Hit:5 http://de.mirrors.clouvider.net/debian trixie-backports InRelease                                                                                                                      
Ign:6 https://ftp-stud.hs-esslingen.de/pub/Mirrors/packages.linuxmint.com gigi InRelease                                                                                  
Get:7 http://security.debian.org trixie-security/main amd64 Packages [98.7 kB]                                                         
Get:8 https://repo.steampowered.com/steam stable InRelease [3622 B]                                                                               
Hit:9 https://ftp-stud.hs-esslingen.de/pub/Mirrors/packages.linuxmint.com gigi Release                                                                       
Hit:10 https://updates.signal.org/desktop/apt xenial InRelease                                                    
Get:12 http://security.debian.org trixie-security/main i386 Packages [90.5 kB]                                    
Get:13 http://security.debian.org trixie-security/main Translation-en [63.2 kB] 
Hit:14 https://liquorix.net/debian trixie InRelease                                                          
Fetched 299 kB in 1s (351 kB/s)
All packages are up to date.    
sudo: apt_install: command not found
downloading uv 0.7.2 x86_64-unknown-linux-gnu
no checksums to verify
installing to /home/jowenzel/.local/bin
  uv
  uvx
everything's installed!

main: Zeile 169: uv: command not found
```

### Steps to reproduce the issue

I tried again with the same result.

### Describe the results you received

[...]

PR #2404, opened 2026-02-05
Fix install.sh on Debian and update PATH for uv

olliewalsh:
Cannot call the apt_install function thru sudo, as it was trivial just remove.

Update path after installing uv.

Fixes: #2387

## Summary by Sourcery

Update Debian-based installation flow to call apt directly and ensure uv is available on the PATH after installation.

Bug Fixes:
- Fix Debian install script by avoiding calling the apt_install helper via sudo and instead invoking apt directly to install podman and docker.

Enhancements:
- Update the uv installation step to prepend the user's local bin directory to PATH when uv is installed there, avoiding missing-command issues.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -132,12 +132,21 @@ install_uv() {
   echo
 
   # Add uv bin dir to PATH if not present
-  if [ -x "$HOME"/.local/bin/uv ]; then
+  local uv_bin_dir=""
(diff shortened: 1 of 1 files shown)


olliewalsh:
Unfortunately the ubuntu github runner environment does not reproduce the issue but I have manually verified it on a LMDE7 live iso VM.

PR #2418, opened 2026-02-10
Add --interactive flag to continue chat after stdin/args

### LINKED ISSUE #2400: Ability to keep a stdin chat going

When using run without a prompt, it is interactive.
But if text is piped or a prompt is passed as an arg, it exits after completion.

The ability to keep prompting on the passed input would be useful.

_No response_

_No response_

### Additional context

_No response_

rhatdan:
Implement the ability to keep a stdin chat going after processing the initial prompt, addressing user feedback that the chat exits immediately after processing arguments or piped input.

Changes:
- Add --interactive/-i flag to chat and run commands
- When flag is set, continue to interactive mode after processing
- Default behavior unchanged (exit after response for scripts)
- Update both ramalama-chat and ramalama-run man pages
- Add comprehensive examples demonstrating the new functionality

The --interactive flag enables use cases like:
- Piping an initial question and continuing the conversation
- Starting with a prompt argument and exploring further
- Building on context from stdin in an interactive session

Fixes #2400

*This PR was created with the assistance of Cursor AI.*

## Summary by Sourcery

Add an interactive mode flag to chat and run commands so sessions can continue after processing initial stdin or prompt arguments.

New Features:
- Introduce an --interactive (-i) flag for ramalama-chat and ramalama-run to continue into interactive chat after handling stdin or prompt arguments.

Documentation:
- Document the --interactive flag semantics and default non-interactive behavior in ramalama-chat and ramalama-run man pages, including new usage examples for stdin and argument-based workflows.

### DIFF
--- a/docs/ramalama-chat.1.md
+++ b/docs/ramalama-chat.1.md
@@ -8,10 +8,13 @@ ramalama\-chat - OpenAI chat with the specified REST API URL
 
 positional arguments:
   ARGS                  overrides the default prompt, and the output is
-                        returned without entering the chatbot
+                        returned without entering the chatbot (unless
+                        --interactive is specified)
 
 ## DESCRIPTION
-Chat with an OpenAI Rest API
+Chat with an OpenAI REST API. By default, if arguments or stdin are provided,
+the response is displayed and the command exits. Use **--interactive** to
+continue to an interactive chat session after processing the initial prompt.
 
 ## OPTIONS
 
@@ -25,6 +28,11 @@ Possible values are "never", "always" and "auto". (default: auto)
 
 #### **--help**, **-h**
 Show this help message and exit
+
+#### **--interactive**, **-i**
+Continue to interactive chat mode after processing stdin or prompt arguments.
+By default, when arguments or piped input are provided, the command exits after
+displaying the response. This flag allows you to continue chatting interactively.
 
 #### **--list**
 List the available models at an endpoint
@@ -77,6 +85,13 @@ $ ramalama chat
 🦭 > Hi \
 🦭 > tell me a funny story \
 🦭 > please
+
+Send an initial prompt via stdin and continue chatting
+$ echo "What is 2+2?" | ramalama chat --interactive
+4
+🦭 > Now multiply that by 3
+12
+🦭 > /bye
 ```
 
 ## SEE ALSO
--- a/ramalama/chat.py
+++ b/ramalama/chat.py
@@ -5,6 +5,7 @@ import cmd
 import copy
 import itertools
 import json
+import os
 import signal
 import subprocess
 import sys
@@ -400,9 +401,16 @@ class RamaLamaShell(cmd.Cmd):
             return None
 
     def handle_args(self, monitor):
+        """Process command-line arguments and/or stdin input.
+
+        Returns True if the program should exit, False if it should continue to interactive mode.
+        """
         prompt = " ".join(self.args.ARGS) if self.args.ARGS else None
+        stdin_was_pipe = False
+
         if not sys.stdin.isatty():
             stdin = sys.stdin.read()
+            stdin_was_pipe = True
             if prompt:
                 prompt += f"\n\n{stdin}"
             else:
@@ -410,6 +418,23 @@ class RamaLamaShell(cmd.Cmd):
 
         if prompt:
             self.default(prompt)
+            # Continue to interactive mode if --interactive flag is set
+            if getattr(self.args, 'interactive', False):
+                # If stdin was a pipe, we need to reopen the terminal for interactive input
+                if stdin_was_pipe:
+                    try:
+                        # Use platform-specific terminal device
+                        # Windows: 'CON', Unix/Linux: '/dev/tty'
+                        terminal_device = 'CON' if os.name == 'nt' else '/dev/tty'
+                        sys.stdin = open(terminal_device, 'r')
+                    except (OSError, FileNotFoundError) as e:
+                        perror(f"Cannot open terminal for interactive mode: {e}")
+                        monitor.stop()
+                        self.kills()
+                        return True
+                print("")
+                return False
+            # Default behavior: exit after displaying response
             monitor.stop()
             self.kills()
             return True
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -1201,6 +1201,12 @@ def chat_run_options(parser):
         metavar="N",
         help="automatically summarize conversation history after N messages to prevent context growth (0=disabled)",
     )
+    parser.add_argument(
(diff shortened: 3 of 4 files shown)


rhatdan:
@engelmi @olliewalsh @bmahabirbu @ieaves @mikebonnet PTAL

docs/ramalama-chat.1.md
```diff
@@ -77,6 +85,18 @@ $ ramalama chat
 🦭 > Hi \
 🦭 > tell me a funny story \
 🦭 > please
+
+Send an initial prompt via stdin and continue chatting
+$ echo "What is 2+2?" | ramalama chat --interactive
```
olliewalsh, line 90: This will not work, stdin is a pipe and will be closed when echo completes

docs/ramalama-chat.1.md
```diff
@@ -77,6 +85,18 @@ $ ramalama chat
...
+4
+🦭 > Now multiply that by 3
+12
+🦭 > /bye
+
+Pass arguments and continue to interactive mode
+$ ramalama chat --interactive "Tell me about AI"
```
olliewalsh: this will work, but doesn't seem necessary - just don't pass the prompt

docs/ramalama-chat.1.md
ieaves: It definitely doesn't work currently, but we could make it work in our chat implementation if we consume the piped stdin for the initial prompt and in interactive mode switch the input from the pipe to the users terminal.

docs/ramalama-chat.1.md
ieaves: I'd second Ollie here. This seems like the cleanest approach. What sort of user feedback did we get though @rhatdan ? Was there something special about the usecase that made them want to seed the initial prompt?

docs/ramalama-chat.1.md
rhatdan: Excellent point.

rhatdan: Should be fixed now.

ieaves (approved):
LGTM

ramalama/chat.py
```diff
@@ -400,16 +400,37 @@ def _select_tools(self):
...
         if prompt:
             self.default(prompt)
+            # Continue to interactive mode if --interactive flag is set
+            if getattr(self.args, 'interactive', False):
+                # If stdin was a pipe, we need to reopen the terminal for interactive input
+                if stdin_was_pipe:
+                    try:
+                        sys.stdin = open('/dev/tty', 'r')
+                    except (OSError, FileNotFoundError) as e:
+                        perror(f"Cannot open terminal for interactive mode: {e}")
```
olliewalsh, line 431: should we handle this better on Windows?

ramalama/chat.py
```diff
@@ -400,16 +401,40 @@ def _select_tools(self):
...
         if prompt:
             self.default(prompt)
+            # Continue to interactive mode if --interactive flag is set
+            if getattr(self.args, 'interactive', False):
+                # If stdin was a pipe, we need to reopen the terminal for interactive input
+                if stdin_was_pipe:
+                    try:
+                        # Use platform-specific terminal device
+                        # Windows: 'CON', Unix/Linux: '/dev/tty'
+                        terminal_device = 'CON' if os.name == 'nt' else '/dev/tty'
```
olliewalsh, line 428: can we have e2e test coverage for this?

ramalama/cli.py
```diff
@@ -1201,6 +1201,12 @@ def chat_run_options(parser):
         metavar="N",
         help="automatically summarize conversation history after N messages to prevent context growth (0=disabled)",
     )
+    parser.add_argument(
```
olliewalsh, line 1204: need this in the ramalama-run docs too

olliewalsh (changes requested):

ramalama/chat.py
rhatdan: You want to test AI within the e2e?  This would be racy and prone to failure, or could you be more specific of what kind of test you want?

ramalama/cli.py
rhatdan: The man pages have been updated for ramalama-run.1?

ramalama/chat.py
olliewalsh: Has this been validated on windows?

rhatdan: Nope, I have no access, and am trusting the AI, any chance  you can check?

olliewalsh: Sure

docs/ramalama-chat.1.md
olliewalsh: @rhatdan WDYT? It's not clear what problem this is solving. Perhaps the user is making an assumption that the prompt cli args is special in some way (e.g system message vs user message, but that is not the case).

olliewalsh: ah, issue comments contains an example. Ok, makes sense I suppose

ramalama/chat.py
olliewalsh: It doesn't work on windows, but just piping a prompt using current main doesn't work either - stdin on line 412 is an empty string. Let me push a fix for that first...

olliewalsh: I'm surprised, since consoles and pipes are a horror show on windows, but it works with https://github.com/containers/ramalama/pull/2458:

```
echo "what is the capital of france" | uv run ramalama run -i ollama://library/deepseek-r1:8b
cf26654527d7e09b7cafaad1f511d20d996849f703ad25608164ee1eb90b331a
The capital of France is **Paris**.

It is also the country's largest city and a major global center for arts, fashion, gastronomy, and culture. Notable landmarks include the **Eiffel Tower** and the **Louvre Museum**.

> what is the timezone?
The time zone for France depends on the part of the country:

1.  **Most of France:** Uses **Central European Time (CET)** and **Central European Summer Time (CEST)**.
    *   Standard Time (Winter): **UTC+1**
    *   Daylight Saving Time (Summer): **UTC+2**

2.  **Overseas Territories:**
    *   **Réunion (offshore):** UTC+1 (same as mainland France during standard time) and UTC+2 (during summer time). However, it does not change with DST.
    *   **French Guiana (South America):** **UTC-3**
    *   **Scattered Islands (e.g., Djibouti, Comorras):** **UTC+1** (same as mainland France during standard time) and **UTC+2** (during summer time). These do change with DST.

**Right now (late June 2024), France is observing Daylight Saving Time, so most parts are using CEST (UTC+2).** This will change again in **August 2024**, when France (following European Union rules) will revert to standard time (CET, UTC+1), ending Daylight Saving Time.
```

PR #2422, opened 2026-02-12
Restores previous llama.cpp jinja behavior

ieaves:
llama.cpp changed the default llama-serve --jinja settings in December 2025: https://github.com/ggml-org/llama.cpp/pull/17911

This causes a regression when using the 0.17.0 base images for some models like`hf://ggml-org/SmolVLM-500M-Instruct-GGUF`

> While executing For at line 1, column 162 in source:
...}{{':'}}{% else %}{{': '}}{% endif %}{% for line in message['content'] %}{% if l...
                                           ^
> Error: Expected iterable or object type in for loop: got String
srv          init: init: please consider disabling jinja via --no-jinja, or use a custom chat template via --chat-template
srv          init: init: for example: --no-jinja --chat-template chatml
srv    operator(): operator(): cleaning up before exit...
main: exiting due to model loading error

This PR explicitly sets our default jinja settings.

## Summary by Sourcery

Set explicit Jinja defaults for llama.cpp, disabling Jinja for multimodal models to restore compatibility with impacted base images.

Enhancements:
- Add a --no-jinja flag for models with mmproj_path to align multimodal behavior with previous llama.cpp defaults.

Tests:
- Update factory command expectation to include --no-jinja for multimodal models, ensuring the new default is covered by tests.


rhatdan:
@engelmi @olliewalsh PTAL

olliewalsh:
MacOS e2e failure I expect is caused by the cli output changing in python 3.14.3 (https://github.com/python/cpython/issues/75949)

PR #2427, opened 2026-02-16
Feat/arg typing

ieaves:
## Summary by Sourcery

Strengthen CLI, transport, and chat handling through stricter argument typing, safer control flow, and more robust edge‑case handling across model management and RAG features.

New Features:
- Introduce typed protocol-based argument models and a narrowing helper for CLI commands to enable safer runtime argument handling.
- Add type-guard utilities for transports (e.g., Ollama vs OCI) to simplify and clarify transport-specific logic.
- Extend RAG support wiring so run/serve and chat flows have explicit runtime-only fields and validation for required options.

Bug Fixes:
- Prevent metadata and inspect operations from misbehaving on hosted API transports by short-circuiting or delegating appropriately.
- Fix chat streaming and summarization edge cases by enforcing required chat args, handling missing delta/content safely, and tightening keepalive/timeout behavior.
- Ensure MLX runtime implicitly disables containers, normalize Docker pull policy, and require required flags such as TARGET for convert and NAME for stop when needed.
- Correct port and engine handling in RAG and daemon flows, including proper exclusion lists for port selection and defaulting engines where missing.
- Harden Hugging Face and HTTP download paths to respect CLI availability, cache and file handling, and to raise clearer errors when downloads or pushes are not possible.
- Improve TOML and config parsing to handle nested sections and non-dict values safely without crashing.

Enhancements:
- Refine CLI post-parse setup to work over typed argument protocols, including normalized model resolution, runtime arg splitting, and logging level selection.
- Generalize container engine helpers with generics and stronger types, simplifying label, network, and device configuration while keeping dry-run semantics intact.
- Tighten typing and initialization across transports, RAG, daemon, file, and HTTP utilities to make assumptions explicit and improve maintainability.

Tests:
- Add unit tests covering chat runtime default fields, post-parse CLI behaviors (runtime args splitting, MLX nocontainer, Docker pull normalization), and RAG argument transformation.
- Introduce tests for Hugging Face push behavior when the HF CLI is unavailable, ensuring a clear NotImplemented error is raised.


ramalama/compat.py
```diff
@@ -22,16 +22,30 @@ def __str__(self) -> str:
 else:
     # Python 3.10 doesn't have NamedTemporaryFile delete_on_close
     @contextmanager
-    def NamedTemporaryFile(*args, **kwargs):  # type: ignore[no-redef]
-        if 'delete_on_close' in kwargs:
```
olliewalsh, line 26: why was this logic removed?

olliewalsh (commented):
Might be worth looking at ty for type checking, if we want to go all-in on the astral tooling

ramalama/compat.py
ieaves: I don't think any logic was removed. The change should just move named arguments out of kwargs and into named variables to make the type checker happy. The underlying logic is still the same though. Specifically,

```python
f = _NamedTemporaryFile(*args, **kwargs, delete=delete and delete_on_close)
```

Became

```python
        f = _NamedTemporaryFile(
            mode=mode,
            buffering=buffering,
            encoding=encoding,
            newline=newline,
            suffix=suffix,
            prefix=prefix,
            dir=dir,
            delete=delete and delete_on_close,
            errors=errors,
        )
```

In both cases, `delete=delete and delete_on_close`. Otherwise the same named arguments are carried forward with the exception of `delete_on_close`.

ieaves:
> Might be worth looking at ty for type checking, if we want to go all-in on the astral tooling

I'm not opposed to this at all. The last time I was playing with it the differences between ty and mypy were a little more apparent than ruff and black though. IIRC ty is a lot more permissive by default, for example, I think in this case


```python
x = ['hello', 'ollie']
```

ty infers x as list[str | Unknown] where mypy will narrow to list[str] where `Unknown` acts like an Any. It's a completely sane default but still a little different.

ramalama/compat.py
olliewalsh: ah, the if block is basically just setting the values to True if not set, so providing arg default achieves the same thing i.e:
```
def NamedTemporaryFile(
    ...
    delete = True,
    delete_on_close = True,
):
```

ramalama/compat.py
```diff
@@ -22,16 +22,30 @@ def __str__(self) -> str:
 else:
     # Python 3.10 doesn't have NamedTemporaryFile delete_on_close
     @contextmanager
-    def NamedTemporaryFile(*args, **kwargs):  # type: ignore[no-redef]
```
olliewalsh, line 25: the `#type:` tag is still necessary I expect

ramalama/compat.py
ieaves: Yeah we might not need anything more than the default values but the current compatibility implementation wants to override delete if either delete or delete_on_class were set to false. I'm not sure if we depend on that behavior anywhere else and I was worried about not carrying it forwards here. Just looking at the [docs](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) there can be strange interactions with mismatched delete settings on windows as well.

ramalama/compat.py
ieaves: We should be good without it. We didn't need it previously either because of the way we handled the import:

```python
from tempfile import NamedTemporaryFile as _NamedTemporaryFile
```

The type checker seems content with the if/else conditional being mutually exclusive paths and isn't complaining at the moment. This might have been a holdover from older versions of mypy or something?

ieaves:
Hey @olliewalsh I'm not sure how close of a look you had at this already but if your concerns were addressed could you signoff on this one? While the actual logic changes are small this PR touches basically every part of the codebase and is going to generate constant merge conflicts if left to sit.

olliewalsh:
> Hey @olliewalsh I'm not sure how close of a look you had at this already but if your concerns were addressed could you signoff on this one? While the actual logic changes are small this PR touches basically every part of the codebase and is going to generate constant merge conflicts if left to sit.

I've not really looked that closely yet - it's vast! I think ci has been consistently failing on this too so was waiting for it to be ready for review.
How expensive is this to rebase when there are conflicts? i.e are you manually picking through the code line-by-line or is AI/tooling doing the heavy-lifting here?

ieaves:
I don't think tests have been failing (other than flaking). They were failing yesterday because I added to the PR reducing reliance on casts in the cli and our tests haven't historically reflected the actual parsed objects produced by the parser. Either way it looks larger than it is since most of this is adding typing to code branches that haven't been typed to date without changing structural behavior. To be honest, I'd rather close this PR than spend the next month and a half constantly manually merging, and squashing. Also happy to close though.

rhatdan:
@ieaves are you still interested in this PR?
@olliewalsh what should we do with it?

ieaves:
I'd be happy to rebase if it was going to be reviewed / merged but if you're trying to clean up outstanding PRs and no one has the bandwidth to put eyes on I can also close. Really up to y'all.

PR #2428, opened 2026-02-16
e2e: install git-core

mikebonnet:
`git` is now required by the `make clean` command.

## Summary by Sourcery

Build:
- Add git-core to the e2e container image dependencies to support commands that now require git.


mikebonnet:
This fixes the permission problems with `py.typed` that are causing the e2e tests to fail.

PR #2429, opened 2026-02-16
Fix to ruff preventing auto import sorting

ieaves:
## Summary by Sourcery

Update Ruff formatting workflow and simplify its configuration.

Enhancements:
- Adjust the format Makefile target to run Ruff import sorting explicitly after general checks.
- Simplify Ruff configuration by removing the file pattern include in favor of existing extend-include and exclude settings.

### DIFF
--- a/test/e2e/conftest.py
+++ b/test/e2e/conftest.py
@@ -70,7 +70,7 @@ def container_registry():
         # Start the registry
         # fmt: off
         subprocess.run(
-            [   
+            [
                 ramalama_container_engine, "run", "-d", "--rm",
                 "--name", registry_name,
                 "-p", f"{registry_port}:5000",


ieaves:
@olliewalsh there was an errant include filter in the ruff toml settings from last year which was masking differences  between ruff and black. In particular it looks like black would tolerate newlines between class definitions and the __init__ which ruff collapses. 


I've fixed the errant pyproject.toml settings here and added isort auto fixes back to `make format` but given the rest of the diff is so large we can revert to the black / isort structure from before if there's a preference for it.

Makefile
olliewalsh, line 124: should run isort here too

pyproject.toml
```diff
@@ -128,7 +128,6 @@ check-hidden = true
 [tool.ruff]
 line-length = 120
 target-version = "py310"
-include = ["\\.pyi?$"]
```
olliewalsh, line 131: issue was that it's a glob, not a regex?

ieaves: Basically yeah - Ruff only supports glob patterns in `include` and this was a regex.

ieaves:
> > I've fixed the errant pyproject.toml settings here and added isort auto fixes back to `make format` but given the rest of the diff is so large
> 
> If it's just trivial whitespace changes then don't think the diff size matters.
> 
> > we can revert to the black / isort structure from before if there's a preference for it.
> 
> IIRC the original motivation was that ruff was much faster than black/isort. Is that still the case?


It's all trivial whitespace changes and it's still just as fast after this fix but I wanted to be cognizant that part of the conversation on the other PR was that black / ruff were generating identical formatting. Given hat background, suffice to say I was very annoyed when I saw the size of this diff yesterday.

pyproject.toml
olliewalsh: ack, I think it would have been good to explain this in the PR description.

@rhatdan WDYT about disabling the sourceryai PR summary? IMO a description of the changes from the author is much better - they can provide the reason for the change, context etc... 
This tends to gets omitted or just lost in the noise when sourcery is adding a summary to the description.
sourcery will still include a summary of the changes in the reviewers guide.

test/e2e/conftest.py
```diff
@@ -70,15 +71,26 @@ def container_registry():
         subprocess.run(
             [
                 # fmt: off
-                ramalama_container_engine, "run", "-d", "--rm",
```
olliewalsh, line 73: previous version was much more readable IMO...  does ruff not respect the `#fmt: off` tag?

olliewalsh: Looks like it should https://docs.astral.sh/ruff/formatter/#format-suppression

olliewalsh:
> It's all trivial whitespace changes and it's still just as fast after this fix but I wanted to be cognizant that part of the conversation on the other PR was that black / ruff were generating identical formatting. Given hat background, suffice to say I was very annoyed when I saw the size of this diff yesterday.

I _think_ the most recent version of black enabled a lot of whitespace formatting rules too, so they are possibly consistent but we were pinning to an older version

test/e2e/conftest.py
ieaves: It respects them generally but only at the statement level. In this case, the comments are nested inside of a `subprocess.run` invocation which ruff doesn't recognize. I can reorganize the format suppression comments so that they are compliant if we prefer the suppressed structure.

ieaves:
That would explain it if so.  I coose to believe if only because it makes me feel better :).

pyproject.toml
rhatdan: I am fine with disabling.

Issue #2343, opened 2026-01-24
ramalama chat should wait for llama.cpp to initiialize


olliewalsh:
### Issue Description

llama-server /models api returns a 503 error while the model is loading. ramalama run handles this but ramalama chat currently does not. For interactive use this isn't a major issue but it would be good to fix this.

Worked around this for CI in https://github.com/containers/ramalama/pull/2342

### Steps to reproduce the issue

Run ramalama chat soon after ramalama serve --detach

### Describe the results you received

ramalama chat fails

### Describe the results you expected

ramalama chat succeeds

### ramalama info output

```yaml
N/A
```

### Upstream Latest Release

Yes

olliewalsh:
Resolve for ramalama run in https://github.com/containers/ramalama/commit/3ee6b796574ff56e0c89de4b781c9139e9523f3e, just need to do something similar for ramalama chat

Kulkarnivaishnavi473:
Hi! I’d like to work on this issue if still available.
Please assign it to me. Thanks!

scoonce:
I have implemented a fix for this issue and submitted a PR (#2433). I was unable to assign it to myself due to permissions.

rhatdan:
Thanks @scoonce

PR #2433, opened 2026-02-17
Add wait_for_server in chat.py and test_chat.py corresponding unit tests

scoonce:
Chat needs to wait for llama-server/model to finish loading before starting, but currently returns a 503 error while model is loading.

Solution: adds wait_for_server in chat.py to poll /health until the server is ready and log errors or timeouts. Unit tests in test/unit/test_chat.py for success, timeout, and retry cases.

Fixes: #2343

Signed-off-by: Shelby Coonce <s.coonce@ufl.edu>

## Summary by Sourcery

Ensure the chat client waits for the llama.cpp HTTP server to become healthy before starting an interactive session.

Bug Fixes:
- Prevent 503 errors at chat startup by polling the server /health endpoint until it reports ready or times out.

Tests:
- Add unit tests covering successful server readiness detection, timeout behavior, and retry logic when the server initially returns errors.

### DIFF
--- a/ramalama/chat.py
+++ b/ramalama/chat.py
@@ -15,6 +15,8 @@ import urllib.request
 from collections.abc import Sequence
 from dataclasses import dataclass
 from datetime import timedelta
+from http.client import HTTPConnection
+from urllib.parse import urlparse
 
 from ramalama.arg_types import ChatArgsType
 from ramalama.chat_providers import ChatProvider, ChatRequestOptions
@@ -772,6 +774,38 @@ def _report_server_exit(monitor):
         perror("Check server logs for more details about why the service exited.")
 
 
+def wait_for_server(args, timeout: int = 30):
+    """Wait for the llama.cpp server to be ready before starting chat."""
+    parsed = urlparse(args.url)
+    host = parsed.hostname or "127.0.0.1"
+    port = parsed.port or 8080
+    start_time = time.time()
+    spinner = Spinner().start()
+    last_error: Exception | None = None
+    while time.time() - start_time < timeout:
+        conn = None
+        try:
+            conn = HTTPConnection(host, port, timeout=3)
+            conn.request("GET", "/health")
+            resp = conn.getresponse()
+            resp.read()
+            if resp.status == 200:
+                spinner.stop()
+                return
+        except Exception as exc:
+            last_error = exc
+        finally:
+            if conn:
+                conn.close()
+        time.sleep(1)
+    spinner.stop()
+
+    if last_error:
+        logger.debug(f"Error waiting for {args.url} /health: {last_error}")
+    else:
+        logger.debug(f"Time out waiting for {args.url} /health")
+
+
 def chat(
     args: ChatArgsType,
     operational_args: ChatOperationalArgs | None = None,
@@ -824,6 +858,9 @@ def chat(
 
     # Assign monitor to operational_args
     operational_args.monitor = monitor
+
+    # Wait until server reports healthy
+    wait_for_server(args)
 
     successful_exit = True
     try:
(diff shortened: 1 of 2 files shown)


olliewalsh:
Thanks for the PR. 

The DCO check is failing:
```
Committer: scoonce; Expected "scoonce s.coonce@ufl.edu", but got "Shelby Coonce s.coonce@ufl.edu".
```

Please re-sign the commit and update the PR. E.g I usually run something along the lines of this when I forget to sign the commit: `git commit -s --amend; git push -f`

rhatdan:
Also please respon to the AI Reviewers.

ramalama/chat.py
```diff
@@ -772,6 +774,38 @@ def _report_server_exit(monitor):
...
+    host = parsed.hostname or "127.0.0.1"
+    port = parsed.port or 8080
+    start_time = time.time()
+    spinner = Spinner().start()
+    last_error: Exception | None = None
+    while time.time() - start_time < timeout:
+        conn = None
+        try:
+            conn = HTTPConnection(host, port, timeout=3)
+            conn.request("GET", "/health")
```
olliewalsh, line 789: This endpoint is specific to llama.cpp afaik. I think we need to hit an endpoint that exists for all/any openai compatible inference server. Maybe a http HEAD request to '/chat/completions' would work

ramalama/chat.py
olliewalsh, line 843: only need to perform the check in this case. In the other two cases the logic to start the server will wait until it is healthy

ramalama/chat.py
```diff
@@ -772,6 +774,38 @@ def _report_server_exit(monitor):
...
+        finally:
+            if conn:
+                conn.close()
+        time.sleep(1)
+    spinner.stop()
+
+    if last_error:
+        logger.debug(f"Error waiting for {args.url} /health: {last_error}")
+    else:
+        logger.debug(f"Time out waiting for {args.url} /health")
```
olliewalsh, line 806: should we fail hard if the check fails?

ramalama/chat.py
olliewalsh: I don't see an openai endpoint that would work, so it may need to be a bit more complex.

This is the models response for llama.cpp:
```
curl http://127.0.0.1:8080/v1/models
{"models":[{"name":"bartowski/Llama-3.2-3B-Instruct-GGUF","model":"bartowski/Llama-3.2-3B-Instruct-GGUF","modified_at":"","size":"","digest":"","type":"model","description":"","tags":[""],"capabilities":["completion"],"parameters":"","details":{"parent_model":"","format":"gguf","family":"","families":[""],"parameter_size":"","quantization_level":""}}],"object":"list","data":[{"id":"bartowski/Llama-3.2-3B-Instruct-GGUF","object":"model","created":1771437861,"owned_by":"llamacpp","meta":{"vocab_type":2,"n_vocab":128256,"n_ctx_train":131072,"n_embd":3072,"n_params":3212749888,"size":2011539712}}]}
```

Could look for the "owned_by" field == "llamacpp" to determine that it is llama.cpp and if so check the /health endpoint (models endpoint will succeed immediately, before the model is ready)..

For comparison this is the models API output for vllm:
```
curl http://127.0.0.1:8080/v1/models
{"object":"list","data":[{"id":"TheBloke/TinyLlama-1.1B-Chat-v1.0-GPTQ","object":"model","created":1771437788,"owned_by":"vllm","root":"/mnt/models","parent":null,"max_model_len":2048,"permission":[{"id":"modelperm-80f69cce9f607ca0","object":"model_permission","created":1771437788,"allow_create_engine":false,"allow_sampling":true,"allow_logprobs":true,"allow_search_indices":false,"allow_view":true,"allow_fine_tuning":false,"organization":"*","group":null,"is_blocking":false}]}]}
```

PR #2436, opened 2026-02-18
Migrates type checker from mypy to ty

ieaves:
## Summary by Sourcery

Migrate the project’s static type checking setup from mypy to ty across configuration, tooling, and documentation references.

Enhancements:
- Switch the Makefile type-check target to use ty check instead of mypy.

Build:
- Replace mypy with ty in development dependencies and configure ty settings in pyproject.toml.

CI:
- Update CI workflow step description to reflect ty-based type checking.

Documentation:
- Update contributor documentation to reference ty instead of mypy for type checking commands and expectations.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
         run: |
           uv run -- make man-check
 
-      - name: Run type checking
+      - name: Run ty type checking
         run: |
           uv run -- make type-check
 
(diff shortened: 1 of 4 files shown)


ieaves:
@olliewalsh thought it would be interesting to see how the type checking results would differ between mypy and ty and it actually looks like ty flags a number of issues mypy passes. The diagnostic information is also much richer. Maybe, like you suggested, it is worth the switch?


There are a few rule changes we could modify as well, so I've added an example section to the pyproject config settings as context.

olliewalsh:
> @olliewalsh thought it would be interesting to see how the type checking results would differ between mypy and ty and it actually looks like ty flags a number of issues mypy passes. The diagnostic information is also much richer. Maybe, like you suggested, it is worth the switch?
> 
> There are a few rule changes we could modify as well, so I've added an example section to the pyproject config settings as context.

yeah, that's interesting. ty has just entered beta though so maybe a bit too early to make the switch

ieaves:
Fair enough! I'll close this one out then.

Issue #2376, opened 2026-01-30
Speach to text support


ieaves:

Crystora:
Hi @ieaves I’m interested in the project, I’d like to contribute. Can you share a bit more background on this issue?

ieaves:
Hey @bittoby. There's some freedom for how best to implement this but there has been some extant work / thought which has gone into using whisper already:

- https://github.com/containers/ramalama/issues/51
- https://github.com/containers/ramalama/pull/2038

The previous efforts implemented a new runtime for whisper.cpp which modified the container entrypoint and might be a good place to start.

rhatdan:
@bittoby I have assigned you to the project.

I think making this simple where you could take in a audio file from stdin and have it processed to an output file would be cool.

mecattaf:
whisper cpp (STT) in ramalama would be amazing!

PR #2438, opened 2026-02-18
feat: add stt support via whisper.cpp with new transcribe subcommand

bittoby:
## Add Speech-to-Text Support via whisper.cpp

Adds a new `transcribe` subcommand and `whisper.cpp` runtime for speech-to-text transcription.

### Usage
```bash
ramalama transcribe huggingface://ggerganov/whisper.cpp/ggml-tiny.bin audio.wav
```

### Changes
- New `transcribe` subcommand using `whisper-cli`
- New `whisper.cpp` runtime with `serve` (HTTP API) and `transcribe` (batch)
- New inference spec `inference-spec/engines/whisper.cpp.yaml`
- Container image: `ghcr.io/ggml-org/whisper.cpp:main`

### Notes
- Audio files must be world-readable in container mode: `chmod o+r audio.wav`
- For serve: `ramalama --runtime whisper.cpp serve MODEL`

## Summary by Sourcery

Add speech-to-text support using a new whisper.cpp runtime and transcribe subcommand integrated into the CLI and inference engine.

New Features:
- Introduce a transcribe CLI subcommand to convert audio files to text using speech-to-text models.
- Add whisper.cpp as a supported runtime with serve (HTTP API) and batch transcription commands.
- Provide a new whisper.cpp inference engine spec defining serve and transcribe commands and options.

Enhancements:
- Extend runtime configuration and image resolution to support a dedicated whisper.cpp container image with sensible defaults.
- Update the shared command context to carry audio transcription parameters such as input file, language, and translate flag.
- Mount audio files into containers automatically when running transcription commands.


Crystora:
@rhatdan @ieaves Could you please review this PR?

Crystora:
@rhatdan I'd appreciate your feedback

rhatdan:
Does this work with STDIN?

```
cat audio.wav | ramalama transcribe huggingface://ggerganov/whisper.cpp/ggml-tiny.bin
```
Or
```
ramalama transcribe huggingface://ggerganov/whisper.cpp/ggml-tiny.bin audio.wav -
```

Not sure if adding a new command is necessary or not. Would doing something like 

```
$ ramalama run huggingface://ggerganov/whisper.cpp/ggml-tiny.bin --transcribe audio.wav
```

Make sense?

Would 
ramalama transcribe run against the whisper server?  What would be the syntax?

Can I override the image with something else besides whisper?

@olliewalsh @engelmi @ieaves @mikebonnet @jtligon @bmahabirbu WDYT?

Crystora:
> Does this work with STDIN?
> 
> ```
> cat audio.wav | ramalama transcribe huggingface://ggerganov/whisper.cpp/ggml-tiny.bin
> ```
> 
> Or
> 
> ```
> ramalama transcribe huggingface://ggerganov/whisper.cpp/ggml-tiny.bin audio.wav -
> ```
> 
> Not sure if adding a new command is necessary or not. Would doing something like
> 
> ```
> $ ramalama run huggingface://ggerganov/whisper.cpp/ggml-tiny.bin --transcribe audio.wav
> ```
> 
> Make sense?
> 
> Would ramalama transcribe run against the whisper server? What would be the syntax?
> 
> Can I override the image with something else besides whisper?
> 
> @olliewalsh @engelmi @ieaves @mikebonnet @jtligon @bmahabirbu WDYT?

==================

STDIN: Not supported right now. It needs an audio file path (the file is mounted into the container). We could add STDIN later by saving it to a temp file first.

Why ramalama transcribe instead of ramalama run --transcribe: Transcribe is a one-shot “audio in, text out” command. run is for interactive chat, so keeping them separate avoids messy argument handling. I can change it if the team prefers run --transcribe.

Using a running Whisper server: Not supported yet. It starts a new container each time. We could add a --server option later to send audio to an existing server.

Custom image: Yes, --image works.

Crystora:
@rhatdan Please let me know your option

rhatdan:
Both of these commands below are single run commands as well:

```
ramalama run MODEL "Explain Nuclear Fision"
echo "Explain Nuclear Fision" | ramalama run MODEL
```

I am just wondering if users would want to something like:

/usr/bin/audiofeed | ramalama run --transcribe MODEL

But if whisper can not handle this, then we would have to live with it.

Does whisper.cpp handle STDIN at all?  If yes then rather then mounting into the container, just handle the input with stdin all the time.

I really want to hear other thoughts on this.

Crystora:
@rhatdan I see your consideration. But whisper.cpp can’t take audio from STDIN. This is whisper.cpp limitation.
whisper-cli needs to jump around inside the audio file (read the header, skip to the middle, etc.). With a file path, it can do that. With a pipe/STDIN, the audio only comes in order, so it can’t do that.
So you have to pass a file path with -f . With the current whisper.cpp, bind-mounting the file is the only option.

ieaves:
This is great! A few thoughts

## Streaming vs. Bulk Transcription

There actually has been some progress on streaming audio with whisper checkout Issue [3521](https://github.com/ggml-org/whisper.cpp/issues/3521). It looks like the ggml team did some work on it [a few years ago](https://github.com/ggml-org/whisper.cpp/tree/master/examples/stream) and there was a release about two weeks ago for streaming [pcm audio](https://github.com/rmorse/whisper.cpp/releases/tag/v1.8.3-stream-pcm) over STDIN/pipes in particular.

From my perspective, something like `ramalama transcribe <path_to_file>` is still useful. That being said, it would be more useful if the transcription were written to STDOUT, so normal shell redirection works. e.g.

```bash
ramalama transcribe /path/to/audio.wave > my_file.txt
```

## CLI structure

For chat interfaces we currently expose three touchpoints

* serve
* chat
* run

Where run is a combo of serve + chat. Should we have a similar functional layout for other modalities (voice + images)?

* serve -> long running whisper server with a mountpoint for audio or STDIN
* transcribe -> pipes STDIN or copies files to mountpoint
* run -> serve + transcribe

Or something else along those lines?

## Misc

* I know we use the `hasattr` / `getattr` flow frequently at the moment but it's a weak pattern and a common source of bugs. Could you please formally types for these CLI paths? There are existing patterns and utilities [here](https://github.com/containers/ramalama/blob/main/ramalama/arg_types.py) to lean on.

* With multimodal capabilities the RUNTIME (e.g. llama.cpp vs. vllm vs. whisper.cpp) starts to seem a little less clear to me. Not sure yet whether this is a problem or not.

olliewalsh:
Similar to the almost identical PR to add stable-diffusion support, the code-base needs to be refactored to allow arbitrary runtimes to be used. I have marked https://github.com/containers/ramalama/issues/2376 as blocked until then. Closing this PR for now.

PR #2445, opened 2026-02-21
Add missing XDG environment variable checks

alyssais:
Just like was already done elsewhere in ramalama/config.py. Otherwise, ramalama would do the wrong thing when XDG_CONFIG_HOME or XDG_DATA_HOME is set to a non-default location.

## Summary by Sourcery

Respect XDG base directory environment variables for Ramalama configuration and data file discovery.

Bug Fixes:
- Use XDG_CONFIG_HOME and XDG_DATA_HOME when resolving the shortnames configuration file path.
- Use XDG_DATA_HOME when determining the default Ramalama data store directory for non-root users.

### DIFF
--- a/ramalama/config.py
+++ b/ramalama/config.py
@@ -83,7 +83,7 @@ def get_default_store() -> str:
     if hasattr(os, 'geteuid') and os.geteuid() == 0:
         return "/var/lib/ramalama"
 
-    return os.path.expanduser("~/.local/share/ramalama")
+    return os.path.expanduser(os.path.join(os.getenv("XDG_DATA_HOME", "~/.local/share"), "ramalama"))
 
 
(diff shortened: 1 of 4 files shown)


rhatdan:
LGTM
Thanks @alyssais

ramalama/config.py
```diff
@@ -83,7 +83,7 @@ def get_default_store() -> str:
     if hasattr(os, 'geteuid') and os.geteuid() == 0:
         return "/var/lib/ramalama"
 
-    return os.path.expanduser("~/.local/share/ramalama")
+    return os.path.expanduser(os.path.join(os.getenv("XDG_DATA_HOME", "~/.local/share"), "ramalama"))
```
olliewalsh, line 86: Thanks for the PR @alyssais. Could you please update the docs too?

Issue #2447, opened 2026-02-22
ramalama serve with llama-stack on ROCM based GPU fails to detect GPU


mkristian:
### Issue Description

Some commands with `--api=llama-stack` can not detect ROCM based GPU

### Steps to reproduce the issue

```
ramalama server some-model --api=llama-stack
```

### Describe the results you received

llama.cpp server fails to detect the GPU
```
ggml_cuda_init: failed to initialize ROCm: no ROCm-capable device is detected
load_backend: loaded ROCm backend from /usr/bin/libggml-hip.so
load_backend: loaded RPC backend from /usr/bin/libggml-rpc.so
load_backend: loaded CPU backend from /usr/bin/libggml-cpu-zen4.so
main: n_parallel is set to auto, using n_parallel = 4 and kv_unified = true
build: 1 (b45ef27) with GNU 15.2.1 for Linux x86_64
```

### Describe the results you expected

The same command without the `--api=llama-stack`
```
ggml_cuda_init: found 1 ROCm devices:
  Device 0: AMD Radeon RX 9060 XT, gfx1200 (0x1200), VMM: no, Wave Size: 32
load_backend: loaded ROCm backend from /usr/bin/libggml-hip.so
load_backend: loaded RPC backend from /usr/bin/libggml-rpc.so
load_backend: loaded CPU backend from /usr/bin/libggml-cpu-zen4.so
main: n_parallel is set to auto, using n_parallel = 4 and kv_unified = true
build: 1 (b45ef27) with GNU 15.2.1 for Linux x86_64

```

### ramalama info output

```yaml
{
    "Accelerator": "hip",
    "Config": {
        "benchmarks": {},
        "provider": {
            "openai": {}
        },
        "settings": {
            "config_files": [
                "/home/christian/.local/share/uv/tools/ramalama/share/ramalama/ramalama.conf"
            ]
        },
        "user": {}
    },
    "Engine": {
        "Info": {
            "host": {
                "arch": "amd64",
                "buildahVersion": "1.42.2",
                "cgroupControllers": [
                    "cpuset",
                    "cpu",
                    "io",
                    "memory",
                    "pids"
                ],
                "cgroupManager": "systemd",
                "cgroupVersion": "v2",
                "conmon": {
                    "package": "conmon-1:2.2.1-1.1",
                    "path": "/usr/bin/conmon",
                    "version": "conmon version 2.2.1, commit: c8cc2c4db27531bd4e084ce7857f73cd21ee639d"
                },
[...]

PR #2448, opened 2026-02-22
Fixes the llama-stack for ROCM based GPUs

mkristian:
Fixes: #2447. Adding /dev/kfd to mounted volume alongside /dev/dri if the device exists

## Summary by Sourcery

Enhancements:
- Mount /dev/dri and /dev/kfd into the container only if they exist on the host when DRI support is enabled


ramalama/stack.py
```diff
@@ -53,9 +53,11 @@ def _gen_volume_mounts(self):
           name: model"""
 
         if self.args.dri == "on" and platform.system() != "Windows":
-            volume_mounts += """
-        - mountPath: /dev/dri
-          name: dri"""
+            for dev in ["dri", "kfd"]:
+                if os.path.exists("/dev/" + dev):
+                    volume_mounts += f"""
+        - mountPath: /dev/{dev}
```
olliewalsh: I suspect this is not necessary for CUDA - nvidia-container-runtime handles it. For ROCm there is an equivalent which, at a glance, suggests it should also just work without explicitly mounting devices - https://github.com/ROCm/container-toolkit?tab=readme-ov-file#docker-runtime-integration

@bmahabirbu WDYT?

rhatdan:
Thanks @mkristian for your contribution, please follow the AI suggestions, for now.

ramalama/stack.py
mkristian: Well this loop is basically stolen from https://github.com/containers/ramalama/blob/main/ramalama/kube.py#L83 as this was the one which just works and it also **fixes** this setup for ROCm with me.

Once you question this then you need to uniform all places where you mount devices, like in the quadlet as well https://github.com/containers/ramalama/blob/main/ramalama/quadlet.py#L62-L64 - the quadlet also works for my ROCm GPU. but as of now this stack setup does not work.

mkristian: the compose generator does the same thing https://github.com/containers/ramalama/blob/main/ramalama/compose.py#L90 and even looks for /dev/accel

probably it makes sense to be consistent across all these 'generators'

ramalama/stack.py
```diff
@@ -69,10 +71,12 @@ def _gen_volumes(self):
           path: {host_model_path}
         name: model"""
         if self.args.dri == "on":
-            volumes += """
+            for dev in ["dri", "kfd"]:
+                if os.path.exists("/dev/" + dev):
```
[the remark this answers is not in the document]
mkristian: I am not sure if this makes sense here to be platform dependent as we look at linux devices anyways

ramalama/stack.py
```diff
@@ -53,9 +53,11 @@ def _gen_volume_mounts(self):
           name: model"""
 
         if self.args.dri == "on" and platform.system() != "Windows":
-            volume_mounts += """
-        - mountPath: /dev/dri
-          name: dri"""
+            for dev in ["dri", "kfd"]:
+                if os.path.exists("/dev/" + dev):
```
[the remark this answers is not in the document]
mkristian: I am not sure if this makes sense here to be platform dependent as we look at linux devices anyways

mkristian:
any ways I added a commit which uniforms all these places where some of devices for GPU are added or not.

I also added the case for stack generate for quadlet and compose at least whey do something now instead of just starting the server. Addressing partly  #2446

Any ways the tests are failing and before I deep dive into these and need a small feedback if this goes into the right direction. thanks

container-images/llama-stack/Containerfile
```diff
@@ -1,21 +1,17 @@
 FROM quay.io/fedora/fedora:43
 
-ARG RAMALAMA_STACK_VERSION=0.2.5
-
-# hack that should be removed when the following bug is addressed
-# https://github.com/containers/ramalama-stack/issues/53
-RUN curl --create-dirs --output ~/.llama/providers.d/remote/inference/ramalama.yaml https://raw.githubusercontent.com/containers/ramalama-stack/refs/tags/v${RAMALAMA_STACK_VERSION}/src/ramalama_stack/providers.d/remote/inference/ramalama.yaml && \
-    curl --create-dirs --output /etc/ramalama/ramalama-run.yaml https://raw.githubusercontent.com/containers/ramalama-stack/refs/tags/v${RAMALAMA_STACK_VERSION}/src/ramalama_stack/ramalama-run.yaml
+ARG RAMALAMA_STACK_VERSION=0.2.6
+COPY ramalama_stack-0.2.6.dev* /
```
mikebonnet: Where is this file coming from? It's not committed to the repo.

container-images/llama-stack/Containerfile
```diff
@@ -1,21 +1,17 @@
...
 
 RUN dnf -y install uv gcc python3-devel && \
     dnf -y clean all
 
-RUN uv venv && \
-    uv pip install ramalama-stack==${RAMALAMA_STACK_VERSION} \
-                   milvus-lite
+RUN --mount=type=cache,target=/root/.cache/uv \
```
mikebonnet: This cache is unhelpful, official builds are run in Konflux on ephemeral VMs.

mikebonnet: If you want to use a cache for local builds, you can modify your `podman build` command to do so without modifying the Containerfile.

container-images/llama-stack/Containerfile
```diff
@@ -1,21 +1,17 @@
...
+    uv venv && \
+    uv pip install ramalama_stack-${RAMALAMA_STACK_VERSION}.dev26.tar.gz
 
-COPY --chmod=755 container-images/llama-stack/entrypoint.sh /usr/bin/entrypoint.sh
+COPY --chmod=755 entrypoint.sh /usr/bin/entrypoint.sh
```
mikebonnet: The top-level directory of the project is the context directory for all builds, so the full path is required. When running local builds, run with `-f container-images/lama-stack/Containerfile .`.

container-images/llama-stack/Containerfile
```diff
@@ -1,21 +1,17 @@
...
 
 ENTRYPOINT [ "/usr/bin/entrypoint.sh" ]
 
-CMD [ "llama",  "stack",  "run", "--image-type", "venv", "/etc/ramalama/ramalama-run.yaml" ]
+CMD [ "llama",  "stack",  "run", "--image-type", "venv", "/root/.llama/distributions/ramalama/ramalama-run.yaml" ]
```
mikebonnet: Where is this file coming from?

container-images/llama-stack/Containerfile
mkristian: sorry should not be part of this PR

container-images/llama-stack/Containerfile
mkristian: sorry should not be part of this PR

container-images/llama-stack/Containerfile
mkristian: sorry should not be part of this PR

container-images/llama-stack/Containerfile
mkristian: sorry should not be part of this PR

mkristian: it comes from here at least if I install it via the tar.gz
https://github.com/containers/ramalama-stack/blob/main/setup.py#L26-L34

rhatdan:
This needs a rebase.

ramalama/stack.py
bmahabirbu: Can confirm for ROCM all you need is /dev/dri for proper gpu passthrough then the library handles the rest sorry for the late reply

mkristian:
@mikebonnet I removed the unintended changes on this Containerfile and adjusted the tests (actually it is the first time I ran a `make test` and it was complete green). Can we move here as well ?

ramalama/engine.py
```diff
@@ -93,9 +93,10 @@ def add_device_options(self):
                 self.exec_args += ["--device", device_arg]
 
         if ramalama.common.podman_machine_accel:
-            self.exec_args += ["--device", "/dev/dri"]
+            for dev in get_gpu_devices().values():
+                self.exec_args += ["--device", dev]
 
-        for path in ["/dev/dri", "/dev/kfd", "/dev/accel", "/dev/davinci*", "/dev/devmm_svm", "/dev/hisi_hdc"]:
+        for path in ["/dev/davinci*", "/dev/devmm_svm", "/dev/hisi_hdc"]:
```
mikebonnet: You need to handle the case when `ramalama.common.podman_machine_accel is False`, which is the general case on non-Apple hardware. Otherwise `/dev/accel`, etc. won't be added.

ramalama/quadlet.py
```diff
@@ -59,9 +59,8 @@ def generate(self) -> list[UnitFile]:
         quadlet_file = UnitFile(container_file_name)
         quadlet_file.add("Unit", "Description", f"RamaLama {self.name} AI Model Service")
         quadlet_file.add("Unit", "After", "local-fs.target")
-        quadlet_file.add("Container", "AddDevice", "-/dev/accel")
-        quadlet_file.add("Container", "AddDevice", "-/dev/dri")
-        quadlet_file.add("Container", "AddDevice", "-/dev/kfd")
+        for dev in get_gpu_devices().values():
+            quadlet_file.add("Container", "AddDevice", "-" + dev)
```
mikebonnet: The leading `-` is no longer necessary, since you've already checked for the existence of the device.

mikebonnet: I would suggest removing this change because it ties the quadlet to the specific machine it was generated on (based on the existence of specific devices) rather than making it more general purpose (including the devices in the quadlet if they exist at runtime).

mkristian: yup - this make sense - reverted it.

ramalama/engine.py
mkristian: no need to change the working logic here so I reverted this. The problem were the incomplete devices in stack.py and this file shares code now with kube.py and compose.py that seems good enough

mikebonnet (approved):
Thanks for the fixes and for your patience working through the issues!

PR #2451, opened 2026-02-23
fixed ctx_size being wired to max_tokens in mlx

ieaves:
We have historically wired the context size into `--max-tokens` in mlx, however, `--max-tokens` controls the max number of tokens to generate not the model context size.

## Summary by Sourcery

Adjust MLX runtime behavior and tests so context size no longer implicitly controls max tokens, and explicitly passing max-tokens is handled and verified.

Tests:
- Update MLX dry-run e2e test to verify that specifying max-tokens propagates correctly.
- Add e2e and system tests ensuring ctx-size does not set or map to max-tokens for the MLX runtime.
- Add system test to validate MLX dry-run run with explicit max-tokens succeeds and includes the corresponding flag in the output.

### DIFF
--- a/inference-spec/engines/mlx.yaml
+++ b/inference-spec/engines/mlx.yaml
@@ -15,7 +15,7 @@ commands:
           description: "Seed the global PRNG"
           value: "{{ args.seed }}"
         - name: "--max-tokens"
-          description: "Maximum number of tokens to generate (0 = unlimited)"
+          description: "Maximum number of tokens to generate (0 = mlx default)"
           value: "{{ args.max_tokens }}"
           if: "{{ (args.max_tokens or 0) > 0 }}"
         - name: "--host"


inference-spec/engines/mlx.yaml
```diff
@@ -15,9 +15,9 @@ commands:
           description: "Seed the global PRNG"
           value: "{{ args.seed }}"
         - name: "--max-tokens"
-          description: "Size of the prompt context"
-          value: "{{ args.ctx_size }}"
-          if: "{{ (args.ctx_size or 0) > 0 }}"
+          description: "Maximum number of tokens to generate (0 = unlimited)"
```
olliewalsh: default is 512

ieaves: What change are you proposing? Are you asking to include a specific reference to 512 as the default? As written, we omit the argument unless provided by the user which defers the choice up to mlx, where they default to 512. Is the goal for our docs to maintain parity with their docs?

olliewalsh: Just the description e.g `description: "Maximum number of tokens to generate (0 = mlx default)"`

ieaves: Done.

PR #2453, opened 2026-02-23
ci: only install podman and crun for podman5 install task

olliewalsh:
apt-get upgrade was upgrading a number of unncessary packages. Switch to apt-get install instead.

Reduces the podman5 install step to ~40s. Previously this would take at least 3-4mins but in some cases could take up to 30mins.

## Summary by Sourcery

Adjust CI workflows to avoid unnecessary package upgrades and explicitly manage Firefox removal and Podman installation.

CI:
- Replace apt-get upgrade with apt-get install for Podman and crun to prevent upgrading unrelated packages in CI jobs.
- Ensure Firefox is explicitly purged in relevant CI job steps before running package installations.
- Remove Podman from generic package install steps, limiting Podman installation to the dedicated Podman 5 setup tasks.


olliewalsh:
@mikebonnet PTAL

ieaves (approved):

olliewalsh:
e2e CI is failing when trying to run a local registry. One of the other package upgrades appears to be significant but not obvious which/why. Setting the PR to draft for now...

ieaves:
@olliewalsh  It's not an issue with this PR. That test is failing everywhere. I was working on this yesterday but can't reproduce it locally. It seems like there's an issue with podman not finding the built imageid in the CI job.

olliewalsh:
> @olliewalsh It's not an issue with this PR. That test is failing everywhere. I was working on this yesterday but can't reproduce it locally. It seems like there's an issue with podman not finding the built imageid in the CI job.

It's passing on main, something about this change seems to be triggering the issue.
Github has issues yesterday though, probably what you were hitting.

ieaves:
*shrug* I hit this exact test failure, with the exact same traceback (modulo the imageid) on two different PRs related to the inference spec ctx size yesterday. I'm skeptical three different PRs covering unrelated functional capabilities all triggered the exact same error in the exact same way but, who knows, maybe so.

PR #2455, opened 2026-02-24
Use hardlink for local model files to save disk space

### LINKED ISSUE #2098: Support local model files without duplicating into model store

Currently when using `ramalama run file://model` the model file will be copied into the ramalama model store. Even when the user provides a symlink to a model file the symlink target will be followed and the actual model file will be copied into the model store.

Allow users to bring their own model files without needing to duplicate the model files.

Users have large models downloaded external to ramalama with lmstudio or other tools.

Support model files continuing to be used by external tools other than ramalama without duplicating these large model files.

Add a command line argument to allow the user to specify whether the model file should be copied into the model store. Should result in a symlink targeting the model file being created in the model store instead of copying the model file. Something like `--no-model-copy`.

Other considerations:

- only for `file://model` type url or also for other urls using a local model cache external to ramalama model store
- support split model urls: `file://models/model-00001-of-00003.gguf`

Tried using a symlink to the model file with a one line change to not follow the symlink. See comment [here](https://github.com/containers/ramalama/issues/1989#issuecomment-3368304212). But that was a lazy implementation that would require the user to manually create symlinks.

Only other option I could think of would be lletting ramalama control where the model is stored and trying to get other tools to use the model in the ramalama model store.

### Additional context

#1989 original related enhancement request in the context of split models

rhatdan:
When pulling local model files (file://), use hardlinks instead of copying the file to avoid duplicating large model files on disk. This addresses issue #2098 where users want to use models from external tools without duplicating multi-GB files.

The implementation uses the existing create_file_link() utility which:
1. First attempts to create a hardlink (most efficient, no extra disk space)
2. Falls back to symlink if hardlink fails (e.g., cross-filesystem)
3. Falls back to copy as last resort (e.g., filesystem doesn't support links)

Benefits:
- No disk space duplication for local models
- Works across all platforms (Linux, macOS, Windows)
- Maintains backward compatibility (copy as fallback)
- Handles edge cases like cross-filesystem scenarios

Changes:
- Updated LocalModelFile.download() to use create_file_link() instead of shutil.copy()
- Added test to verify hardlink behavior
- Updated comment to reflect new hardlink-first approach

Fixes #2098

*This PR was created with the assistance of Cursor AI.*

## Summary by Sourcery

Use file linking instead of copying for local model downloads to avoid duplicating large files on disk.

Bug Fixes:
- Prevent unnecessary duplication of local model files by replacing copies with hardlinks or symlinks when downloading.

Enhancements:
- Update LocalModelFile downloads to use a link-creation utility that prefers hardlinks, then symlinks, and falls back to copying if needed.

Tests:
- Add a unit test verifying that LocalModelFile.download() creates a hardlink for local model files and preserves file contents.

### DIFF
--- a/ramalama/transports/url.py
+++ b/ramalama/transports/url.py
@@ -1,11 +1,10 @@
 import os
 import re
-import shutil
 from pathlib import Path
 
 from ramalama.common import SPLIT_MODEL_PATH_RE, generate_sha256, is_split_file_model
 from ramalama.model_store.snapshot_file import SnapshotFile, SnapshotFileType
-from ramalama.path_utils import normalize_host_path_for_container
+from ramalama.path_utils import create_file_link, normalize_host_path_for_container
 from ramalama.transports.base import Transport
 from ramalama.transports.huggingface import HuggingfaceRepository
 from ramalama.transports.modelscope import ModelScopeRepository
@@ -36,8 +35,8 @@ class LocalModelFile(SnapshotFile):
     def download(self, blob_file_path, snapshot_dir):
         if not os.path.exists(self.url):
             raise FileNotFoundError(f"No such file: '{self.url}'")
-        # moving from the local location to blob directory so the model store "owns" the data
-        shutil.copy(self.url, blob_file_path)
+        # Use hardlink first (space-efficient), fallback to symlink, then copy if needed
+        create_file_link(self.url, blob_file_path)
         return os.path.relpath(blob_file_path, start=snapshot_dir)
 
 
--- a/test/unit/test_url.py
+++ b/test/unit/test_url.py
@@ -1,9 +1,10 @@
+import os
 from dataclasses import dataclass, field
 
 import pytest
 
 from ramalama.model_store.snapshot_file import SnapshotFile
-from ramalama.transports.url import URL
+from ramalama.transports.url import URL, LocalModelFile
 
 
 @dataclass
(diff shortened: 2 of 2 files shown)


rhatdan:
@olliewalsh @engelmi do you think the risk of symlinks being broken is worth copying all the time, if hard links do not work.  I think the risk of users changing permissions on hard linked files is not worth the additional space saved by this patch.

rhatdan:
@engelmi @olliewalsh @ieaves thoughts on this PR?

olliewalsh:
> @engelmi @olliewalsh @ieaves thoughts on this PR?

LGTM. We could try to be clever with this, catch cases where the ownership has changed etc...  but it's already an edge case. An e2e test to ensure it fails well in those conditions might be a good idea.

ramalama/transports/url.py
```diff
@@ -36,8 +35,8 @@ def __init__(
     def download(self, blob_file_path, snapshot_dir):
         if not os.path.exists(self.url):
             raise FileNotFoundError(f"No such file: '{self.url}'")
-        # moving from the local location to blob directory so the model store "owns" the data
-        shutil.copy(self.url, blob_file_path)
+        # Use hardlink first (space-efficient), fallback to symlink, then copy if needed
+        create_file_link(self.url, blob_file_path)
```
giuseppe, line 39: should you still try to copy if the hardlink fails?  Could source and destination be on different file systems?

olliewalsh: create_file_link will fallback to copy

olliewalsh:
@rhatdan all of the test_artifact e2e tests are failing on main since this merged

Issue #2379, opened 2026-01-30
Make MODEL runtimes pluggable


ieaves:
Define a syntax that simplifies adding new runtimes.
Current runtimes:

    llama.cpp
    vllm
    mlx

olliewalsh:
Note: there is runtime dependant logic beyond just the command line to be run (e.g health check logic, model switching apis etc...). The inference engine spec approach may not be the best method to represent this. Alternatives such as python plugins could be considered.

PR #2456, opened 2026-02-24
Major refactor to use pluggable modules for inference runtimes

olliewalsh:
Related to #2379 

Inference runtime logic is spread throughout the code-base, particularly in `cli.py` and `transport/base.py`. Much of the functionality is runtime specific, in many cases the entire subcommand is llama.cpp specific.

The data-driven inference spec approach is too rigid to represent this. This change attempts to switch to a python plugin approach instead, and refactor the runtime specific logic out of the core modules and into the relevant runtime plugin. This leaves the core modules to primarily handle managing the models and interacting with the container engine

There is still more work to be done, e.g I've largely ignored the daemon. Convert handling in the oci transport needs further refactoring.

However this is already a very significant refactor so I'm creating a draft PR for review/feedback on the general approach.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

# Summary from Claude

plugins: replace YAML+Jinja2 inference specs with Python runtime plugin system

Replaces the YAML+Jinja2 inference specification system with a Python plugin architecture for runtime inference engine management.

## Plugin system

Each inference engine is now a self-contained Python plugin discovered via importlib entry points (group: "ramalama.runtimes"):

- ramalama/plugins/interface.py — RuntimePlugin(ABC) and InferenceRuntimePlugin(ABC): the public contracts every plugin must satisfy.  InferenceRuntimePlugin adds @abstractmethod _cmd_run and three inference-specific hooks (chat_prefix, handle_nocontainer_chat, api_model_name).
- ramalama/plugins/runtimes/common.py — BaseInferenceRuntime: concrete base with run/serve subcommand registration and _run_handler / _serve_handler.  ContainerizedInferenceRuntimePlugin extends it with the RAG subcommand (only registered when containers are enabled).
- ramalama/plugins/runtimes/llama_cpp_commands.py — LlamaCppCommands mixin: pure command builders (_cmd_run, _cmd_serve, _cmd_perplexity, _cmd_bench, _cmd_rag, _cmd_run_rag, _cmd_serve_rag, _cmd_convert, _cmd_quantize) plus llama.cpp-specific constants and _default_threads.
- ramalama/plugins/runtimes/llama_cpp.py — LlamaCppPlugin inherits from LlamaCppCommands + ContainerizedInferenceRuntimePlugin; owns image selection, health checks, argparse registration, container operations, and command handlers for bench/perplexity/convert/benchmarks.
- ramalama/plugins/runtimes/vllm.py — VllmPlugin: vLLM support with its own image defaults and max_model_len arg.
- ramalama/plugins/runtimes/mlx.py — MlxPlugin: macOS/Apple Silicon only, always --nocontainer.  Implements all four inference hooks to keep MLX-specific behaviour out of shared code.
- ramalama/plugins/loader.py — get_runtime(name), get_all_runtimes(), assemble_command() via PluginRegistry.
- ramalama/plugins/registry.py — generic PluginRegistry[T] backed by entry points.

## Class hierarchy

```
  RuntimePlugin (ABC)                      interface.py
    └── InferenceRuntimePlugin (ABC)       interface.py
          └── BaseInferenceRuntime         runtimes/common.py
                └── ContainerizedInferenceRuntimePlugin
                      └── LlamaCppPlugin (+ LlamaCppCommands mixin)
                      └── VllmPlugin
                └── MlxPlugin
```
## Transport changes

- bench, perplexity, serve, run subcommand handlers moved from Transport base class into plugin _*_handler methods.
- _convert_to_gguf and _quantize moved from Transport into LlamaCppPlugin.
- Runtime-specific CLI args (--ngl, --threads, --cache-reuse, --thinking, --webui, --model-draft, --max-model-len) moved into each plugin's _add_*_inference_args and register_subcommands methods.
- MLX-specific branches (== "mlx" checks) removed from chat.py and transports/base.py; behaviour now lives in MlxPlugin hooks.
- build_command dispatches via _cmd_<subcommand> naming convention.

## Config changes

- RamalamaImages and RamalamaRagImages dataclasses removed from config.py; images and rag_images are now plain dict[str, str]. Plugin-owned image defaults live in each plugin (LlamaCppPlugin, VllmPlugin); RAG image defaults live in rag.py (_DEFAULT_RAG_IMAGES).
- accel_image() bug fixed: plugin delegation now only runs when no explicit images= dict is passed (previously it ran even for RAG image selection, returning the wrong image).
- stack.py: fix args.context → args.ctx_size in _gen_llama_args.

## arg_types cleanup

- ServeRunArgsType removed: it was a monolithic Protocol listing every possible serve/run field across all runtimes, with stale field names and llama.cpp-specific fields.  RagArgsType now extends DefaultArgsType directly with only the fields RAG actually needs.

## Removed

- inference-spec/ directory (YAML specs + JSON schema)
- ramalama/command/ directory (factory, context, schema, error)
- test/unit/command/ directory

### DIFF
--- a/ramalama/plugins/runtimes/inference/common.py
+++ b/ramalama/plugins/runtimes/inference/common.py
@@ -2,7 +2,25 @@ import argparse
 from abc import abstractmethod
 from typing import Any
 
+from ramalama.cli import (
+    GENERATE_OPTIONS,
+    OverrideDefaultAction,
+    chat_run_options,
+    local_models,
+    parse_generate_option,
+    parse_port_option,
+    runtime_options,
+    suppressCompleter,
+)
+from ramalama.common import ContainerEntryPoint, set_accel_env_vars
+from ramalama.config import get_config
+from ramalama.logger import logger
 from ramalama.plugins.interface import InferenceRuntimePlugin
+from ramalama.plugins.loader import assemble_command
+from ramalama.stack import Stack
+from ramalama.transports.api import APITransport
+from ramalama.transports.base import compute_serving_port
+from ramalama.transports.transport_factory import New, TransportFactory
 
 
 class BaseInferenceRuntime(InferenceRuntimePlugin):
@@ -37,9 +55,6 @@ class BaseInferenceRuntime(InferenceRuntimePlugin):
 
     def _add_inference_args(self, parser: "argparse.ArgumentParser", command: str) -> None:
         """Add inference-specific args shared across all runtimes for run/serve/perplexity."""
-        from ramalama.cli import OverrideDefaultAction, parse_port_option, suppressCompleter
-        from ramalama.config import get_config
-
         config = get_config()
         parser.add_argument(
             "-c",
@@ -101,8 +116,6 @@ class BaseInferenceRuntime(InferenceRuntimePlugin):
             self._register_serve_subcommand(subparsers)
 
     def _register_run_subcommand(self, subparsers: "argparse._SubParsersAction") -> "argparse.ArgumentParser":
-        from ramalama.cli import chat_run_options, local_models, runtime_options, suppressCompleter
-
         parser = subparsers.add_parser("run", help="run specified AI Model as a chatbot")
         runtime_options(parser, "run")
         self._add_inference_args(parser, "run")
@@ -124,8 +137,6 @@ class BaseInferenceRuntime(InferenceRuntimePlugin):
         return parser
 
     def _register_serve_subcommand(self, subparsers: "argparse._SubParsersAction") -> "argparse.ArgumentParser":
-        from ramalama.cli import local_models, runtime_options
-
         parser = subparsers.add_parser("serve", help="serve REST API on specified AI Model")
         runtime_options(parser, "serve")
         self._add_inference_args(parser, "serve")
@@ -135,10 +146,6 @@ class BaseInferenceRuntime(InferenceRuntimePlugin):
 
     def _do_run(self, args: argparse.Namespace, model: "Any") -> None:
         """Execute run after the model is resolved. Override to inject pre-run logic."""
-        from ramalama.common import ContainerEntryPoint
-        from ramalama.plugins.loader import assemble_command
-        from ramalama.transports.api import APITransport
-
         if isinstance(model, APITransport):
             model.run(args, [])
             return
@@ -152,9 +159,6 @@ class BaseInferenceRuntime(InferenceRuntimePlugin):
 
     def _do_serve(self, args: argparse.Namespace, model: "Any") -> None:
         """Execute serve after the model is resolved. Override to inject pre-serve logic."""
-        from ramalama.common import set_accel_env_vars
-        from ramalama.plugins.loader import assemble_command
-
         set_accel_env_vars()
         cmd = assemble_command(args)
         if getattr(args, "generate", None):
@@ -167,10 +171,6 @@ class BaseInferenceRuntime(InferenceRuntimePlugin):
             raise e
 
     def _run_handler(self, args: argparse.Namespace) -> None:
-        from ramalama.logger import logger
-        from ramalama.transports.base import compute_serving_port
-        from ramalama.transports.transport_factory import New, TransportFactory
-
         try:
             # detect available port and update arguments
             args.port = compute_serving_port(args)
@@ -188,11 +188,6 @@ class BaseInferenceRuntime(InferenceRuntimePlugin):
         self._do_run(args, model)
 
     def _serve_handler(self, args: argparse.Namespace) -> None:
-        from ramalama.stack import Stack
-        from ramalama.transports.api import APITransport
-        from ramalama.transports.base import compute_serving_port
-        from ramalama.transports.transport_factory import New, TransportFactory
-
         if not args.container:
             args.detach = False
 
@@ -234,9 +229,6 @@ class ContainerizedInferenceRuntimePlugin(BaseInferenceRuntime):
     """Base class for inference plugins that support container-dependent args"""
 
     def _add_containerized_inference_args(self, parser: "argparse.ArgumentParser", command: str) -> None:
-        from ramalama.cli import GENERATE_OPTIONS, parse_generate_option
-        from ramalama.config import get_config
-
         config = get_config()
         parser.add_argument(
             "--api",
@@ -268,16 +260,12 @@ class ContainerizedInferenceRuntimePlugin(BaseInferenceRuntime):
             )
 
     def _register_run_subcommand(self, subparsers: "argparse._SubParsersAction") -> "argparse.ArgumentParser":
-        from ramalama.config import get_config
-
         parser = super()._register_run_subcommand(subparsers)
         if get_config().container:
             self._add_containerized_inference_args(parser, "run")
         return parser
 
     def _register_serve_subcommand(self, subparsers: "argparse._SubParsersAction") -> "argparse.ArgumentParser":
-        from ramalama.config import get_config
-
         parser = super()._register_serve_subcommand(subparsers)
         if get_config().container:
             self._add_containerized_inference_args(parser, "serve")
--- a/ramalama/chat.py
+++ b/ramalama/chat.py
@@ -36,6 +36,8 @@ from ramalama.file_loaders.file_manager import OpanAIChatAPIMessageBuilder
 from ramalama.logger import logger
 from ramalama.mcp.mcp_agent import LLMAgent
 from ramalama.mcp.mcp_client import PureMCPClient
+from ramalama.plugins.interface import InferenceRuntimePlugin
+from ramalama.plugins.loader import get_runtime
 from ramalama.proxy_support import setup_proxy_support
 
 # Setup proxy support on module import
@@ -272,9 +274,6 @@ class RamaLamaShell(cmd.Cmd):
         return self.provider.create_request(messages, options)
 
     def _resolve_model_name(self) -> str | None:
-        from ramalama.plugins.interface import InferenceRuntimePlugin
-        from ramalama.plugins.loader import get_runtime
-
         runtime = getattr(self.args, "runtime", None)
         if runtime:
             plugin = get_runtime(runtime)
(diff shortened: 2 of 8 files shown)


olliewalsh:
@engelmi PTAL

olliewalsh:
Draft until #2468 merges

olliewalsh:
@mikebonnet FYI

ramalama/plugins/interface.py
```diff
@@ -0,0 +1,47 @@
+import argparse
+from abc import ABC, abstractmethod
+from http.client import HTTPConnection
+from typing import Any
+
+
+class RuntimePlugin(ABC):
```
olliewalsh, line 7: this is the interface for a runtime plugin e.g whisper etc..., where it can add subcommands to the CLI, and define handlers for the subcommands

ramalama/plugins/interface.py
```diff
@@ -0,0 +1,47 @@
...
+    def service_ready_check_timeout(self) -> int:
+        """Seconds to wait for the runtime server to become ready."""
+        return 180
+
+    def service_ready_check(self, conn: HTTPConnection, args: Any, model_name: str | None = None) -> bool:
+        """Check if the service is ready to receive requests."""
+        return True
+
+
+class InferenceRuntimePlugin(RuntimePlugin, ABC):
```
olliewalsh, line 41: this is the interface for inference runtimes specifically i.e runtimes that provide an openai API that the chat client can connect to

ramalama/cli.py
```diff
@@ -323,24 +307,23 @@ def configure_subcommands(parser):
...
     chat_parser(subparsers)
     containers_parser(subparsers)
-    convert_parser(subparsers)
     help_parser(subparsers)
     info_parser(subparsers)
     inspect_parser(subparsers)
     list_parser(subparsers)
     login_parser(subparsers)
     logout_parser(subparsers)
-    perplexity_parser(subparsers)
```
olliewalsh, line 337: all of the llama.cpp specific CLI parsers and handlers move to the llama.cpp plugin. cli.py just has the logic for container and model management

ramalama/cli.py
```diff
@@ -401,12 +384,6 @@ def map_https_to_transport(input: str) -> str:
     if hasattr(args, "runtime_args"):
         args.runtime_args = shlex.split(args.runtime_args)
 
-    # MLX runtime automatically requires --nocontainer
```
olliewalsh, line 404: ... similarly, mlx specific logic moves the to mlx plugin

ramalama/cli.py
```diff
@@ -1287,111 +967,12 @@ def _rag_args(args):
     # select a random port for the model
     args.port = None
     rag_args.model_port = args.port = compute_serving_port(args, exclude=[rag_args.port])
+    args.rag = None
     rag_args.model_args = args
     rag_args.generate = ""
     return rag_args
 
 
-def run_parser(subparsers):
```
olliewalsh, line 1296: the logic that is common to all inference runtimes moves to plugins/runtimes/inference/common.py - BaseInferenceRuntime or ContainerizedInferenceRuntime (which adds functionality for runtimes that support running in containers i.e not mlx)

pyproject.toml
```diff
@@ -100,6 +100,11 @@ Matrix = "https://matrix.to/#/#ramalama:fedoraproject.org"
 [project.scripts]
 ramalama = "ramalama.cli:main"
 
+[project.entry-points."ramalama.runtimes.v1alpha"]
```
olliewalsh, line 103: any python module can register these entry points, e.g we could package the vllm plugin as a separate optional python module or rpm.

A 3rd party could create a plugin but I've named the entrypoint group "v1alpha" to signal that it should not be considered a stable interface yet

ramalama/plugins/runtimes/inference/common.py
```diff
@@ -0,0 +1,284 @@
...
+        from ramalama.cli import local_models, runtime_options
+
+        parser = subparsers.add_parser("serve", help="serve REST API on specified AI Model")
+        runtime_options(parser, "serve")
+        self._add_inference_args(parser, "serve")
+        parser.add_argument("MODEL", completer=local_models)  # positional argument
+        parser.set_defaults(func=self._serve_handler)
+        return parser
+
+    def _do_run(self, args: argparse.Namespace, model: "Any") -> None:
```
olliewalsh, line 147: this exists to support rag. llama.cpp plugin logic overrides the method when args.rag is True

olliewalsh:
@bmahabirbu @rhatdan  PTAL

ramalama/daemon/handler/daemon.py
```diff
@@ -14,6 +13,7 @@
 from ramalama.daemon.logging import DEFAULT_LOG_DIR, logger
 from ramalama.daemon.service.model_runner import ManagedModel, ModelRunner, generate_model_id
 from ramalama.model_store.global_store import GlobalModelStore
+from ramalama.plugins.loader import assemble_command
```
engelmi, line 16: Not related to this PR, but I think the outcome of one discussion was that the concept of a "daemon" is something we do not want to use in RamaLama, so the whole directory can be removed (so less unnecessary maintenance)

pyproject.toml
engelmi: This should probably be added to the [ramalama.ai (docsite)](https://github.com/containers/ramalama/tree/main/docsite), incl. a how-to create a new runtime plugin. Can be in a follow-up, of course.

ramalama/plugins/runtimes/inference/common.py
```diff
@@ -0,0 +1,284 @@
...
+        )
+        parser.add_argument(
+            "--max-tokens",
+            dest="max_tokens",
+            type=int,
+            default=config.max_tokens,
+            help="maximum number of tokens to generate (0 = unlimited)",
+            completer=suppressCompleter,
+        )
+        if command in ("run", "serve"):
```
engelmi, line 76: Since this is explicit for these two commands and there are now dedicated register functions, I'd move these there. The same applies for the `--host` further below and in the containerized plugin class.

ramalama/transports/api.py
```diff
@@ -51,9 +51,6 @@ def _get_chat_template_path(self, use_container: bool, should_generate: bool, dr
     def remove(self, args):
         raise NotImplementedError("Hosted API transports do not support removing remote models.")
 
-    def bench(self, args, cmd: list[str]):
-        raise NotImplementedError("bench is not supported for hosted API transports.")
-
     def run(self, args, server_cmd: list[str]):
```
engelmi, line 54: Was `run` left here on purpose?

pyproject.toml
olliewalsh: Yes, was planning to do that

ramalama/daemon/handler/daemon.py
olliewalsh: :+1: some renewed interest in https://github.com/containers/ramalama/discussions/2282 so can follow up there

ramalama/transports/api.py
olliewalsh: yea, APITransport is an odd case where run == chat

olliewalsh: I have concerns about the APITransports in general, and I definitely had concerns about overloading the run command when reviewing the PR that added it. Probably best to leave this for a follow-up.

ramalama/plugins/runtimes/inference/common.py
olliewalsh: yes, that would change the CLI arg order though so would rather keep it for a follow-up (trying to keep the CLI interface and e2e tests unchanged in this PR)

engelmi: Ah ok, makes sense. This PR is already big enough :grin:

ramalama/transports/api.py
engelmi: Yeah, it evolved over time and became quite bloated with little separation of concerns (among others). Lets tackle it in a follow-up.

olliewalsh: It (i.e support for chat connecting to frontier models) was added quite recently, just openai initially, but there were follow-up PRs to add support for anthropic etc...
I'm not sure if allowing the chat client to connect to openai, anthropic etc... is really in scope for a project focused on local LLM use cases. Also would need to whip out my credit-card to test/develop/debug it! Definitely needs further discussion...

ieaves: The run == chat dynamic just reveals a degenerate case where serve is a noop (i.e. run = serve + chat) and would have required a larger refactor of `chat` to accommodate transport resolution at the model level in order to `ramalama chat openai://gpt-5.1` without passing in an explicit `--url https://api.openai.com/v1`. 

The hosted components could probably be refactored into a runtime plugin (first or third party) rather than a transport but you lose the ability to easily take advantage of shortnames.

ramalama/plugins/runtimes/inference/common.py
```diff
@@ -0,0 +1,284 @@
...
+        runtime_options(parser, "serve")
+        self._add_inference_args(parser, "serve")
+        parser.add_argument("MODEL", completer=local_models)  # positional argument
+        parser.set_defaults(func=self._serve_handler)
+        return parser
+
+    def _do_run(self, args: argparse.Namespace, model: "Any") -> None:
+        """Execute run after the model is resolved. Override to inject pre-run logic."""
+        from ramalama.common import ContainerEntryPoint
+        from ramalama.plugins.loader import assemble_command
```
ieaves: There are a few places with deferred imports from ramalama.plugins and I can't tell why. They don't seem to introduce a circular import and elsewhere plugins are imported at the top level so they don't defer any long running processes so far as I can tell.

ieaves:
This is a great extension, looks good to me.

rhatdan:
Needs a rebase, then we can merge.

ramalama/plugins/runtimes/inference/common.py
olliewalsh: :+1:  will clean this up

engelmi (approved):
LGTM

PR #2476, opened 2026-02-27
Add --chat-template-file to mount custom chat template (fixes #1783)

### LINKED ISSUE #1783: Allow passing additional config files (e.g. chat templates) into the model container

I was poking at ramalama today, and wound up trying to run the unsloth version of the new Qwen3 coder model. There's a [bug in it](https://huggingface.co/unsloth/Qwen3-Coder-30B-A3B-Instruct-GGUF/discussions/1), though. They sent out a tentative fix - an updated chat template file - with instructions to just run with `--chat-template-file <path/to/thatfile>`, no need to redownload the whole model!

So, OK, I try with `--runtime-args="--jinja --chat-template-file <path/to/thatfile>"`...and it complains about not being able to read the file. I assume this is because the model container doesn't have my home directory mounted into it or anything like that.

Is there any way to handle this kind of situation? I poked about in the docs but couldn't find anything. If I'm understanding it all right, it seems like it would be useful to have an option to mount some arbitrary file or directory into the model container for cases like this. If I'm not understanding it all right, please let me know. :D

_No response_

_No response_

### Additional context

_No response_

rhatdan:
Users can pass a host path to a chat template file; it is mounted into the container at /mnt/models/chat_template.file and passed to the inference engine. Enables using fixed or custom templates without redownloading the model.

Made-with: Cursor

Fixes: #1783 

## Summary by Sourcery

Add support for providing a host chat template file that is mounted into the container and passed through to the inference engine.

New Features:
- Accept a --chat-template-file CLI option for run, serve, and perplexity commands to supply a custom chat template from the host.
- Propagate the chat template file path through the runtime context so it is available to the inference engine configuration.
- Allow llama.cpp engine configs to prioritize a user-specified chat template file over the model’s default template when present.

Enhancements:
- Extend container mount setup to bind-mount a user-provided chat template file into a fixed in-container path for use by runtimes.

### DIFF
--- a/ramalama/transports/base.py
+++ b/ramalama/transports/base.py
@@ -28,6 +28,7 @@ if TYPE_CHECKING:
     from ramalama.transports.oci.oci import OCI
 
 from ramalama.common import (
+    MNT_CHAT_TEMPLATE_FILE,
     MNT_DIR,
     MNT_FILE_DRAFT,
     exec_cmd,
@@ -410,6 +411,15 @@ class Transport(TransportBase):
         self.engine.exec(stdout2null=args.noout)
         return True
 
+    def _add_chat_template_file_mount(self, args):
+        """Bind-mount --chat-template-file from host path into MNT_CHAT_TEMPLATE_FILE when set."""
+        chat_template_file = getattr(args, "chat_template_file", None)
+        if not chat_template_file or not os.path.isfile(chat_template_file):
+            return
+        # Host path (resolved and normalized for container engine, e.g. /c/Users/... on Windows)
+        src = get_container_mount_path(chat_template_file)
+        self.engine.add([f"--mount=type=bind,src={src},destination={MNT_CHAT_TEMPLATE_FILE},ro{self.engine.relabel()}"])
+
     def setup_mounts(self, args):
         if args.dryrun:
             return
@@ -428,6 +438,7 @@ class Transport(TransportBase):
             else:
                 raise NotImplementedError(f"No compatible oci mount method for engine: {self.engine.args.engine}")
             self.engine.add([mount_cmd])
+            self._add_chat_template_file_mount(args)
             return None
 
         ref_file = self.model_store.get_ref_file(self.model_tag)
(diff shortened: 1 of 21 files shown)


ramalama/transports/base.py
```diff
@@ -443,6 +444,12 @@ def setup_mounts(self, args):
             else:
                 raise NotImplementedError(f"No compatible oci mount method for engine: {self.engine.args.engine}")
             self.engine.add([mount_cmd])
+            chat_template_file = getattr(args, "chat_template_file", None)
+            if chat_template_file and os.path.isfile(chat_template_file):
+                container_chat_template = get_container_mount_path(chat_template_file)
+                self.engine.add(
+                    [f"--mount=type=bind,src={container_chat_template},destination={MNT_CHAT_TEMPLATE_FILE},ro{self.engine.relabel()}"]
```
olliewalsh: IIRC it's more complex than just bind mounting - have to consider oci images, kube/quadlet etc...

olliewalsh: Another option might be to support overriding the chat template in the `ramalama convert` cli - use it to generate an oci image/artifact with an overridden template? That would work in all cases I expect

rhatdan:
@AdamWill @olliewalsh let's either merge this or close it if it is no loinger necessary.

AdamWill:
oof, this is pretty far back. I haven't been actively using ramalama for this purpose lately. But I'll leave it up to you @rhatdan (hi dan! how's it going? missing you at devconf, I was telling some young'uns stories about you earlier) and @olliewalsh to decide what you want to do with it.

olliewalsh:
> oof, this is pretty far back. I haven't been actively using ramalama for this purpose lately. But I'll leave it up to you @rhatdan (hi dan! how's it going? missing you at devconf, I was telling some young'uns stories about you earlier) and @olliewalsh to decide what you want to do with it.

The upstream llama.cpp jinja/chat-template logic has been completely rewritten, most of the quirks in old model chat templates have been addressed. Unless there is a specific reason I would just drop this PR. We can resurrect it later if the need arises

PR #2498, opened 2026-03-07
ramalama sandbox: run an AI agent in a sandbox, backed by a local AI Model

mikebonnet:
The sandbox command runs an agent (currently Goose) in a container, configured to communicate with a local AI model via container networking. Instructions to the agent may be be passed as command-line arguments or via stdin. If neither are provided, an interactive session is started.

The agent has no access to the local filesystem by default. A directory can be mounted into the agent image with the `-w/--workdir` option.

On completion of the agent session, the model server will be removed.

## Summary by Sourcery

Add a sandbox command to run AI agents (currently Goose) in an isolated container wired to a local model server, with CLI, engine orchestration, and documentation support.

New Features:
- Introduce the ramalama sandbox CLI with a Goose subcommand to run an AI agent against a locally served model in containers.
- Add a Goose-specific sandbox engine that connects to the model server via container networking and supports interactive, stdin-driven, or argument-driven sessions.
- Allow mounting a host work directory into the Goose sandbox container for file access at /work.

Enhancements:
- Refactor runtime inference argument handling so llama.cpp-specific options can be reused by the sandbox serve flow and restrict certain flags to run/serve commands.
- Extend the generic run_cmd helper to support configurable stdin streams instead of always using /dev/null.
- Adjust TTY detection in the engine to be reusable across commands and support sandbox behavior.

Documentation:
- Add man pages for ramalama-sandbox and ramalama-sandbox-goose, including usage, options, and examples, and link sandbox into the main ramalama documentation index.

Tests:
- Add unit tests for the sandbox CLI and GooseEngine behavior, covering argument parsing, environment variables, networking, TTY modes, and workdir handling.
- Add end-to-end tests for sandbox dry-run output and live agent execution using different input modes, custom models, images, and workdir mounts.

Chores:
- Register the sandbox subcommand in the main CLI wiring and add shortnames for several Qwen3 GGUF models for convenient use with the sandbox.

### DIFF
--- a/ramalama/sandbox.py
+++ b/ramalama/sandbox.py
@@ -48,16 +48,10 @@ class SandboxEngineArgsType(BaseEngineArgsType):
     ARGS: list[str]
 
 
-class GooseEngineArgsType(SandboxEngineArgsType):
-    goose_image: str
-    workdir: str | None
-
-
 class SandboxEngine(Engine):
     """Engine for running sandbox containers."""
 
-    def __init__(self, args: SandboxEngineArgsType, *, model_name: str) -> None:
-        self.model_name = model_name
+    def __init__(self, args: SandboxEngineArgsType) -> None:
         super().__init__(args)
 
     def base_args(self) -> None:
@@ -82,54 +76,50 @@ class SandboxEngine(Engine):
         pass
 
 
-class GooseEngine(SandboxEngine):
-    """Goose-specific sandbox engine.
+class GooseArgsType(SandboxEngineArgsType):
+    goose_image: str
+    workdir: str | None
 
-    Extends SandboxEngine with environment variables required by Goose, and
-    workdir handling. If args are provided, they will be passed to Goose to
-    process non-interactively. If there are no arguments and stdin is a tty,
-    an interactive session will be started. Otherwise, instructions will be
-    read from stdin.
+
+class Goose:
+    """
+    Run Goose in a sandbox.
+    Environment variables required by Goose will be set, and any workdir specified will be mounted into the container.
+    If args are provided, they will be passed to Goose to process non-interactively. If there are no arguments and stdin
+    is a tty, an interactive session will be started. Otherwise, instructions will be read from stdin.
     """
 
-    def __init__(self, args: GooseEngineArgsType, *, model_name: str) -> None:
-        super().__init__(args, model_name=model_name)
-        self.add_name(f"goose-{self.args.name}")  # type: ignore[attr-defined]
-        self.add_workdir()
-        self.add_args(args.goose_image)
-        if args.ARGS:
-            self.add_args("run", "-t", " ".join(args.ARGS))
-        elif self.use_tty():
-            self.add_args("session")
-        else:
-            self.add_args("run", "-i", "-")
-
-    def add_env_options(self) -> None:
-        super().add_env_options()
-        self.add_env_option("GOOSE_PROVIDER=openai")
-        self.add_env_option(f"OPENAI_HOST=http://localhost:{self.args.port}")
-        self.add_env_option("OPENAI_API_KEY=ramalama")
-        self.add_env_option(f"GOOSE_MODEL={self.model_name}")
-        self.add_env_option("GOOSE_TELEMETRY_ENABLED=false")
-        self.add_env_option("GOOSE_CLI_SHOW_THINKING=true")
-
-    def add_privileged_options(self) -> None:
-        super().add_privileged_options()
-        # The goose image needs to run as uid 1000 so it can write state to the
-        # home directory in the image. Map uid 1000 to the local user (which is
-        # mapped to intermediate uid/gid 0) so files can be written to /work if
-        # required.
-        if self.use_podman:
+    def __init__(self, args: GooseArgsType, model_name: str) -> None:
+        self.engine = SandboxEngine(args)
+        if self.engine.use_podman:
             if platform.system() != "Windows":
-                self.add_args("--uidmap=+1000:0")
+                self.engine.add_args("--uidmap=+1000:0")
+        self.engine.add_name(f"goose-{args.name}")  # type: ignore[attr-defined]
+        self.add_env_options(args, model_name)
+        self.add_workdir(args)
+        self.engine.add_args(args.goose_image)
+        if args.ARGS:
+            self.engine.add_args("run", "-t", " ".join(args.ARGS))
+        elif self.engine.use_tty():
+            self.engine.add_args("session")
+        else:
+            self.engine.add_args("run", "-i", "-")
 
-    def add_workdir(self):
-        if self.args.workdir:
-            self.add_volume(self.args.workdir, "/work", opts="rw")
-            self.add_args("--workdir=/work")
+    def add_env_options(self, args: GooseArgsType, model_name: str) -> None:
+        self.engine.add_env_option("GOOSE_PROVIDER=openai")
+        self.engine.add_env_option(f"OPENAI_HOST=http://localhost:{args.port}")
+        self.engine.add_env_option("OPENAI_API_KEY=ramalama")
+        self.engine.add_env_option(f"GOOSE_MODEL={model_name}")
+        self.engine.add_env_option("GOOSE_TELEMETRY_ENABLED=false")
+        self.engine.add_env_option("GOOSE_CLI_SHOW_THINKING=true")
+
+    def add_workdir(self, args: GooseArgsType):
+        if args.workdir:
+            self.engine.add_volume(args.workdir, "/work", opts="rw")
+            self.engine.add_args("--workdir=/work")
 
     def run(self) -> None:
-        run_cmd(self.exec_args, stdout=None, stdin=None)
+        run_cmd(self.engine.exec_args, stdout=None, stdin=None)
 
 
 def run_sandbox(args):
@@ -145,13 +135,10 @@ def run_sandbox(args):
 
     model.serve_nonblocking(args, cmd)
 
-    goose = GooseEngine(
-        args,
-        model_name=model.model_alias,
-    )
+    goose = Goose(args, model.model_alias)
 
     if args.dryrun:
-        goose.dryrun()
+        goose.engine.dryrun()
         return
 
     try:
--- a/test/unit/test_sandbox_cmd.py
+++ b/test/unit/test_sandbox_cmd.py
@@ -3,13 +3,13 @@ from types import SimpleNamespace
 import pytest
 
 from ramalama.cli import parse_args_from_cmd, sandbox_cli
-from ramalama.sandbox import GooseEngine
+from ramalama.sandbox import Goose
 
 TEST_MODEL = "qwen3:4b"
 
 
 def _make_args(engine="podman"):
-    """Create minimal args for GooseEngine tests."""
+    """Create minimal args for Goose tests."""
     return SimpleNamespace(
         engine=engine,
         dryrun=False,
@@ -78,14 +78,11 @@ def test_goose_custom_image():
     assert args.goose_image == "myimage:v1"
 
 
-def test_goose_engine_env_vars():
-    """GooseEngine should set various env vars"""
+def test_goose_env_vars():
+    """Goose should set various env vars"""
     args = _make_args()
-    goose = GooseEngine(
-        args,
-        model_name="Qwen3-4B-Q4_K_M",
-    )
-    cmd = goose.exec_args
+    goose = Goose(args, "Qwen3-4B-Q4_K_M")
+    cmd = goose.engine.exec_args
     assert cmd[0] == "podman"
     assert "run" in cmd
     assert "--rm" in cmd
@@ -95,57 +92,42 @@ def test_goose_engine_env_vars():
     assert "GOOSE_MODEL=Qwen3-4B-Q4_K_M" in cmd
 
 
-def test_goose_engine_network():
-    """GooseEngine should setup container networking"""
+def test_goose_network():
+    """Goose should setup container networking"""
     args = _make_args()
-    goose = GooseEngine(
-        args,
-        model_name="test-model",
-    )
-    assert "--network=container:ramalama_model_abc" in goose.exec_args
+    goose = Goose(args, "test-model")
+    assert "--network=container:ramalama_model_abc" in goose.engine.exec_args
 
 
-def test_goose_engine_interactive():
-    """GooseEngine should set the -i option"""
+def test_goose_interactive():
+    """Goose should set the -i option"""
     args = _make_args()
-    goose = GooseEngine(
-        args,
-        model_name="test-model",
-    )
-    assert "-i" in goose.exec_args
+    goose = Goose(args, "test-model")
+    assert "-i" in goose.engine.exec_args
 
 
-def test_goose_engine_with_tty(monkeypatch):
-    """GooseEngine should run the session command when run with a tty"""
+def test_goose_with_tty(monkeypatch):
+    """Goose should run the session command when run with a tty"""
     monkeypatch.setattr("ramalama.engine.sys.stdin.isatty", lambda: True)
     args = _make_args()
-    goose = GooseEngine(
-        args,
-        model_name="test-model",
-    )
-    assert goose.exec_args[-1] == "session"
+    goose = Goose(args, "test-model")
+    assert goose.engine.exec_args[-1] == "session"
 
 
-def test_goose_engine_no_tty(monkeypatch):
-    """GooseEngine should run "run -i -" when run without a tty, to read commands from stdin"""
+def test_goose_no_tty(monkeypatch):
+    """Goose should run "run -i -" when run without a tty, to read commands from stdin"""
     monkeypatch.setattr("ramalama.engine.sys.stdin.isatty", lambda: False)
     args = _make_args()
-    goose = GooseEngine(
-        args,
-        model_name="test-model",
-    )
-    assert goose.exec_args[-3:] == ["run", "-i", "-"]
+    goose = Goose(args, "test-model")
+    assert goose.engine.exec_args[-3:] == ["run", "-i", "-"]
 
 
-def test_goose_engine_args():
-    """GooseEngine should run "run -t" when args are passed on the command-line"""
+def test_goose_args():
+    """Goose should run "run -t" when args are passed on the command-line"""
     args = _make_args()
     args.ARGS = ["hello", "ramalama"]
-    goose = GooseEngine(
-        args,
-        model_name="test-model",
-    )
-    assert goose.exec_args[-3:] == ["run", "-t", " ".join(args.ARGS)]
+    goose = Goose(args, "test-model")
+    assert goose.engine.exec_args[-3:] == ["run", "-t", " ".join(args.ARGS)]
 
 
 def test_sandbox_workdir_default_none():
@@ -163,26 +145,20 @@ def test_sandbox_workdir_option():
     assert args.workdir == "/tmp"
 
 
-def test_goose_engine_workdir():
-    """GooseEngine should add -v and --workdir=/work when workdir is set."""
+def test_goose_workdir():
+    """Goose should add -v and --workdir=/work when workdir is set."""
     args = _make_args()
     args.workdir = "/tmp/myproject"
-    goose = GooseEngine(
-        args,
-        model_name="test-model",
-    )
-    cmd = goose.exec_args
+    goose = Goose(args, "test-model")
+    cmd = goose.engine.exec_args
     assert "--workdir=/work" in cmd
     assert "/tmp/myproject:/work:rw" in cmd
 
 
-def test_goose_engine_no_workdir():
-    """GooseEngine should not add volume or --workdir when workdir is not set."""
+def test_goose_no_workdir():
+    """Goose should not add volume or --workdir when workdir is not set."""
     args = _make_args()
-    goose = GooseEngine(
-        args,
-        model_name="test-model",
-    )
-    cmd = goose.exec_args
+    goose = Goose(args, "test-model")
+    cmd = goose.engine.exec_args
     assert "--workdir=/work" not in cmd
     assert "-v" not in cmd


docs/ramalama-agent.1.md
```diff
@@ -0,0 +1,241 @@
...
+process non-interactively. Commands may also be passed via stdin.
+
+Two containers are started: a model server (llama-server) and the Goose
+agent. They communicate via container networking. When the Goose session
+exits, the model server container is automatically stopped and removed.
+
+## OPTIONS
+
+#### **--agent-image**=*image*
+Agent container image to use (default: ghcr.io/block/goose)
```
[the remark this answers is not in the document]
mikebonnet: Excluding the version tag so it doesn't have to be kept in sync every time the default value is updated.

docs/ramalama-agent.1.md
```diff
@@ -0,0 +1,241 @@
...
+#### **--help**, **-h**
+show this help message and exit
+
+#### **--host**="0.0.0.0"
+IP address for llama.cpp to listen on.
+
+#### **--image**=IMAGE
+OCI container image to run with specified AI model. RamaLama defaults to using
+images based on the accelerator it discovers. For example:
+`quay.io/ramalama/ramalama`. See the table above for all default images.
```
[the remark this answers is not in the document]
mikebonnet: Fixed.

docs/ramalama-agent.1.md
```diff
@@ -0,0 +1,241 @@
...
+
+#### **--temp**="0.8"
+Temperature of the response from the AI Model.
+llama.cpp explains this as:
+
+    The lower the number is, the more deterministic the response.
+
+    The higher the number is the more creative the response is, but more likely to hallucinate when set too high.
+
+	Usage: Lower numbers are good for virtual assistants where we need deterministic responses. Higher numbers are good for roleplay or creative tasks like editing stories
```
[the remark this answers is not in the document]
mikebonnet: Fixed.

docs/ramalama.conf.5.md
```diff
@@ -60,6 +60,14 @@ The ramalama table contains settings to configure and manage the OCI runtime.
 
 `[[ramalama]]`
 
+**agent_image** = "ghcr.io/block/goose:latest"
```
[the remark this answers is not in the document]
mikebonnet: Using `latest` instead of a specific version in the documentation to avoid having to update the docs every time the default is updated.

docs/ramalama.conf
```diff
@@ -17,6 +17,14 @@
 
 [ramalama]
 
+# Image used by "ramalama agent"
+#
+#agent_image = "ghcr.io/block/goose:latest"
```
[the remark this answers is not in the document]
mikebonnet: Using `latest` instead of a specific version in the documentation to avoid having to update the docs every time the default is updated.

ramalama/common.py
```diff
@@ -171,7 +172,7 @@ def run_cmd(
         env = os.environ | env
 
     result = subprocess.run(
-        args, check=True, cwd=cwd, stdout=sout, stderr=serr, stdin=subprocess.DEVNULL, encoding=encoding, env=env
+        args, check=True, cwd=cwd, stdout=sout, stderr=serr, stdin=stdin, encoding=encoding, env=env
     )
```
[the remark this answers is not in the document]
mikebonnet: Not an issue.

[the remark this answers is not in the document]
mikebonnet: @sourcery-ai dismiss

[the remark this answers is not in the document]
mikebonnet: @sourcery-ai dismiss

ramalama/config.py
```diff
@@ -155,6 +157,8 @@ def __post_init__(self):
 
 @dataclass
 class BaseConfig:
+    agent_image: str = DEFAULT_AGENT_IMAGE
```
olliewalsh: I would avoid placing this in the global BaseConfig as it is specific to a subcommand (I'm working on cleaning this up to remove the llama.cpp specific conf).

ramalama/engine.py
```diff
@@ -197,12 +197,15 @@ def add_privileged_options(self) -> None:
         else:
             super().add_privileged_options()
 
+    def is_tty_cmd(self) -> bool:
+        return getattr(self.args, "subcommand", "") == "run"
+
     def use_tty(self) -> bool:
         if not sys.stdin.isatty():
             return False
         if getattr(self.args, "ARGS", None):
```
olliewalsh: I expect this should move to the default is_tty_cmd method too -  ARGS is the prompt for the run command. If the agent also accepts a prompt then it should use the same condition in the overridden is_tty_cmd.

ramalama/cli.py
```diff
@@ -974,6 +975,51 @@ def _rag_args(args):
     return rag_args
 
 
+def agent_parser(subparsers):
+    config = get_config()
+    runtime = get_runtime(config.runtime)
+    parser = subparsers.add_parser("agent", help="run an AI agent backed by a local AI Model")
```
olliewalsh: maybe should be named something more generic? e.g sandbox. It could run any type of client, and the "agent" terminology may not persist.

ramalama/config.py
```diff
@@ -31,6 +31,8 @@
     "Q8_0",
 ]
 DEFAULT_GGUF_QUANTIZATION_MODE: GGUF_QUANTIZATION_MODES = "Q4_K_M"
+DEFAULT_AGENT_IMAGE: str = "ghcr.io/block/goose:1.27.1"
+DEFAULT_AGENT_MODEL: str = "hf://Qwen/Qwen3-1.7B-GGUF/Qwen3-1.7B-Q8_0.gguf"
```
olliewalsh: why have a default model when there is no default model for the run and serve commands?

ramalama/plugins/runtimes/inference/llama_cpp.py
```diff
@@ -160,8 +160,9 @@ def get_container_image(self, config: Any, gpu_type: str) -> str | None:
 
     # --- subcommand registration ---
 
-    def _add_llama_cpp_inference_args(self, parser: "argparse.ArgumentParser", command: str) -> None:
+    def _add_inference_args(self, parser: "argparse.ArgumentParser", command: str) -> None:
```
olliewalsh, line 197: Are these changes required for this PR?

ramalama/cli.py
olliewalsh: also could add an arg to specify which agent to run, assuming more will be added soon. E.g ramalama agent goose ....

ramalama/agent.py
```diff
@@ -0,0 +1,128 @@
...
+            self.add_args("run", "-i", "-")
+
+    def add_env_options(self) -> None:
+        super().add_env_options()
+        self.add_env_option("GOOSE_PROVIDER=openai")
+        self.add_env_option(f"OPENAI_HOST=http://localhost:{self.args.port}")
+        self.add_env_option("OPENAI_API_KEY=ramalama")
+        self.add_env_option(f"GOOSE_MODEL={self.model_name}")
+        self.add_env_option("GOOSE_TELEMETRY_ENABLED=false")
+        if getattr(self.args, "thinking", True):
```
olliewalsh: It this necessary? If thinking is False then there would be no thinking to show, so could just always set GOOSE_CLI_SHOW_THINKING

ramalama/agent.py
```diff
@@ -0,0 +1,128 @@
...
+    model.ensure_model_exists(args)
+
+    runtime = get_runtime(get_config().runtime)
+    cmd = runtime.handle_subcommand("serve", args)
+
+    model.serve_nonblocking(args, cmd)
+
+    goose = GooseEngine(
+        args,
+        model_name=model.model_name,
```
olliewalsh: this probably needs to be the model_alias to ensure it exactly matches the name the inference engine expects

ramalama/plugins/runtimes/inference/llama_cpp.py
mikebonnet: Without this, `agent_parser()` would also have to call `_add_llama_cpp_inference_args()` to setup the parser with the appropriate flags for running a `llama.cpp` inference server. And likewise would have to call runtime-specific methods for setting up the parser to support other runtime plugins. This approach seemed a lot cleaner. I also suggest possibly adding `add_inference_args()` to the runtime plugin interface for this reason.

ramalama/agent.py
mikebonnet: Yes, that could work. I thought it might be confusing if that env var was set even if the user passed `--thinking=off` on the command-line, but its simpler to always set it.

ramalama/agent.py
mikebonnet: This is working as-is in local testing and the e2e tests. I'll re-test with `model_alias`.

ramalama/cli.py
mikebonnet: I considered adding the arg, but it seemed unnecessary when only one agent is supported. I was planning to support future agents via a `--agent/--type` arg.

I'm open to a different terminology here, suggestions? `client` seems a little too generic.

ramalama/config.py
mikebonnet: I was thinking about the user interface. I really want `ramalama agent` to do something useful, without the user needing to know anything about model names, quantizations, etc.

ramalama/config.py
mikebonnet: The subcommand is not specific to a runtime plugin though, it could work with `vllm` just as well.

Where would this config go, if not the `BaseConfig`?

ramalama/agent.py
olliewalsh: Might need special handling for some runtimes too, e.g mlx does not expect a model name - not sure if it's possible to tell goose to omit it from the request

ramalama/engine.py
mikebonnet: It does use the same condition, `AgentEngine` is relying on the `Engine` checking `use_tty()` in the `add_tty_option()` method. `is_tty_cmd()` allows commands besides `run` to request a tty. If you think it would be better to just change the condition in `use_tty()` to `getattr(self.args, "subcommand", "") in ["run", "agent"]` I can do that.

ramalama/config.py
olliewalsh: Seems inconsistent with the current cli, and not likely to be a good choice in many cases e.g there are much better models for tool calling if you have the VRAM, also wouldn't work with vllm etc...
I think if we want to automatically choose/suggest a model to use it should be a feature in itself, and be based on the available hardware etc...

ramalama/engine.py
olliewalsh: I just mean the the ARGS handling should move from use_tty to is_tty_cmd in both cases because they are the only commands where ARGS is the prompt.

E.g the default here (for run)

```
def is_tty_cmd(self) -> bool:
  return getattr(self.args, "subcommand", "") == "run" and not getattr(self.args, "ARGS", None)
```

and overridden for agent:
```
def is_tty_cmd(self) -> bool:
  return getattr(self.args, "subcommand", "") == "agent" and not getattr(self.args, "ARGS", None)
```

olliewalsh: (mostly just a nit, since I don't think ARGS is currently used for any other command)

mikebonnet: Fixed.

ramalama/agent.py
mikebonnet: Fixed.

ramalama/agent.py
mikebonnet: Fixed to use `model_alias`.

ramalama/common.py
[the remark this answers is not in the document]
mikebonnet: @sourcery-ai dismiss

ramalama/config.py
mikebonnet: Updated to remove the default model and require the model as the first argument on the command-line, for consistency with `run` and `serve`.

ramalama/common.py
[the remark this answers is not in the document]
mikebonnet: @sourcery-ai dismiss

mikebonnet:
@sourcery-ai summary

ramalama/common.py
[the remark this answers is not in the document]
mikebonnet: @sourcery-ai dismiss

mikebonnet:
Ready for re-review

test/unit/test_agent_cmd.py
```diff
@@ -0,0 +1,156 @@
...
+        name="ramalama_model_abc",
+        port="8080",
+        thinking=False,
+        workdir=None,
+        subcommand="agent",
+        ARGS=[],
+    )
+
+
+def test_agent_model_positional():
```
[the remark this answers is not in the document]
mikebonnet: Unnecessary.

test/unit/test_sandbox_cmd.py
```diff
@@ -0,0 +1,156 @@
...
+    monkeypatch.setattr("ramalama.engine.sys.stdin.isatty", lambda: True)
+    args = _make_args()
+    goose = GooseEngine(
+        args,
+        model_name="test-model",
+    )
+    assert goose.exec_args[-1] == "session"
+
+
+def test_goose_engine_no_tty(monkeypatch):
```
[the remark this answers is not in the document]
mikebonnet: Unnecessary.

docs/ramalama-agent.1.md
```diff
@@ -0,0 +1,234 @@
...
+Add a host device to the container. Optional permissions parameter can
+be used to specify device permissions by combining r for read, w for
+write, and m for mknod(2).
+
+Example: --device=/dev/dri/renderD128:/dev/xvdc:rwm
+
+The device specification is passed directly to the underlying container engine. See documentation of the supported container engine for more information.
+
+Pass '--device=none' explicitly add no device to the container, eg for
+running a CPU-only performance comparison.
```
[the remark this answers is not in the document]
mikebonnet: Will fix in a future PR.

docs/ramalama-sandbox-goose.1.md
```diff
@@ -0,0 +1,234 @@
...
+
+#### **--name**, **-n**
+Name of the container to run the Model in.
+
+#### **--network**=*""*
+set the network mode for the container
+
+#### **--ngl**
+number of gpu layers, 0 means CPU inferencing, 999 means use max layers (default: -1)
+The default -1, means use whatever is automatically deemed appropriate (0 or 999)
```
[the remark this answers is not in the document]
mikebonnet: Unnecessary.

docs/ramalama-agent.1.md
```diff
@@ -0,0 +1,234 @@
...
+number of gpu layers, 0 means CPU inferencing, 999 means use max layers (default: -1)
+The default -1, means use whatever is automatically deemed appropriate (0 or 999)
+
+#### **--oci-runtime**
+
+Override the default OCI runtime used to launch the container. Container
+engines like Podman and Docker, have their own default oci runtime that they
+use. Using this option RamaLama will override these defaults.
+
+On Nvidia based GPU systems, RamaLama defaults to using the
```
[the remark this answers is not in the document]
mikebonnet: Unnecessary.

docs/ramalama-agent.1.md
```diff
@@ -0,0 +1,234 @@
...
+- **always**: Always pull the image and throw an error if the pull fails.
+- **missing**: Only pull the image when it does not exist in the local containers storage. Throw an error if no image is found and the pull fails.
+- **never**: Never pull the image but use the one from the local containers storage. Throw an error when no image is found.
+- **newer**: Pull if the image on the registry is newer than the one in the local containers storage. An image is considered to be newer when the digests are different. Comparing the time stamps is prone to errors. Pull errors are suppressed if a local image was found.
+
+#### **--runtime-args**="*args*"
+Add *args* to the runtime (llama.cpp or vllm) invocation.
+
+#### **--seed**=
+Specify seed rather than using random seed model interaction
```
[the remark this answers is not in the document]
mikebonnet: Unnecessary.

docs/ramalama-agent.1.md
```diff
@@ -0,0 +1,234 @@
...
+
+    The higher the number is the more creative the response is, but more likely to hallucinate when set too high.
+
+    Usage: Lower numbers are good for virtual assistants where we need deterministic responses. Higher numbers are good for roleplay or creative tasks like editing stories
+
+#### **--thinking**=*true*
+Enable or disable thinking mode in reasoning models
+
+#### **--threads**, **-t**
+Maximum number of cpu threads to use.
```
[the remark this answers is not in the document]
mikebonnet: Unnecessary.

ramalama/common.py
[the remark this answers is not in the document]
mikebonnet: @sourcery-ai dismiss

ramalama/cli.py
olliewalsh: Reason I mention having the agent name as an arg is that we will almost certainly support more in future, so better to address this now in the cli api design as it's pretty sticky.

e.g by accepting up to 2 positional args:
`ramalama agent goose qwen3:1.7b -- I am a prompt`

or by using the single positional arg for the agent name and switching the model to be an optional named param:
`ramalama agent goose --model qwen3:1.7b I am a prompt`

mikebonnet: Both those options feel like they're a bit unwieldy, and not consistent with cli conventions. Would it make more sense to make the command `ramalama goose <MODEL>`?

olliewalsh: > Both those options feel like they're a bit unwieldy, and not consistent with cli conventions. Would it make more sense to make the command `ramalama goose <MODEL>`?

I don't think that makes sense or is consistent. The cli syntax is broadly `ramalama verb subject`. goose is not a verb, it is the subject.
'agent' isn't a verb either though. `ramalama sandbox` is probably the best fit, even if docker already uses it.

If you were to create a new cli then sure, then e.g `ramalama-sandbox goose` would make sense in that case.

ramalama/config.py
olliewalsh: Would have to split the config. Can leave it as is for now, I'll try to cleanup all of the config later in a follow-up

ramalama/cli.py
mikebonnet: I don't think I want to create a whole new cli for this experiment. For now...`ramalama run-agent`? `ramalama launch-agent`? Or just leave as-is?

olliewalsh: ramalama sandbox? Docker doesn't own the term, we can use it if it makes sense

olliewalsh: NVIDIA openshell uses this too: `openshell sandbox create ...`

ramalama/common.py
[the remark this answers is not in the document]
mikebonnet: @sourcery-ai dismiss

ramalama/cli.py
mikebonnet: Agreed on `ramalama sandbox`.

ramalama/config.py
mikebonnet: No changes to config are required now.

ramalama/common.py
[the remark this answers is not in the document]
mikebonnet: @sourcery-ai dismiss

mikebonnet:
@sourcery-ai guide

ramalama/engine.py
```diff
@@ -197,12 +197,13 @@ def add_privileged_options(self) -> None:
         else:
             super().add_privileged_options()
 
+    def is_tty_cmd(self) -> bool:
+        return getattr(self.args, "subcommand", "") == "run" and not getattr(self.args, "ARGS", None)
```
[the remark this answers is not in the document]
mikebonnet: This behavior is intentional, `goose` only displays `thinking` output when connected to a tty. It is also desirable for a tty to be connected so `goose` can use colors, etc. in its output, even if commands have been passed via the cli.

docs/ramalama-sandbox-goose.1.md
```diff
@@ -0,0 +1,240 @@
...
+Add a host device to the container. Optional permissions parameter can
+be used to specify device permissions by combining r for read, w for
+write, and m for mknod(2).
+
+Example: --device=/dev/dri/renderD128:/dev/xvdc:rwm
+
+The device specification is passed directly to the underlying container engine. See documentation of the supported container engine for more information.
+
+Pass '--device=none' to explicitly add no device to the container, eg for
+running a CPU-only performance comparison.
```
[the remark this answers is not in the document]
mikebonnet: Unnecessary.

docs/ramalama-sandbox-goose.1.md
```diff
@@ -0,0 +1,240 @@
...
+
+#### **--env**=
+
+Set environment variables inside of the container.
+
+This option allows arbitrary environment variables that are available for the
+process to be launched inside of the container. If an environment variable is
+specified without a value, the container engine checks the host environment
+for a value and set the variable only if it is set on the host.
```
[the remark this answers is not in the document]
mikebonnet: Unnecessary.

docs/ramalama-sandbox-goose.1.md
```diff
@@ -0,0 +1,240 @@
...
+#### **--model-draft**
+
+A draft model is a smaller, faster model that helps accelerate the decoding
+process of larger, more complex models, like Large Language Models (LLMs). It
+works by generating candidate sequences of tokens that the larger model then
+verifies and refines. This approach, often referred to as speculative decoding,
+can significantly improve the speed of inferencing by reducing the number of
+times the larger model needs to be invoked.
+
+Use --runtime-arg to pass the other draft model related parameters.
```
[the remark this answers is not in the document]
mikebonnet: Unnecessary.

docs/ramalama-sandbox-goose.1.md
```diff
@@ -0,0 +1,240 @@
...
+
+#### **--name**, **-n**
+Name of the container to run the Model in.
+
+#### **--network**=*""*
+set the network mode for the container
+
+#### **--ngl**
+number of GPU layers, 0 means CPU inferencing, 999 means use max layers (default: -1)
+The default -1, means use whatever is automatically deemed appropriate (0 or 999)
```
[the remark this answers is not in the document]
mikebonnet: Unnecessary.

ramalama/common.py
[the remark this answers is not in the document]
mikebonnet: @sourcery-ai dismiss

docs/ramalama-sandbox-goose.1.md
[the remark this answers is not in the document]
mikebonnet: Fixed.

docs/ramalama-sandbox-goose.1.md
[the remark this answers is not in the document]
mikebonnet: Fixed.

docs/ramalama-sandbox-goose.1.md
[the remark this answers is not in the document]
mikebonnet: Fixed.

ramalama/sandbox.py
```diff
@@ -0,0 +1,130 @@
...
+        pass
+
+    def add_detach_option(self) -> None:
+        pass
+
+    def add_device_options(self) -> None:
+        pass
+
+
+class GooseEngine(SandboxEngine):
```
olliewalsh: Is inheritance the right approach here?
It's typically used for specialization ("is a" relationship). SandboxEngine looks like a specialization of an Engine, but goose looks like a consumer of a sandbox so composition seems like the correct approach to reuse code.

ramalama/plugins/runtimes/inference/llama_cpp.py
```diff
@@ -328,33 +328,6 @@ def register_subcommands(self, subparsers: "argparse._SubParsersAction") -> None
         perplexity_parser = subparsers.add_parser("perplexity", help="calculate perplexity for specified AI Model")
         runtime_options(perplexity_parser, "perplexity")
         self._add_inference_args(perplexity_parser, "perplexity")
-        # llama.cpp-specific perplexity args
```
olliewalsh, line 343: why were these removed? Are they duplicates? Were they wrong (the args were taken from the inferences specs so not necessarily correct)?

ramalama/cli.py
```diff
@@ -981,6 +982,51 @@ def _rag_args(args):
     return rag_args
 
 
+def sandbox_parser(subparsers):
```
olliewalsh, line 985: Could we keep the subcommand parsers in sandbox.py?
I've been trying to trim this module down to just the core container logic e.g by splitting out the inference runtimes.

olliewalsh:
> * The TTY detection logic for sandbox commands (`SandboxEngine.is_tty_cmd`) ignores the presence of `ARGS`, so `ramalama sandbox goose ... ARGS...` will still request a TTY when stdin is a TTY; if non-interactive runs with positional instructions should behave like `run` (no TTY), you may want to mirror the original `use_tty`/`ARGS` gating.

Is that accurate?

ramalama/plugins/runtimes/inference/llama_cpp.py
mikebonnet: They are duplicates. They're now added by the preceding call to `_add_inference_args()`.

mikebonnet:
> > * The TTY detection logic for sandbox commands (`SandboxEngine.is_tty_cmd`) ignores the presence of `ARGS`, so `ramalama sandbox goose ... ARGS...` will still request a TTY when stdin is a TTY; if non-interactive runs with positional instructions should behave like `run` (no TTY), you may want to mirror the original `use_tty`/`ARGS` gating.
> 
> Is that accurate?

Yes, that is the desired behavior. Consider the case where the user is running the command interactively from their console:

```
ramalama sandbox goose -w $PWD please analyze the code in the current directory
```

The user benefits by seeing console colors and thinking output (which is only displayed by `goose` if the console is a tty).

ramalama/sandbox.py
mikebonnet: Switched from `GooseEngine` -> `Goose` (standalone class) which constructs a `SandboxEngine` to launch `goose` in the container. Please take a look.

ramalama/cli.py
mikebonnet: It's a little bit awkward because it uses `runtime_options()`, `local_images()`, and `local_models()`, but I took a shot at it, let me know what you think.

ramalama/common.py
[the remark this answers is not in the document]
mikebonnet: @sourcery-ai dismiss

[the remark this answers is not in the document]
mikebonnet: @sourcery-ai dismiss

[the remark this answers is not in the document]
mikebonnet: @sourcery-ai dismiss

PR #2506, opened 2026-03-10
version is a subcommand

hulet:
See https://ramalama.ai/docs/commands/ramalama/version

## Summary by Sourcery

Update version check instructions to use the `version` subcommand instead of the `--version` flag across macOS docs and installer assets.

Build:
- Update macOS package build script messaging to reference `ramalama version` for installation verification.

Deployment:
- Adjust macOS installer conclusion screen to show `ramalama version` as the version verification command.

Documentation:
- Change macOS installation docs to instruct users to run `ramalama version` for version checks.

### DIFF
--- a/docs/MACOS_INSTALL.md
+++ b/docs/MACOS_INSTALL.md
@@ -40,7 +40,7 @@ The installer places files in:
 
 ```bash
 # Check version
-ramalama --version
+ramalama version
 
 # Get help
 ramalama --help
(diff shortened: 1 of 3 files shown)


engelmi:
Thanks for your contribution! @hulet 

I am wondering if we should support both, `--version` and `version`. WDYT? @olliewalsh 
Otherwise LGTM

olliewalsh:
> Thanks for your contribution! @hulet
> 
> I am wondering if we should support both, `--version` and `version`. WDYT? @olliewalsh Otherwise LGTM

Yeah, would be good if -v/--version worked too

mikebonnet:
I filed #2537 to support the new behavior. For now, this change looks appropriate, thanks for the submission!

PR #2518, opened 2026-03-15
update llama.cpp commit for the container images to b8352

mkristian:
just needed to run the new qwen-3.5 models which will not work otherwise

## Summary by Sourcery

Build:
- Bump the default llama.cpp commit hash in the container image build script from b8256 to b8352.


olliewalsh:
Closing this - maintainers manage bumping llama.cpp. Please raise an issue instead if there is a specific issue.

FWIW qwen3.5 should work in the current version (b8256) but we are waiting for some other fixes to land in llama.cpp before cutting a new ramalama release and publishing the new images

PR #2521, opened 2026-03-16
Prevent vLLM runtime from being used with run command

Samarthuday:
Fix crash when using `ramalama run --runtime vllm`.

The vLLM runtime currently only supports `serve`, but `ramalama run`
still attempts to launch it, which causes the server to exit unexpectedly.

This change prevents the run command from being used with the vLLM runtime
and provides a clear error message suggesting:

ramalama serve --runtime vllm

## Summary by Sourcery

Bug Fixes:
- Prevent crashes when invoking `ramalama run` with the vLLM runtime by rejecting this unsupported usage with a clear error message.

### DIFF
--- a/ramalama/plugins/runtimes/inference/common.py
+++ b/ramalama/plugins/runtimes/inference/common.py
@@ -171,11 +171,6 @@ class BaseInferenceRuntime(InferenceRuntimePlugin):
             raise e
 
     def _run_handler(self, args: argparse.Namespace) -> None:
-        # Prevent unsupported runtime mode
-        if getattr(args, "runtime", None) == "vllm":
-            raise ValueError(
-                "vLLM runtime does not support 'run'. Use `ramalama serve --runtime vllm <MODEL>`."
-            )
         try:
             # detect available port and update arguments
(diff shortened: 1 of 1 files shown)


olliewalsh:
run should work with vllm, please just raise an issue if it is not

Samarthuday:
> run should work with vllm, please just raise an issue if it is not

Thanks for the clarification!

I added the guard because the current behavior crashes when using `run` with the `vllm` runtime in my local tests. If `run` is expected to work with `vllm`, I'll remove the restriction and instead open an issue describing the crash scenario so it can be investigated properly.

Samarthuday:
> run should work with vllm, please just raise an issue if it is not

Thanks for the clarification. I’ve opened an issue describing the crash when running `ramalama run` with the `vllm` runtime:

#2523

mikebonnet:
There are now no code changes associated with this PR, can it be closed?

Samarthuday:
> There are now no code changes associated with this PR, can it be closed?

The behavior is expected (vLLM requires CUDA and isn’t supported on macOS yet), I’ll close this PR.

PR #2525, opened 2026-03-16
Support using the upstream llama.cpp container images

### LINKED ISSUE #2213: Add --entrypoint support to handle GGML images
### Issue Description

@briansmith0 was running into an issue with the version of llama.cpp in the Ramalama images and GPT-OSS 20b, which led us towards trying the GGML images (which worked under podman) under Ramalama for ease of use and cleaner commands. That led us to a seperate issue where the entrypoint of the GGML images conflicted with what Ramalama expected.

The issue is a double command injection:
1. The `ghcr.io/ggml-org/llama.cpp:server-cuda` image and others have an ENTRYPOINT set to the server binary (e.g., ENTRYPOINT ["/app/llama-cli"]).
2. Ramalama constructs the execution command by appending llama-server + arguments to the container run string.

The container attempts to run: `/app/llama-cli llama-server --host ...` which is not great.

Potential fixes could be to:
1. Always have the ramalama llama-server/vllm/binary execution within a container `--entrypoint` arg
2. Expose an `--entrypoint` argument that allows the user to pass a custom entrypoint via the CLI: `ramalama run --entrypoint ""`

Any thoughts on the proper way to go about this from the maintainers?

### Steps to reproduce the issue

```
ramalama --debug serve \
  --image ghcr.io/ggml-org/llama.cpp:server-cuda \
  file:///path/to/some/model.gguf
```

### Describe the results you received

The container fails immediately with error: invalid argument: llama-server.

### Describe the results you expected

A seamless swap of a rapid llama.cpp or other runtime build versus the default Ramalama image.

### ramalama info output

```yaml
N/A, affects all versions
```

### Upstream Latest Release

Yes

### Additional environment details

_No response_

### Additional information

This issue also affects the `ghcr.io/ggml-org/llama.cpp:full-cuda` image, though the error differs (Unknown command: llama-server) because the binary is not in the global $PATH for that specific image tag. 

Using --entrypoint logic should resolve both scenarios!

olliewalsh:
Detect when the llama.cpp image is the upstream ggml full image by inspecting the entrypoint and adjusting
the container cli as necessary.

Implementing this exposed a number of issues in the config handling logic and image pulling logic. Each has been addressed in a separate commit in this PR, so when reviewing it may be easier to look at the diff for each commit instead of the combined diff:

Assisted-by: Claude <noreply@anthropic.com>

Fixes #2213

## Summary by Sourcery

Support upstream llama.cpp container images and simplify image selection and pulling while centralizing configuration access.

New Features:
- Detect and adapt to upstream ggml llama.cpp "full" container images by inspecting the image entrypoint and adjusting CLI commands accordingly.

Bug Fixes:
- Fix configuration layering so later layers and runtime CLI overrides correctly take precedence and are tracked via is_set().
- Correct default container engine help output on macOS when no engine is available.

Enhancements:
- Refactor image tagging and pulling into reusable helpers, aligning default RamaLama and RAG images on version-based tags and improving fallback behavior when pulls fail.
- Centralize global configuration access through ActiveConfig/DefaultConfig and update callers across CLI, runtimes, transports, daemon, and utilities.
- Adjust inference runtimes to ensure container images are inspected or pulled on-demand based on pull policy, including RAG, llama.cpp, and vLLM flows.

Documentation:
- Update run and serve manpages to document support for upstream llama.cpp "full" images and automatic CLI adjustment.

Tests:
- Extend unit and e2e tests to cover ggml vs non-ggml llama.cpp images, new image selection behavior, updated config layering semantics, and revised error messages for missing images.

### DIFF
--- a/container-images/scripts/build_llama.sh
+++ b/container-images/scripts/build_llama.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-DEFAULT_LLAMA_CPP_COMMIT="f76565db92272d98976d5b8b1809ebe1e079f8cd" # b8256
-MESA_VULKAN_VERSION=25.2.3-101.fc43
+DEFAULT_LLAMA_CPP_COMMIT="a69d54f990d0cd88786d5943632d6426dc9660b7" # b8401
+MESA_VULKAN_VERSION=25.3.6-102.fc43
 
 dnf_install_intel_gpu() {
   local intel_rpms=("intel-oneapi-mkl-sycl-devel" "intel-oneapi-dnnl-devel"


ramalama/plugins/runtimes/inference/llama_cpp_commands.py
```diff
@@ -43,8 +43,7 @@ def _get_model_name(self, args: argparse.Namespace) -> str:
     def _cmd_run(self, args: argparse.Namespace) -> list[str]:
         if getattr(args, 'rag', None):
             return self._cmd_run_rag(args)
-
-        cmd = ["llama-server"]
+        cmd = ["llama-server"] if not self._container_image_is_ggml(args) else ["--server"]  # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
olliewalsh: still have the call the method in all places, no benefit

ramalama/common.py
```diff
@@ -676,46 +685,57 @@ def accel_image(config: Config, images: dict[str, str] | None = None, conf_key:
...
+    # This is a fallback for main when the version has been bumped but the
+    # tagged image for the version has not been pushed yet
+    # Only try this for quay.io/ramalama/ images
+    base, _, _ = image.rpartition(":")
+    if base and base.startswith("quay.io/ramalama/"):
+        latest = latest_tagged_image(base)
+        try:
+            run_cmd([conman, "pull", latest], ignore_stderr=True)
+            return latest
+        except Exception as e:
```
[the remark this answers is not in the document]
olliewalsh: AFAICT ignore_error is always false so I can just remove this conditional

ramalama/common.py
```diff
@@ -676,46 +685,53 @@ def accel_image(config: Config, images: dict[str, str] | None = None, conf_key:
...
-
-    # If the image from the config is specified by tag or digest, return it unmodified
-    if ":" in image:
-        return image
+    # Get image based on detected GPU type; tag user overrides with :latest if untagged
+    return latest_tagged_image(images.get(gpu_type, getattr(config, f"default_{conf_key}")))
 
-    vers = minor_release()
 
-    should_pull = config.pull in ["always", "missing"] and not config.dryrun
```
olliewalsh, line 691: this was never true, the default is pull=newer

ramalama/config.py
```diff
@@ -304,11 +304,20 @@ def load_env_config(env: Mapping[str, str] | None = None) -> dict[str, Any]:
...
-    return Config(load_env_config(env), load_file_config())
+def load_config(env: Mapping[str, str] | None = None) -> Config:
+    """Returns a Config object with layers initialized from config file and environment."""
+    return Config(load_file_config(), load_env_config(env))
 
 
-@lru_cache(maxsize=1)
-def get_config() -> Config:
-    return default_config()
+def ActiveConfig() -> Config:
```
mikebonnet, line 312: Camel-cased function names are inconsistent with the coding style of the rest of the codebase. Is there a reason for this choice?

ramalama/config.py
```diff
@@ -304,11 +304,20 @@ def load_env_config(env: Mapping[str, str] | None = None) -> dict[str, Any]:
     return config
 
 
-def default_config(env: Mapping[str, str] | None = None) -> Config:
-    """Returns a default Config object with all layers initialized."""
-    return Config(load_env_config(env), load_file_config())
+def load_config(env: Mapping[str, str] | None = None) -> Config:
```
mikebonnet, line 307: The diff would have been much smaller if the behavior of `default_config()` and `get_config()` had been changed, rather than introduce `DefaultConfig()` and `ActiveConfig()`. What's the rationale for the change?

ramalama/config.py
olliewalsh: it's a singleton - behaves like a class/object with only one instance so it uses a class-style name, the fact that it is a function is just how it's implemented

ramalama/config.py
olliewalsh: It's for clarity. default_config() did not return the "default" config, it returned the current active config.

get_config also behaved like a singleton by abusing the lru_cache with a limit of 1, just implementing a simple singleton pattern is much clearer

PR #2532, opened 2026-03-19
llama-cpp: add flash attention for benchmarks for openvino backend

kannon92:
**Problem**

I was trying out the newly released upstream images PR to do some benchmarking of the OpenVino backend on my laptop.

```
ramalama benchmark --runtime-args "-fa 1" --image ghcr.io/ggml-org/llama.cpp:full-openvino deepseek-r1:1.5b
```

Ref: https://github.com/ggml-org/llama.cpp/pull/15307#issuecomment-4056962095

Turns out that the OpenVino runtime requires flash attention. Its not that well documented.

**Solution**

Add `--runtime-args` support to the bench subcommand so users can pass additional arguments (like `-fa 1` for flash attention) directly to `llama-bench`.

**Usage**

```bash
# Enable flash attention for OpenVino benchmarking
ramalama benchmark --runtime-args "-fa 1" --image ghcr.io/ggml-org/llama.cpp:full-openvino deepseek-r1:1.5b

# Pass any additional llama-bench arguments
ramalama benchmark --runtime-args "<args>" <model>
```

**Note:** Flash attention is enabled by default for `llama-server` (run/serve), but not for `llama-bench`. The `--runtime-args` flag allows users to pass `-fa 1` when benchmarking with backends like OpenVino that require it.

**AI Disclosure**

Claude code was used


ramalama/plugins/runtimes/inference/llama_cpp_commands.py
```diff
@@ -201,6 +201,9 @@ def _cmd_bench(self, args: argparse.Namespace) -> list[str]:
         if threads is not None:
             cmd += ["--threads", str(threads)]
 
+        if getattr(args, 'flash_attn', False):
+            cmd += ["-fa", "1"]
```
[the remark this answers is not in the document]
kannon92: This is not true. https://github.com/ggml-org/llama.cpp/blob/master/tools/llama-bench/llama-bench.cpp#L444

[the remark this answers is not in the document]
kannon92: addressed

ramalama/plugins/runtimes/inference/llama_cpp.py
```diff
@@ -311,6 +311,13 @@ def register_subcommands(self, subparsers: "argparse._SubParsersAction") -> None
             ),
             completer=suppressCompleter,
         )
+        bench_parser.add_argument(
```
olliewalsh: I would just add the `--runtime-args` for bench (not sure why it's not there already).
Since we don't have an explicit arg for this in serve/run it wouldn't make sense to add it just for bench.

olliewalsh:
> Turns out that the OpenVino runtime requires flash attention. Its not that well documented.

If that's true then best to raise an issue in https://github.com/ggml-org/llama.cpp/issues to enable it by default in that backend

olliewalsh:
> > Turns out that the OpenVino runtime requires flash attention. Its not that well documented.
> 
> If that's true then best to raise an issue in https://github.com/ggml-org/llama.cpp/issues to enable it by default in that backend

I checked with the llama.cpp devs - flash attention should already be enabled by default for lama-server, it's just no enabled by default for benchmarks, which makes some sense I guess (although might be better for it to default to it on too).

ramalama/plugins/runtimes/inference/llama_cpp.py
olliewalsh: Can address in a follow-up PR if you don't want to do this here...

olliewalsh: Since fa is default for serve, but not for bench, it might make sense to add this after all, but I would default to True so the results are more representative of the expected performance with serve.

kannon92:
@olliewalsh can you approve the CI?

ramalama/plugins/runtimes/inference/llama_cpp.py
kannon92: I can make this default to true.

kannon92: When you say "default to True" — do you mean we should just always enable flash attention for bench (hardcode it), or should we default the flag to True but allow users to disable it with `--no-flash-attn`? With `store_true` + `default=True` the flag becomes a no-op since there's no way to turn it off.

olliewalsh: Default the arg to true. Need to change it to store.
This would be consistent with llama.cpp too as requires a value now (was originally just a flag IIRC).

Might just be easier and more useful in-general to add `--runtime-args` for bench, and just document this gotcha for openvino.

olliewalsh:
> @olliewalsh can you approve the CI?

done, can run the github action CI on your fork too if you just enable actions for the fork

kannon92:
Addressed @olliewalsh's feedback — switched from a dedicated `--flash-attn` flag to adding `--runtime-args` support for the bench subcommand (commit 657086bc). This is more flexible and consistent with how run/serve handle extra args.

Updated the PR description with usage examples:

```bash
# Enable flash attention for OpenVino benchmarking
ramalama benchmark --runtime-args "-fa 1" --image ghcr.io/ggml-org/llama.cpp:full-openvino deepseek-r1:1.5b
```

Re: the bot comments:
- **sourcery-ai** (`getattr` vs direct access): Moot now since the `--flash-attn` arg was removed in favor of `--runtime-args`.
- **gemini-code-assist** (`-fa` doesn't take a value): Incorrect — `llama-bench` does accept `-fa` with a value (`0` or `1`) as shown in the [llama-bench source](https://github.com/ggml-org/llama.cpp/blob/master/tools/llama-bench/llama-bench.cpp#L444).

ramalama/plugins/runtimes/inference/llama_cpp.py
kannon92: Went ahead and added `--runtime-args`.

Not sure where you want the documentation though.

kannon92:
> > @olliewalsh can you approve the CI?
> 
> done, can run the github action CI on your fork too if you just enable actions for the fork

I'm not actually sure how to do that. I tried to enable actions on my fork but not sure if it runs.

olliewalsh:
> > > @olliewalsh can you approve the CI?
> > 
> > 
> > done, can run the github action CI on your fork too if you just enable actions for the fork
> 
> I'm not actually sure how to do that. I tried to enable actions on my fork but not sure if it runs.

Should be triggered when you push to a branch

ramalama/plugins/runtimes/inference/llama_cpp.py
```diff
@@ -295,6 +295,7 @@ def register_subcommands(self, subparsers: "argparse._SubParsersAction") -> None
         # bench / benchmark
         bench_parser = subparsers.add_parser("bench", aliases=["benchmark"], help="benchmark specified AI Model")
         runtime_options(bench_parser, "bench")
+        self._add_inference_args(bench_parser, "bench")
```
olliewalsh: Do these args apply to bench?

kannon92: I'm not sure I really follow you.

The parameters ngl and threads are valid options for bench.

rhatdan:
Looks like this needs a rebase.

docs/ramalama-bench.1.md
```diff
@@ -30,6 +30,7 @@ URL support means if a model is on a web site or even on your local system, you
 #### **--authfile**=*password*
 path of the authentication file for OCI registries
 
+<<<<<<< HEAD
```
mikebonnet: Looks like these rebase markers need to be cleaned up.

docs/ramalama-bench.1.md
```diff
@@ -71,6 +72,13 @@ ramalama bench --backend vulkan granite
 ramalama bench --backend vulkan granite
 ramalama bench --backend rocm granite
 ```
+=======
+#### **--cache-reuse**=256
+Min chunk size to attempt reusing from the cache via KV shifting
+
+#### **--ctx-size**, **-c**
+size of the prompt context. This option is also available as **--max-model-len**. Applies to llama.cpp and vllm regardless of alias (default: 0, 0 = loaded from model)
+>>>>>>> e0390d90 (llama-cpp: Use --runtime-args for bench subcommand)
```
mikebonnet: Another rebase marker that needs to be cleaned up.

kannon92: done.

docs/ramalama-bench.1.md
kannon92: done.

ramalama/plugins/runtimes/inference/llama_cpp.py
mikebonnet: `_add_inference_args()` unconditionally adds a `--cache-reuse` option, which is not supported by `llama-bench`:
```
$ podman run -it --rm quay.io/ramalama/ramalama:0.18 llama-bench --help 2>&1 | grep cache
  -ctk, --cache-type-k <t>                    (default: f16)
  -ctv, --cache-type-v <t>                    (default: f16)
```
That should be conditionalized the same way the `--logfile` option is handled.

mikebonnet: I also don't see `--ctx-size`, `--max-tokens`, or `--temp`:
```
$ podman run -it --rm quay.io/ramalama/ramalama:0.18 llama-bench --help 2>&1 | grep -Ec 'ctx|max|temp'
0
```

mikebonnet: It would be simpler to remove the call to `_add_inference_args()` and add `--runtime-args` directly to `bench_parser`.

test/e2e/test_bench.py
```diff
@@ -12,3 +12,11 @@ def test_model_and_params_columns(test_model):
     result = check_output(["ramalama", "bench", "-t", "2", test_model])
 
     assert re.search(r"\|\s+model\s+\|\s+params", result)
+
+
+@pytest.mark.e2e
+@skip_if_no_llama_bench
```
mikebonnet: This could be `skip_in_no_container` instead. If using containers, `llama-bench` doesn't need to be available on the host.

test/e2e/test_bench.py
```diff
@@ -12,3 +12,11 @@ def test_model_and_params_columns(test_model):
...
+def test_bench_runtime_args(test_model):
+    result = check_output(["ramalama", "bench", "-t", "2", "--runtime-args", "-fa 1", test_model])
```
mikebonnet: This could be a `--dryrun` test.

test/e2e/test_bench.py
mikebonnet: If testing with `--dryrun` (recommended) then the skip can be removed entirely.

test/e2e/test_bench.py
```diff
@@ -12,3 +12,11 @@ def test_model_and_params_columns(test_model):
...
+
+    assert re.search(r"\|\s+model\s+\|\s+params", result)
```
mikebonnet: If this is converted to a `--dryrun` test, then you should be checking for `-fa 1` in the result.

test/e2e/test_bench.py
```diff
@@ -2,13 +2,18 @@
 
 import pytest
 
-from test.conftest import skip_if_no_llama_bench
 from test.e2e.utils import check_output
 
 
 @pytest.mark.e2e
-@skip_if_no_llama_bench
```
mikebonnet, line 10: Please revert the change to this test, it's unrelated to the issue being fixed.

ramalama/plugins/runtimes/inference/llama_cpp.py
kannon92: updated.

test/e2e/test_bench.py
kannon92: thanks for that. updated.

test/e2e/test_bench.py
kannon92: updated.

test/e2e/test_bench.py
kannon92: updated.

PR #2563, opened 2026-03-27
feat: add multi-model router mode for ramalama serve

bmahabirbu:
## Summary
- When `ramalama serve` is invoked without a MODEL argument, it starts llama.cpp in router mode
- Automatically discovers and mounts all locally stored GGUF models into a container, serving them via llama.cpp's `--models-dir` feature
- Adds `--models-max` flag to control how many models can be loaded concurrently (default: 4)
- Router mode requires a container runtime; raises `NotImplementedError` with `--nocontainer`

## Details
This enables users to serve multiple models through a single endpoint with automatic model swapping, leveraging llama.cpp's built-in router/model-directory support. The server will load models on demand as requests come in, up to the configured `--models-max` limit.

### Files changed
- `ramalama/plugins/runtimes/inference/common.py` — Made MODEL optional in serve registration, added `_serve_router()` handler and `_enumerate_store_gguf_models()` helper
- `ramalama/plugins/runtimes/inference/llama_cpp_commands.py` — Updated `_cmd_run` to emit `--models-dir`/`--models-max` instead of single-model flags when in router mode
- `ramalama/plugins/runtimes/inference/llama_cpp.py` — Added `--models-max` CLI argument, updated `service_ready_check` for router mode

## Test plan
- [ ] `ramalama serve` with no model argument (container mode) starts llama.cpp with `--models-dir`
- [ ] `ramalama serve` with no model and `--nocontainer` raises a clear error
- [ ] `ramalama serve <model>` continues to work as before (single model mode)
- [ ] `ramalama serve --models-max 2` limits concurrent models
- [ ] `ramalama serve --dryrun` shows the full container command with model mounts

## Summary by Sourcery

Enable container-only multi-model router mode for `ramalama serve` using llama.cpp’s models directory support.

New Features:
- Allow `ramalama serve` to run in router mode when invoked without a MODEL argument, serving multiple GGUF models from the local model store through a single endpoint.
- Add a `--models-max` CLI option to limit the number of models that can be loaded concurrently in router mode.

Enhancements:
- Automatically discover GGUF models in the global model store and mount them into the inference container for llama.cpp router mode.
- Update llama.cpp command construction to use `--models-dir`/`--models-max` in router mode instead of single-model flags and to skip per-model GPU/draft settings when routing.
- Treat router-mode servers as ready as soon as any model is available, adjusting the readiness check accordingly.
- Disallow router mode without a container runtime by raising a clear `NotImplementedError` when `ramalama serve` is run without a model and `--nocontainer` is set.

### DIFF
--- a/docs/ramalama-serve.1.md
+++ b/docs/ramalama-serve.1.md
@@ -4,11 +4,17 @@
 ramalama\-serve - serve REST API on specified AI Model
 
 ## SYNOPSIS
-**ramalama serve** [*options*] _model_
+**ramalama serve** [*options*] [_model_]
 
 ## DESCRIPTION
 Serve specified AI Model as a chat bot. RamaLama pulls specified AI Model from
 registry if it does not exist in local storage.
+
+When invoked without a model argument, starts in **router mode**: all locally
+stored GGUF models are mounted into a container and served via llama.cpp's
+multi-model directory feature. Requests are routed to the appropriate model
+based on the model name in the API request. Router mode requires a container
+runtime.
 
 ## MODEL TRANSPORTS
 
@@ -211,6 +217,10 @@ times the larger model needs to be invoked.
 
 Use --runtime-arg to pass the other draft model related parameters.
 Make sure the sampling parameters like top_k on the web UI are set correctly.
+
+#### **--models-max**=*integer*
+Maximum number of models to load concurrently in router mode (default: 4).
+Only used when `ramalama serve` is invoked without a model argument.
 
 #### **--name**, **-n**
 Name of the container to run the Model in.
--- a/ramalama/plugins/runtimes/inference/common.py
+++ b/ramalama/plugins/runtimes/inference/common.py
@@ -1,4 +1,6 @@
 import argparse
+import os
+import sys
 from abc import abstractmethod
 from typing import Any
 
@@ -12,9 +14,21 @@ from ramalama.cli import (
     runtime_options,
     suppressCompleter,
 )
-from ramalama.common import ContainerEntryPoint, accel_image, ensure_image, set_accel_env_vars
+from ramalama.common import (
+    ContainerEntryPoint,
+    accel_image,
+    ensure_image,
+    genname,
+    sanitize_filename,
+    set_accel_env_vars,
+)
 from ramalama.config import ActiveConfig
+from ramalama.engine import Engine
 from ramalama.logger import logger
+from ramalama.model_store.constants import DIRECTORY_NAME_BLOBS, DIRECTORY_NAME_REFS, DIRECTORY_NAME_SNAPSHOTS
+from ramalama.model_store.global_store import GlobalModelStore
+from ramalama.model_store.reffile import RefJSONFile, StoreFileType, migrate_reffile_to_refjsonfile
+from ramalama.path_utils import get_container_mount_path
 from ramalama.plugins.interface import InferenceRuntimePlugin
 from ramalama.plugins.loader import assemble_command
 from ramalama.stack import Stack
@@ -140,7 +154,7 @@ class BaseInferenceRuntime(InferenceRuntimePlugin):
         parser = subparsers.add_parser("serve", help="serve REST API on specified AI Model")
         runtime_options(parser, "serve")
         self._add_inference_args(parser, "serve")
-        parser.add_argument("MODEL", completer=local_models)  # positional argument
+        parser.add_argument("MODEL", nargs="?", default=None, completer=local_models)  # positional argument
         parser.set_defaults(func=self._serve_handler)
         return parser
 
@@ -202,6 +216,9 @@ class BaseInferenceRuntime(InferenceRuntimePlugin):
         if not args.container:
             args.detach = False
 
+        if args.MODEL is None:
+            return self._serve_router(args)
+
         if getattr(args, "api", None) == "llama-stack":
             if not args.container:
                 raise ValueError(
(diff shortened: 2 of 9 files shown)


bmahabirbu:
So one thing i noticed if we default to context size as normal the first model loads with the max vram available and the next model will be forced to run on cpu. So do you think we should add a secret --context = 2048 so that doesnt happen since this is an experimental feature?? 

What i did notice is if you correctly unload the first model before using the second llama.cpp will correctly load the new model with gpu

mikebonnet:
Very cool! Could you add unit/e2e tests for this feature?

docs/ramalama-serve.1.md
```diff
@@ -4,12 +4,18 @@
 ramalama\-serve - serve REST API on specified AI Model
 
 ## SYNOPSIS
-**ramalama serve** [*options*] _model_
+**ramalama serve** [*options*] [_model_]
```
rhatdan: Would it make sense to move to 0 or more models specified?

`**ramalama serve** [*options*] [_model_ ...]`

This would allow me t launch a server with only a couple of the models I downloaded to be available within the container instead of all of them?

ramalama/plugins/runtimes/inference/common.py
```diff
@@ -235,6 +252,88 @@ def _serve_handler(self, args: argparse.Namespace) -> None:
...
+
+        engine.add([args.image] + cmd)
+
+        if args.dryrun:
+            engine.dryrun()
+            return
+        engine.exec()
+
+
+def _enumerate_store_gguf_models(store, refs_dir_name, snapshots_dir_name, blobs_dir_name, RefJSONFile, migrate_fn):
```
mikebonnet: This kind of feels like it should be a method of `GlobalModelStore`. It shares a decent amount of code with `list_models()`, maybe there's an opportunity for future consolidation there?

mikebonnet (approved):
One minor comment, but otherwise LGTM!

docs/ramalama-serve.1.md
mikebonnet: This could be a future enhancement, but I don't think it needs to block the current implementation.

bmahabirbu: This makes sense to me! If possible I can get this in and make a small pr to add that!

ramalama/plugins/runtimes/inference/common.py
bmahabirbu: Nice catch I can see if we can consolidate

rhatdan:
@bmahabirbu this was ready for merge, I believe but now needs a rebase.

ramalama/plugins/runtimes/inference/common.py
bmahabirbu: coming back to this! I think once this gets in i can consolidate after

bmahabirbu:
@olliewalsh ready-to-test

ramalama/plugins/runtimes/inference/common.py
```diff
@@ -238,6 +247,51 @@ def _serve_handler(self, args: argparse.Namespace) -> None:
         self._do_serve(args, model)
 
 
+def _enumerate_store_gguf_models(
+    store: Any,
+    refs_dir_name: str,
+    snapshots_dir_name: str,
+    blobs_dir_name: str,
+    ref_json_cls: Any,
+    migrate_fn: Callable[[str, str], Any],
```
olliewalsh: This seems to be doing too much, an enumeration function should not have a side-effect.

ramalama/plugins/runtimes/inference/common.py
```diff
@@ -238,6 +247,51 @@ def _serve_handler(self, args: argparse.Namespace) -> None:
         self._do_serve(args, model)
 
 
+def _enumerate_store_gguf_models(
```
olliewalsh: Why does this have a leading underscore? The suggests it's private to the module.

ramalama/plugins/runtimes/inference/llama_cpp.py
```diff
@@ -501,7 +602,20 @@ def _register_run_subcommand(self, subparsers: "argparse._SubParsersAction") ->
 
     def _register_serve_subcommand(self, subparsers: "argparse._SubParsersAction") -> "argparse.ArgumentParser":
         parser = super()._register_serve_subcommand(subparsers)
+        for action in parser._actions:
+            if hasattr(action, 'dest') and action.dest == 'MODEL':
```
olliewalsh: This looks a bit evil. Could you remove it from the base parser and add it in the subclasses?

olliewalsh: Or pull it out into a method that can be overridden by this class

ramalama/plugins/runtimes/inference/common.py
bmahabirbu: makes sense will update!

ramalama/plugins/runtimes/inference/common.py
bmahabirbu: gotcha! updated

ramalama/plugins/runtimes/inference/llama_cpp.py
bmahabirbu: updated!

bmahabirbu:
To test

```
ramalama serve model1 model2

or

ramalama serve

(which serves all models)

```
to list all models

`curl http://localhost:8080/v1/models`

To switch a model without needing the ui or specifically calling the serve load api u can just switch the name

```
curl http://localhost:8080/v1/chat/completions \
    -H "Content-Type: application/json" \
    -d '{
      "model": "model-name",
      "messages": [{"role": "user", "content": "What is the capital of France?"}]
    }'
```

then switch to a different name

```
curl http://localhost:8080/v1/chat/completions \
    -H "Content-Type: application/json" \
    -d '{
      "model": "model-name",
      "messages": [{"role": "user", "content": "What is the capital of France?"}]
    }'
```

the names will be from the curl models

bmahabirbu:
@olliewalsh ready-to-test again/review

olliewalsh:
LGTM but one of the e2e tests failing in konflux:

```FAILED test/e2e/test_sandbox.py::test_sandbox_run_stdin[opencode] - AssertionError: assert '42' in '8bb486f0cebcf4c697d7c5007321549f69da7cccaf700ac68b6f03c95e328364\n'```

Re-running the tests...

olliewalsh:
> LGTM but one of the e2e tests failing in konflux:
> 
> `FAILED test/e2e/test_sandbox.py::test_sandbox_run_stdin[opencode] - AssertionError: assert '42' in '8bb486f0cebcf4c697d7c5007321549f69da7cccaf700ac68b6f03c95e328364\n'`
> 
> Re-running the tests...

@bmahabirbu might be a genuine issue?

bmahabirbu:
Thanks ill give it a look!

bmahabirbu:
@olliewalsh I ran the test locally and it passed it seems to be a flaky llm response

PR #2567, opened 2026-03-27
fix: Always retain explicit RAMALAMA_CONTAINER_ENGINE=podman on MacOS

### LINKED ISSUE #2568: config: explicit RAMALAMA_CONTAINER_ENGINE=podman is discarded on macOS when _finalize_engine() falls back
### Issue Description

On macOS, an explicit `RAMALAMA_CONTAINER_ENGINE=podman` can be incorrectly treated like an implicit default during config finalization.

The special macOS handling in `Config._finalize_engine()` is intended to fall back from Podman only when the engine was not explicitly set. But during initialization, `self.is_set("engine")` is evaluated too early, so an explicitly configured engine may be treated as unset.

That can rewrite:

- `engine = "podman"` -> `engine = None` (or `"docker"` if available)
- and then `container` becomes `False` because it is derived from `engine is not None`

### Failing test / log snippet

This was reproduced by adding a macOS unit-test CI leg. The failure is:

```text
__________________________ test_cfg_container_not_set __________________________
test/unit/test_config.py:161: in test_cfg_container_not_set
    assert cfg.container is True
E   AssertionError: assert False is True
E    +  where False = Config(api='none', api_key=None, benchmarks=Benchmarks(storage_folder='/Users/runner/.local/share/ramalama/benchmarks', disable=False), carimage='registry.access.redhat.com/ubi10-micro:latest', container=False, ctx_size=0, convert_type='raw', default_image='quay.io/ramalama/ramalama:0.18', default_rag_image='quay.io/ramalama/ramalama-rag:0.18', dryrun=False, engine=None, env=[], gguf_quantization_mode='Q4_K_M', host='0.0.0.0', http_client=HTTPClientConfig(max_retries=5, max_retry_delay=30), image='quay.io/ramalama/ramalama:0.18', images={}, rag_image=None, rag_images={}, keep_groups=False, log_level=None, max_tokens=0, ocr=False, port='8080', prefix=None, pull='newer', rag_format='qdrant', runtime='llama.cpp', selinux=False, settings=RamalamaSettings(config_files=None), stack_image='quay.io/ramalama/llama-stack:0.18', store='/Users/runner/.local/share/ramalama', summarize_after=4, temp='0.8', transport='ollama', user=UserConfig(no_missing_gpu_prompt=True), verify=True, provider=
============================= slowest 10 durations =============================
```

Relevant test:

```python
def test_cfg_container_not_set():
    with patch.dict(os.environ, {"RAMALAMA_CONTAINER_ENGINE": "podman"}):
        cfg = load_config()
        assert cfg.is_set("container") is False
        assert cfg.container is True
```

### Why this happens

The macOS-specific code in `Config._finalize_engine()` does:

```python
if self.engine is not None and os.path.basename(self.engine) == "podman" and sys.platform == "darwin":
    run_with_podman_engine = apple_vm(self.engine, self)
    if not run_with_podman_engine and not self.is_set("engine"):
        self.engine = "docker" if available("docker") else None
```

`RAMALAMA_CONTAINER_ENGINE=podman` is loaded into `engine`, but during initialization `self.is_set("engine")` is not yet reflecting that explicit layer correctly, so `_finalize_engine()` can wrongly enter the fallback branch.

After that, `BaseConfig.__post_init__()` computes:

```python
self.container = coerce_to_bool(self.container) if self.container is not None else self.engine is not None
```

So once `engine` has been rewritten to `None`, `container` becomes `False`.

### Expected behavior

On macOS, if `RAMALAMA_CONTAINER_ENGINE=podman` is explicitly set, `_finalize_engine()` should preserve that explicit choice instead of treating it as an implicit default and falling back.

### Steps to reproduce

One way to reproduce is via the existing unit test on macOS:

```bash
pytest test/unit/test_config.py -k test_cfg_container_not_set -vv
```

The failure shows up on macOS CI where the Podman fallback path is exercised.

### Additional context

This seems specific to the macOS `_finalize_engine()` path and the timing of `is_set("engine")` during layered config initialization.

booxter:
Fixes #2568

This PR also adds a MacOS unit test job that [would](https://github.com/booxter/ramalama/actions/runs/23667747532/job/68953859160) expose the issue.

## Summary by Sourcery

Add a macOS variant to the unit test job matrix and parameterize the job runner per matrix variant in the CI workflow.

CI:
- Extend the unit test job matrix to include a macOS run alongside existing Ubuntu runs.
- Configure the unit test job to select its GitHub Actions runner from the matrix instead of a hard-coded Ubuntu runner.

## Summary by Sourcery

Ensure explicit container engine configuration is preserved on macOS and extend CI coverage to macOS for unit tests.

Bug Fixes:
- Preserve an explicitly configured podman engine on macOS even when podman machine detection fails.

Enhancements:
- Adjust layered configuration initialization so that explicit attribute writes are tracked only after configuration layers are finalized, keeping is_set() behavior consistent.

CI:
- Extend the unit test matrix to run on both Ubuntu and macOS runners and parameterize the runner and setup commands per matrix variant.

### DIFF
--- a/test/e2e/test_help.py
+++ b/test/e2e/test_help.py
@@ -57,7 +57,7 @@ def test_help_command_flags():
     # Test for regression of #7273 (spurious "--remote" help on output)
     for help_opt in ["help", "-h", "--help"]:
         result = check_output(["ramalama", help_opt])
-        assert re.search(r"^usage: ramalama \[-h] \[--debug.*] \[--dryrun] \[--engine {podman,docker}]", result)
+        assert re.search(r"^usage: ramalama \[-h] \[-v] \[--debug.*] \[--dryrun] \[--engine {podman,docker}]", result)
 
 
 @pytest.mark.e2e
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -26,7 +26,7 @@ jobs:
       artifact-metadata: write
     steps:
       - name: Harden Runner
-        uses: step-security/harden-runner@fa2e9d605c4eeb9fcad4c99c224cee0c6c7f3594 # v2.16.0
+        uses: step-security/harden-runner@fe104658747b27e96e4f7e80cd0a94068e53901d # v2.16.1
         with:
           egress-policy: audit
 
@@ -65,7 +65,7 @@ jobs:
     needs: build-package
     steps:
       - name: Harden Runner
-        uses: step-security/harden-runner@fa2e9d605c4eeb9fcad4c99c224cee0c6c7f3594 # v2.16.0
+        uses: step-security/harden-runner@fe104658747b27e96e4f7e80cd0a94068e53901d # v2.16.1
         with:
           egress-policy: audit
 
@@ -99,7 +99,7 @@ jobs:
     needs: build-package
     steps:
       - name: Harden Runner
-        uses: step-security/harden-runner@fa2e9d605c4eeb9fcad4c99c224cee0c6c7f3594 # v2.16.0
+        uses: step-security/harden-runner@fe104658747b27e96e4f7e80cd0a94068e53901d # v2.16.1
         with:
           egress-policy: audit
 
--- a/.tekton/pipelines/pull-request-pipeline.yaml
+++ b/.tekton/pipelines/pull-request-pipeline.yaml
@@ -165,7 +165,7 @@ spec:
       - name: name
         value: prefetch-dependencies-oci-ta
       - name: bundle
-        value: quay.io/konflux-ci/tekton-catalog/task-prefetch-dependencies-oci-ta:0.3@sha256:a579d00fe370b6d9a1cb1751c883ecd0ec9f663604344e2fd61e1f6d5bf4e990
+        value: quay.io/konflux-ci/tekton-catalog/task-prefetch-dependencies-oci-ta:0.3@sha256:2229dbc5e15acc0a6d8aec526465aeb0ad54e269c311ac3d0aba88013845e308
       - name: kind
         value: task
       resolver: bundles
@@ -365,7 +365,7 @@ spec:
       - name: name
         value: clair-scan
       - name: bundle
-        value: quay.io/konflux-ci/tekton-catalog/task-clair-scan:0.3@sha256:59dec3000fa35c714d0209dc61e6ec9de15d4d45172246cf5eeff796c5cff65f
+        value: quay.io/konflux-ci/tekton-catalog/task-clair-scan:0.3@sha256:3fa03be0280f33d7070ea53f26d53e727199737a7a2b9a59a95071ae40a999ac
       - name: kind
         value: task
       resolver: bundles
@@ -558,7 +558,7 @@ spec:
       - name: name
         value: apply-tags
       - name: bundle
-        value: quay.io/konflux-ci/tekton-catalog/task-apply-tags:0.3@sha256:510b6d2a3b188adeb716e49566b57d611ab36bd69a2794b5ddfc11dbf014c2ca
+        value: quay.io/konflux-ci/tekton-catalog/task-apply-tags:0.3@sha256:aa62b41861c09e2e59c69cc6e9a1f740bf0c81e6a1eb03f57f59dfda0f65840e
       - name: kind
         value: task
       resolver: bundles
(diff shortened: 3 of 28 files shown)


.github/workflows/ci.yml
```diff
@@ -70,19 +70,30 @@ jobs:
...
+            runs-on: ubuntu-24.04
             uv-options: {}
           - title: " (python 3.10)"
+            runs-on: ubuntu-24.04
             uv-options:
               python-version: '3.10'
+          - title: " (MacOS)"
+            runs-on: macos-26
+            uv-options: {}
+            install-command: brew install podman
```
booxter: honestly not sure if it's reasonable to expect podman installed just to run unit tests, but that's what it demands otherwise a e2e test import triggers `ramalama help` that [fails](https://github.com/booxter/ramalama/actions/runs/23667524192/job/68953094062).

booxter:
@olliewalsh FYI

rhatdan:
@mikebonnet PTAL
@booxter Please respond to the AI questions.

.github/workflows/ci.yml
mikebonnet: This shouldn't be necessary, and I'd rather fix the underlying issue with `ramalama help`. Can someone reproduce this on a Mac without `podman` or `docker` installed?

booxter: You can see the exact crash without this patch here: https://github.com/booxter/ramalama/actions/runs/23667528260/job/68953182754 To reproduce, just pick the very first patch in the series and post it to github to trigger the workflow.

.github/workflows/ci.yml
```diff
@@ -70,19 +70,30 @@ jobs:
...
         variant:
           - title: ""
+            runs-on: ubuntu-24.04
             uv-options: {}
           - title: " (python 3.10)"
+            runs-on: ubuntu-24.04
             uv-options:
               python-version: '3.10'
+          - title: " (MacOS)"
+            runs-on: macos-26
```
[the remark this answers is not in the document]
booxter: Evidently, the bot doesn't know that MacOS 26 was released a while ago and is [available](https://github.blog/changelog/2026-02-26-macos-26-is-now-generally-available-for-github-hosted-runners/) as GA runner.

.github/workflows/ci.yml
```diff
@@ -70,19 +70,30 @@ jobs:
...
     steps:
       - uses: actions/checkout@v6
       - name: Install the latest version of uv
         uses: astral-sh/setup-uv@v7
         with: ${{ matrix.variant.uv-options }}
 
+      - name: Install test prerequisites
+        if: ${{ matrix.variant.install-command }}
+        run: |
+          ${{ matrix.variant.install-command }}
```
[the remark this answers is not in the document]
booxter: I don't think this is true, as can be seen in the actual CI run: https://github.com/containers/ramalama/actions/runs/23669442048/job/68959384706?pr=2567

.github/workflows/ci.yml
booxter: I've replaced installation of `podman` with a change to e2e test file to delay `ramalama help`.

test/e2e/test_help.py
```diff
@@ -61,27 +61,24 @@ def test_help_command_flags():
...
-    else:
-        usage_cmd_name = subcommand
-
-    # Check if usage exists in the help output
-    assert result.startswith(f"usage: ramalama {usage_cmd_name}")
-
-    # Check if the option section is rendered
-    assert re.search("^options:$", result, re.MULTILINE)
+def test_help_output():
+    # Avoid running `ramalama help` during pytest collection for non-e2e jobs.
```
olliewalsh: I take it that means `get_ramalama_subcommands()` runs `ramalama help` under the hood... we should change that now that the e2e tests are pytest vs bats.

For now though could we just not collect the e2e tests when running unit tests?

ramalama/layered_config.py
```diff
@@ -48,21 +48,25 @@ class LayeredMixin:
     def __init__(self, *layers: dict[str, Any]):
         self._fields = {f.name for f in fields(self.__class__) if not f.name.startswith("_")}  # type: ignore[arg-type]
         _layers = [{k: layer[k] for k in layer.keys() & self._fields} for layer in layers]
+        # Expose source layers during dataclass __post_init__ so is_set() works,
+        # but don't record attribute writes until initialization is complete.
+        self._layers = _layers
+        self._layers_finalized = False
```
olliewalsh, line 54: Logic looks good to me. I had done exactly this initially in #2525 (which cased this issue I assume), but then simplified it because it didn't appear to be necessary 🙄

test/e2e/test_help.py
olliewalsh: #2578 ~~should fix the `get_ramalama_subcommands()` issue~~ fixes the unhandled exception

mikebonnet: But why is `ramalama help` failing on MacOS if `podman` or `docker` isn't installed? Shouldn't `ramalama` run on MacOS in `nocontainer` mode just fine with no container engine installed?

olliewalsh: yea, figured that out ... was just not catching FileNotFoundError, updated #2578

olliewalsh: Confirmed it works in https://github.com/olliewalsh/ramalama/actions/runs/23827204646/job/69452884861

@booxter could you drop the e2e test_help commit from this PR?

PR #2575, opened 2026-03-31
Add Intel Windows GPU detection and OpenVINO GPU acceleration

### LINKED ISSUE #2529: Integrate new llama.cpp OpenVINO container

Once the llama.cpp OpenVINO container is done #2528, it should be used by the ramalama cli. 

Can be based on Intel GPU detection or another flag.

_No response_

### Additional context

_No response_

bmahabirbu:
## Summary
- Add Windows Intel GPU detection via PowerShell `Get-CimInstance Win32_VideoController` query (no additional packages required)
- Pass correct container arguments for OpenVINO GPU acceleration: `/dev/dxg` device and `/usr/lib/wsl` bind mount on Windows, `/dev/dri` on Linux (already handled)
- Set `GGML_OPENVINO_DEVICE=GPU` environment variable when Intel GPU is detected
- Add `OPENVINO_VISIBLE_DEVICES` to `GPUEnvVar` and `GGML_OPENVINO_DEVICE` to `AccelEnvVar` so they get passed to containers

Fixes: https://github.com/containers/ramalama/issues/2529

## Test plan
- [ ] Verify Intel GPU detection on native Linux via sysfs (existing behavior unchanged)
- [ ] Verify Intel GPU detection on Windows via PowerShell Win32_VideoController
- [ ] Verify `/dev/dxg` and `/usr/lib/wsl` mount are added to container args on Windows
- [ ] Verify `GGML_OPENVINO_DEVICE=GPU` is passed to container when Intel GPU detected
- [ ] Unit tests pass (`make unit-tests`)

## Summary by Sourcery

Add Intel GPU detection support on Windows and wire it into container execution for OpenVINO GPU acceleration.

New Features:
- Detect Intel GPUs on Windows via PowerShell Win32_VideoController queries.
- Expose Intel/OpenVINO GPU selection to containers through new environment variables and device mounts.

Enhancements:
- Refactor Intel GPU detection to share core logic while separating Linux and Windows implementations.
- Extend container device and mount configuration to support Intel/OpenVINO GPU acceleration on Windows.

### DIFF
--- a/ramalama/common.py
+++ b/ramalama/common.py
@@ -505,6 +505,21 @@ def is_arm() -> bool:
 
 def check_intel() -> Literal["intel"] | None:
     igpu_num = 0
+
+    if platform.system() == "Windows":
+        igpu_num = _check_intel_windows()
+    else:
+        igpu_num = _check_intel_linux()
+
+    if igpu_num:
+        os.environ["INTEL_VISIBLE_DEVICES"] = str(igpu_num)
+        return "intel"
+
+    return None
+
+
+def _check_intel_linux() -> int:
+    igpu_num = 0
     # Device IDs for select Intel GPUs.  See: https://dgpu-docs.intel.com/devices/hardware-table.html
     intel_gpus = (
         b"0xe20b",
@@ -526,11 +541,18 @@ def check_intel() -> Literal["intel"] | None:
             for gpu_id in intel_gpus:
                 if gpu_id in content:
                     igpu_num += 1
-    if igpu_num:
-        os.environ["INTEL_VISIBLE_DEVICES"] = str(igpu_num)
-        return "intel"
+    return igpu_num
 
-    return None
+
+def _check_intel_windows() -> int:
+    """Detect Intel GPUs on Windows via WMI Win32_VideoController."""
+    try:
+        import wmi  # type: ignore
+
+        w = wmi.WMI()
+        return sum(1 for gpu in w.Win32_VideoController() if "intel" in (gpu.Name or "").lower())
+    except Exception:
+        return 0
 
 
 def check_mthreads() -> Literal["musa"] | None:
--- a/ramalama/plugins/runtimes/inference/llama_cpp.py
+++ b/ramalama/plugins/runtimes/inference/llama_cpp.py
@@ -344,6 +344,23 @@ class LlamaCppPlugin(LlamaCppCommands, ContainerizedInferenceRuntimePlugin):
                 help="enable or disable the web UI (default: on)",
             )
 
+    @staticmethod
+    def _set_openvino_env(args: argparse.Namespace) -> None:
+        """Set OpenVINO env vars when Intel GPU is detected (llama.cpp-specific)."""
+        if not os.environ.get("INTEL_VISIBLE_DEVICES"):
+            return
+
+        openvino_env = [
+            f"GGML_OPENVINO_DEVICE={os.environ.get('GGML_OPENVINO_DEVICE', 'GPU')}",
+            f"GGML_OPENVINO_STATEFUL_EXECUTION={os.environ.get('GGML_OPENVINO_STATEFUL_EXECUTION', 1)}",
+        ]
+        args.env = openvino_env + getattr(args, "env", [])
+
+    def handle_subcommand(self, command: str, args: argparse.Namespace) -> list[str]:
+        set_accel_env_vars()
+        self._set_openvino_env(args)
+        return super().handle_subcommand(command, args)
+
     def _do_run(self, args: argparse.Namespace, model: Any) -> None:
         if getattr(args, "rag", None):
             if isinstance(model, APITransport):
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,6 +26,7 @@ dependencies = [
   "pyyaml",
   "jsonschema",
   "jinja2",
+  "wmi ; sys_platform == 'win32'",
 ]
 maintainers = [
    { name = "Dan Walsh", email = "dawalsh@redhat.com" },
--- a/ramalama/engine.py
+++ b/ramalama/engine.py
@@ -1,6 +1,7 @@
 import glob
 import json
 import os
+import platform
 import subprocess
 import sys
 import time
@@ -104,6 +105,7 @@ class BaseEngine(ABC):
             for dev in glob.glob(path):
                 self.exec_args += ["--device", dev]
 
+        intel_windows_added = False
         for k, v in get_accel_env_vars().items():
             # Special case for Cuda
             if k == "CUDA_VISIBLE_DEVICES":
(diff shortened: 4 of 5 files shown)


ramalama/common.py
```diff
@@ -526,11 +542,21 @@ def check_intel() -> Literal["intel"] | None:
...
 
-    return None
+
+def _check_intel_windows() -> int:
+    """Detect Intel GPUs on Windows via PowerShell and Win32_VideoController."""
+    try:
+        command = [
+            "powershell",
+            "-Command",
+            "Get-CimInstance Win32_VideoController | Select-Object -ExpandProperty AdapterCompatibility",
```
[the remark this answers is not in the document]
mikebonnet: Is this a concern?

mikebonnet:
Could you provide a unit test which mocks the `powershell` result?

ramalama/common.py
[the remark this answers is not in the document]
bmahabirbu: hmm seems legit i will update

bmahabirbu:
@mikebonnet  gave this a test on my intel windows laptop and the windows detection worked i.e find the word "inte" and we know we have some intel hardware

once the openvino backend is selected i default to gpu for the env var but it can be changed via pre populating the env var "GGML_OPENVINO_DEVICE" with gpu cpu or npu

ramalama/common.py
```diff
@@ -586,6 +613,7 @@ def set_gpu_type_env_vars():
     "HIP_VISIBLE_DEVICES",
     "INTEL_VISIBLE_DEVICES",
     "MUSA_VISIBLE_DEVICES",
+    "OPENVINO_VISIBLE_DEVICES",
```
mikebonnet: Should `OPENVINO_VISIBLE_DEVICES` be set somewhere, the way `INTEL_VISIBLE_DEVICES` is, or is the expectation that the user will set it before calling `ramalama`?

bmahabirbu: hmm so intel_visible_devices is what we rely on for intel and thats where its set after detection is done. Really openvino_visible_devices can be set implicitly based on INTEL_VISIBLE_DEVICES which is what Ive done!

bmahabirbu:
https://github.com/containers/ramalama/pull/2548

This pr is not blocking the current one but when the above does in we'll want to use this as the target image rather than the upstream llama.cpp for now

ramalama/common.py
mikebonnet: Sorry, I see `INTEL_VISIBLE_DEVICES` and `GGML_OPENVINO_DEVICE` set in `check_intel()`, but I don't see `OPENVINO_VISIBLE_DEVICES` set anywhere.

bmahabirbu: sorry yes its not being set! I wasnt clear in that i meant im just relying on the intel visible devices and the openvino visible devices can be set implicily based on that. Right now though since i dont have a good way to detect if the intel hardware can be used for openvino its not being set

ramalama/common.py
```diff
@@ -526,11 +545,21 @@ def check_intel() -> Literal["intel"] | None:
...
-        return "intel"
+    return igpu_num
 
-    return None
+
+def _check_intel_windows() -> int:
+    """Detect Intel GPUs on Windows via PowerShell and Win32_VideoController."""
+    try:
+        command = [
+            "powershell",
```
olliewalsh: Process creation is very slow on windows. Can we use the python win32 module?

ramalama/common.py
```diff
@@ -586,6 +615,7 @@ def set_gpu_type_env_vars():
     "HIP_VISIBLE_DEVICES",
     "INTEL_VISIBLE_DEVICES",
     "MUSA_VISIBLE_DEVICES",
+    "OPENVINO_VISIBLE_DEVICES",
```
olliewalsh: Is there a reason to add this var at all? It's used in the llama.cpp plugin to allow the image to be overridden but I don't think it's required for anything else.

ramalama/common.py
```diff
@@ -604,6 +634,8 @@ def get_gpu_type_env_vars() -> dict[GPUEnvVar, str]:
 
 AccelEnvVar: TypeAlias = Literal[
     "CUDA_LAUNCH_BLOCKING",
+    "GGML_OPENVINO_DEVICE",
```
olliewalsh: llama.cpp specific vars should not be referenced outside of the llama.cpp plugin

ramalama/common.py
```diff
@@ -23,6 +23,9 @@
 from ramalama.logger import logger
 from ramalama.version import version
 
+if platform.system() == "Windows":
+    import wmi
```
olliewalsh: append `# type: ignore` for mypy

olliewalsh (changes requested):

ramalama/plugins/runtimes/inference/llama_cpp.py
```diff
@@ -344,6 +344,27 @@ def _add_inference_args(self, parser: "argparse.ArgumentParser", command: str) -
                 help="enable or disable the web UI (default: on)",
             )
 
+    @staticmethod
+    def _set_openvino_env(args: argparse.Namespace) -> None:
+        """Set OpenVINO env vars when Intel GPU is detected (llama.cpp-specific)."""
+        if not os.environ.get("INTEL_VISIBLE_DEVICES"):
+            return
+
+        if "GGML_OPENVINO_DEVICE" not in os.environ:
```
olliewalsh: could just do:
```
env = getattr(args, "env", None) or []
env["GGML_OPENVINO_DEVICE"] = os.environ.get("GGML_OPENVINO_DEVICE", "GPU")
env["GGML_OPENVINO_STATEFUL_EXECUTION"] = os.environ.get("GGML_OPENVINO_STATEFUL_EXECUTION", 1)
 ```

of if you want the env from args to take precedence:
```
env = getattr(args, "env", None) or []
env.setdefault("GGML_OPENVINO_DEVICE", os.environ.get("GGML_OPENVINO_DEVICE", "GPU"))
env.setdefault("GGML_OPENVINO_STATEFUL_EXECUTION", os.environ.get("GGML_OPENVINO_STATEFUL_EXECUTION", 1))
```

Comment (olliewalsh) [reply ramalama/plugins/runtimes/inference/llama_cpp.py]: ^H^H missed that args.env is a list... so, I would prepend these args so that user supplied args.env wins:

```
openvino_env = [
    f"GGML_OPENVINO_DEVICE={os.environ.get('GGML_OPENVINO_DEVICE', 'GPU')}", 
    f"GGML_OPENVINO_STATEFUL_EXECUTION={os.environ.get('GGML_OPENVINO_STATEFUL_EXECUTION', 1)}"
]
args.env = openvino_env + getattr(args, "env", [])
```

bmahabirbu: Thanks! ill take your suggestion I was having a tough time figuring out the right way to do it

bmahabirbu:
Thanks Ollie for the reviews this should be ok to go now!

bmahabirbu:
Had to move the wmi import to _check_intel_windows since it was giving warning errors for ramalama list which was odd

bmahabirbu:
@olliewalsh read to test again! thanks

olliewalsh (approved):
LGTM

PR #2584, opened 2026-04-03
Add list-devices subcommand

bmahabirbu:
## Summary
- Adds `ramalama list-devices` subcommand that runs `llama-server --list-devices` inside the default vulkan container image (`quay.io/ramalama/ramalama`)
- Bypasses GPU detection internally — always mounts `/dev/dri` (Linux) or `/dev/dxg` with WSL bind mount (Windows/WSL)
- Only exposes `--image` flag for overriding the container image
- Filters output to show only the "Available devices:" section

## Test plan
- [x] `ramalama list-devices` shows available GPU devices
- [x] `ramalama --dryrun list-devices` shows correct container command
- [x] `ramalama list-devices --help` shows only `--image` argument
- [ ] Verify on WSL system that `/dev/dxg` and `/usr/lib/wsl` mount are included

## Summary by Sourcery

Add a new CLI subcommand to list available GPU/accelerator devices using llama.cpp within the configured container engine.

New Features:
- Introduce a `list-devices` CLI subcommand that runs `llama-server --list-devices` inside the configured container image to enumerate GPU/accelerator devices.

Enhancements:
- Automatically mount common GPU device paths, including WSL-specific mounts when present, and filter the command output to only display the available devices section.
- Support overriding the container image used for device detection via a dedicated `--image` flag while respecting existing dry-run behavior.

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -327,6 +327,7 @@ def configure_subcommands(parser):
     info_parser(subparsers)
     inspect_parser(subparsers)
     list_parser(subparsers)
+    list_devices_parser(subparsers)
     login_parser(subparsers)
     logout_parser(subparsers)
     pull_parser(subparsers)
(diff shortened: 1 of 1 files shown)


rhatdan:
Not necessarily agreeing this needs a subcommand, but if we agree on it, the command should just be `devices`.

Would it make more sense to just do this with `ramalama server --devices` rather then add a new command that will be used by almost no one?

rhatdan:
Another option would be `ramalama list --devices`

mikebonnet:
`ramalama info --devices`?

bmahabirbu:
@rhatdan @mikebonnet that makes sense to me I can figure out what sub command works best but really this is useful for hardware detection without relying on nvidia-smi rocm amd or intel stuff we have right now. So it could be an argument that only developers use for now

the idea is we rely on the vulkan drivers inside the base container to detect the hardware which is super handy! It also gives available memory which is nice to have

Ive been recently looking at llmfit which suggests a model based on hardware. With this technique it would cost us very little to start playing around with that idea

`brian@fedora:~/ramalama$ uv run bin/ramalama list-devices
Available devices:
  Vulkan0: NVIDIA GeForce RTX 3080 (NVK GA102) (10240 MiB, 6237 MiB free)
brian@fedora:~/ramalama$ 
`

bmahabirbu:
In Windows native (remote doesnt have the change wil update shortly) it also looks like this, which is super neat

`PS C:\Users\bmahabir\Desktop\bri\ramalama> uv run python .\bin\ramalama list-devices
Available devices:
  Vulkan0: Microsoft Direct3D12 (NVIDIA GeForce RTX 3080) (10051 MiB, 10051 MiB free)
PS C:\Users\bmahabir\Desktop\bri\ramalama>`

I also found out that a lot of the arguments that are needed for ramalama unnecessarily go through gpu detection like list help etc that should be fixed

ramalama/cli.py
```diff
@@ -540,6 +540,68 @@ def list_containers(args):
...
+        raise ValueError("no container manager (Podman, Docker) found")
+
+    conman_args = [conman, "run", "--rm"]
+    conman_args += ["--security-opt=label=disable"]
+    conman_args += ["--cap-drop=all", "--security-opt=no-new-privileges"]
+
+    import platform
+
+    if platform.system() == "Windows":
+        conman_args += ["--device", "/dev/dxg"]
```
mikebonnet, line 570: Are `/dev/dxg` and `/usr/lib/wsl` guaranteed to exist in every Windows environment? Would it make sense to check for their existence first, like the non-Windows case?

mikebonnet:
Agreed with @bmahabirbu that this looks useful. I don't have strong feelings on naming, but if @rhatdan prefers `ramalama devices` for consistency, that works for me. Otherwise the implementation looks good.

ramalama/cli.py
```diff
@@ -539,6 +540,57 @@ def list_containers(args):
...
+
+    conman_args += [image, "llama-server", "--list-devices"]
+
+    if args.dryrun:
+        from ramalama.engine import dry_run
+
+        dry_run(conman_args)
+        return
+
+    result = subprocess.run(conman_args, capture_output=True, text=True, check=False)
```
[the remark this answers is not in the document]
mikebonnet: @sourcery-ai dismiss

mikebonnet:
The new subcommand is going to need a reference in `docs/ramalama.1.md` and a whole new `.md` file under `docs`.

ramalama/cli.py
bmahabirbu: Good point so /dev/dxg is only available on wsl enviroments and not sure about hyperv. If podman machine is running I could check this since it wont be available on host

ramalama/cli.py
```diff
@@ -540,6 +540,72 @@ def list_containers(args):
...
+        conman_args += ["-e", "LD_LIBRARY_PATH=/usr/lib/wsl/lib"]
+    elif platform.system() == "Darwin":
+        # Assuming user is using a libkrun container with /dev/dri available
+        conman_args += ["--device", "/dev/dri"]
+    else:
+        for path in ["/dev/dri", "/dev/kfd"]:
+            if os.path.exists(path):
+                conman_args += ["--device", path]
+
+    conman_args += [image, "llama-server", "--list-devices"]
```
olliewalsh, line 581: This needs to be in the llama.cpp plugin. Also can't assume the CLI is llama-server, the upstream images use a different entrypoint.

olliewalsh (changes requested):

rhatdan:
I would prefer not to have a subcommand and make this an option, I am fine with ramalama info, if there is not a lot of overhead, then it should be just added to the json file output there.

PR #2589, opened 2026-04-04
Skip GPU detection for subcommands that don't need it

bmahabirbu:
## Summary
- Subcommands like `list`, `version`, `pull`, `rm`, etc. don't need runtime plugin registration or GPU detection, but were paying the cost because all subcommand parsers were built together
- Detects the subcommand early in phase 1 parsing and skips runtime/sandbox/daemon parser registration for lightweight subcommands
- Avoids expensive GPU detection calls (e.g. `nvidia-smi`, CDI config lookups) for commands that never use them

## Test plan
- [ ] Verify lightweight subcommands (`list`, `version`, `pull`, `rm`, `containers`, `login`, `logout`, `push`, `inspect`, `info`, `help`, `stop`, `chat`) work without GPU detection
- [ ] Verify GPU-dependent subcommands (`run`, `serve`, `sandbox`, `daemon`) still perform GPU detection as before
- [ ] Run `make validate` to check lint/format/type checks

## Summary by Sourcery

Optimize CLI argument parsing to avoid unnecessary runtime plugin registration and GPU detection for lightweight subcommands.

Enhancements:
- Add early subcommand detection to classify lightweight commands and skip expensive parser setup for them.
- Restrict runtime plugin subcommand registration and GPU-related setup to GPU-dependent commands while keeping lightweight commands fully functional.

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -171,10 +171,10 @@ def get_initial_parser():
     return parser
 
 
-def get_parser():
+def get_parser(lightweight=False):
     description = get_description()
     parser = create_argument_parser(description, add_help=True)
-    configure_subcommands(parser)
+    configure_subcommands(parser, lightweight=lightweight)
     return parser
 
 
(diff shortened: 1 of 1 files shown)


mikebonnet (approved):
LGTM!

ramalama/cli.py
```diff
@@ -312,16 +318,39 @@ def configure_arguments(parser):
     )
 
 
-def configure_subcommands(parser):
+# Subcommands that don't need runtime plugin registration or GPU detection.
```
olliewalsh, line 338: @bmahabirbu it probably makes more sense to move the gpu detection call into the runtime plugin code path that actually depends on it. I can take a look at this in a follow-up, but might be a good idea to add some basic tests in this PR to ensure it doesn't regress?

rhatdan: Is this still true, @oll did  you ever work on this?

olliewalsh: No, was planning to look at it once this merges

ramalama/cli.py
```diff
@@ -329,16 +335,39 @@ def configure_arguments(parser):
...
+# Only run/serve/bench (via runtime plugin), sandbox, and daemon need it.
+LIGHTWEIGHT_SUBCOMMANDS = {
```
olliewalsh, line 340: would it be easier to list the heavyweight subcommands?

ramalama/cli.py
```diff
@@ -329,16 +335,39 @@ def configure_arguments(parser):
...
+    "list",
+    "login",
+    "logout",
+    "ls",
+    "pull",
+    "push",
+    "rm",
+    "stop",
+    "version",
+}
```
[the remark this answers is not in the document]
olliewalsh: the help subcommand requires runtime plugin registration

ramalama/cli.py
olliewalsh: @bmahabirbu  did you measure how expensive is the GPU detection is? Doesn't seem to be significant. Maybe just on windows where process creation is painfully slow?

rhatdan:
@bmahabirbu What is going on with this PR?

bmahabirbu:
@rhatdan originally this was needed because ramalama would try to detect gpu using list etc which wasnt needed

@olliewalsh still interested in this one i can update it soon

PR #2596, opened 2026-04-05
Fix: handle backslash line continuation without a preceding space (Is…

javierveny:
I've updated the code to be consistent as you suggested. I'm now using .endswith("\") for detection to catch the backslash even without a preceding space.

Also, I replaced rstrip(" \") with .removesuffix("\") to ensure only the trailing backslash is removed, avoiding accidental stripping of other characters. I've verified this logic with unit tests and it's working as expected.
Issue #2437

## Summary by Sourcery

Bug Fixes:
- Fix backslash line-continuation detection so inputs ending with a bare backslash are recognized and processed correctly without stripping other characters.


mikebonnet:
Is there an existing test that covers this functionality? If not, could you add one?

mikebonnet:
Please sign-off on your commits with `git commit -s`.

ramalama/chat.py
```diff
@@ -483,10 +480,14 @@ def default(self, user_content):
             return False
 
         # Handle multi-line input (backslash continuation)
-        self.content.append(user_content.rstrip(" \\"))
-        if user_content.endswith(" \\"):
+        self.content.append(user_content.removesuffix("\\"))
+        if user_content.endswith("\\"):
             return False
 
+        # self.content.append(user_content.rstrip(" \\"))
```
mikebonnet: No need to include this commented-out code.

test/unit/test_chat.py
```diff
@@ -0,0 +1,68 @@
...
+    # instance of RamaLamaShell with the fake args
+    ramalama = RamaLamaShell(mock_args)
+    
+    # disable network function
+    ramalama._req = MagicMock(return_value="Answer")
+    
+    # Case 1, with backslash
+    assert ramalama.default("Hola\\") is False
+    assert ramalama.content == ["Hola"]
+
```
mikebonnet: Worth checking that `ramalama._req` was not called.

test/unit/test_chat.py
```diff
@@ -0,0 +1,68 @@
...
+    ramalama._req = MagicMock(return_value="Answer") # disable network function
+
+    # case 2, we add text after backslash
+    assert ramalama.default("Hola \\") is False
+    assert ramalama.content == ["Hola "]
+
+    # we add text after
+    assert ramalama.default("Mundo") is not False
+    assert ramalama.content == []
+
```
mikebonnet, line 56: Worth checking that `ramalama._req` was called once.

test/unit/test_chat.py
```diff
@@ -0,0 +1,80 @@
...
+    # instance of RamaLamaShell with the fake args
+    ramalama = RamaLamaShell(mock_args)
+    ramalama._req = MagicMock(return_value="Answer") # disable network function
+    
+    # case 3, no backslash
+    result = ramalama.default("Hola")
+    assert result is not False # it does not return True
+    assert ramalama.content == [] # the message has been sended
+
+    # Verify that _req was not called
```
mikebonnet: Comment doesn't match test

mikebonnet:
Please sign-off on all your commits with `git commit -s`.

mikebonnet:
The unused `pytest` import needs to be removed. You can run `make validate` locally to test your changes before pushing.

PR #2629, opened 2026-04-13
Improve macOS install instructions

nickjwhite:
Today I tried to install ramalama for the first time, on a macOS machine that was freshly re-imaged a week ago, and I found quite a few issues with the install instructions, which I have addressed here. I am new to ramalama, so please review carefully in case I've made mistakes. Also maybe worth noting is that the latest released package doesn't seem to work at all on my computer, failing with `unknown runtime: 'llama.cpp`, but the version I built from latest main seems to be working fine.

Changes this PR makes:

- krunkit is a required dependency on macOS, but it was missing from the list of dependencies.
- Also, the --provider argument for `podman machine init` does not exist, so instead recommend adding the provider to the config file, which works.
- pip3 does not work out of the box from brew on macOS, whereas pipx does, so use that in the instructions.
- Installing from source was covered differently in 2 sections, now it is just mentioned once.
- Reorganised to put Prerequisites section before the install section, and the Verify Installation section after all the install sections.


docs/MACOS_INSTALL.md
```diff
@@ -2,9 +2,39 @@
 
 This guide covers the different ways to install RamaLama on macOS.
 
+## Prerequisites
+
+Before using RamaLama, you'll need a container engine and krunkit:
```
rhatdan: You don't necessarily need this. RamaLama will work with MLX locally as well, I believe.

nickjwhite: Oh OK, interesting, I was going off `ramalama-macos.7.md`, which suggests you do need it (and indeed it does complain if `podman machine` is left to the default). https://docs.ramalama.com/oss-docusaurus/docs/getting-started/installation#mlx-runtime-macos-with-apple-silicon suggests that MLX works with it, so long as you use `--nocontainer`.

As a new user, both to ramalama and local LLMs in general, I'm not in a position to say what new users should be suggested to use. My intuition says that `--nocontainer` feels bad, because ramalama is for people who love containers, but I'll have to defer to someone else on this.

olliewalsh: It should default to mlx/--nocontainer if podman/docker are not installed

olliewalsh: Running in a container is particularly challenging on macOS, so `--nocontainer` is an option to consider in this case.

There is a good article on challenges and current solution here https://developers.redhat.com/articles/2025/06/05/how-we-improved-ai-inference-macos-podman-containers#comparison_of_the_container_performance_and_native_performance

There is also work in progress to close the gap to native performance - https://developers.redhat.com/articles/2025/09/18/reach-native-speed-macos-llamacpp-container-inference

nickjwhite: Really interesting reads, thanks for the explanation & links! OK, the macOS documentation needs more changes, then, really. I'm happy to do that, but I won't have time for a couple of days.

nickjwhite: I just pushed an update which mentions MLX, thanks for your patience!

docs/MACOS_INSTALL.md
```diff
@@ -2,9 +2,59 @@
...
+# Install krunkit
+brew tap slp/krun
+brew install krunkit
+
+brew install podman
+
+# Configure Podman machine to use libkrun for GPU access
+mkdir -p $HOME/.config/containers
+printf '[machine]\nprovider = "libkrun"\n' >> $HOME/.config/containers/containers.conf
+
```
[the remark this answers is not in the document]
nickjwhite: This is a fair comment, though this is an install guide, so I was thinking this would be done by new users who don't already have this file. I guess we should either:
- Use `>`, to ensure the config file generated is valid, at a cost of overwriting the config of a user (though only if they'd already deliberately created one, in which case I'd imagine they'd know enough to not blindly copy this)
- Do as the AI review suggests and just ask the user to manually paste  in some config.
I slightly prefer the former, because you can then blindly copy the commands and get a working build, if you're new / confused, but I'm happy with either.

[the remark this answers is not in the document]
nickjwhite: I changed my mind and went with just providing the config excerpts in the updated version.

docs/MACOS_INSTALL.md
nickjwhite: I did some testing, and without podman installed nocontainer is used automatically, but llama.cpp is defaulted to  rather than MLX. ~~Which may not be an unreasonable default; the version of MLX installed with brew seems to not work OOTB with ramalama (created an issue #2639).~~ I just added a note to add the `--runtime` flag /config option for mtx.

nickjwhite:
Note that I haven't done much with python for ages, maybe `uv` should be recommended rather than `pipx` for the install, I don't know. `pipx` definitely worked straightforwardly, though.

rhatdan:
@nickjwhite Still working on this or should it be closed?

nickjwhite:
> @nickjwhite Still working on this or should it be closed?

Hi @rhatdan, I'm not still working on it, as far as I'm concerned it's good (at least better than what was there) and worth merging. If there's anything you're not happy with let me know and I'm happy to rework it.

nickjwhite:
I just rebased against main to resolve a small conflict, made no other changes.

rhatdan:
@olliewalsh @slp @bmahabirbu @mikebonnet WDYT?

slp:
> @olliewalsh @slp @bmahabirbu @mikebonnet WDYT?

LGTM.

olliewalsh (approved):
LGTM

rhatdan (approved):

rhatdan:
Thanks @nickjwhite

nickjwhite:
Thanks all <3

PR #2632, opened 2026-04-13
feat: implemented Ramalama  List shortname functionality

ub-mishra14:


olliewalsh:
@ub-mishra14 how is this intended to work exactly and what is the motivation for doing this?

I'm not sure you can reliably map from a given model back to a short name. When pulling a model from Hugging Face without specifying a quant it will use some defaulting logic to pick a quant. 
For example shortnames.conf has `"smolvlm" = "hf://ggml-org/SmolVLM-500M-Instruct-GGUF"`, but the actual model pulled is `hf://ggml-org/SmolVLM-500M-Instruct-GGUF:Q8_0`.

Also this feature mimics the podman shortname pulling functionality, which does not map the image name back to the shortname AFAIK.

rhatdan:
Theoretically this will be impossible. Or at least difficult to do in general.  Podman and Docker do not do this for container images, so perhaps this is not a great idea.

ub-mishra14:
Hi @olliewalsh @rhatdan sir! 
After looking into the codebase, I noticed `ramalama info --shortnames` already exists and does something similar. it lists all known shortnames from config.

If the goal is to show which shortnames are locally pulled, we could just add a `LOCAL` column to that existing command rather than touching `list` at all. No reverse mapping needed.

Given that Podman and Docker don't do this either maybe this issue isn't really worth pursuing? Happy to close the PR if that makes more sense.

Sorry for the delayed response! appreciate your patience.

rhatdan:
If we had a good way of guaranteeing the reverse mapping, it would be fine, but without that, I think we will just end up with errors and bug reports.

PR #2673, opened 2026-04-23
fix: use model_alias for INFERENCE_MODEL in llama-stack pod

### LINKED ISSUE #2668: llama-stack: INFERENCE_MODEL mismatch with llama-server --alias
## Description

When serving a model via `ramalama serve --api llama-stack`, the generated pod YAML sets the `INFERENCE_MODEL` environment variable to `model_name` (e.g., `granite3.1-dense`), but `llama-server` reports the model via `--alias` as `model_alias` (e.g., `library/granite3.1-dense`).

This causes the llama-stack provider to fail with:

```
ValueError: Model granite3.1-dense is not being served. Available models: library/granite3.1-dense
```

## Steps to Reproduce

1. `ramalama serve --port 8085 --api llama-stack --name granite-service -d granite`
2. Wait for the service to start
3. `curl http://localhost:8085/v1/openai/v1/models`
4. Observe the model name mismatch in logs

## Root Cause

In `ramalama/stack.py` line 78, the `INFERENCE_MODEL` value uses `self.model.model_name` instead of `self.model.model_alias`. The llama-server `--alias` flag (set in `llama_cpp_commands.py:91`) uses `model.model_alias`, so the stack env var must match.

## Expected Behavior

`INFERENCE_MODEL` should use `model_alias` to match what `llama-server --alias` reports.

AlexonOliveiraRH:
## Summary

- Fix `INFERENCE_MODEL` env var in llama-stack pod to use `model_alias` instead of `model_name`, matching what `llama-server --alias` reports

## Problem

The llama-stack provider fails with:
```
ValueError: Model granite3.1-dense is not being served. Available models: library/granite3.1-dense
```

The `--alias` flag in `llama-server` uses `model.model_alias` (e.g., `library/granite3.1-dense`), but `INFERENCE_MODEL` was set to `model.model_name` (e.g., `granite3.1-dense`).

## Test plan

- [ ] `ramalama serve --port 8085 --api llama-stack --name granite-service -d granite`
- [ ] `curl http://localhost:8085/v1/openai/v1/models` returns the model
- [ ] `curl http://localhost:8085/v1/openai/v1/chat/completions` with inference request succeeds

Fixes #2668

### DIFF
--- a/ramalama/stack.py
+++ b/ramalama/stack.py
@@ -75,7 +75,7 @@ class Stack:
         - name: RAMALAMA_URL
           value: http://127.0.0.1:{self.model_port}
         - name: INFERENCE_MODEL
-          value: {self.model.model_name}""",
+          value: {self.model.model_alias}""",
             "port_string": f"""\
         ports:
         - containerPort: 8321


olliewalsh:
Already fixed in #2659

AlexonOliveiraRH:
Good catch, @olliewalsh. #2659 indeed already addresses this. Thank you.

PR #2712, opened 2026-04-30
fix: sync args.rag with formatted model after OCI init

### LINKED ISSUE #2711: RAG: bare image names still resolve to docker.io despite localhost/ prefix fix
## Description

After the fix in PR #2676 that adds a `localhost/` prefix to bare RAG image names via `RagTransport.format_model()`, the image mount command still resolves against `docker.io` because `args.rag` is not updated with the formatted model name.

## Root Cause

`RagTransport.__init__()` calls `super().__init__(args.rag, ...)` which triggers `format_model()` and updates `self.model` to `localhost/myrag:latest`. However, `args.rag` still holds the original bare name (`myrag`).

Later, `RagEngine.add_rag()` uses `args.rag` directly in the container mount command:

```python
self.add_args(f"--mount=type=image,source={rag},destination=/rag,rw=true")
```

Since `args.rag` was never synced, the mount source is the bare name, which Podman/Docker resolves against `docker.io` - producing errors like:

```
Error: docker.io/test:latest does not exist.
```

## Steps to Reproduce

1. Build a RAG database: `ramalama rag doc.md test:latest`
2. Run with the RAG image: `ramalama run --rag test:latest phi4`
3. Observe: `Error: docker.io/test:latest does not exist.`

## Fix

Add `args.rag = self.model` after `super().__init__()` in `RagTransport.__init__()` so that the formatted model name (with `localhost/` prefix) propagates to `RagEngine.add_rag()`.

AlexonOliveiraRH:
## Summary

- Fixes #2711
- After `RagTransport.format_model()` adds the `localhost/` prefix to bare image names, `args.rag` was not updated to match `self.model`. Since `RagEngine.add_rag()` uses `args.rag` directly in the container mount command, the bare name resolved against `docker.io` instead of the local registry, causing `Error: docker.io/<name>:latest does not exist`
- Syncs `args.rag = self.model` after `super().__init__()` so the formatted name propagates to the mount command
- Adds a unit test verifying that `args.rag` reflects the `localhost/` prefix after initialization

## Why

PR #2676 correctly added a `localhost/` prefix via `format_model()` to prevent bare OCI names from defaulting to `docker.io`. However, `format_model()` only updates `self.model` - the `args.rag` field passed to `RagEngine.add_rag()` at mount time was left unchanged, so the fix had no effect on the actual container mount source.

## Test plan

- [x] Unit test `test_bare_name_syncs_args_rag` asserts that `args.rag == "localhost/myrag:latest"` after `RagTransport` init
- [x] All 9 RAG unit tests pass
- [x] All 36 transport base tests pass (conftest fixture shared correctly)

### DIFF
--- a/ramalama/rag.py
+++ b/ramalama/rag.py
@@ -170,6 +170,7 @@ class RagTransport(OCI):
     def __init__(self, imodel: Transport, cmd: list[str], args: RagArgsType):
         self.kind = RagSource.DB if os.path.exists(args.rag) else RagSource.IMAGE
         super().__init__(args.rag, args.store, args.engine)
+        args.rag = self.model
         self.imodel = imodel
         self.model_cmd = cmd
 
--- a/test/e2e/test_serve.py
+++ b/test/e2e/test_serve.py
@@ -851,7 +851,7 @@ def test_serve_with_rag():
         assert re.search(r".*llama-server", result_a)
         assert re.search(r".*quay.io/.*-rag(@sha256)?:", result_a)
         assert re.search(r".*rag_framework serve", result_a)
-        assert re.search(r".*--mount=type=image,source=quay.io/ramalama/rag,destination=/rag,rw=true", result_a)
+        assert re.search(r".*--mount=type=image,source=quay.io/ramalama/rag:latest,destination=/rag,rw=true", result_a)
 
         result_b = ctx.check_output(
             [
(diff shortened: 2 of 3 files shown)


mikebonnet:
Nice catch! Looks like you just need to update `test/e2e/test_serve.py:test_serve_with_rag()` to add `:latest` to the `--mount=` regex.

AlexonOliveiraRH:
> Nice catch! Looks like you just need to update `test/e2e/test_serve.py:test_serve_with_rag()` to add `:latest` to the `--mount=` regex.

@mikebonnet Excellent. Thank you. Done

mikebonnet (approved):
Thanks for the fix!

PR #2444, opened 2026-02-21
Add --engine-args flag for custom container engine arguments

rhatdan:
Implement a new --engine-args flag that allows users to pass additional arguments directly to the container engine (podman/docker). This provides flexibility for advanced users who need to customize container behavior beyond what RamaLama's built-in flags offer.

The flag is named --engine-args (not --container-args) to clarify that arguments are passed to the container engine, not the container itself.

Changes:
- Add --engine-args CLI argument in runtime_options()
- Can be specified multiple times to add multiple arguments
- Implement add_engine_args() method in Engine class
- Arguments are appended directly to exec_args before execution
- Add validation to conflict with --nocontainer flag
- Add engine_args to BaseConfig for ramalama.conf support
- Document flag in ramalama-run and ramalama-serve man pages
- Include usage examples, config file examples, and conflict documentation

Configuration file support:
Users can set default engine args in ramalama.conf:
```toml
[ramalama]
engine_args = ["--read-only", "--tmpfs /tmp"]
```

Use cases enabled:
- Custom mounts: --engine-args '--mount type=bind,src=/data,dst=/data'
- Additional env vars: --engine-args '--env CUSTOM_VAR=value'
- Security options: --engine-args '--security-opt label=disable'
- Read-only containers: --engine-args '--read-only'
- Any other podman/docker run flags

Tests:
- Unit tests verify arguments are added to exec_args correctly
- E2E dryrun tests verify arguments appear in engine command
- Conflict tests verify --engine-args raises error with --nocontainer
- All tests passing (5 new unit tests, 2 e2e tests)

The arguments are passed directly to the container engine, allowing users to leverage the full power of podman/docker without RamaLama needing to explicitly support every possible container option.

Fixes #2440

*This PR was created with the assistance of Cursor AI.*

## Summary by Sourcery

Add support for passing custom arguments to the container engine via CLI and configuration, and ensure they are applied to engine execution only when container mode is enabled.

New Features:
- Introduce a --engine-args CLI flag for run and serve commands to pass additional arguments directly to the container engine.
- Allow configuring default engine_args in ramalama.conf and the BaseConfig so custom engine flags can be set globally.

Enhancements:
- Extend the Engine initialization to append user-specified engine arguments into the container execution command.
- Add validation to prevent using --engine-args when containers are disabled via --nocontainer.

Documentation:
- Document the --engine-args option and configuration usage, including examples, in ramalama-run, ramalama-serve, and ramalama.conf manuals.

Tests:
[...]

PR #2728, opened 2026-05-06
feat: add global --engine-args for podman/docker (discussion #2703)

rhatdan:
Expose a shell-quoted --engine-args flag (same parsing as --runtime-args) so users can pass extra container-engine flags such as additional --mount bindings for mmproj files. Tokens are appended after RamaLama-generated options and model mounts, immediately before the container image.

Also apply engine args to podman kube play/down (llama-stack), and to convert/quantize helper containers.

Replaces: https://github.com/containers/ramalama/pull/2444

Docs: ramalama(1) GLOBAL OPTIONS. E2E dry-run coverage for --engine-args.


ramalama/engine.py
olliewalsh, line 185: could the engine args be added once here with all of the other podman/docker specific args?

rhatdan:
@olliewalsh PTANL

test/e2e/utils.py
```diff
@@ -1,14 +1,44 @@
...
 import re
 import shutil
 import subprocess
+import sys
 import tempfile
 from pathlib import Path
 
 from test.conftest import ramalama_container, ramalama_container_engine
 
+_REPO_ROOT = Path(__file__).resolve().parents[2]
```
olliewalsh: what is this for? Doesn't look related to the PR

rhatdan: Cursor added it when generating tests, if you don't think we need to _ensure_repo_on_pythonpath is necessary I can remove it.

rhatdan:
@olliewalsh Waiting on you I believe.

test/e2e/utils.py
olliewalsh: If neither of us understand why it's there then it shouldn't be merged

rhatdan:
@bmahabirbu @mikebonnet PTAL

rhatdan:
@olliewalsh PTAL

test/e2e/utils.py
```diff
@@ -1,15 +1,34 @@
...
 import shutil
 import subprocess
+import sys
 import tempfile
 from pathlib import Path
 
 from test.conftest import ramalama_container, ramalama_container_engine
 
 
+def _rewrite_ramalama_cmd(args: tuple) -> tuple:
```
olliewalsh: I don't understand this. It mentions issues with the pytest collection but we know that collection works. Cursor slop?

olliewalsh:
@coderabbitai review

olliewalsh (approved):
LGTM

ramalama/plugins/runtimes/inference/llama_cpp.py
```diff
@@ -216,8 +217,7 @@ def _convert_to_gguf(self, outdir, source_model, args):
                 args.tools_image = ensure_image(
                     args.engine, args.tools_image, should_pull=should_pull, quiet=getattr(args, "quiet", False)
                 )
-            engine.add_args(args.tools_image)
-            engine.add_args(*self._cmd_convert(args))
+            engine.add_container_image(args.tools_image, *self._cmd_convert(args))
```
olliewalsh:
```
File "/tmp/.tox/slow/lib/python3.14/site-packages/ramalama/plugins/runtimes/inference/llama_cpp.py", line 220, in _convert_to_gguf
    engine.add_container_image(args.tools_image, *self._cmd_convert(args))
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: BaseEngine.add_container_image() takes from 2 to 3 positional arguments but 6 were given
```

ramalama/plugins/runtimes/inference/llama_cpp.py
```diff
@@ -268,7 +268,7 @@ def _quantize(self, source_model, args, model_dir):
         engine.add_args(args.image)
```
olliewalsh, line 268: I expect this is the cause of #2843, adding the image args twice

Issue #2398, opened 2026-02-05
Change default model repo to Hugging Face


olliewalsh:
When not explicitly provided, the default model repository should change from Ollama to Hugging Face.

Ollama has diverged from upstream llama.cpp. As a result of this, an increasing number of newer ollama models are not compatible with lama.cpp (e.g gpt-oss).

While we can workaround this be using shortnames to override specific models, this is not sustainable.

rhatdan:
Making this change requires the RamaLama project to greatly improve our handling of shortnames.conf,  

We need to go through the most popular AI models, and update the shortnames.conf file to pick a default. 

```
$ ramalama run hf://deepseek
Downloading hf:///deepseek ...
Trying to pull hf:///deepseek ...
URL pull failed and hf not available
Error: Failed to pull model: 'No metadata found
```
The average user getting this error is going to be at a loss and will switch very quickly to ollama.

When a new popular model hits the press, RamaLama needs to quickly release a new version.

If huggingface or llama.cpp or a different entity provided an easy API for ramalama to call into for shortnames expansion then I am 100% for this change.

olliewalsh:
Yes, as I commented on #2405 this isn't a straightforward case of changing the default which is why I raised an issue for it.

> Making this change requires the RamaLama project to greatly improve our handling of shortnames.conf,
> We need to go through the most popular AI models, and update the shortnames.conf file to pick a default.

Yes, preserving the current ollama-style shortname behaviour is an issue.

> When a new popular model hits the press, RamaLama needs to quickly release a new version.

Flipping that around - when a new popular model hits the press, the ollama model will not be compatible with llama.cpp (only legacy model are IIUC), so will fail and the average user will be at a loss and switch very quickly to ollama...

akashWhoCodes:
Hi @olliewalsh @rhatdan,

Thanks for opening this issue and for your guidance on why Ollama was the original default. I agree that the divergence between Ollama and upstream `llama.cpp` makes the current default increasingly brittle.

I’ve opened PR #2405 which implements the switch to Hugging Face. To avoid surprising users who rely on Ollama shortnames, my plan is to stage the change across releases:

• In the next minor release we keep `ollama` as the default, emit a deprecation warning at start‑up and update the docs to explain how to set `RAMALAMA_TRANSPORT=ollama` or use an explicit `ollama://` prefix to retain current behaviour.

• Meanwhile we can document or maintain short‑name mappings for popular Hugging Face models so that common shortnames remain functional.

[...]

PR #2743, opened 2026-05-11
Begin deprecation of ollama model support

olliewalsh:
Related to #2398.

If a model name includes a slash then default to Hugging Face instead of ollama - none of the listed models in https://ollama.com/library contain a slash (some unlisted user contributed models do).
Add a deprecation warning when pulling an ollama model.
Update shortnames.conf to remove any references to ollama models and cleanup the existing entries for consistency.

Assisted-by: Claude Opus 4.6 <noreply@anthropic.com>


ramalama/transports/transport_factory.py
```diff
@@ -25,19 +26,22 @@
...
-        transport: str = "ollama",
+        transport: Optional[str] = None,
         ignore_stderr: bool = False,
     ):
 
         self.model = model
         self.store_path = args.store
-        self.transport = transport
+        self._transport_is_default = transport is None
+        self.transport = transport if transport is not None else "ollama"
```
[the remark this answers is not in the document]
olliewalsh: That behaviour is correct. If ollama is explicitly specified in the config file then it should be used regardless of the existence of the slash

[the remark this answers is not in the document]
olliewalsh: Perhaps `_transport_is_default` should be renamed to `_transport_is_implicit`?

README.md
```diff
@@ -271,7 +271,8 @@ RamaLama supports multiple AI model registries types called transports.
 |                          |Examples: [`quay.io`](https://quay.io),  [`Docker Hub`](https://docker.io), [`Pulp`](https://pulpproject.org), and [`Artifactory`](https://jfrog.com/artifactory/)|
 
 ### Default Transport
-RamaLama uses the Ollama registry transport by default
+
+Models specified in the Hugging Face `<org>/<model>` format are automatically pulled from Hugging Face. For models specified without an organization (e.g. `granite-code`), RamaLama currently defaults to the Ollama transport. **Note:** Ollama models are no longer compatible with llama.cpp, and support for the Ollama transport will be removed in a future release. Users should migrate to Hugging Face models.
```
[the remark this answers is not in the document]
olliewalsh: Only real issue is some example shortnames which are no longer accurate, will amend

ramalama/transports/transport_factory.py
```diff
@@ -74,18 +79,38 @@ def detect_model_model_type(self) -> tuple[type[CLASS_MODEL_TYPES], Callable[[],
...
+    def _warn_ollama_deprecated(self):
+        global _ollama_default_warned
+        if _ollama_default_warned:
+            return
+        _ollama_default_warned = True
+        msg = ""
+        if self._transport_is_default:
+            msg = f"Defaulting to Ollama transport for '{self.model}'.\n"
+        warnings.warn(
+            f"{msg}Note: Ollama models are no longer compatible with llama.cpp. "
```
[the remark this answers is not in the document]
olliewalsh: The model normalization logic is split between cli.py and transport_factory.py, will address this

ramalama/transports/transport_factory.py
```diff
@@ -74,18 +79,38 @@ def detect_model_model_type(self) -> tuple[type[CLASS_MODEL_TYPES], Callable[[],
...
+        if self._transport_is_implicit:
+            msg = f"Defaulting to Ollama transport for '{self.model}'.\n"
+        warnings.warn(
+            f"{msg}Note: Ollama models are no longer compatible with llama.cpp. "
+            "Support for Ollama models will be removed in a future release. "
+            "Use the Hugging Face <org>/<model> shorthand format, "
+            "or specify a transport explicitly, e.g. huggingface:// or oci://.",
+            FutureWarning,
+            stacklevel=3,
+        )
```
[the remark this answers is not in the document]
olliewalsh: It's just a warning, lets not over-engineer it

shortnames/shortnames.conf
```diff
@@ -1,69 +1,69 @@
...
-  "qwen3:1.7b" = "hf://Qwen/Qwen3-1.7B-GGUF/Qwen3-1.7B-Q8_0.gguf"
-  "qwen3:4b" = "hf://Qwen/Qwen3-4B-GGUF/Qwen3-4B-Q4_K_M.gguf"
-  "qwen3:8b" = "hf://Qwen/Qwen3-8B-GGUF/Qwen3-8B-Q4_K_M.gguf"
-  "qwen3:32b" = "hf://Qwen/Qwen3-32B-GGUF/Qwen3-32B-Q4_K_M.gguf"
-  "qwen3.5:0.8b" = "hf://bartowski/Qwen_Qwen3.5-0.8B-GGUF/Qwen_Qwen3.5-0.8B-Q4_K_M.gguf"
-  "qwen3.5:2b" = "hf://bartowski/Qwen_Qwen3.5-2B-GGUF/Qwen_Qwen3.5-2B-Q4_K_M.gguf"
-  "qwen3.5:4b" = "hf://bartowski/Qwen_Qwen3.5-4B-GGUF/Qwen_Qwen3.5-4B-Q4_K_M.gguf"
-  "qwen3.5:9b" = "hf://bartowski/Qwen_Qwen3.5-9B-GGUF/Qwen_Qwen3.5-9B-Q4_K_M.gguf"
-  "qwen3.5:27b" = "hf://bartowski/Qwen_Qwen3.5-27B-GGUF/Qwen_Qwen3.5-27B-Q4_K_M.gguf"
+  "qwen3:0.6b" = "hf://Qwen/Qwen3-0.6B-GGUF"
```
rhatdan, line 57: What about 
"qwen3", 
"qwen3.5"

These should default to the size Ollama is pulling now, except from HF.

ramalama/transports/transport_factory.py
```diff
@@ -74,18 +79,38 @@ def detect_model_model_type(self) -> tuple[type[CLASS_MODEL_TYPES], Callable[[],
...
         if self.transport == "ollama":
+            self._warn_ollama_deprecated()
             return Ollama, self.create_ollama
         if self.transport == "rlcr":
             return RamalamaContainerRegistry, self.create_rlcr
         if self.transport == "oci":
             return OCI, self.create_oci
         raise KeyError(f'transport "{self.transport}" not supported. Must be oci, huggingface, modelscope, or ollama.')
 
+    def _warn_ollama_deprecated(self) -> None:
```
rhatdan, line 97: Need to be able to disable this in ramalama.conf, if I know what I am doing, I don't want to be warned.

shortnames/shortnames.conf
olliewalsh: Yes, HF defaults to Q4_K_M, and Ollama always uses Q4_K_M

ramalama/transports/transport_factory.py
```diff
@@ -171,5 +196,6 @@ def create_api_transport(self) -> APITransport:
 
 def New(name, args, transport: Optional[str] = None) -> CLASS_MODEL_TYPES:
     if transport is None:
-        transport = ActiveConfig().transport
+        config = ActiveConfig()
+        transport = config.transport if config.is_set("transport") else None
```
olliewalsh, line 200: @rhatdan this handles the case where it's set in the config file: transport is not None so _transport_is_implicit==False, and the ~~warning~~ "Defaulting to..." is not emitted

ramalama/transports/transport_factory.py
olliewalsh: ~~Yes, that works. See other comment~~

olliewalsh: Actually no, I changed it to always warn that ollama is deprecated but only add the "Defaulting to..." prefix when it's implicit. Ollama has broken compatibility with llama.cpp for some time now so ultimately I think we should drop support for it's models

olliewalsh: So yeah, if you are explicitly choosing ollama I still want the nagging warning to nudge you away from doing this. There isn't really a compelling reason to stick with the old ollama models that were compatible with llama.cpp - they all just seem to be taken directly from Hugging Face anyway so better to go direct to the source

mikebonnet (approved):
LGTM!

ramalama/transports/transport_factory.py
olliewalsh: @rhatdan WDYT? It has merged already but will wait for your feedback before cutting a new release

ramalama/transports/transport_factory.py
rhatdan: Well we will see what people's reaction is, I just often like a `Shut the F Up mechanism` when someone is constandly nagging me.

olliewalsh: ah, there is a global STFU env var for python warnings e.g:
```
$ PYTHONWARNINGS="ignore::FutureWarning" ramalama pull smollm
Using cached ollama://library/smollm:latest ...
```

PR #2762, opened 2026-05-21
Enforce sorted shortnames.conf and prefer bartowski quantizations

olliewalsh:
Add Shortnames.sort_file() and make ramalama/shortnames.py directly
runnable to sort shortnames.conf by key while preserving comments.

Prefer bartowski quantizations where they exist. They are more likely to be updated to take advantage of llama.cpp features e.g his qwen3.6 quantizations have been updated to support MTP.


olliewalsh:
@rhatdan PTAL

hack/sort-shortnames
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+# Usage: hack/sort-shortnames [--check]
+# Sorts shortnames/shortnames.conf entries by key.
+# With --check, exits non-zero if the file is not already sorted.
+
+set -euo pipefail
+
+CONF="shortnames/shortnames.conf"
+
+header=$(head -n1 "$CONF")
```
olliewalsh: this is too naive

rhatdan:
You beat me to this.  Please rebase.

ramalama/shortnames.py
```diff
@@ -66,3 +67,60 @@ def _strip_quotes(self, s) -> str:
...
+        header: list[str] = []
+        entries: list[str] = []
+        in_entries = False
+        for line in lines:
+            if not in_entries:
+                header.append(line)
+                if line.strip().lower() == '[shortnames]':
+                    in_entries = True
+            else:
+                entries.append(line)
```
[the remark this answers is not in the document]
olliewalsh: If the file does not contain a `[shortnames]` section it is still valid - just ignored when parsing

olliewalsh:
@rhatdan WDYT? Would like to get this merged before cutting a new version

shortnames/shortnames.conf
```diff
@@ -16,194 +16,195 @@
...
+  "qwen2.5-coder:14b" = "hf://bartowski/Qwen2.5-Coder-14B-Instruct-GGUF"
+  "qwen2.5-coder:32b" = "hf://bartowski/Qwen2.5-Coder-32B-Instruct-GGUF"
+  "qwen2.5-coder:7b" = "hf://bartowski/Qwen2.5-Coder-7B-Instruct-GGUF"
+  "qwen2.5:0.5b" = "hf://bartowski/Qwen2.5-0.5B-Instruct-GGUF"
+  "qwen2.5:1.5b" = "hf://bartowski/Qwen2.5-1.5B-Instruct-GGUF"
+  "qwen2.5:14b" = "hf://bartowski/Qwen2.5-14B-Instruct-GGUF"
+  "qwen2.5:32b" = "hf://bartowski/Qwen2.5-32B-Instruct-GGUF"
+  "qwen2.5:3b" = "hf://bartowski/Qwen2.5-3B-Instruct-GGUF"
+  "qwen2.5:72b" = "hf://bartowski/Qwen2.5-72B-Instruct-GGUF"
+  "qwen2.5:7b" = "hf://bartowski/Qwen2.5-7B-Instruct-GGUF"
```
rhatdan: This looks wrong, it should be sorted
  "qwen2.5:7b" = "hf://bartowski/Qwen2.5-7B-Instruct-GGUF"
 "qwen2.5:72b" = "hf://bartowski/Qwen2.5-72B-Instruct-GGUF"

rhatdan: Need to sort by version.

olliewalsh: the tags are just strings though, not numeric or symver

olliewalsh: ...so lexical sort is the correct approach

rhatdan: I am just looking at it from a Human perspective.  If I see them I expect them to be sorted in smallest to largest.

olliewalsh: Could probably do something to achieve this, but really seems like over-engineering. This file just needs to be sorted consistently to avoid things like duplicate entries etc... Humans are not going to be reading it very often (mostly just us I expect)

rhatdan: I am fine with whatever you decide.  Cursor built the following python code for my version of this PR which handled it, if you want to take a look:

[sort-shortnames-conf.py](https://github.com/user-attachments/files/28144747/sort-shortnames-conf.py.txt)

Also Python seems to have a version sort function, which I have not attempted.

```
from packaging.version import parse

versions = ["1.1", "1.1b1", "1.1a1", "1.0"]
versions.sort(key=parse)

print(versions) 
# Output: ['1.0', '1.1a1', '1.1b1', '1.1']
```

olliewalsh: Yeah, cursor has really over-engineered it IMO. I've an idea for something much simpler/robust...

olliewalsh: Ok, I think it makes sense to sort by tag numerically - most follow that same pattern.
I don't think it's worth trying to extract a version from the name though, it's very messy.

olliewalsh: Or trying to deal with gemma-4 vs gemma4 (Why do we have both?). It  just gets hairy very very quickly. If we want to help users get to the right shortname there is probably a better way to do this than having duplicate entries - some sort of "did you mean: gemma4" if the user asks for gemma-4.

rhatdan (approved):
LGTM

PR #2763, opened 2026-05-21
Sort shortnames as versions

rhatdan:


rhatdan:
We can go with @olliewalsh solution.

Issue #2734, opened 2026-05-10
make -ncmoe a first class parameter for llama.cpp runtime


mkristian:
### Feature request description

Since I have a limited VRAM setup I am mainly looking for MoE models, with them I try to keep as many experts on the GPU. This setup produces good results for me. The bench sub-command is great to find the sweet spot of how many experts running on the CPU only. But this feature forces me to use the runtime-args argument and then the output of the bench command does not show ncmoe argument used, unless I use the verbose json format. So it is tricky to use them.

Since for me this tuning is the main way to setup a model I would like to see the ngl and ncmoe parameter treated in the same manner. 

Ideally the table output of the bench/benchmark command could suppress either ngl or ncmoe if the defaults are used to keep the table width small.

### Suggest potential solution

Add the ncmoe parameter alongside ngl parameter (run, server, bench, etc) and improve the table output of the benchmark.

rhatdan:
Do these options make sense for other runtimes beside llama.cpp?  

Interested in opening a PR to add this feature?

mkristian:
No, this would be llama.cpp runtime only and the the bench mark commands.

Yes, I can polish my working branch and submit a PR, but due to holidays it can take a week or two to finish.

rhatdan:
Ok look forward to seeing it.

mkristian:
#2778 merged

PR #2778, opened 2026-06-02
Add ncmoe as parameter

mkristian:
This PR adds the ncmoe as parameter for llama.cpp for convenience and improves a little bit the out of the table by hiding columns which just shows default values.

Addresses #2734

### DIFF
--- a/docs/ramalama.conf.5.md
+++ b/docs/ramalama.conf.5.md
@@ -262,6 +262,8 @@ Options: `Q2_K`, `Q3_K_S`, `Q3_K_M`, `Q3_K_L`, `Q4_0`, `Q4_K_S`, `Q4_K_M`, `Q5_0
 
 **ngl**=-1: Number of layers to offload to the GPU. Set to -1 to offload all layers.
 
+**ncmoe**=0: Keep the Mixture of Experts (MoE) weights of the first N layers in the CPU.  
+
 **temp**="0.8": Response sampling temperature.
 - Lower values: more deterministic output
 - Higher values: more creative output (higher hallucination risk)
--- a/ramalama/benchmarks/utilities.py
+++ b/ramalama/benchmarks/utilities.py
@@ -40,8 +40,9 @@ def print_bench_results(records: list[BenchmarkRecord]):
         model = result.model_filename or ""
         params = f"{result.model_n_params / 1e9:.2f} B" if result.model_n_params else "-"
         backend = result.gpu_info or result.cpu_info or "CPU"
-        ngl = str(result.n_gpu_layers) if result.n_gpu_layers else "-"
-        threads = str(result.n_threads) if result.n_threads else "-"
+        ngl = str(result.n_gpu_layers) if result.n_gpu_layers is not None else "-"
+        ncmoe = str(result.n_cpu_moe) if result.n_cpu_moe is not None else "-"
+        threads = str(result.n_threads) if result.n_threads is not None else "-"
 
         # Format test type
         if result.n_prompt and result.n_gen:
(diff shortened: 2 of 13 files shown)


rhatdan:
Please respond to the AI Reviewers.

If we add this new option, then it needs to have defaults set in ramalama.conf file as well.

ramalama/benchmarks/utilities.py
```diff
@@ -82,7 +84,13 @@ def print_bench_results(records: list[BenchmarkRecord]):
             for row in rows:
                 row.pop(field, None)
 
-    column_order = ["id", "model", "params", "backend", "ngl", "threads", "test", "t/s", "engine", "date"]
+    fields_with_defaults = {"ngl": "99", "ncmoe": "0"}
+    for key, default in fields_with_defaults.items():
+        if all((row.get(key) in [value, '-']) for row in rows):
+            for row in rows:
+                row.pop(key, None)
+
```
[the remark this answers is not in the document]
mkristian: fixed

docs/ramalama.conf.5.md
```diff
@@ -262,6 +262,8 @@ Options: `Q2_K`, `Q3_K_S`, `Q3_K_M`, `Q3_K_L`, `Q4_0`, `Q4_K_S`, `Q4_K_M`, `Q5_0
 
 **ngl**=-1: Number of layers to offload to the GPU. Set to -1 to offload all layers.
 
+**ncmoe**=0: Keep the Mixture of Experts (MoE) weights of the first N layers in the CPU.
```
olliewalsh, line 265: As @rhatdan mentioned, this should be added to `docs/ramalama.conf` too

ramalama/benchmarks/utilities.py
```diff
@@ -82,7 +84,13 @@ def print_bench_results(records: list[BenchmarkRecord]):
             for row in rows:
                 row.pop(field, None)
 
-    column_order = ["id", "model", "params", "backend", "ngl", "threads", "test", "t/s", "engine", "date"]
+    fields_with_defaults = {"ngl": "99", "ncmoe": "0"}
+    for key, default in fields_with_defaults.items():
```
olliewalsh, line 90: looks like this can this move after `headers` is set and then just pop the field from `headers` instead of iterating all rows

mkristian: yup. done.

docs/ramalama.conf.5.md
mkristian: added in ramalama.conf as well

mkristian:
The last time there were errors here: https://github.com/containers/ramalama/runs/79366125195
I do not have access to the details so any hint what went wrong to address them is appreciated.

olliewalsh:
> The last time there were errors here: https://github.com/containers/ramalama/runs/79366125195 I do not have access to the details so any hint what went wrong to address them is appreciated.

Looks like unrelated infra issues, re-running.

ramalama/benchmarks/utilities.py
olliewalsh: I meant:
```
column_order = ["id", "model", "params", "backend", "ngl", "ncmoe", "threads", "test", "t/s", "engine", "date"]
headers = [column for column in column_order if column in rows[0]]

fields_with_defaults = {"ngl": "99", "ncmoe": "0"}
for key, default in fields_with_defaults.items():
    if all((row.get(key) in [default, '-']) for row in rows):
        headers.remove(key)
```

Dropping the field from row[0] has the same effect but it's not as clear

olliewalsh:
@mkristian I've rebased this but still failing on the DCO check because the author email does not match the signing email on the commits

mkristian:
@olliewalsh  thanks, I was offline until today, I will take care of this once I am back at my laptop

mkristian:
@olliewalsh all rebased and fixed this one commit I made through the github GUI.

olliewalsh (approved):
LGTM

PR #2779, opened 2026-06-03
Improve flaky e2e push tests

olliewalsh:
The e2e push tests are often failing with a timeout. I cannot reproduce this locally or in github CI runners. Make the registry init more robust and add some logging to help debug this if/when it occurs again:

Replace sleep(2) with active readiness poll on the registry's /v2/ endpoint
Add timeout=180 to push calls so hung processes fail fast
Dump registry container logs on teardown for debugging


rhatdan (approved):
LGTM

PR #2781, opened 2026-06-04
Add ramalama.conf tempdir setting for TMPDIR selection

### LINKED ISSUE #2731: `/tmp` still used for storage-heavy operations
### Issue Description

The problem is somewhat related to #1778, which was closed few months ago, but I still observe space errors when running e.g. `ramalama convert ...` due to it using the directory `/tmp` instead of `/var/tmp` despite `TMPDIR` being unset. When I explicitly set the variable, then everything works as expected, but when unset or empty it defaults to `/tmp`.

### Steps to reproduce the issue

1. Make sure `TMPDIR` environment variable is not set (`$ unset TMPDIR`)
2. In a separate terminal window run `$ watch -n 1 ls /tmp` to observe content of `/tmp` directory
3. Launch a space extensive operation, e.g. `$ ramalama convert --gguf ...` with model larger than the tmpfs partition

### Describe the results you received

* Temporary directories appear in `/tmp`:
  ```
  $ ls -1 /tmp
  ...
  RamaLama_convert_src_abc
  RamaLama_convert_xyz
  ...
  ```

* The convert itself fails with:
  ```
  Error: [Errno 122] Disk quota exceeded: '/.../ramalama/store/.../model/blobs/sha256-...' -> '/tmp/RamaLama_convert_src_abc/model
  ```

### Describe the results you expected

The directory `/var/tmp` should have been used instead of `/tmp`

### Upstream Latest Release

Yes

### Additional environment details

* Arch Linux
* Podman 5.8.2
* RamaLama 0.20.0

### Additional information

On a related note, since there are many valid cases for setting up `TMPDIR` variable in the environment, a switch to a dedicated variable `RAMALAMA_TMPDIR` and/or option in config file would be highly beneficial.

rhatdan:
Packaged installs call cli.main directly, so default /var/tmp was never applied. Apply tempdir from config (overriding host TMPDIR when set), otherwise honor host TMPDIR, then fall back to /var/tmp on Unix.

Fixes https://github.com/containers/ramalama/issues/2731

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ uv.lock
 !.pixi/config.toml
 *~
 *#
+.mypy_cache
--- a/bin/ramalama
+++ b/bin/ramalama
@@ -1,16 +1,13 @@
 #!/usr/bin/env python3
 
-import os
 import sys
+
+from ramalama.config import ActiveConfig, ensure_tmpdir
 
 
 def main():
     if not sys.stdout.isatty():
         sys.stdout.reconfigure(line_buffering=True)
-
-    if sys.platform != 'win32':
-        # default TMPDIR environment to /var/tmp on non Windows platforms
-        os.environ["TMPDIR"] = os.getenv("TMPDIR", "/var/tmp")
 
     sys.path.insert(0, './')
     try:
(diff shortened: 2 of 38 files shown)


rhatdan:
@olliewalsh @bmahabirbu @mikebonnet PTAL

mikebonnet (approved):
LGTM

PR #2794, opened 2026-06-11
test: use simpler math prompt in sandbox stdin test

bmahabirbu:
## Summary
- Replace "What is the atomic number of molybdenum?" with "What is 6 times 7?" in `test_sandbox_run_stdin`


mikebonnet (approved):
Hopefully this reduces the flakes!

PR #2796, opened 2026-06-11
Add popular Hugging Face and Ollama shortname mappings

rhatdan:
Map trending Ollama library models and recent HF GGUF releases to shortnames, including DeepSeek V3/V4, GLM 4.7/5, Ministral 3, Qwen3 variants, Llama 4, and Granite 4.1.


shortnames/shortnames.conf
```diff
@@ -70,6 +81,9 @@
   "glm-4.6" = "hf://unsloth/GLM-4.6-GGUF"
   "glm-4.6-vl" = "hf://ggml-org/GLM-4.6V-GGUF"
   "glm-4.6v-flash" = "hf://bartowski/zai-org_GLM-4.6V-Flash-GGUF"
+  "glm-4.7" = "hf://bartowski/zai-org_GLM-4.7-GGUF"
+  "glm-5" = "hf://DevQuasar/zai-org.GLM-5-GGUF"
```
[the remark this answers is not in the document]
rhatdan: Cursor responds.
```
Gemini flagged the glm-5 entry on line 85:


shortnames.conf
Lines 85-86
  "glm-5" = "hf://unsloth/GLM-5-GGUF"
  "glm-5.1" = "hf://bartowski/zai-org_GLM-5.1-GGUF"
It thought DevQuasar/zai-org.GLM-5-GGUF was a typo and suggested zai-org_GLM-5-GGUF instead.

Gemini’s suggested fix would be wrong — DevQuasar/zai-org.GLM-5-GGUF is a real repo (DevQuasar uses . in names), while DevQuasar/zai-org_GLM-5-GGUF does not exist. There’s also no bartowski/zai-org_GLM-5-GGUF yet (only the 5.1 quant).

What I changed: point glm-5 at unsloth/GLM-5-GGUF instead, which:

Uses normal repo naming (no confusing dot)
Matches our secondary-source preference when bartowski isn’t available
Has more downloads (~4.4k vs ~146)
Pushed as commit 00460514 on shortnames-popular-models for PR #2796. You can reply on the review that the dot wasn’t a typo, but we switched to unsloth for clearer naming and better maintenance.
```

bmahabirbu (approved):

shortnames/shortnames.conf
```diff
@@ -136,6 +167,10 @@
   "nemotron:51b" = "hf://bartowski/Llama-3_1-Nemotron-51B-Instruct-GGUF"
   "nemotron:70b" = "hf://bartowski/Llama-3.1-Nemotron-70B-Instruct-HF-GGUF"
   "nemotron:120b" = "hf://lmstudio-community/NVIDIA-Nemotron-3-Super-120B-A12B-GGUF"
+  "nemotron-3-nano" = "hf://lmstudio-community/NVIDIA-Nemotron-3-Nano-30B-A3B-GGUF"
```
olliewalsh: can we use the bartowski models?

olliewalsh (changes requested):

rhatdan:
@olliewalsh How does this look?  Is it too much?

olliewalsh:
> @olliewalsh How does this look? Is it too much?

The huge models are too much maybe. That's into datacenter-class hardware territory.

rhatdan:
I don't know why we should limit the size of the models the user would want to pull.  If the user wants to pull a HUGE model and run it with llama.cpp, I see no reason to block it.

olliewalsh:
> I don't know why we should limit the size of the models the user would want to pull. If the user wants to pull a HUGE model and run it with llama.cpp, I see no reason to block it.

We are not limiting the model size, just not sure if adding shortnames for huge models makes sense - anyone with that class of hardware needs to know what they are doing.

... but it's just a few entries so doesn't matter in the grand scheme of things :-)

rhatdan:
Fine, tell me which shortnames you want to remove and then add documentation and a description so that cursor will know not to add them in the future.

rhatdan:
@olliewalsh what do you want removed, I want to get the new GLP model in, since it is in the news quite a bit.

rhatdan:
GLM not GLP, AI never looses weight.

olliewalsh:
> GLM not GLP, AI never looses weight.

Who is going to be able to run it? GLM 5.2 @ Q4_K_M needs around 500GB

rhatdan:
So users will not be able to run glm-5.2 on local laptops without HUGE memory.

rhatdan:
Ok I removed the default to GLM go point at 5.2.  Any other you would want removed.

olliewalsh:
> Ok I removed the default to GLM go point at 5.2. Any other you would want removed.

Do we have a target user in mind?

cogito-671b-v2.1-GGUF is huge too... I haven't check the other models but anything with hundreds of billions of params is likely to require datacenter hardware

rhatdan:
Sure but if a user wants to use the tool to pull a specific model via the name cogito-2.1 I don't see a reason not to pull it.

I am not sure there is much value in making it more difficult for users to pull models on platforms that could not run them, especially when we don't know the heuristics of what can and can not run. Also these numbers will change over time.

olliewalsh:
At least I would follow to recommendations and use a 2bit quant -  https://unsloth.ai/docs/models/tutorials/cogito-v2-how-to-run-locally

For novice/casual users with consumer-level hardware I think it's find to choose a sensible default. I don't see the point in adding shortnames for these datacenter class models though.

shortnames/shortnames.conf
```diff
@@ -32,67 +43,75 @@
   "deepseek-coder:6.7b" = "hf://TheBloke/deepseek-coder-6.7b-instruct-GGUF"
   "deepseek-coder:33b" = "hf://TheBloke/deepseek-coder-33b-instruct-GGUF"
   "deepseek-r1" = "hf://bartowski/DeepSeek-R1-GGUF"
-  "deepseek-v3" = "hf://unsloth/DeepSeek-V3-GGUF"
```
olliewalsh, line 35: I would not remove any existing shortnames. It is potentially a breaking change for users.

shortnames/shortnames.conf
```diff
@@ -1,25 +1,36 @@
...
 # - Models may have specific requirements (temperature, context length, etc.)
 #
+# TARGET USERS:
+# - Consumer and prosumer hardware (laptops, desktops, single-GPU workstations)
+#
+# EXCLUDED (do not add shortnames without maintainer review):
+# - Datacenter-class models (typically 100B+ parameters or requiring 100GB+ RAM/VRAM)
+#   Examples: 671B MoE, 675B, 550B Ultra, 480B coder, 397B, 235B VL, 120B+
+# - When a trending Ollama tag maps only to a datacenter model, omit the shortname
+# - Users with datacenter hardware can pull models directly via hf:// URLs
```
olliewalsh: The hf:// isn't necessary now, can just pull `<org>/<model>`

rhatdan: I don't think this is correct, if the user changes the default to ollama or oci, then the lack of hf:// will cause shortnames to blow up, I believe.

olliewalsh: Yeah, not suggest the shortnames drop the prefix

shortnames/shortnames.conf
```diff
@@ -1,98 +1,126 @@
...
   "deepseek-coder" = "hf://TheBloke/deepseek-coder-6.7b-instruct-GGUF"
   "deepseek-coder:6.7b" = "hf://TheBloke/deepseek-coder-6.7b-instruct-GGUF"
   "deepseek-coder:33b" = "hf://TheBloke/deepseek-coder-33b-instruct-GGUF"
   "deepseek-r1" = "hf://bartowski/DeepSeek-R1-GGUF"
-  "deepseek-v3" = "hf://unsloth/DeepSeek-V3-GGUF"
+  "deepseek-v3" = "hf://bartowski/deepseek-ai_DeepSeek-V3-0324-GGUF"
+  "deepseek-v3.1" = "hf://bartowski/deepseek-ai_DeepSeek-V3.1-GGUF"
+  "deepseek-v3.1:671b" = "hf://bartowski/deepseek-ai_DeepSeek-V3.1-GGUF"
+  "deepseek-v3.2" = "hf://unsloth/DeepSeek-V3.2-GGUF"
+  "deepseek-v4-flash" = "hf://antirez/deepseek-v4-gguf"
```
olliewalsh, line 44: model card says "This quants are specific for the DS4 inference engine."

shortnames/shortnames.conf
```diff
@@ -1,98 +1,126 @@
...
+  "deepseek-v4-pro" = "hf://teamblobfish/DeepSeek-V4-Pro-GGUF"
```
olliewalsh, line 45: "GGUF quantizations of [deepseek-ai/DeepSeek-V4-Pro](https://huggingface.co/deepseek-ai/DeepSeek-V4-Pro) for use with the V4-aware llama.cpp fork at [cchuter/llama.cpp @ feat/v4-port-cuda](https://github.com/cchuter/llama.cpp/tree/feat/v4-port-cuda)."

olliewalsh: @rhatdan I don't think delegating this to Cursor is working well

rhatdan: Ok Closing.

PR #2797, opened 2026-06-12
huggingface: use the HF_TOKEN env var if it's defined

mikebonnet:
Read the `HF_TOKEN` environment variable and include it in requests to huggingface if it's defined. Fall back to reading the cached token from the home directory.

Update CI to pass `HF_TOKEN` through to the `pytest` invocation, both locally and in Github Actions. Note that the `HF_TOKEN` `secret` will only be available if it has been defined in the repo where CI is currently running, and will not be available for PRs submitted from other repos.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,6 +97,8 @@ jobs:
         with: ${{ matrix.variant.uv-options }}
 
       - name: Run unit tests
+        env:
+          HF_TOKEN: ${{ secrets.HF_TOKEN }}
         run: |
           uv run -- make unit-tests
 
@@ -159,6 +161,8 @@ jobs:
            sudo apt-get upgrade -y podman crun
 
       - name: run e2e-tests
+        env:
+          HF_TOKEN: ${{ secrets.HF_TOKEN }}
         run: |
            make e2e-tests
 
@@ -219,6 +223,8 @@ jobs:
            sudo apt-get upgrade -y podman crun
 
       - name: run e2e-tests-nocontainer
+        env:
+          HF_TOKEN: ${{ secrets.HF_TOKEN }}
         run: |
            make e2e-tests-nocontainer
 
@@ -298,6 +304,8 @@ jobs:
            df -h
 
       - name: run e2e-tests-docker
+        env:
+          HF_TOKEN: ${{ secrets.HF_TOKEN }}
         run: |
            docker info
            make e2e-tests-docker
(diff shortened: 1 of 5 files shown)


ramalama/transports/huggingface.py
```diff
@@ -27,8 +27,12 @@
 """
 
 
-def huggingface_token():
-    """Return cached Hugging Face token if it exists otherwise None"""
+def huggingface_token() -> str | None:
+    """Return Hugging Face token from HF_TOKEN env var or cached token file, otherwise None"""
+    token = os.environ.get("HF_TOKEN")
+    if token:
+        return token
```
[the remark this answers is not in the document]
mikebonnet: Unnecessary, pass the token from the env var as it exists.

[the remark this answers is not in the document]
mikebonnet: Fixed.

.github/workflows/ci.yml
```diff
@@ -73,6 +73,8 @@ jobs:
       github.event_name == 'pull_request' ||
       github.ref == 'refs/heads/main' ||
       github.repository_owner != 'containers'
+    env:
+      HF_TOKEN: ${{ secrets.HF_TOKEN }}
```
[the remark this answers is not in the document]
mikebonnet: Fixed.

bmahabirbu (approved):

.github/workflows/ci.yml
```diff
@@ -97,6 +97,8 @@ jobs:
         with: ${{ matrix.variant.uv-options }}
 
       - name: Run unit tests
+        env:
+          HF_TOKEN: ${{ secrets.HF_TOKEN }}
```
olliewalsh, line 101: If the secret does not exist does this set it to an empty string?

mikebonnet: That's correct.

ramalama/transports/huggingface.py
```diff
@@ -27,8 +27,11 @@
 """
 
 
-def huggingface_token():
-    """Return cached Hugging Face token if it exists otherwise None"""
+def huggingface_token() -> str | None:
+    """Return Hugging Face token from HF_TOKEN env var or cached token file, otherwise None"""
+    if "HF_TOKEN" in os.environ:
```
olliewalsh, line 32: should this fall through to the token file when the env var is an empty string?

mikebonnet: CodeRabbit actually suggested the opposite, to make a defined-but-empty token authoritative and avoid an unexpected fallback to the cached token in that case. I could change the behavior if you feel strongly about it.

olliewalsh: No, don't feel strongly about it - the cached token code path just cannot be tested in CI if we are always setting this but not a major issue

olliewalsh (approved):
LGTM

PR #2803, opened 2026-06-15
Relax ollama deprecation now that is it pivoting back to llama.cpp

olliewalsh:
Remove the ollama model deprecation warning as ollama is pivoting back to llama.cpp.

However a number of ollama models remain incompatible so update the logic for unknown shortnames to raise a clear error instead of silently defaulting to ollama transport. Explicit ollama:// prefix and --transport ollama still work.


rhatdan (approved):
LGTM

Issue #2385, opened 2026-01-30
Expanded hosted provider support


ieaves:
* Anthropic
* Gemini
* Bedrock

PR #2808, opened 2026-06-20
feat: add LiteLLM as chat provider for 100+ LLM providers

RheagalFire:
## Summary

- Add LiteLLM as a chat provider, enabling access to 100+ LLM providers (Anthropic, Bedrock, Vertex, Gemini, Cohere, Mistral, etc.) through a single `litellm://` transport scheme
- The provider extends `OpenAICompletionsChatProvider` since the LiteLLM proxy speaks the OpenAI Chat Completions format natively
- Users point at their running LiteLLM proxy and can use any model it serves

## Motivation

Addresses #2385 ("Expanded hosted provider support") which requests Anthropic, Gemini, and Bedrock providers. LiteLLM solves all of these with one integration instead of individual provider implementations.

## Changes

- `ramalama/chat_providers/litellm.py` - new `LiteLLMChatProvider` extending `OpenAICompletionsChatProvider`
- `ramalama/chat_providers/api_providers.py` - registered `litellm` in `DEFAULT_PROVIDERS` and `PROVIDER_API_KEY_RESOLVERS`
- `ramalama/chat_providers/__init__.py` - exported `litellm` module
- `ramalama/config.py` - added `LitellmProviderConfig` with `api_key` and `base_url` fields
- `test/unit/providers/test_litellm_provider.py` - unit tests
- `test/unit/test_api_providers.py` - provider registry test

## Tests

```python
from ramalama.chat_providers.litellm import LiteLLMChatProvider
from ramalama.chat_providers.base import ChatRequestOptions
from ramalama.chat_utils import UserMessage

provider = LiteLLMChatProvider(base_url='http://localhost:4000', api_key='sk-test')
messages = [UserMessage(text='What is 2+2? Answer in one word.')]
options = ChatRequestOptions(model='anthropic/claude-sonnet-4-6', stream=False, max_tokens=10)
request = provider.create_request(messages, options)
# Send via urllib, parse response
```

```
Non-streaming: 'Four'
Streaming: '1, 2, 3.'
E2E PASSED
```

## Example usage

```bash
# Start a LiteLLM proxy with your providers
litellm --config config.yaml --port 4000

# Chat with any model through LiteLLM
ramalama run litellm://anthropic/claude-sonnet-4-6
ramalama run litellm://gpt-4o
```

## Risk / Compatibility

- Additive only. Existing OpenAI provider untouched.
- No new dependencies - the provider talks to a LiteLLM proxy via HTTP.
- Config fields are optional - defaults to `http://localhost:4000` if no base_url set.

### DIFF
--- a/ramalama/chat_providers/__init__.py
+++ b/ramalama/chat_providers/__init__.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from ramalama.chat_providers import api_providers, openai
+from ramalama.chat_providers import api_providers, litellm, openai
 from ramalama.chat_providers.base import (
     ChatProvider,
     ChatProviderError,
@@ -8,4 +8,12 @@ from ramalama.chat_providers.base import (
     ChatStreamEvent,
 )
 
-__all__ = ["ChatProvider", "ChatProviderError", "ChatRequestOptions", "ChatStreamEvent", "openai", "api_providers"]
+__all__ = [
+    "ChatProvider",
+    "ChatProviderError",
+    "ChatRequestOptions",
+    "ChatStreamEvent",
+    "openai",
+    "litellm",
+    "api_providers",
+]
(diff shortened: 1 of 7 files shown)


olliewalsh (changes requested):
I don't see what functionality this PR adds or why it is necessary. Since litellm provides an OpenAI compatible endpoint it should already work.

PR #2812, opened 2026-06-24
Add Pi coding agent to ramalama sandbox

bmahabirbu:
## Summary
- Add Pi (pi.dev) as a third sandbox agent alongside Goose and OpenCode
- Uses `docker.io/michaelwadman/pi-agent` container image with git, ripgrep, and fd pre-installed
- Integrates with the local model server via `pi-llama-cpp` extension for native llama.cpp support
- Includes `pi-web-access` extension for web search with curator workflow disabled for local model compatibility
- Configures `LLAMA_SERVER_URL` env var so pi auto-discovers models from the ramalama server

## Test plan
- [x] All 50 sandbox unit tests pass (`tox -- test/unit/test_sandbox_cmd.py`)
- [x] `ruff check` and `ruff format` pass
- [x] `ramalama sandbox --help` shows pi subcommand
- [x] `ramalama sandbox pi --help` shows all options including `--pi-image`
- [x] Manually tested `ramalama sandbox pi qwen3.5:9b` — pi launches, connects to model server, web search works

### DIFF
--- a/container-images/pi-agent/Containerfile
+++ b/container-images/pi-agent/Containerfile
@@ -1,0 +1,52 @@
+FROM quay.io/fedora/fedora:44
+
+ARG PI_VERSION=0.80.2
+ARG NODE_VERSION=26.4.0
+ARG PI_LLAMA_CPP_VERSION=0.7.2
+ARG PI_WEB_ACCESS_VERSION=0.12.0
+
+ENV TERM=xterm-256color \
+    HOME=/home/node \
+    GIT_CONFIG_GLOBAL=/home/node/.gitconfig
+
+RUN dnf -y --nodocs --setopt=install_weak_deps=false install \
+        bash \
+        ca-certificates \
+        curl \
+        fd-find \
+        file \
+        git-core \
+        jq \
+        libatomic \
+        openssh-clients \
+        ripgrep \
+        xz \
+    && dnf -y clean all \
+    && rm -rf /var/cache/dnf
+
+RUN curl -fsSL https://nodejs.org/dist/v${NODE_VERSION}/node-v${NODE_VERSION}-linux-x64.tar.xz \
+    | tar -xJf - --strip-components=1 -C /usr/local
+
+RUN mkdir -p /app /home/node/.pi /home/node/.ssh \
+    && chgrp -R 0 /app /home/node \
+    && chmod -R g=u /app /home/node \
+    && chmod 700 /home/node/.ssh
+
+RUN echo "Include /home/node/.ssh/config" >> /etc/ssh/ssh_config \
+    && echo "IdentityFile /home/node/.ssh/id_rsa" >> /etc/ssh/ssh_config \
+    && echo "IdentityFile /home/node/.ssh/id_ed25519" >> /etc/ssh/ssh_config \
+    && echo "IdentityFile /home/node/.ssh/id_ecdsa" >> /etc/ssh/ssh_config \
+    && echo "IdentityFile /home/node/.ssh/id_dsa" >> /etc/ssh/ssh_config \
+    && echo "UserKnownHostsFile /home/node/.ssh/known_hosts" >> /etc/ssh/ssh_config \
+    && echo "StrictHostKeyChecking accept-new" >> /etc/ssh/ssh_config
+
+RUN npm install -g --ignore-scripts @earendil-works/pi-coding-agent@${PI_VERSION} \
+    && pi install npm:pi-llama-cpp@${PI_LLAMA_CPP_VERSION} \
+    && pi install npm:pi-web-access@${PI_WEB_ACCESS_VERSION} \
+    && npm cache clean --force \
+    && echo '{"workflow":"none"}' > /home/node/.pi/web-search.json \
+    && rm -rf /tmp/* /root/.npm
+
+WORKDIR /app
+
+ENTRYPOINT [ "pi" ]
--- a/container-images/pi-agent/Containerfile
+++ b/container-images/pi-agent/Containerfile
@@ -11,9 +11,6 @@ ENV TERM=xterm-256color \
     GIT_CONFIG_GLOBAL=/home/node/.gitconfig
 
 RUN dnf -y --nodocs --setopt=install_weak_deps=false install \
-        bash \
-        ca-certificates \
-        curl \
         fd-find \
         file \
         git-core \
@@ -21,7 +18,6 @@ RUN dnf -y --nodocs --setopt=install_weak_deps=false install \
         libatomic \
         openssh-clients \
         ripgrep \
-        xz \
     && dnf -y clean all \
     && rm -rf /var/cache/dnf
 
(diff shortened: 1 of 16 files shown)


ramalama/sandbox.py
```diff
@@ -67,6 +67,20 @@ def add_sandbox_subparsers(subparsers: argparse._SubParsersAction, img_comp: Cal
     parser.set_defaults(func=run_sandbox_opencode)
     yield parser
 
+    parser = subparsers.add_parser("pi", help="run Pi in a sandbox, backed by a local AI Model")
+    if getattr(runtime, "_add_inference_args", None):
+        runtime._add_inference_args(parser, "serve")  # type: ignore[attr-defined]
+    parser.add_argument("MODEL", completer=model_comp)
+    parser.add_argument(
+        "--pi-image",
+        default="docker.io/michaelwadman/pi-agent:latest",
```
olliewalsh: Is that an official image?

bmahabirbu: sadly no there isnt an official image right now. Should we build it ourselves?

bmahabirbu: I'll work on adding a container

mikebonnet:
I'll need to add a new Konflux `component` before the image will be built. In the meantime, can you see about adding `PipelineRun` definitions for the new `pi-agent` component? You can use `.tekton/asahi/asahi-pull-request.yaml` and `.tekton/asahi/asahi-push.yaml` as good templates to start from.

bmahabirbu:
@mikebonnet thanks! Do u think we should build images for all the agents using fedora? The only reason I built this one was because before i was relying on a communty image. The other agents have official images so maybe we dont need to 

Also the ServiceAccountName should be build-pipeline-pi-agent

mikebonnet:
@bmahabirbu If there are specific features/dependencies missing from the upstream images we should talk about building them locally. Otherwise I don't see a lot of value in it.

Yes, that's the correct `serviceAccountName`, it just hadn't been deployed to the cluster yet. It has now, I'll re-trigger the `PipelineRun`.

ramalama/sandbox.py
olliewalsh: Do we know anything about that image? I don't think we should depend on an image from some random guy - it's begging for a supply chain attack

rhatdan: Definitely not, it would be better to build our own or get this from an official site.

container-images/pi-agent/Containerfile
```diff
@@ -0,0 +1,52 @@
...
+        git-core \
+        jq \
+        libatomic \
+        openssh-clients \
+        ripgrep \
+        xz \
+    && dnf -y clean all \
+    && rm -rf /var/cache/dnf
+
+RUN curl -fsSL https://nodejs.org/dist/v${NODE_VERSION}/node-v${NODE_VERSION}-linux-x64.tar.xz \
```
mikebonnet: The image is also being built for `arm64`, could you conditionalize the URL to download the correct version when building on that arch?

mikebonnet: Alternately, is there a reason to not install `node` via `rpm`?

mikebonnet:
You'll want to handle `DEFAULT_PI_IMAGE` in `ramalama/config.py`, similar to how `DEFAULT_TOOLS_IMAGE` is handled, including adding it to the `BaseConfig`. This will allow it to be overridden by a env var. I'll provide a snippet that will add that env var to the test environment so the built image can be tested in CI.

mikebonnet:
@bmahabirbu You can add these changes to your branch to add the `pi-agent` image to the e2e tests:

https://github.com/mikebonnet/ramalama/commit/3fcf24a3f7139ec11c209fb3d6c343090ec1296a

And don't forget to add `pi` as a parameter to `test/e2e/test_sandbox.py`.

container-images/pi-agent/Containerfile
bmahabirbu: Sadly I ran into a weird bug where the rpm version of node would segfault the pi agent? It was really weird so I had to use this method for installing node

bmahabirbu: will work on the arm64 change!

ramalama/sandbox.py
```diff
@@ -204,6 +225,36 @@ def add_env_options(self, args: OpenCodeArgsType) -> None:
...
+        self.engine.add_args(args.pi_image)
+        pi_args = ["--provider", f"llama-server=http://localhost:{args.port}", "--model", self.model_name]
+        if args.ARGS:
+            pi_args += ["-p", " ".join(args.ARGS)]
+        elif not self.engine.use_tty():
+            pi_args += ["-p", "-"]
+        self.engine.add(pi_args)
+
+    def add_env_options(self, args: PiArgsType) -> None:
+        self.engine.add_env_option(f"LLAMA_SERVER_URL=http://localhost:{args.port}")
```
mikebonnet, line 261: Are both `--provider llama-server=` and `LLAMA_SERVER_URL` required?

ramalama/sandbox.py
```diff
@@ -204,6 +225,36 @@ def add_env_options(self, args: OpenCodeArgsType) -> None:
...
+        super().__init__(args, model_name)
+        self.engine.add_name(f"pi-{args.name}")  # type: ignore[attr-defined]
+        self.add_env_options(args)
+        self.engine.add_workdir(args)
+        self.engine.add_args(args.pi_image)
+        pi_args = ["--provider", f"llama-server=http://localhost:{args.port}", "--model", self.model_name]
+        if args.ARGS:
+            pi_args += ["-p", " ".join(args.ARGS)]
+        elif not self.engine.use_tty():
+            pi_args += ["-p", "-"]
```
mikebonnet: It looks like `pi` doesn't accept `-` as an argument to `-p` to mean "read from stdin". You might need to do something clever like `xargs` to convert stdin into cmdline args.

container-images/pi-agent/Containerfile
```diff
@@ -0,0 +1,58 @@
...
+        ca-certificates \
+        curl \
+        fd-find \
+        file \
+        git-core \
+        jq \
+        libatomic \
+        openssh-clients \
+        ripgrep \
+        xz \
```
mikebonnet: `bash`, `ca-certificates`, `curl`, and `xz` are included in the base `fedora` image and don't need to be listed here.

ramalama/sandbox.py
mikebonnet: Oh, actually looks like `pi` handles `stdin` correctly already, you can remove the `-` (and in fact can remove the `use_tty()` check altogether, it switches to `print` mode automatically if not on a tty:

https://github.com/earendil-works/pi/issues/708

container-images/pi-agent/Containerfile
bmahabirbu: Your right thanks these were artifcats from trying to do a ubi10 micro image but the fedora 44 one was only lke 50mb more

ramalama/sandbox.py
bmahabirbu: They do different things so the --provider is for pi and the url is for ramalama i believe but i think i can refractor it to make it easier to understand

mikebonnet (approved):
This looks great, nice addition!

ramalama/sandbox.py
rhatdan: You can always use /dev/stdin also if a file is acceptable and - is not

Issue #2819, opened 2026-06-29
"bad CPU type in executable" when running in Intel mac


mandarvaze:
### Issue Description

Installer worked. No errors.
But when I try to run the command, I get this error

```sh
$ ramalama --version
zsh: bad CPU type in executable: ramalama
$ ramalama version
zsh: bad CPU type in executable: ramalama
```
Obviously, this is coming from the fact that I'm using Intel mac, while binary is for Apple Silicon/ARM
Search yielded same error for llama.cpp

### Steps to reproduce the issue

1. Install on Intel mac (macOS 15.7.7 - Sequoia)
2. Run `ramalama` from the terminal

### Describe the results you received

zsh: bad CPU type in executable: ramalama

### Describe the results you expected

Prints version

### ramalama info output

```yaml
Does not run
```

### Upstream Latest Release

No

### Additional information

Intel mac (macOS 15.7.7 - Sequoia)

rhatdan:
Yes RamaLama currently doe not support intel macs. Not sure there is any GPUs to take advantage of.

mandarvaze:
https://github.com/containers/ramalama/blob/main/docs/MACOS_INSTALL.md#system-requirements says `Intel or Apple Silicon (M1/M2/M3) processor`

Maybe documentation needs update ?
Also update the installer - detect Intel CPU and refuse to install

rhatdan:
Interested in opening a PR?

mandarvaze:
> Yes RamaLama currently doe not support intel macs. Not sure there is any GPUs to take advantage of.

What about 

> - **Intel GPUs**: vulkan (Linux/macOS) or sycl (Windows); openvino available as explicit option

in docs/options/backend.md

mandarvaze:
> Interested in opening a PR?

If you confirm here that Intel macs are definitely **not** supported, I can create a PR for documentation.
I still think refusing to install on Intel macs is better/fool-proof option, but I do not have enough knowledge to update the installer.

olliewalsh:
See https://pyinstaller.org/en/stable/feature-notes.html#macos-multi-arch-support

olliewalsh:
> Not sure there is any GPUs to take advantage of.

The Mac Pro 2019 had AMD GPUs. IIUC ROCm isn't supported, metal possibly works,  Vulkan probably works. Without hardware to play with though it's very difficult to officially support.

PR #2821, opened 2026-06-29
Document Intel Mac CPU support and clarify platform limits

rhatdan:
Intel Macs already work with the default llama.cpp runtime (containers or --nocontainer) but MLX and GPU acceleration remain Apple Silicon only. Update macOS docs, README hardware table, install.sh mlx-lm skip on x86_64, and add an Intel Mac MLX rejection unit test.

Fixes: https://github.com/containers/ramalama/pull/2820

### DIFF
--- a/docs/MACOS_INSTALL.md
+++ b/docs/MACOS_INSTALL.md
@@ -2,15 +2,101 @@
 
 This guide covers the different ways to install RamaLama on macOS.
 
-## GPU acceleration options
+## Platform support
+
+RamaLama runs on both **Intel** and **Apple Silicon** Macs. What differs is
+which runtimes and accelerators are available:
+
+| Platform | Default runtime | GPU acceleration |
+| :------- | :-------------- | :--------------- |
+| Apple Silicon (M1/M2/M3+) | llama.cpp (container, native Metal, or MLX) | Native llama.cpp Metal (`--nocontainer`), libkrun GPU passthrough in Podman, or MLX |
+| Intel Mac | llama.cpp (container or `--nocontainer`) | Not supported (CPU only) |
+
+On Intel Macs, do **not** use `--runtime=mlx` or install `mlx-lm`; MLX requires
+Apple Silicon hardware.
+
+## Intel Mac (CPU only)
+
+Intel Macs use the default **llama.cpp** runtime with CPU inference only.
+There is no GPU passthrough or MLX support on this hardware.
+
+### Option A: Containers (recommended)
+
+Install Podman or Docker Desktop, then use RamaLama normally.
+
+**Podman:**
+
+```bash
+brew install podman
+podman machine init && podman machine start
+```
+
+**Docker Desktop** (alternative; `brew install docker` installs only the CLI and is
+not sufficient—you need the Desktop app):
+
+```bash
+brew install --cask docker
+open -a Docker   # wait until Docker Desktop is running
+```
+
+Then pull and run a model:
+
+```bash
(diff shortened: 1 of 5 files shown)


rhatdan:
@mandarvaze I asked Claude to develop a PR to add support for intel macs, could you check if this fixes your issues on the MAC?

bmahabirbu (approved):
LGTM

mandarvaze:
> @mandarvaze I asked Claude to develop a PR to add support for intel macs, could you check if this fixes your issues on the MAC?

I downloaded the installer from the URL I found on https://github.com/containers/ramalama/actions/runs/28384410088/job/84095198616

I still get the same error when I run `ramalama version`

rhatdan:
@mandarvaze Could you output 

ramalama --debug version

mandarvaze:
> @mandarvaze Could you output
> 
> ramalama --debug version

$ ramalama --debug version
zsh: bad CPU type in executable: ramalama

@rhatdan I wonder if I downloaded correct installer 🤷‍♂️
Can you share the correct link ?

rhatdan:
ramalama should be a python script. which is executing:

/usr/bin/env python3


There are many ways to get ramalama installed onto your system. This link should show you how to install on a MAC.

https://github.com/containers/ramalama/blob/main/docs/MACOS_INSTALL.md

mandarvaze:
> ramalama should be a python script. which is executing:
> There are many ways to get ramalama installed onto your system. This link should show you how to install on a MAC.

Yess!

I was able to get it (at least `ramalama version`) working 🎉 via the following command:

`uv tool install --python python3.14 ramalama`

Thanks!

rhatdan:
Are you able to run a model?

mandarvaze:
> Are you able to run a model?

Yess!

I tried `ramalama run ollama://llama3.2:latest` and it worked! 🎉

rhatdan:
@olliewalsh @mikebonnet @bmahabirbu PTAL

olliewalsh:
I would lean toward not supporting this unless we can get hardware or somebody with hardware volunteers to maintain support.

The issue with the installer CPU type should be fixable - https://github.com/containers/ramalama/issues/2819#issuecomment-4866118989

Also the Vulkan or Metal llama.cpp backend probably work on the Intel based Mac - https://github.com/containers/ramalama/issues/2819#issuecomment-4866449732.

We just don't have any hardware to validate it.

rhatdan:
Since Intel MAC is slowly disappearing, I don't think we should overly invest in it.  But should we document it and note that it is not supported?

rhatdan:
@olliewalsh Better?

docs/MACOS_INSTALL.md
```diff
@@ -2,12 +2,75 @@
...
 
-## GPU acceleration options
+## Platform support
+
+RamaLama runs on both **Intel** and **Apple Silicon** Macs. What differs is
+which runtimes and accelerators are available:
+
+| Platform | Default runtime | GPU acceleration |
+| :------- | :-------------- | :--------------- |
+| Apple Silicon (M1/M2/M3+) | llama.cpp (container) or MLX (`--nocontainer`) | MLX native, or libkrun GPU passthrough in Podman |
```
olliewalsh: also have the native llama.cpp metal backend

olliewalsh:
> @olliewalsh Better?

Just noticed we don't mention llama.cpp metal, that arguably the best option on macos, mlx-lm doesn't seem to be very active these days.

PR #2822, opened 2026-06-30
Add router mode support to ramalama sandbox

bmahabirbu:
## Summary

- Allow sandbox subcommands (pi, goose, opencode) to run with zero or multiple models using llama.cpp router mode
  - `ramalama sandbox pi` (no model) — serves all GGUF models from the store
  - `ramalama sandbox pi model1 model2` — serves specified models
  - `ramalama sandbox pi model1` — unchanged single-model behavior
- Refactor `_serve_router` into `_build_router_engine` + `serve_router_nonblocking` for sandbox reuse
- Convert positional `ARGS` to `--prompt` named option (required since MODEL is now `nargs="*"`)
- Add `--models-max` option to sandbox subparsers
- Fix health check to handle both `models`/`data` keys and `name`/`id` fields in `/models` response
- After server is healthy, query `/v1/models` to discover available model IDs and configure agents accordingly

https://github.com/user-attachments/assets/0a9205ea-268e-451d-8877-6b84ee140727

This works best with the pi agent with the build in support for llama.cpp new router mode. if you want to test with this use container_build.sh build pi-agent so that u can have the container to test

right now the container isnt built to test yet

Signed-off-by: Brian <brian@fedora>

### DIFF
--- a/container-images/pi-agent/Containerfile
+++ b/container-images/pi-agent/Containerfile
@@ -2,7 +2,6 @@ FROM quay.io/fedora/fedora:44
 
 ARG PI_VERSION=0.80.7
 ARG NODE_VERSION=26.5.0
-ARG PI_LLAMA_CPP_VERSION=0.8.1
 ARG PI_WEB_ACCESS_VERSION=0.13.0
 ARG TARGETARCH
 
@@ -43,7 +42,7 @@ RUN echo "Include /home/node/.ssh/config" >> /etc/ssh/ssh_config \
     && echo "StrictHostKeyChecking accept-new" >> /etc/ssh/ssh_config
 
 RUN npm install -g --ignore-scripts @earendil-works/pi-coding-agent@${PI_VERSION} \
-    && pi install npm:pi-llama-cpp@${PI_LLAMA_CPP_VERSION} \
+    && pi install npm:pi-llama-server@1.1.0 \
     && pi install npm:pi-web-access@${PI_WEB_ACCESS_VERSION} \
     && npm cache clean --force \
     && echo '{"workflow":"none"}' > /home/node/.pi/web-search.json \
--- a/container-images/pi-agent/Containerfile
+++ b/container-images/pi-agent/Containerfile
@@ -2,6 +2,7 @@ FROM quay.io/fedora/fedora:44
 
 ARG PI_VERSION=0.80.7
 ARG NODE_VERSION=26.5.0
+ARG PI_LLAMA_SERVER_VERSION=1.1.0
 ARG PI_WEB_ACCESS_VERSION=0.13.0
 ARG TARGETARCH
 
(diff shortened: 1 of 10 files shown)


bmahabirbu:
@mikebonnet had to change ARGS to --prompt for sandbox wondering if thats ok open to suggestion how how to better fix the 2 greedy arguments which are now serve then args

mikebonnet:
> @mikebonnet had to change ARGS to --prompt for sandbox wondering if thats ok open to suggestion how how to better fix the 2 greedy arguments which are now serve then args

Sure, that sounds reasonable. It's probably not necessary to make `--prompt` greedy, the whole prompt could be passed as a single argument, which seems more intuitive.

You'll need to update `test/e2e/test_sandbox.py` to use the new argument format.

bmahabirbu:
@mikebonnet @olliewalsh ready-to-test if you have the time! @rhatdan too I think this one is pretty cool cause it has the baked in model spawing loading + pi already does the chat summarizations as well

bmahabirbu:
look at https://github.com/am17an/pi-llama-server/issues/2 

resolved so this pr is ready-to-test @mikebonnet @olliewalsh! Thanks @am17an for the changes on the pi extension

container-images/pi-agent/Containerfile
```diff
@@ -43,7 +42,7 @@ RUN echo "Include /home/node/.ssh/config" >> /etc/ssh/ssh_config \
     && echo "StrictHostKeyChecking accept-new" >> /etc/ssh/ssh_config
 
 RUN npm install -g --ignore-scripts @earendil-works/pi-coding-agent@${PI_VERSION} \
-    && pi install npm:pi-llama-cpp@${PI_LLAMA_CPP_VERSION} \
+    && pi install npm:pi-llama-server@1.1.0 \
```
mikebonnet: It would be nice to define the `pi-llama-server` version as an `ARG` for consistency with the rest of the component versions.

ramalama/sandbox.py
```diff
@@ -178,7 +192,8 @@ def add_env_options(self, args: GooseArgsType) -> None:
         self.engine.add_env_option("GOOSE_PROVIDER=openai")
         self.engine.add_env_option(f"OPENAI_HOST=http://localhost:{args.port}")
         self.engine.add_env_option("OPENAI_API_KEY=ramalama")
-        self.engine.add_env_option(f"GOOSE_MODEL={self.model_name}")
+        if self.model_name:
```
mikebonnet: `model_name` is always defined and non-empty (except in the `dryrun` case). I'd suggest removing the conditionals around `model_name` to avoid confusion.

ramalama/sandbox.py
```diff
@@ -296,10 +330,56 @@ def run_sandbox(args: SandboxEngineArgsType, agent_cls: type[Agent]):
...
+    runtime = get_runtime(ActiveConfig().runtime)
+
+    serve_router = getattr(runtime, "serve_router_nonblocking", None)
+    if serve_router is None:
+        raise ValueError("Router mode (zero or multiple models) is only supported with the llama.cpp runtime.")
+
+    serve_router(cast(argparse.Namespace, args))
+
+    if args.dryrun:
+        agent = agent_cls(args, "")
```
mikebonnet: The second arg here should be a non-empty string (`dummy` or something) to match the non-`dryrun` behavior.

ramalama/sandbox.py
```diff
@@ -296,10 +330,56 @@ def run_sandbox(args: SandboxEngineArgsType, agent_cls: type[Agent]):
...
+        return
+
+    try:
+        wait_for_healthy(args, partial(is_healthy))
+
+        if args.port is None:
+            raise ValueError("Router mode requires a resolved serving port")
+        model_ids = _query_router_models(args.port)
+        args.router_model_ids = model_ids  # type: ignore[attr-defined]
+        first_model = model_ids[0] if model_ids else ""
```
mikebonnet: This could be an `assert` that `model_ids` is non-empty (as enforced by `_build_router_engine()`), then an unconditional reference to `model_ids[0]`.

bmahabirbu:
@mikebonnet updated!

mikebonnet (approved):
This is working great for me, really nice work!

PR #2824, opened 2026-06-30
Fix import order in bin/ramalama to use local package

bmahabirbu:
## Summary
- Moved the `from ramalama.config import ActiveConfig, ensure_tmpdir` import to after `sys.path.insert(0, './')` so the local development version is resolved instead of a system-installed package that may not have `ensure_tmpdir` yet.

## Test plan
- [ ] Run `bin/ramalama list` from the repo root and confirm no `ImportError`
- [ ] Verify `bin/ramalama` still works when installed via pip (no regression)

### DIFF
--- a/bin/ramalama
+++ b/bin/ramalama
@@ -2,8 +2,6 @@
 
 import sys
 
-from ramalama.config import ActiveConfig, ensure_tmpdir
-
 
 def main():
     if not sys.stdout.isatty():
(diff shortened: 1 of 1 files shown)


bmahabirbu:
I noticed bin/ramalama wasnt working anymore I think this will fix it. Tested on mac still need some help testing on windows/linux

just clone the repo cd into it and use bin/ramalama it be used normally and not have an error

mikebonnet (approved):
LGTM!

PR #2826, opened 2026-07-01
Add HF_ENDPOINT environment variable support to specify Hugging Face base URL

olliewalsh:
Allow users to override the default Hugging Face API endpoint (https://huggingface.co) via the HF_ENDPOINT environment variable, matching the convention used by the official huggingface_hub library. This enables use of HF mirrors and proxies.

### DIFF
--- a/ramalama/transports/huggingface.py
+++ b/ramalama/transports/huggingface.py
@@ -27,6 +27,11 @@ sudo dnf install python3-huggingface-hub
 """
 
 
+def huggingface_endpoint() -> str:
+    """Return Hugging Face base URL from HF_ENDPOINT env var, defaulting to https://huggingface.co"""
+    return os.environ.get("HF_ENDPOINT", HuggingfaceRepository.REGISTRY_URL).rstrip("/")
+
+
 def huggingface_token() -> str | None:
     """Return Hugging Face token from HF_TOKEN env var or cached token file, otherwise None"""
     if "HF_TOKEN" in os.environ:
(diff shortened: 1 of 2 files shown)


ramalama/transports/huggingface.py
```diff
@@ -27,6 +27,11 @@
 """
 
 
+def huggingface_endpoint() -> str:
+    """Return Hugging Face base URL from HF_ENDPOINT env var, defaulting to https://huggingface.co"""
+    return os.environ.get("HF_ENDPOINT", "https://huggingface.co").rstrip("/")
```
rhatdan: Why not use HuggingfaceRepository.REGISTRY_URL here?

olliewalsh: was to avoid a circular reference but I _think_ it would work, will test it out

ramalama/transports/huggingface.py
```diff
@@ -87,7 +92,7 @@ def fetch_repo_manifest(repo_name: str, tag: str = "latest"):
 def fetch_repo_files(repo_name: str, revision: str = "main"):
     """Fetch the list of files in a HuggingFace repository using the Files API with pagination support."""
     token = huggingface_token()
-    base_api_url = f"https://huggingface.co/api/models/{repo_name}/tree/{revision}"
+    base_api_url = f"{huggingface_endpoint()}/api/models/{repo_name}/tree/{revision}"
```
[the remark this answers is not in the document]
olliewalsh: No, the mirror would be generating the headers

olliewalsh:
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> ⚠️ Outside diff range comments (1)
> > ramalama/transports/huggingface.py (1)> `154-155`: _🎯 Functional Correctness_ | _🟡 Minor_ | _⚡ Quick win_
> > > **Update the Hugging Face URL matcher to use `huggingface_endpoint()`.** `ramalama/transports/url.py:80` still reads `HuggingfaceRepository.REGISTRY_URL`, so custom `HF_ENDPOINT` hosts can be misparsed. `Huggingface.REGISTRY_URL` at `ramalama/transports/huggingface.py:266` looks unused and can be removed.

No, the Hugging Face URL provided by the user does not change when they are using a proxy/mirror.

mikebonnet (approved):
LGTM

PR #2830, opened 2026-07-06
Add ramalama models command to list server-loaded models

rhatdan:
Query a running inference server's /v1/models or llama.cpp /models endpoints so users can see which models are currently served, distinct from ramalama list which shows locally downloaded models.

### DIFF
--- a/ramalama/cli.py
+++ b/ramalama/cli.py
@@ -348,6 +348,7 @@ def configure_subcommands(parser):
     list_parser(subparsers)
     login_parser(subparsers)
     logout_parser(subparsers)
+    models_parser(subparsers)
     pull_parser(subparsers)
     push_parser(subparsers)
     rm_parser(subparsers)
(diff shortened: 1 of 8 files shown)


ramalama/model_server.py
```diff
@@ -0,0 +1,98 @@
...
+
+def normalize_server_url(url: str) -> tuple[str, str]:
+    """Return (server_root, openai_base) for the given server URL."""
+    url = url.rstrip("/")
+    if url.endswith("/v1"):
+        return url[:-3], url
+    return url, f"{url}/v1"
+
+
+def parse_llama_cpp_models(payload: Any) -> list[str]:
```
olliewalsh: Can we move this to the llama.cpp plugin?

ramalama/model_server.py
```diff
@@ -0,0 +1,98 @@
...
+    except urllib_error.URLError as exc:
+        raise ModelServerError(f"Could not connect to model server at {url}: {exc.reason}") from exc
+
+
+def list_server_models(url: str, api_key: Optional[str] = None) -> list[str]:
+    """Return model identifiers exposed by a running inference server."""
+    server_root, openai_base = normalize_server_url(url)
+
+    try:
+        return OpenAICompletionsChatProvider(openai_base, api_key).list_models()
```
olliewalsh, line 57: Does this work with mlx?

rhatdan: Cursor says:
```
Yes for MLX via the primary path: mlx_lm.server exposes an OpenAI-compatible /v1/models endpoint, which is what we query first.

The llama.cpp-native /models fallback only runs when /v1/models fails (e.g. llama.cpp router mode). MLX servers don't use that fallback in practice.

Documented this in ramalama-models(7).
```

ramalama/model_server.py
```diff
@@ -0,0 +1,80 @@
+from __future__ import annotations
+
+import json
+from typing import Any, Optional
+from urllib import error as urllib_error
+from urllib import request as urllib_request
+
+from ramalama.chat_providers.base import ChatProviderError
+from ramalama.chat_providers.openai import OpenAICompletionsChatProvider
+from ramalama.plugins.runtimes.inference.llama_cpp import parse_models_payload
```
olliewalsh, line 10: nit: this is bypassing the abstraction layer of the plugin system interface. I have some ideas to address it but can wait for a follow-up PR

ramalama/cli.py
```diff
@@ -595,6 +596,61 @@ def list_parser(subparsers):
     parser.set_defaults(func=list_cli)
 
 
+def _format_model_server_host(host: str) -> str:
+    host = host.strip("[]")
```
olliewalsh, line 600: nit: the looks very familiar, I think we do this in multiple places -  would be good to refactor it all into a single method in a follow-up PR

olliewalsh (approved):
LGTM

PR #2848, opened 2026-07-17
Add speculative decoding CLI args 

olliewalsh:
Enable MTP and other speculative decoding modes from the CLI by exposing llama.cpp's --spec-type and the most commonly tuned --spec-draft-* knobs.

### DIFF
--- a/ramalama/plugins/runtimes/inference/llama_cpp.py
+++ b/ramalama/plugins/runtimes/inference/llama_cpp.py
@@ -86,6 +86,10 @@ class LlamaCppConfig:
     gguf_quantization_mode: GGUF_QUANTIZATION_MODES = DEFAULT_GGUF_QUANTIZATION_MODE  # type: ignore[assignment]
     ngl: Optional[str] = None
     ncmoe: Optional[int] = None
+    spec_type: Optional[str] = None
+    spec_draft_n_max: Optional[int] = None
+    spec_draft_n_min: Optional[int] = None
+    spec_draft_p_min: Optional[float] = None
     temp: float = 0.8
     thinking: Optional[bool] = None
     threads: int = field(default_factory=_default_threads)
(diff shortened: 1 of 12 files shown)


mikebonnet (approved):
Looks good!

ramalama/plugins/runtimes/inference/llama_cpp.py
```diff
@@ -441,6 +453,37 @@ def _add_inference_args(self, parser: "argparse.ArgumentParser", command: str) -
...
+                dest="spec_type",
+                default=None,
+                help="speculative decoding type(s), comma-separated (e.g. draft-mtp, draft-simple, ngram-simple)",
+                completer=suppressCompleter,
+            )
+            parser.add_argument(
+                "--spec-draft-n-max",
+                dest="spec_draft_n_max",
+                type=int,
+                default=None,
```
rhatdan, line 467: Why not set the default here to 3, or better yet allow this to be specified in ramalama.conf?

ramalama/plugins/runtimes/inference/llama_cpp.py
```diff
@@ -441,6 +453,37 @@ def _add_inference_args(self, parser: "argparse.ArgumentParser", command: str) -
...
+                help="max number of tokens to draft for speculative decoding (default: 3)",
+                completer=suppressCompleter,
+            )
+            parser.add_argument(
+                "--spec-draft-n-min",
+                dest="spec_draft_n_min",
+                type=int,
+                default=None,
```
rhatdan, line 475: Ditto.

ramalama/plugins/runtimes/inference/llama_cpp.py
```diff
@@ -441,6 +453,37 @@ def _add_inference_args(self, parser: "argparse.ArgumentParser", command: str) -
...
+                help="min number of draft tokens for speculative decoding (default: 0)",
+                completer=suppressCompleter,
+            )
+            parser.add_argument(
+                "--spec-draft-p-min",
+                dest="spec_draft_p_min",
+                type=float,
+                default=None,
+                help="min speculative decoding probability (default: 0.0)",
```
rhatdan, line 484: Ditto.

ramalama/plugins/runtimes/inference/llama_cpp.py
olliewalsh: Setting in ramalama.conf should already work, everything in LlamaCppConfig can be set this way. Defaults to None just allows it to be omitted by default, since it would just add noise if not using speculative decoding

PR #2849, opened 2026-07-17
Auto-pull draft model in run/serve handlers

olliewalsh:
When --model-draft is specified, ensure_model_exists() was only called on the primary model, requiring users to manually pull the draft model beforehand.
Call ensure_model_exists() on the draft model too so it is automatically pulled like the primary model.


mikebonnet (approved):
LGTM

PR #2868, opened 2026-07-27
fix(container): install ffmpeg in Fedora-based llama.cpp images (#2867)

### LINKED ISSUE #2867: Missing FFmpeg dependency in the Llama.cpp container
I appears that FFmpeg is used by Llama.cpp for some inputs to multi-modal models.

When I try to input a video file, it errors out with a message of:
```
E probe: failed to launch ffprobe
E mtmd_helper_video_init_from_buf: ffprobe failed on buffer (is ffprobe in PATH?)
E mtmd_helper_bitmap_init_from_buf: failed to decode buffer as either image/audio/video
```
When testing by using the official Llama.cpp container and manually loading the same model + mmproj, there error does not occur and looking at their repo it seems that ffmpeg is included in the dockerfile.

Looking in this repo, it seems like vLLM has ffmpeg included in it's build process but not Llama.cpp

Anai-Guo:
## Problem

`llama.cpp`'s `mtmd` helpers shell out to the **`ffprobe` / `ffmpeg` binaries** to
decode audio and video inputs for multimodal models. None of the RamaLama
llama.cpp images install them, so passing a video file fails with:

```
E probe: failed to launch ffprobe
E mtmd_helper_video_init_from_buf: ffprobe failed on buffer (is ffprobe in PATH?)
E mtmd_helper_bitmap_init_from_buf: failed to decode buffer as either image/audio/video
```

The same model + mmproj works in the upstream llama.cpp container, which does
install ffmpeg. In this repo ffmpeg is only pulled in by the **vLLM** build
(`container-images/scripts/build-vllm.sh:25`), never by `build_llama.sh`.

Fixes #2867

## Change

Add `ffmpeg-free` to the runtime package set in `dnf_install_runtime_deps()`
for Fedora-based images:

```bash
if [ "${ID}" = "fedora" ]; then
  runtime_pkgs+=(ffmpeg-free)
fi
```

`ffmpeg-free` is the ffmpeg build shipped in the **stock Fedora repos** (the
fully-featured `ffmpeg` package only exists in RPM Fusion, which these builds
do not enable). It provides `/usr/bin/ffmpeg` and `/usr/bin/ffprobe`, which is
exactly what `mtmd` looks for.

This covers the images whose runtime stage is Fedora-based — `ramalama`
(the default Vulkan image), `rocm*`, `asahi`.

## Deliberately out of scope

- **RHEL/UBI-based images** (e.g. `cuda`, built on `nvidia/cuda:*-runtime-ubi9`):
  there `ffmpeg-free` lives in EPEL, and EPEL is only enabled in the *build*
  stage (`dnf_install`), not in the runtime stage. Adding it would mean calling
  `dnf_install_epel` + `rm_non_ubi_repos` inside `dnf_install_runtime_deps` too.
  Happy to extend the PR that way if you want the CUDA image covered — I left it
  out because I can't build/verify those images locally and didn't want to risk
  the runtime stage of an image that currently builds fine.
- **openEuler (`cann`) and other non-Fedora bases** are untouched for the same
  reason — the guard means their package set is byte-for-byte what it is today.

## Note on image size

`ffmpeg-free` plus its `libav*` dependencies adds roughly 40–50 MB to the
Fedora-based images. That is noise next to the CUDA/ROCm images but is a real
bump for the plain `ramalama` image; if you'd rather gate it behind a build arg
or restrict it to specific containerfiles, say the word and I'll rework it.

## Verification

Static only — I don't have a builder for these images. `bash -n` passes on the
modified script, and the change is a single guarded `runtime_pkgs+=(...)` that
follows the same `[ "${ID}" = "fedora" ]` pattern already used by
`dnf_install_mesa()` in the same file.


Anai-Guo:
The E2E failures (`sudo: ./container-images/scripts/build_llama.sh: command not found`) were my fault, not the change itself — the commit rewrote the script through the GitHub API, which dropped its mode from `100755` to `100644`, so the workflow could no longer execute it. Restored the executable bit in `39e3bda`; the file content is unchanged (same blob SHA).

PR #2872, opened 2026-07-29
add hosts entry to fix RAG and sandbox under Docker

mikebonnet:
Add a `host.docker.internal` hosts entry when running under Docker, enabling the container to connect to services running on the host. This matches the behavior of Docker Desktop and Podman. Only add it if `--network` is not specified, otherwise assume the user has configured the custom network as desired.

Enable sandbox tests to run under Docker.

### DIFF
--- a/test/e2e/test_sandbox.py
+++ b/test/e2e/test_sandbox.py
@@ -8,7 +8,6 @@ import pytest
 import requests
 
 from test.conftest import (
-    skip_if_docker,
     skip_if_no_container,
     skip_if_not_windows,
     skip_if_ppc64le,
@@ -265,7 +264,6 @@ def test_sandbox_run_stdin(sandbox_ctx, tmp_path, agent):
 
 @pytest.mark.e2e
 @pytest.mark.slow
-@skip_if_docker
 @skip_if_no_container
 @skip_if_ppc64le
 @skip_if_s390x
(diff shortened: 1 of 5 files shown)


bmahabirbu (approved):
Lgtm codewise but I dont have docker to test atm

mikebonnet:
Thanks! Ran the `e2e` and `slow` tests under Docker locally, and confirmed all tests are passing.

test/e2e/test_sandbox.py
```diff
@@ -310,7 +308,7 @@ def test_sandbox_using_url(sandbox_ctx, caplog, agent):
                 "sandbox",
                 agent,
                 "--url",
-                f"http://host.containers.internal:{container_port}",
+                f"http://localhost:{container_port}",
```
olliewalsh, line 311: why has this been changed?

mikebonnet: So that `run_sandbox` can convert it to `host.containers.internal` or `host.docker.internal` as appropriate.

mikebonnet: https://github.com/containers/ramalama/blob/main/ramalama/sandbox.py#L343

olliewalsh: ah, ok, that probably needs handle loopback IPs too but not related to this PR

olliewalsh (approved):
LGTM

PR #2885, opened 2026-08-05
add ffmpeg as a runtime dependency for all container images

### LINKED ISSUE #2867: Missing FFmpeg dependency in the Llama.cpp container

olliewalsh:
ffmpeg is needed to decode user input video to raw rgb24 frames.
Installs ffmpeg-free on Fedora and UBI9 (via EPEL + CentOS Stream repos), and ffmpeg on openEuler (cann) and Ubuntu (musa).

Fixes: #2867

### DIFF
--- a/container-images/cann/Containerfile
+++ b/container-images/cann/Containerfile
@@ -10,6 +10,7 @@ WORKDIR /src/
 RUN ./build_llama.sh cann
 
 FROM quay.io/ascend/${ASCEND_VERSION}
+RUN dnf install -y --setopt=install_weak_deps=false ffmpeg && dnf -y clean all
 RUN --mount=type=bind,from=builder,source=/tmp/install,target=/tmp/install,relabel=private \
     cp -a /tmp/install/bin/ /usr/ && \
     cp -a /tmp/install/lib64/*.so* /usr/lib64/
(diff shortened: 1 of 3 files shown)


olliewalsh:
@mikebonnet PTAL

mikebonnet (approved):
LGTM

rhatdan (approved):
LGTM, Remove draft and merge.

bmahabirbu (approved):
LGTM

So now with this video input is enabled by default?

olliewalsh:
> LGTM
> 
> So now with this video input is enabled by default?

Video input was already enabled (e.g in the llama.cpp web ui) if the model supported it but would fail with a runtime error. It now works, although I've only tried the `cuda` and `ramalama` images.

PR #2887, opened 2026-08-11
Update podman-machine-list call to work on MacOS with podman 6

nickjwhite:
As of v6.0.0, Podman no longer supports the `--all-providers` flag for the `podman machine list` command.


mikebonnet:
@nickjwhite This behavior change should be conditional based on a version check: `podman --version`

nickjwhite:
> @nickjwhite This behavior change should be conditional based on a version check: `podman --version`

Good point, done.

bmahabirbu (commented):
LGTM codewise

I do have an intel mac and a apple silicon one so i can test

olliewalsh:
> LGTM codewise
> 
> I do have an intel mac and a apple silicon one so i can test

The macos e2e CI jobs install podman 6, can we add test coverage there?

nickjwhite:
> The macos e2e CI jobs install podman 6, can we add test coverage there?

Hi Ollie, thanks for taking a look. I hadn't looked at the e2e tests before, I'd thought the existing coverage in `e2e/test_list.py` would cover this functionality. That said, if that were the case it should have been failing before this patch, and it wasn't - I guess a few more asserts to cover the specific failure messages would do the trick. I'll take a look and sort it in the next day or two.

nickjwhite:
> The macos e2e CI jobs install podman 6, can we add test coverage there?

Apologies it took me a while to get to this. I added an extra e2e list test which just checks that 'WARNING' is not printed to stderr, which is enough to catch the error before this fix. If you had something else in mind let me know and I'm happy to rework as appropriate.

olliewalsh (approved):
LGTM

olliewalsh:
konflux failure is not related (testing a fix in https://github.com/containers/ramalama/pull/2922)

PR #2888, opened 2026-08-15
Fix broken star history chart

OctoBored:
The star history chart shown in the README is broken because of GitHub stargazer API restrictions. This PR updates the chart links to star-history.dera.page, which uses an alternative data source that requires no API token, so the chart renders correctly again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1266,11 +1266,11 @@ This project wouldn't be possible without the help of other projects like:
 
 so if you like this tool, give some of these repos a :star:, and hey, give us a :star: too while you are at it.
 
-<a href="https://www.star-history.com/#containers/ramalama&type=date&legend=top-left">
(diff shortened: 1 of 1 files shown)


olliewalsh:
@Mubelotix as the owner of the star-history.dera.page project, is @OctoBored connected to your project in any way? Looks like a bot submitting PRs to many repos to switch from star-history.com.

OctoBored:
I do send lots of PRs but I ain't a bot

olliewalsh:
> I do send lots of PRs but I ain't a bot

I can see over a hundred commits from your account in the past ~week, all switching the star history url. No commits before that. There is no identifiable human associated with the account and the account activity is private. If you are not a bot then I suggest you stop behaving like one.

Closing this. The star history graph in the README really isn't important or particularly relevant to users. I'll just remove it.

PR #2889, opened 2026-08-16
Fix engine-args in router-mode

mkristian:
This PR allows to set engine-args when using router-mode, i.e. none or many models with serve and sandbox command.

Use case: setup a presets.ini and use it in router-mode, for this you need to mount a volume (engine-args) and add a runtime-args for the presets file

### DIFF
--- a/ramalama/plugins/runtimes/inference/llama_cpp.py
+++ b/ramalama/plugins/runtimes/inference/llama_cpp.py
@@ -580,7 +580,7 @@ class LlamaCppPlugin(LlamaCppCommands, ContainerizedInferenceRuntimePlugin):
             container_host_path = get_container_mount_path(host_path)
             engine.add([f"--mount=type=bind,src={container_host_path},destination={mount_path},ro{engine.relabel()}"])
 
-        engine.add([args.image] + cmd)
+        engine.add_container_image(args.image, cmd)
         return engine
 
     def _serve_router(self, args: argparse.Namespace) -> None:
--- a/test/unit/test_router_mode.py
+++ b/test/unit/test_router_mode.py
@@ -123,6 +123,25 @@ class TestServeRouter:
         with pytest.raises(SystemExit):
             self.plugin._serve_router(args)
 
+    @patch("ramalama.plugins.runtimes.inference.llama_cpp.enumerate_store_gguf_models", return_value=[("a", "model")])
+    @patch.object(LlamaCppPlugin, "_migrate_store_ref_files")
+    @patch("ramalama.plugins.runtimes.inference.llama_cpp_commands.should_colorize", return_value=False)
+    def test_engine_args_get_applied(self, mock_colorize, mock_migrate, mock_enum):
+        args = argparse.Namespace(
+            container=True,
+            store="/fake/store",
+            MODEL=[],
+            engine_args=["--volume", "./presets.ini:/config/presets.ini"],
+            dryrun=True,
+            engine='podman',
+            image="quay.io/ramalama/rocm",
+        )
+        engine = self.plugin._build_router_engine(args)
+
+        volume_index = engine.exec_args.index("--volume")
+        assert engine.exec_args[volume_index + 1] == "./presets.ini:/config/presets.ini"
+        assert volume_index < engine.exec_args.index(args.image)
+
 
 # ---------------------------------------------------------------------------
 # service_ready_check router mode


test/unit/test_router_mode.py
```diff
@@ -123,6 +123,25 @@ def test_no_models_exits(self, mock_port, mock_accel, mock_migrate, mock_enum):
...
 
+    @patch("ramalama.plugins.runtimes.inference.llama_cpp.enumerate_store_gguf_models", return_value=[("a", "model")])
+    @patch.object(LlamaCppPlugin, "_migrate_store_ref_files")
+    @patch("ramalama.plugins.runtimes.inference.llama_cpp_commands.should_colorize", return_value=False)
+    def test_engine_args_get_applied(self, mock_colorize, mock_migrate, mock_enum):
+        args = argparse.Namespace(
+            container=True,
+            store="/fake/store",
+            MODEL=[],
+            engine_args=["--volume", "./presets.ini:/config/presets.ini"],
```
olliewalsh, line 134: IIUC the intention is to use an engine_args bind mount to provided an [ini file](https://github.com/ggml-org/llama.cpp/blob/master/docs/preset.md). We probably should look at managing this config file in the longer term... @bmahabirbu WDYT?

mkristian: I started on this ini file already https://github.com/containers/ramalama/pull/2905 but wanted this fix to be in first, and the other PR is not ready yet.

I am OK to drop this one in favor of having only one PR

test/unit/test_router_mode.py
```diff
@@ -123,6 +123,25 @@ def test_no_models_exits(self, mock_port, mock_accel, mock_migrate, mock_enum):
...
+            dryrun=True,
+            engine='podman',
+            image="quay.io/ramalama/rocm",
```
olliewalsh, line 137: why use the rocm image here? It does not seem significant to the test.
The preset.ini mount also doesn't seem significant. I would just set an arbitrary engine arg and confirm that it is passed through.

test/unit/test_router_mode.py
olliewalsh: Lets merge this first, looks like it was missed when engine-args was added

mkristian:
@olliewalsh in the test I just used my use-case which made me run into this bug. Happy to make the test more neutral ?

olliewalsh:
> @olliewalsh in the test I just used my use-case which made me run into this bug. Happy to make the test more neutral ?

Yes, I would make it more generic, keep the tests as simple as possible

test/unit/test_router_mode.py
```diff
@@ -123,6 +123,25 @@ def test_no_models_exits(self, mock_port, mock_accel, mock_migrate, mock_enum):
         with pytest.raises(SystemExit):
             self.plugin._serve_router(args)
 
+    @patch("ramalama.plugins.runtimes.inference.llama_cpp.enumerate_store_gguf_models", return_value=[("a", "model")])
+    @patch.object(LlamaCppPlugin, "_migrate_store_ref_files")
+    @patch("ramalama.plugins.runtimes.inference.llama_cpp_commands.should_colorize", return_value=False)
```
olliewalsh, line 128: Does this need to patch `ramalama.plugins.runtimes.inference.llama_cpp.set_accel_env_vars`?

test/unit/test_router_mode.py
```diff
@@ -123,6 +123,25 @@ def test_no_models_exits(self, mock_port, mock_accel, mock_migrate, mock_enum):
...
+            store="/fake/store",
+            MODEL=[],
+            engine_args=["--annotation", "some=thing"],
+            dryrun=True,
+            engine='podman',
+            image='quay.io/ramalama/ramalama',
+        )
+        engine = self.plugin._build_router_engine(args)
+
+        volume_index = engine.exec_args.index("--annotation")
```
olliewalsh, line 141: nit: `volume_index` is no longer the correct name

test/unit/test_router_mode.py
```diff
@@ -123,6 +123,25 @@ def test_no_models_exits(self, mock_port, mock_accel, mock_migrate, mock_enum):
         with pytest.raises(SystemExit):
             self.plugin._serve_router(args)
 
+    @patch("ramalama.plugins.runtimes.inference.llama_cpp.enumerate_store_gguf_models", return_value=[("a", "model")])
+    @patch.object(LlamaCppPlugin, "_migrate_store_ref_files")
+    @patch("ramalama.plugins.runtimes.inference.llama_cpp_commands.should_colorize", return_value=False)
+    def test_engine_args_get_applied(self, mock_colorize, mock_migrate, mock_enum):
+        args = argparse.Namespace(
```
olliewalsh, line 130: use `make_ns` for consistency with the other tests

ramalama/plugins/runtimes/inference/llama_cpp.py
```diff
@@ -601,7 +601,7 @@ def _build_router_engine(self, args: argparse.Namespace) -> Engine:
             container_host_path = get_container_mount_path(host_path)
             engine.add([f"--mount=type=bind,src={container_host_path},destination={mount_path},ro{engine.relabel()}"])
 
-        engine.add([args.image] + cmd)
+        engine.add_container_image(args.image, cmd)
```
olliewalsh, line 604: does this need to be done in `sandbox.py` too?

mkristian: not sure, but I did use this method for adding the image in `sandbox.py` instead of `add_args`

test/unit/test_router_mode.py
olliewalsh: @mkristian ? I think this needs to be patched

test/unit/test_router_mode.py
```diff
@@ -117,12 +117,21 @@ def test_nocontainer_exits(self):
...
-    def test_no_models_exits(self, mock_port, mock_accel, mock_migrate, mock_enum):
-        args = argparse.Namespace(container=True, store="/fake/store", port="8080", MODEL=[])
+    def test_no_models_exits(self, mock_port, mock_accel, mock_migrate):
+        args = make_ns(MODEL=[])
         with pytest.raises(SystemExit):
             self.plugin._serve_router(args)
 
+    @patch("ramalama.plugins.runtimes.inference.llama_cpp.enumerate_store_gguf_models", return_value=[("a", "model")])
+    @patch.object(LlamaCppPlugin, "_migrate_store_ref_files")
+    def test_engine_args_get_applied(self, mock_colorize, mock_migrate):
```
olliewalsh, line 127: (from previous review) Does this need to patch ramalama.plugins.runtimes.inference.llama_cpp.set_accel_env_vars?

olliewalsh (changes requested):

test/unit/test_router_mode.py
mkristian: I removed all occurances of `ramalama.plugins.runtimes.inference.llama_cpp.set_accel_env_vars` in this file now.

olliewalsh: You need to add it for this test, not remove it from the other tests. It's to disable GPU detection during the unit tests.

olliewalsh: If you don't then this test will set environment variables based on the host hardware which could affect other unit tests

mkristian: OK - I did not understand what this mock_accel was/is doing. Still a bit uncertain though I got the idea behind it now.

olliewalsh: ack - unit tests should not be affected by the specific host that they are run on, and not affect subsequent unit tests by altering the state e.g by setting the GPU specific env vars for this case. For CI it's possibly not an issue as we don't currently have any jobs with a real GPU but it could affect developers running unit tests on their host that do have GPUs.

PR #2894, opened 2026-08-17
gguf_inspect: support NVFP4 format

ueno:
When pulling a model in the NVFP4 format, it fails with: "Error:
Failed to pull model: 40 is not a valid GGML_TYPE".

It is registered in gguf-py as:
https://github.com/ggml-org/llama.cpp/blob/7077abbe14c510cb829c93a1328c2815b5805ebd/gguf-py/gguf/constants.py#L5293

### DIFF
--- a/ramalama/model_inspect/gguf_parser.py
+++ b/ramalama/model_inspect/gguf_parser.py
@@ -54,6 +54,9 @@ class GGML_TYPE(IntEnum):
     # GGML_TYPE_IQ4_NL_4_8 = 37,
     # GGML_TYPE_IQ4_NL_8_8 = 38,
     GGML_TYPE_MXFP4 = (39,)  # MXFP4 (1 block)
+    GGML_TYPE_NVFP4 = (40,)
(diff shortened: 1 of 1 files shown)


ramalama/model_inspect/gguf_parser.py
```diff
@@ -54,6 +54,7 @@ class GGML_TYPE(IntEnum):
     # GGML_TYPE_IQ4_NL_4_8 = 37,
     # GGML_TYPE_IQ4_NL_8_8 = 38,
     GGML_TYPE_MXFP4 = (39,)  # MXFP4 (1 block)
+    GGML_TYPE_NVFP4 = (40,)
```
olliewalsh, line 57: @ueno why not add the other new types too?

ueno: Added `Q1_0` and `Q2_0` as well.

olliewalsh (approved):
LGTM

PR #2899, opened 2026-08-19
update RAG deps and sandbox images to their latest versions

olliewalsh:
llama-cpp-scripts is updated to b10485 for consistency with the current llama.cpp version. All python deps in the -rag and -tools images are updated to their latest versions.

The pi-agent image is updated to 0.84.2 running on nodejs 26.7.0. pi-llama-cpp and pi-web-access are updated to their latest versions.

Goose is updated to 1.46.0.

OpenCode is updated to 1.18.18.


olliewalsh:
@bmahabirbu PTAL

bmahabirbu (approved):
Overall LGTM

haven't tested for the pi agent update dependencies but it should be fine!

olliewalsh:
> Overall LGTM
> 
> haven't tested for the pi agent update dependencies but it should be fine!

IIUC konflux runs the slow e2e tests which should cover this

PR #2900, opened 2026-08-19
ci: work around hanging Azure apt mirror

olliewalsh:
GitHub-hosted Ubuntu runners default to azure.archive.ubuntu.com in /etc/apt/apt-mirrors.txt, which intermittently hangs. Replace it with the canonical archive.ubuntu.com before the first apt use in each Linux job so apt keeps a working mirror.

Assisted-by: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,9 @@ jobs:
       # is unreliable. The distro package is sufficient.
       - name: Install go-md2man
         run: |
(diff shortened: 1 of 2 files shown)


olliewalsh:
@bmahabirbu PTAL

olliewalsh:
Force merging this to avoid CI hangs e.g https://github.com/containers/ramalama/actions/runs/32279112521/job/96153405821?pr=2899

PR #2896, opened 2026-08-18
Bump llama.cpp to v0.1.2

olliewalsh:

PR #2904, opened 2026-08-20
openvino: bump OpenVINO to 2026.2.1 to match llama.cpp v0.1.2

olliewalsh:
The llama.cpp OpenVINO backend at v0.1.2 uses the GatherMatmul and GatedDeltaNet internal ops, which are only available in OpenVINO 2026.2.1. Building against 2026.0.0 fails at link time with undefined references to these symbols. Bump to the same OpenVINO version pinned by the upstream llama.cpp.


olliewalsh:
This was broken by https://github.com/containers/ramalama/pull/2896 because the openvino-on-pull-request konflux job was not triggered on that PR

bmahabirbu (approved):
LGTM

PR #2905, opened 2026-08-20
Feat: easy presets for router mode

mkristian:
This usecase already manifested in the other PR #2889 and adds

- [x] make the aliases llama.cpp and mounted files equal, i.e. remove all `.gguf` from the file name not only the extension
- [x] adds a presets.ini file
- [ ] <del>find out whether the presets file works with single model mode</del>
   - [ ] <del>make sure we use the same file names for mounts as in router mode which helps to figure the model name to write a such presets.ini</del>
- [x] test cases
- [x] fix the perplexity subcommand to relabel the bind mount for SE-Linux and podman (extra commit)

### DIFF
--- a/ramalama/plugins/runtimes/inference/common.py
+++ b/ramalama/plugins/runtimes/inference/common.py
@@ -285,7 +285,7 @@ def enumerate_store_gguf_models(
             ref_file = ref_json_cls.from_path(ref_file_path)
 
             tag, _ = os.path.splitext(ref_file_name)
-            model_rel = root.replace(store.path, "").lstrip(os.sep)
+            model_rel = root.replace(store.path, "").lstrip(os.sep).replace(".gguf", "")
             parts = model_rel.split(os.sep)
             readable = "-".join(parts + [tag])
 
--- a/ramalama/plugins/runtimes/inference/llama_cpp_commands.py
+++ b/ramalama/plugins/runtimes/inference/llama_cpp_commands.py
@@ -170,7 +170,14 @@ class LlamaCppCommands:
 
         return cmd
 
-    _cmd_serve = _cmd_run
+    def _cmd_serve(self, args: argparse.Namespace) -> list[str]:
+        cmd = self._cmd_run(args)
+
+        presets_file = getattr(args, 'presets_file', None)
+        if presets_file:
+            cmd += ["--presets", "/etc/presets.ini"]
+
+        return cmd
 
     def _cmd_perplexity(self, args: argparse.Namespace) -> list[str]:
         cmd = ["llama-perplexity"] if not self._container_image_is_ggml(args) else ["--perplexity"]  # type: ignore[attr-defined]
--- a/docs/options/presets-file.md
+++ b/docs/options/presets-file.md
@@ -1,0 +1,6 @@
+####> This option file is used in:
+####>   ramalama sandbox goose, ramalama sandbox opencode, ramalama sandbox pi, ramalama serve
+####> If this file is edited, make sure the changes
+####> are applicable to all of those.
+#### **--presets-file**=*path*
+Presets ini-file for router mode.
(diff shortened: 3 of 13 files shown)


ramalama/plugins/runtimes/inference/common.py
```diff
@@ -285,7 +285,7 @@ def enumerate_store_gguf_models(
             ref_file = ref_json_cls.from_path(ref_file_path)
 
             tag, _ = os.path.splitext(ref_file_name)
-            model_rel = root.replace(store.path, "").lstrip(os.sep)
+            model_rel = root.replace(store.path, "").lstrip(os.sep).replace(".gguf", "")
```
rhatdan, line 288: model_rel = root.replace(store.path, "").lstrip(os.sep).strip(".gguf")

ramalama/plugins/runtimes/inference/llama_cpp_commands.py
```diff
@@ -170,7 +170,16 @@ def _cmd_run(self, args: argparse.Namespace) -> list[str]:
...
 
-    _cmd_serve = _cmd_run
+    def _cmd_serve(self, args: argparse.Namespace) -> list[str]:
+        cmd = self._cmd_run(args)
+
+        router_mode = getattr(args, 'router_mode', False)
+        if router_mode:
+            presets_file = getattr(args, 'models_preset', None)
+            if presets_file:
+                cmd += ["--models-preset", "/etc/presets.ini"]
```
rhatdan, line 180: This looks like it is hard coded to /etc/presets.ini?

ramalama/plugins/runtimes/inference/common.py
olliewalsh, line 288: what if it's a safetensors model? Or a split model?

ramalama/plugins/runtimes/inference/llama_cpp_commands.py
mkristian: yes, this hardcoded to the place the bind_mount will put the presets.ini file inside the container. any other place will do as well.

Or the remark on I shall put this string into a constant using it at these two place ? I think I should to this to make this clearer.

ramalama/plugins/runtimes/inference/common.py
mkristian: This just a convenient so the model name llama.cpp prints into console on startup matches the file name ramalama uses for the gguf file inside the container. Example

llama.cpp output
```
0.00.520.965 I srv   load_models: Loaded 17 local model presets from /mnt/models
0.00.521.346 I srv    operator(): Available models (17) (*: custom preset)
0.00.521.348 I srv    operator():     https-huggingface.co-luxuansang-Qwen3.8-9B-heretic-uncensored-NVFP4-GGUF-Qwen3.8-9B-distill-heretic_nvfp4_q4_k_m-main
0.00.521.349 I srv    operator():     huggingface-Jackrong-Qwopus3.5-4B-Coder-MTP-GGUF-Qwopus3.5-4B-Coder-MTP-IQ4_XS-latest
```
inside the container
```
bash-5.1# ls /mnt/models/
https-huggingface.co-luxuansang-Qwen3.8-9B-heretic-uncensored-NVFP4-GGUF-Qwen3.8-9B-distill-heretic_nvfp4_q4_k_m.gguf-main.gguf
huggingface-Jackrong-Qwopus3.5-4B-Coder-MTP-GGUF-Qwopus3.5-4B-Coder-MTP-IQ4_XS.gguf-latest.gguf
```

So the intention is to cleanup these `gguf-main.gguf` to `main.gguf` and `gguf-latest.gguf` to `latest.gguf`
so the filename on the mount matches the name llama.cpp prints out on the console. This again helps to setup your presets.ini file where you need to mention the complete path of the mounted file as used inside the container.

But I see that there are some side effects on the other runtimes which can use safetensors or split models.

BTW I using mostely huggingface urls from the `download with hf cli` which is usually a single gguf file.

If this is beyond this feature to establish presets - I am happy to move it to another PR for this topic.

PR #2909, opened 2026-08-21
Packages skills, agents, and plugins as OCI artifacts.

### LINKED ISSUE #2859: RFE: OCI artifacts for skills, agents and plugins

Using `ramalama sandbox` I can start a containerized (goose|opencode|pi) and use the skills, agents, plugins in the `/work/.agents` directory (or equivalent). Instead of copy&pasting these files between projects and relying solely on the presence in the mounted directory, it would be great to leverage OCI artifacts for cross-cutting skills etc.

AI skills, agents, plugins, etc. could be packaged as OCI artifacts - similar to AI models - and mounted on demand into the sandbox container. This way the same distribution channel as for images and models could be used.

Similar to the `shortnames.conf` for AI models, another file (e.g.`shortnames-skills.conf`) could be used to define shortnames for these OCI artifacts. 

#### Proposed CLI changes

- `ramalama [skill | agent | plugin] build -d <path> -t <tag>`
-> Creates a .tar.gz for the <path> and adds it to the artifact (overwriting existing one locally). The tarball is required since skills etc. can consist of multiple files and (nested) directories.  

- `ramalama [skill | agent | plugin] ls`
-> Lists locally available OCI artifacts.

- `ramalama sandbox goose --skill <full name or shortname> <model>`
-> The `--skill <name>` option can be repeated n times
-> Run the local model and AI agent, (optionally) resolve the skill shortname and mount the OCI artifact into the sandbox container. In order to make the skill inside the tarball usable, it needs to be extracted into the respective directory (e.g. `~/.agents/skills` for goose). This could be done by overwriting the entrypoint and running a plain bash script first, unpacking these tarballs.

#### shortnames for skill artifacts

```toml
[shortnames.skills]
"wiki-kb" = "quay.io/ramalama/wiki-kb"
...
[shortnames.agents]
...
```

_No response_

### Additional context

_No response_

Mennatullah122:
## Summary

Implements #2859 — packages skills, agents, and plugins as OCI artifacts,
mirroring how RamaLama already distributes AI models, and makes them
mountable into `ramalama sandbox` containers.

## What's included

- `ramalama [skill|agent|plugin] build -d <path> -t <tag>` — tars a directory
  and adds it as a local OCI artifact via the container engine
- `ramalama [skill|agent|plugin] ls` — lists locally available artifacts of
  that kind (queries the engine's own artifact storage)
- `ramalama [skill|agent|plugin] push/pull` — publish/fetch artifacts via a
  remote registry, reusing the existing OCI transport machinery used for
  models
- `ramalama sandbox <agent> --skill/--agent/--plugin <name>` (each
  repeatable) — resolves shortnames, then mounts the artifact into the
  sandbox container, either via native OCI artifact mounting (Podman ≥5.7.0)
  or a pull+extract+bind-mount fallback on older engines/Docker
- Per-kind shortnames support: `shortnames-skills.conf`,
  `shortnames-agents.conf`, `shortnames-plugins.conf`, each with a matching
  `[shortnames.<kind>]` section, matching the syntax proposed in the issue
- New CNAI artifact type (`application/vnd.cncf.skill.manifest.v1+json`) so
  skill/agent/plugin artifacts are distinguishable from AI model artifacts,
  including in the pull-side manifest validation

## Deviation from the issue's suggested implementation

The issue suggested unpacking tarballs via an entrypoint-overwrite script.
Instead, this mounts artifacts natively via `podman artifact` mount support
where available, falling back to host-side extraction + bind-mount
otherwise — avoids entrypoint hacking and preserves file metadata when the
engine supports it.

## Known limitations

- `<kind> pull` currently caches pulled artifacts via the same storage used
  for AI models, rather than a dedicated skills/agents/plugins location.
  Functionally correct, but not yet cleanly separated.

## Testing

- Unit tests added/updated: shortnames parameterization, artifact
  build/push/pull/ls CLI, sandbox artifact preparation
- Manually verified against real Podman 6.0.2 (Windows/WSL2): `build`
  correctly creates a tagged artifact with the new artifact type, `ls`
  correctly filters and lists it, and the sandbox skill-resolution/mount
  logic runs cleanly against real `podman artifact` output. Full
  container-runtime verification (confirming mounted files are visible
  inside a running sandbox container) was not completed due to local
  Podman-on-Windows connectivity instability unrelated to this change.
- Noted but unrelated to this PR: `test_agent_workdir` in
  `test_sandbox_cmd.py` fails on current `main` as well (pre-existing,
  confirmed via `git stash`), unrelated to this change.

Closes #2859

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,5 @@ uv.lock
 *~
 *#
 .mypy_cache
+
+.venv/
--- a/docs/skills_agents_plugins.md
+++ b/docs/skills_agents_plugins.md
@@ -1,0 +1,87 @@
+# Skills / Agents / Plugins OCI Artifacts
+
+This document describes the `skill`, `agent` and `plugin` CLI subcommands and how to use OCI-style tarball artifacts with `ramalama sandbox`.
+
+## Building artifacts
+
+Package a directory into a local artifact tarball and tag it:
+
+```bash
+ramalama skill build -d ./my-skill -t quay.io/ramalama/wiki-kb
+ramalama agent build -d ./my-agent -t quay.io/ramalama/my-agent
+ramalama plugin build -d ./my-plugin -t quay.io/ramalama/my-plugin
+```
+
+Artifacts are stored under your RamaLama store path in `artifacts/{skills,agents,plugins}`.
+
+## Listing artifacts
+
+List local artifacts:
+
+```bash
+ramalama skill ls
+ramalama agent ls --json
+ramalama plugin ls --path
+```
+
+- `--json` prints structured JSON with `tag`, `file`, `size`, `modified`.
+- `--path` prints the full file paths for scripting.
+
+## Using artifacts in sandbox
+
+You can pass `--skill` (repeatable) to `ramalama sandbox` subcommands to mount skills into the sandboxed agent:
+
+```bash
+ramalama sandbox goose --skill quay.io/ramalama/wiki-kb <model>
+```
+
+Shortnames may be defined in `shortnames.conf` using a `[shortnames.skills]` section. Example:
+
+```ini
+[shortnames.skills]
+"wiki-kb" = "quay.io/ramalama/wiki-kb"
+```
+
+The CLI resolves shortnames before locating artifact tarballs in the local artifact store.
+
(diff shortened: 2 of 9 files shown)


rhatdan:
@Mennatullah122 Please respond to the AI Agents and rebase your PR?

Mennatullah122:
@rhatdan, I finished working on the comments and rebase the PR.

PR #2920, opened 2026-09-05
Rag private network

olliewalsh:
The ramalama rag ingestion pipeline reached its llama.cpp helper servers (docling VLM, embedding, optional captioning) from the doc2rag container via host.containers.internal, which required publishing those ports on a wildcard host interface (the expose_to_containers stopgap from PR #2919).
This creates a private container network for the run instead, and has the containers reach each other by container name.

Depends on PR #2919


olliewalsh:
@bmahabirbu @mikebonnet @rhatdan PTAL

bmahabirbu (approved):
Gave it a quick test looks good

PR #2922, opened 2026-09-08
e2e: tolerate read timeouts while waiting for ollama server

olliewalsh:
The ollama_server fixture polls GET / until the server responds, but only caught requests.exceptions.ConnectionError. ReadTimeout derives from Timeout, not ConnectionError, so once ollama binds the port and is slow to answer (e.g. while it runs GPU discovery at startup) the 0.5s read timeout escaped the retry loop and aborted fixture setup with an error instead of retrying.

Catch Timeout as well, matching the container_registry fixture, and give the final readiness assertion a timeout so it cannot hang indefinitely.

### DIFF
--- a/test/e2e/conftest.py
+++ b/test/e2e/conftest.py
@@ -133,12 +133,12 @@ class OllamaServer:
             try:
                 requests.get(self.url, timeout=0.5)
                 break
-            except requests.exceptions.ConnectionError:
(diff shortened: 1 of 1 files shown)


olliewalsh:
@bmahabirbu @mikebonnet @rhatdan PTAL, fixes a flaky test in konflux CI

olliewalsh:
Merging this to get CI green

PR #2927, opened 2026-09-14
feat: add start command and allow stopping multiple containers

### LINKED ISSUE #2555: Ability to start containers + Ability to stop or start multiple containers

1. I use the stop command(`ramalama stop <c-id>`) when I don't use the containers. But to start them back up, i have to run `podman start` command instead. Wouldn't start command getting included compliment the already existing stop command, making it easier to bring containers back up instead of going to the container engine.
2. Sometimes i stop multiple containers and currently you can stop only one, or all containers via `ramalama stop`. An enhancement to make it similar to podman would be making it have variable length args so you can run `ramalama stop <c-id1> <c-id2>`.

1. For start of containers, creating equivalent parser and functions to call respective engine to bring up the container as implemented here by [stop](https://github.com/containers/ramalama/blob/main/ramalama/cli.py#L984).
2. Currently stop containers takes only one name as a parameter [here](https://github.com/containers/ramalama/blob/main/ramalama/engine.py#L388), which can be optimized by updating argparse and the function to accept variable length args.

Let me know if this makes sense and I can try to pick it up when I get some time.

This is an enhancement and alternatives is always directly talking to the engine instead of ramalama.

### Additional context

_No response_

mmustafasenoglu:
Complements 'ramalama stop' (which only stops one container or all) in two ways asked for in #2555:

- new 'ramalama start NAME [NAME...]' command that starts stopped containers via the engine, so there's no need to drop down to podman/docker directly
- 'ramalama stop' now accepts multiple names: 'ramalama stop a b'

Also added man pages (new ramalama-start.1, updated ramalama-stop.1) and unit tests for the engine calls plus CLI parsing/dispatch.

Verified with the full unit suite (1127 passed), ruff check/format clean, codespell clean. Couldn't run the container e2e here since there's no engine in this environment.

Fixes #2555

Assisted-by: Claude <noreply@anthropic.com>


mmustafasenoglu:
Good catch on the pod round trip — fixed. stop_container with remove=False now runs pod stop instead of pod rm --force, so pod-backed containers survive stop/start. The cleanup callers passing remove=True (rag, sandbox, chat) keep the removal behavior. Added tests for both paths. Also added the -> None annotations. Full unit suite (1129) plus ruff pass.

rhatdan:
Looks like  you need a rebase.

mmustafasenoglu:
Rebased onto current main (resolved a conflict with the new create_network helpers — kept both) and force-pushed. Full unit suite (1141) plus ruff pass on the rebased branch.

mmustafasenoglu:
Skipping the annotation nitpick deliberately: every sibling (stop_parser, stop_container, rm_parser, containers, inspect, logs) leaves args/subparsers unannotated, and there's no fitting Protocol (BaseEngineArgsType has no ignore/NAME). create_network/remove_network are upstream's, not mine. The -> None returns from the last round stay. Makefile already excludes tests from mypy, and --check-untyped-defs still checks the new bodies.

olliewalsh:
> Skipping the annotation nitpick deliberately: every sibling (stop_parser, stop_container, rm_parser, containers, inspect, logs) leaves args/subparsers unannotated, and there's no fitting Protocol (BaseEngineArgsType has no ignore/NAME). create_network/remove_network are upstream's, not mine. The -> None returns from the last round stay. Makefile already excludes tests from mypy, and --check-untyped-defs still checks the new bodies.

Please do not use AI to generate responses  to review comment, see https://github.com/containers/ramalama/blob/main/CONTRIBUTING.md#ai-assisted-contributions

mmustafasenoglu:
You're right, that reply was AI-polished and I shouldn't have pasted it as-is. In my own words: I skipped the annotations because none of the functions around them annotate those args either, so annotating only mine would stick out. But it's your call — if you'd rather have the new functions annotated, I'll add them.

olliewalsh:
> You're right, that reply was AI-polished and I shouldn't have pasted it as-is. In my own words: I skipped the annotations because none of the functions around them annotate those args either, so annotating only mine would stick out. But it's your call — if you'd rather have the new functions annotated, I'll add them.

That looks even more AI generated!

PR #2932, opened 2026-09-16
Enable vulkan on NVIDIA

olliewalsh:
The nvidia-container-toolkit supports Vulkan.
This adds the missing deps to the ramalama image to allow it to be used, enables the `graphics` driver capability (required for docker support), and refactors the device pass-through logic to work with Vulkan.
The `--backend` cli arg now allows vulkan on NVIDIA.


bmahabirbu (approved):
Very cool!

PR #2933, opened 2026-09-16
fix: refcount blobs by hash when removing snapshots

boaz2026:
Snapshot removal used file names when counting blob references. If two refs pointed at the same blob hash under different file names, removing one ref could delete the shared blob while the other ref still referenced it.

This patch counts blob references by hash instead, matching how blobs are stored on disk.

A script to reproduce the bug:
```bash
STORE="$(mktemp -d)"
BASE="$STORE/store/huggingface/acme/shared-model"
mkdir -p "$BASE/refs" "$BASE/blobs" "$BASE/snapshots"

echo "shared model bytes" > "$BASE/blobs/sha256-shared"

cat > "$BASE/refs/A.json" <<EOF
{
  "version": "v1.0.1",
  "hash": "snap-a",
  "path": "$BASE/refs/A.json",
  "files": [
    {"hash": "sha256-shared", "name": "model-a.gguf", "type": "gguf"}
  ]
}
EOF

cat > "$BASE/refs/B.json" <<EOF
{
  "version": "v1.0.1",
  "hash": "snap-b",
  "path": "$BASE/refs/B.json",
  "files": [
    {"hash": "sha256-shared", "name": "renamed-model.gguf", "type": "gguf"}
  ]
}
EOF

echo "Before rm:"
find "$BASE" -type f -print

ramalama --store "$STORE" --nocontainer rm huggingface://acme/shared-model:a

echo "After rm:"
find "$BASE" -type f -print
test -e "$BASE/blobs/sha256-shared" && echo "OK" || echo "BUG: shared blob deleted"

rm -rf $BASE
```


olliewalsh:
@boaz2026 have you actually encountered this issue in practice?

PR #2935, opened 2026-09-18
run, server: add option to enable/disable multimodal support

ueno:
Enabling multimodal support with llama-server consumes VRAM (or host memory if --no-mmproj-offload is used). ramalama currently always enables multimodal support if a mmproj file is detected and does not provide a way of disabling it. This patch adds an --mtmd option to allow users to explicitly disable multimodal support.


ramalama/plugins/runtimes/inference/llama_cpp.py
```diff
@@ -506,6 +509,12 @@ def _add_inference_args(self, parser: "argparse.ArgumentParser", command: str) -
                 help="min speculative decoding probability (default: 0.0)",
                 completer=suppressCompleter,
             )
+            parser.add_argument(
+                "--mtmd",
+                default=None,
```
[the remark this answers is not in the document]
olliewalsh: @ueno this looks valid

ramalama/plugins/runtimes/inference/llama_cpp_commands.py
```diff
@@ -72,7 +72,14 @@ def _cmd_run(self, args: argparse.Namespace) -> list[str]:
             model_path = model._get_entry_model_path(is_container, should_generate, dry_run)
             cmd += ["--model", model_path]
 
-            mmproj_path = model._get_mmproj_path(is_container, should_generate, dry_run)
+            mtmd = getattr(args, 'mtmd', None)
+            if mtmd is None or mtmd:
```
olliewalsh: Can simplify this to:
```
mmproj_path = None
if mtmd is not False:
   ...
   ...
```

olliewalsh:
@ueno also need to handle this in the generated quadlet/kube etc...

ramalama/plugins/runtimes/inference/llama_cpp_commands.py
```diff
@@ -72,7 +72,14 @@ def _cmd_run(self, args: argparse.Namespace) -> list[str]:
...
+                mmproj_path = model._get_mmproj_path(is_container, should_generate, dry_run)
+                if mtmd and not mmproj_path:
```
olliewalsh: will this raise for a --dry-run?

ramalama/plugins/runtimes/inference/llama_cpp_commands.py
ueno: Thanks, done.

ramalama/plugins/runtimes/inference/llama_cpp_commands.py
ueno: Good catch, removed the `raise` and adjusted the test.

ueno:
> @ueno also need to handle this in the generated quadlet/kube etc...

I haven't tried kube, but for quadlet, it already generates a file according to the option.

ramalama/plugins/runtimes/inference/llama_cpp.py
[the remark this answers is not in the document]
olliewalsh: Is this fixed, or not valid?

